PR #99, opened 2017-03-06
People Page

alanmoo:
@gvn @gideonthomas @cadecairos I made some changes to the schema, are these going to be an issue?


source/json/people.schema.json
```diff
@@ -36,7 +36,25 @@
         }
       },
       "image": {
-        "type": "string"
+        "type": "string",
+        "faker": "internet.avatar"
+      },
+      "internetHealthIssues": {
```
gideonthomas: can we stick to an `_` convention that Django uses? Would rather not write a bunch of extra code to get it to do camelcasing.

gideonthomas:
other than what I mention above, don't think any of these changes will be hard to implement

source/json/people.schema.json
alanmoo: lowercased then,`internet_health_issues`?

gvn:
FYI: We're probably going to have to add another image field for partner logo.

alanmoo:
@gvn this is ready for review. It's a bit WET but trying to be very DRY (between featured and unfeatured cards) was a huge time-sink. 

Filtering has been put on hold for now, and I put in placeholders based on best guesses as much as I could. This might need another pass once we're further along but it felt like a good point to bring things in then re-evaluate.

alanmoo:
@cadecairos and @gideonthomas I added a few more fields in the schema based on what was in the spreadsheet. This *should* be all I need for this page. I'm not sure if you use the "required" property, but I had to put a few extra items in there for faker to generate things properly; there's actually some logic about when items are required or not. Does the schema as it exists in this PR present a problem?

gideonthomas:
@alanmoo yeah I think this should be fine.

source/js/components/people/person.jsx
```diff
@@ -3,16 +3,112 @@ import React from 'react';
...
+              </div>
+              <div className="row">
+                <div className="col">
+                  <img className="partnership_logo" src={this.props.metadata.partnership_logo} alt="Logo of partnered organization"/>
+                </div>
+              </div>
+            </div>
+            <div className="col d-flex flex-column">
+              <div className="justify-content-end row no-gutters">
+                <div className="col col-auto" tabIndex="1" onClick={this.flip} >
```
gvn: This should be a `button`

source/js/components/people/person.jsx
```diff
@@ -3,16 +3,112 @@ import React from 'react';
...
+            </div>
+          </div>
+        </div>
+      );
+    } else {
+      return (
+        <div className="col-md-6 col-12 p-3 mb-4">
+          <div className="person-card row no-gutters">
+            <div className="col col-sm-4 mr-3">
+              <img src={this.props.metadata.image} style={{"width":`100%`}} alt="Headshot" />
```
gvn: Why is this an inline style?

source/js/components/people/person.jsx
```diff
@@ -3,16 +3,112 @@ import React from 'react';
...
+              <div className={`row flex-1 flippable-card flip-${this.state.flipped}`}>
+                <div className="col-12 bio-content">
+                  <h2 className="h5-black my-2">{this.props.metadata.name}</h2>
+                  <div className="small">{this.props.metadata.role} / {this.props.metadata.location}</div>
+                  <div className="person-issues">{issues}</div>
+                  <div className="person-affiliations small-gray mt-2">{this.props.metadata.affiliations}</div>
+                  <div className="person-bio body-black">
+                    <ul>
+                      {this.props.metadata.bio.map((item, index)=>{
+                        console.log(item);
```
gvn: remove console

source/js/components/people/person.jsx
```diff
@@ -3,16 +3,112 @@ import React from 'react';
...
+    });
+
+    if (this.props.metadata.featured) {
+      return (
+        <div className="col-12 p-3 mb-4">
+          <div className="person-card person-card-featured row no-gutters">
+            <div className="col-12 col-sm-4 mr-3">
+              <div className="row">
+                <div className="col">
+                  <img src={this.props.metadata.image} alt="Headshot" />
```
gvn: Let's add a width: 100% to these

alanmoo: Makes sense, but I changed the column classes to (always 4 columns, even at xs) because we were getting *giant* faces at xs sizes.

source/js/components/people/person.jsx
alanmoo: Debugging. Consolidated it.

source/js/components/people/person.jsx
alanmoo: Done, but since buttons already have an arrow in their `::after` pseudo-content, should we roll with that?

alanmoo:
Changes made, please check out the new commit.

gideonthomas:
@alanmoo weren't we gonna change the bio to a single string?

gvn:
R+!

PR #101, opened 2017-03-06
WIP Slanted background classes

alanmoo:


source/sass/utilities.scss
```diff
@@ -5,3 +5,39 @@
...
+    content: '';
+    width: 0.6em;
+    background-color: inherit;
+    transform: skew(-20deg);
+    display: block;
+    position: absolute;
+    height: 100%;
+    margin-right: -0.3em;
+    right: 0;
+    top: 0;
```
gvn: Can we make this a little DRY-er?

alanmoo:
Do you have an example of what it's supposed to look like on multiple lines?

gvn:
I don't... @taisdesouzalessa can you provide an example of that?

taisdesouzalessa:
Hey @gvn @alanmoo 

@sabrinang  and I came up with 2 options for that. We would like to hear your thoughts on that in terms of feasibility:

screen shot 2017-03-07 at 4 17 39 pm

screen shot 2017-03-07 at 4 17 16 pm

alanmoo:
My instinct says the first one will be much easier to implement

taisdesouzalessa:
@alanmoo that is what we thought as well. Let's run with the first option then. We expect the cases in which we need 2 lines to be rare since those dividers are supposed to be easy to get at a glimpse.

gvn:
I think all we need to do for these is create repeating SVG background tiles. We shouldn't need any fancy CSS if we have that.

taisdesouzalessa:
@gvn @alanmoo do you need anything from design here?

gvn:
@taisdesouzalessa Yeah, can you make a repeating SVG background tile for the slashes?

alanmoo:
As Gavin said, we need a repeating image. For an example, I made the following 
strokes

and it resulted in
screenshot 2017-03-27 12 45 38

But it was just roughly eyed, I'm not sure the best way to make a properly spaced and repeating image.

gvn:
@taisdesouzalessa This might be useful: https://medium.com/sketch-app-sources/creating-patterns-in-sketch-12cc612ef09a#.9miyigb80

taisdesouzalessa:
Thanks for the link @gvn. I usually do my patterns in Illustrator and it is nice to see an article for Sketch. 
@alanmoo could you send me the link for the page you are working on? I would like to see the code.

alanmoo:
@taisdesouzalessa while the code will change (getting it to appear on a page right now is just for demonstration purposes), this PR has been updated with the code used to generate that screenshot

taisdesouzalessa:
@alanmoo here is the slanting line pattern:
https://drive.google.com/a/mozilla.com/file/d/0B3b1KWRwjxtocVdaV1Z3dWFBRTg/view?usp=sharing

I did some tests in the browser, playing with background-size property and it seems to be working seamlessly. 

screen shot 2017-03-27 at 1 49 05 pm

Let me know if you run into any issues or need any design tweak.

alanmoo:
The pattern works for me, I'll continue chipping away at this puzzle as I have a chance- it's more complicated than it might seem given the variables of screen sizes

gvn:
We resolved this in a different PR. Closing.

Issue #98, opened 2017-03-06
URLs for Landing Pages


xmatthewx:
Landing Pages will serve our ~~Offerings~~ Opportunities and, in the future, incoming traffic from Advocacy, Fellows etc.

@hannahkane proposes this URL structure. Looks good to me.

/opportunity/foobar
/opportunity/ → redirect home (for now)

gvn:
@cadecairos How do we want to handle redirects given our hosting?

hannahkane:
Is this still blocked?

gvn:
Still waiting on an answer. @cadecairos any thoughts on my q above?

cadecairos:
We can fairly easily use S3 redirect rules to send `/opportunity` and `/opportunity/doesnotexist` to the homepage

gvn:
Cool, can you set that up?

gvn:
Blocked pending discovery of how Mezzanine handles routes.

cadecairos:
My plan here is to have someone duplicate the existing landingpages in mezzanine, and then we update the CDN to proxy requests to mezzanine.

- [X] Land 'landingpages' on staging
- [ ] Land 'landingpages' on production
- [ ] add existing pages to mezzanine as 'landingpage' contenttypes
- [ ] update the Cloudfront distribution so that `/opportunity` proxies to mezzanine

xmatthewx:
@cadecairos – team suggests you add this redirect while working other related things today....

alanmoo:
@cadecairos most of this is done, just the redirect at /opportunity/ that doesn't...redirect, right? It shows the homepage but the URL remains.

cadecairos:
> @cadecairos most of this is done, just the redirect at /opportunity/ that doesn't...redirect, right? It shows the homepage but the URL remains.

This is done.

I used Cloudfront behaviours and S3 Redirect rules to implement the desired redirection.

PR #105, opened 2017-03-06
changing URL

gvn:
related #98


gvn:
We're planning to redirect it to home (see #98).

PR #107, opened 2017-03-07
News Page

### LINKED ISSUE #36: Implement News page
This page is linked from button below News on the homepage. Not in the primary Nav menu, but should sit at /news. 

Note. We will want staff to be able to edit this page content. We should probably discuss pros/cons of components before we implement that instead of html content.

Kevin has identified earned media that populate this page. @taisdesouzalessa can share copy for implementation.

gvn:
Closes #36


gvn:
Adding icons as a follow-on #108

Issue #31, opened 2017-01-30
Implement Upcoming page


xmatthewx:
Note. We will want staff to be able to edit this page content. We should probably discuss pros/cons of components before implement that instead of html content.

UX may still be a work in progress. 

- [x] finish styling
- [x] add images

taisdesouzalessa:
Wireframe of Upcoming page: https://invis.io/ZJ9YTAGEA#/216715367_Upcoming

xmatthewx:
Depends on https://github.com/mozilla/network/issues/57

taisdesouzalessa:
@xmatthewx here is the page. Please take a look at that and let me know your feedback in InVision.

https://invis.io/ZJ9YTAGEA#/216715367_Upcoming

cc: @sabrinang

xmatthewx:
Looks great. Added some comments on copy in invision. 

We will need to grab images for each item. Yikes. And, let's decide with @gvn that it's okay if we launch without images. This page can survive without.

I filed a ticket to gather images (https://github.com/mozilla/network/issues/67). Please don't hesitate to mention program people to gather some options for you. Let's distribute the work so you can focus on more critical things.

gvn:
Looks like images might have been removed from the comp? Either way, I'd be fine with launching without images. 👍

taisdesouzalessa:
The comp has just placeholders for now (solid color squares) because I am focusing on building the structure to pass it on to build. 

I would highly recommend we launch with images to make this page more exciting on launch, to take opportunity of the momentum when the network sees it for the first time. 

I think we can find more generic images for launch in case we don't hear from all program people and we can replace them as we receive better ones from staff. I agree with you @xmatthewx to focus on critical things first but I think we can focus on finding those images after the critical work is done, what do you think?

sabrinang:
These content dependent images are gathered on issue https://github.com/mozilla/network/issues/67

Some feedback to consider is the location or label of "recent" at the bottom of the upcoming page especially if it gets long. I understand the purpose of archiving/noting things that has happened but I think this page is framed for future things to expect to get involved with and people may not expect to find past things in "upcoming".

gvn:
Blocked by review in #67

xmatthewx:
Calling this unblocked. New images look good here: https://github.com/mozilla/network/issues/67#issuecomment-284091011

gvn:
@xmatthewx This pages is landed, but the links are placeholder URLs. Do we have the URLs for all the items?

[...]

PR #136, opened 2017-03-13
adding links to Upcoming page

gvn:
related #31


gvn:
Still need the rest of the links, but this is better than none...

PR #152, opened 2017-03-14
adding validation to pulled json

gvn:
**DON'T MERGE YET** – Django "People" data is currently malformed.


gvn:
This will validate the JSON being pulled from Django as a test as well as validating that `npm build` succeeds as a test.

source/json/people.schema.json
```diff
@@ -9,8 +9,7 @@
         "chance": "name"
       },
       "role": {
-        "type": "string",
-        "pattern": "Staff|OWF Host|Partner|Hive NYC Partner|Advisor|Member, MLN Advisory Group"
```
gvn: @gideonthomas @alanmoo – I had to remove this pattern to get the Django-returned JSON to validate. Is `role` being treated as a freeform string in Django? If so, is that going to be problematic for any future front-end functionality you're aware of, @alanmoo ?

gideonthomas: I can fix that if we have a specified list of acceptable roles and we want to confine it to those
cc @hannahkane

gvn: Based on stand conversation: Roles will be freeform.

Issue #64, opened 2017-02-07
Analytics: identify events and goals


xmatthewx:
A primary goal of the site is to increase membership of people with leadership potential. Let's plan our analytics to help us track efforts toward that. 

First step: identify key events to track

Second step: gather advice from @vojtechmzl

xmatthewx:
Most user action will be of three types:
 - Join form
 - Featured "Offering" to Join form
 - Outgoing links to strategic distinations  


### Events 
 - User taps Join button in header 
   - User submits Join form
 - User submits Join form on homepage 
 - User taps a featured "offering"
   - User submits Join form connected with offering
   - and/or... User follows outgoing link to the offering

@vojtechmzl – is it useful to track some outgoing links as events? For example, our member profiles will include links to their twitter/linkedin, a useful path to building network strength. Can we gain insight by tracking these as events? To see how this action trends in aggregate? Or do we have all that info without event tracking?

Here are some key outgoing links
   - Link to a member's Linkedin or Twitter (contributes to network strength)
   - Link to a featured "Offering" (e.g. Internet Health report)
   - Link to Pulse or a Pulse item
   - Other links? Probably fine tracked just as links. 


cc @taisdesouzalessa @sabrinang @gvn

vojtechmzl:
Hey Matthew,

The list of events around the sign up process makes sense. Given the site's role in serving as a directory of a lot of different activities within Mozilla I think it's a good idea to set up events that track outbound links, so that we can distinguish between abandoned visits (i.e. people who simply left) and people who engaged and followed onto a purposefully linked entity. There are some built-in functionalities to analyze outbound links, but events are I think the best option.

Events in GA can take on three attributes -> category, action and label. That means things can be neatly organized into sections. For instance 

Category: People
   Action: Clicked on Profile
      Label 1: LinkedIn
      Label 2: Twitter

Category: Programs
   Action: Visited an offering page
      Label: [name of offering]
   Action: Clicked through to offering
      Label:  [name of offering]

Once events are passed in this way to GA, we can set up specific Goals in order to track for instance how many people clicked through to people's social profiles. We can then also utilize those events in building funnels to best understand how people engage with the site before they sign up.

xmatthewx:
Thanks @vojtechmzl, this is great. It'll be especially useful for our "offerings"

For some, the flow will be a simple experience: tap b/c you're interested, join to get access (e.g. apply for a fellowship, or propose a mozfest session). 

[...]

PR #154, opened 2017-03-14
adding analytics snippet

gvn:
related #64


cadecairos:
Yes please!

https://gist.github.com/cadecairos/416e7abe6aab11b8317833924ed1e45d

gvn:
@cadecairos Unless I'm missing something that code will never run unless you invoke `WebAnalytics`, which isn't done in the snippet. I'm making it an IIFE and you might want to do that to your gist as well.

PR #158, opened 2017-03-14
Make header join button a link

### LINKED ISSUE #155: "Join Us" button in Top Nav should go to /join/
Currently our "Join Us" button on Top Nav is linking to the hidden form on the same page
screen-shot-2017-03-14-at-2 05 52-pm

**Steps to reproduce:**

- Go to: https://network.mofostaging.net/

- Click on "Join Us" from the top nav

Note: This happens on all pages.

-----

We would like it go to the standalone page: https://network.mofostaging.net/join/

alanmoo:
And not a slide-out

Closes #155

### DIFF
--- a/source/pug/templates/master.pug
+++ b/source/pug/templates/master.pug
@@ -46,7 +46,7 @@ html
                     a.logo.text-hide(href=`${env.virtualRoot}`) Mozilla Network
                   .center-nav-title.hidden-xs-down
                     block centerNav
-                  a#join-trigger.btn.btn-normal(href="/join") Join Us
+                  a.btn.btn-normal(href="/join") Join Us
       #hero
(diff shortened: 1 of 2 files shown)


source/pug/templates/master.pug
```diff
@@ -46,7 +46,7 @@ html
                     a.logo.text-hide(href=`${env.virtualRoot}`) Mozilla Network
                   .center-nav-title.hidden-xs-down
                     block centerNav
-                  button#join-trigger.btn.btn-normal Join Us
+                  a#join-trigger.btn.btn-normal(href="/join") Join Us
```
gvn: Since this ID was just used as a hook for JS we can drop it.

PR #205, opened 2017-03-20
MLN -> MN

### LINKED ISSUE #88: Remove Leadership from title and MLN acronym
Site title is currently Mozilla Leadership Network.
https://github.com/mozilla/network/search?utf8=%E2%9C%93&q=mln

Putting this in a future milestone to address after we land the final name. 

The rationale: We use the word Leadership internally, and inherited the acronym MLN from the mozilla learning network. We'll move away from both. Externally, we've found that the word Leadership is often ambiguous and exclusive. It's also implied that a network will be lead by leaders. MLN and other acronyms don't transmit info to audience who haven't been initiated.

gvn:
closes #88

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,9 @@
 {
-  "name": "mozilla-leadership-network",
+  "name": "network",
   "repository": "https://github.com/mozilla/network",
   "issues": "https://github.com/mozilla/network/issues",
   "version": "1.0.0",
-  "description": "Mozilla Leadership Network",
+  "description": "Mozilla Network",
   "scripts": {
     "autoprefix": "node scripts/autoprefix.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,9 +1,9 @@
 {
-  "name": "mozilla-leadership-network",
+  "name": "network",
```
alanmoo: Just "network"?

gvn: Yeah, that's the name of what would be the package if we published this project to npm. I'm making that consistent with the repo name per convention.

alanmoo: I guess...since it's not actually used it doesn't matter, but that's not a name that would be useful with a package manager. As it is I think it's a bit weird to just be a `mozilla/network` repo

Issue #173, opened 2017-03-15
Home Page/ News Page - missing byliner on titles


taisdesouzalessa:
**Issue:** the byliner is missing in the Home Page and the News page.

Example
![byliner-news](https://cloud.githubusercontent.com/assets/22157921/23960697/e95a2346-0965-11e7-8a87-aa913cc40f28.png)

The rule was that if we had the name of the author in Medium we would add it as a byliner. If it was just Mozilla we wouldn't add. This is what I got from our past conversations on that. @xmatthewx @sabrinang @KevZawacki please correct me if I am wrong here :).

hannahkane:
Note: Also applies to the News page itself.

gvn:
I think this info was missing from the spreadsheet when I pulled it.

hannahkane:
@gvn - is it still missing? Does someone need to add them now?

gvn:
I'd suggest that we add them after we hook News up to the CMS, otherwise it's just more data to migrate later and extra work now that will be replaced soon.

gvn:
Related: #206

xmatthewx:
Fine by me as long as @gideonthomas is aware of this for the News content-type model.

gideonthomas:
author has been added to the CMS for News

gvn:
@hannahkane Are you planning to add the authors? Once they're in the CMS I can add them to the front end.

hannahkane:
Sure. I will add them now.

hannahkane:
Where do I add them? I don't see a News module in the admin.

gvn:
@gideonthomas ^ ?

PR #206, opened 2017-03-20
updating News schema

gvn:
Related to #173

### DIFF
--- a/source/json/news.json
+++ b/source/json/news.json
@@ -61,7 +61,7 @@
     "topic": "Decentralization",
     "date": "2016-09-01T07:00:00.000Z",
     "link": "http://bit.ly/2bGMj65",
-    "featured": false
+    "featured": true
   },
   {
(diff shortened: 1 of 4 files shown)


gvn:
@gideonthomas – I guess I spoke too soon about this schema being final. One more field for `author`.

gvn:
Ok, also another field for `glyph`, which is an image URL.

gvn:
And...also a `featured` boolean.

gvn:
That's a good question for design I suppose. Currently there are 4, but I don't know if that's the max. @xmatthewx thoughts?

gvn:
Based on Slack: There will be 4 featured items.

gideonthomas:
@gvn, will they be uploading `glyph` through the admin interface or is it just a url to an image somewhere on the web?

gvn:
@gideonthomas It should be uploaded via the CMS

PR #211, opened 2017-03-21
closes #183

### LINKED ISSUE #183: People Page - Hero sentence break
**Issue:** in some screen widths the sentence breaks in a way that leaves "One" separated from "Mission"
screen shot 2017-03-15 at 12 01 10 pm

**Fix:** we were wondering if a `&nbsp;`  would fix this problem.  I tested it in browser and the only breaking point that seems a little weird to me is this one:
screen shot 2017-03-15 at 12 11 05 pm

However I think it looks better than the "One" separated from "Mission". Curious to see what others think.

cc: @sabrinang

gvn:


source/pug/templates/master.pug
```diff
@@ -1,7 +1,7 @@
 mixin subpageHero(header, subhead)
   .row.py-5.d-flex.align-items-center
     .col-md-6.pr-5
```
alanmoo: Make this a `.col-md-7.col-lg-6`

source/pug/templates/master.pug
```diff
@@ -1,7 +1,7 @@
...
-      h1.h1-white=header
+      h1.h1-white !{header}
     .col-md-6
```
alanmoo: Make this a `.col-md-5.col-lg-6`

PR #228, opened 2017-03-27
removing `topic` from news schema

gvn:


gideonthomas:
you probably want to remove it from the news.json file as well

gvn:
I'm working on a PR that deletes that file and uses the CMS json.

PR #232, opened 2017-03-27
Pulling News from Django

### LINKED ISSUE #231: Connect News to Django data

gvn:
Closes #231


gvn:
@gideonthomas Is there some reason dates aren't coming back from Django in ISO format? I'm seeing dates like: `2017-12-13` in News JSON.

scripts/fetch-json.js
```diff
@@ -6,24 +6,18 @@ let chalk = require(`chalk`);
...
-request(`https://network-pulse-api-production.herokuapp.com/entries/?ordering=-created&page_size=996&format=json&featured=True&page=1`, (error, response, body) => {
-  if (!error && response.statusCode === 200) {
-    console.log(chalk.green(`Pulled Pulse JSON`));
-    shell.ShellString(body).to(`source/json/temp/pulse.json`);
-  } else {
-    console.log(chalk.red(`Failed to pull Pulse JSON!`));
-    shell.exit(1);
-  }
-});
+let fetch = (shortName, source) => {
```
alanmoo: Can you name this something else so it isn't confusing in the future? (https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API)

gvn: Do you mean the function name?

gvn:
@gideonthomas – I had to add another field to the schema: `excerpt`, which is just a string.

gvn:
Related / Follow-on: #236

Issue #164, opened 2017-03-14
Opportunity Landing Pages: CTA should have a form


taisdesouzalessa:
**Issue:**
Our Opportunity landing Pages on staging don't have a form embedded on the CTA area:

<img width="556" alt="screen shot 2017-03-14 at 3 43 56 pm" src="https://cloud.githubusercontent.com/assets/22157921/23925581/1b72162a-08cd-11e7-9105-743a91e773f3.png">

Here are the links to the pages on staging:
- https://network.mofostaging.net/opportunity/fellowships/
- https://network.mofostaging.net/opportunity/training/
- https://network.mofostaging.net/opportunity/internet-health-report/

**Fix**
Those pages should have a form like in the mockups:
- [Fellowships](https://invis.io/ZJ9YTAGEA#/222644514_Landing-Page-Fellowships)
- [Training](https://invis.io/ZJ9YTAGEA#/222644515_Landing-Page-Leadership-Training)
- [Internet Health Report](https://invis.io/ZJ9YTAGEA#/216715361_Landing_Page_Internet_Health)

**Reason:** we want users to have quick access to the forms from these pages. Our goal is to add as little friction as possible to sign up.

This ticket is mentioned in a bigger one related to forms #156

xmatthewx:
@hannahkane – now that we have stable Sign-up (and a focus on sign up, not join) does this change any Landing Page form copy?

alanmoo:
Are all 3 of these forms adding people to the same mailing list?

xmatthewx:
Yes. It's a single list. The CRM should capture the form's origin page. It would also be ideal to capture an ID for category/interest/campaign, but I'm not sure if that's in the plans now or later (cc @arlissc).

Important to note: We plan to have staff create these simple landing pages via the CMS. Authors should have a toggle to include the form or not, and then a couple fields to add the form header and copy. I believe Gavin had some thoughts on how to compose the component so we could inject these strings. 

Make sense? Any questions?

alanmoo:
I spoke to Scott earlier and the ID isn't something that's baked in to Basket right now, so we're going to have to rely on the URL of the submission page.

The rest makes sense, I'm going to build a component that accepts arguments for the text, and when we integrate a CMS we can utilize that.

alanmoo:
@gvn I'd like to get your thoughts on implementing this today based on the way pages are currently composed.

gvn:
[...]

PR #250, opened 2017-03-29
Componentize form across different formats

alanmoo:
This addresses #164.

I know it probably needs a little design adjustment but I can't look at this alone anymore. The logic to meet all the [design variations](https://projects.invisionapp.com/share/ZJ9YTAGEA#/screens/226344662_Join-Us) was intense

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -77,21 +77,21 @@ export default class JoinUs extends React.Component {
                 : <p>{this.props.thankYouMessage}</p>
               }
             </div>
-          { !this.state.signupSuccess ?
+          { !this.state.signupSuccess &&
           <div className="col join-form">
             <form onSubmit={this.submitForm}>
               <div className={inputGroupClass}>
                 <div className="mb-2">
                   <input type="email" className="form-control" placeholder="EMAIL ADDRESS" ref="email"/>
                 </div>
-                {this.state.userSubmitted && !this.refs.email.value ? <small className="form-check form-control-feedback">Please enter your email</small> : null }
-                {this.state.signupFailed? <small className="form-check form-control-feedback">Something went wrong. Please check your email address and try again</small> : null }
+                {this.state.userSubmitted && !this.refs.email.value && <small className="form-check form-control-feedback">Please enter your email</small>}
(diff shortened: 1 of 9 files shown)


source/js/components/join/join.jsx
```diff
@@ -64,23 +51,34 @@ export default class JoinUs extends React.Component {
...
+              </div>
+              <div className="join-heading">
+                <h2 className="h3-black">{!this.state.signupSuccess ? `${this.props.ctaHeader}` : `Thank You`}</h2>
+              </div>
+              {!this.state.signupSuccess ?
+                <p className="body-black" dangerouslySetInnerHTML={{__html:this.props.ctaDescription}}></p>
+                : <p>{this.props.thankYouMessage}</p>
+              }
+            </div>
+          { !this.state.signupSuccess ?
```
gvn: I find these massive ternary blocks difficult to read/parse. Can you use [this method](https://facebook.github.io/react/docs/conditional-rendering.html#inline-if-with-logical--operator) instead?

source/js/components/join/join.jsx
```diff
@@ -5,32 +5,19 @@ import basketSignup from '../../basket-signup.js';
 export default class JoinUs extends React.Component {
   constructor(props) {
     super(props);
+    // Default props defined at end of file
```
gvn: A little odd, but ok.

source/js/components/join/join.scss
```diff
@@ -3,7 +3,121 @@
...
+        max-width: 5em;
+      }
+    }
+  }
+
+  &.full-width {
+    background: $off-white;
+
+    .join-graphic {
+      @include make-col(12);
```
gvn: Are these `make-col`s alternatives to adding `col` classes inline?

source/js/components/join/join.jsx
alanmoo: I couldn't actually get this to work any other way with ES6 classes. Is there something I was missing?

source/js/components/join/join.scss
alanmoo: Yeah, they're what Bootstrap is doing under the hood. But because this component has different variations, I needed to include this in the stylesheet instead of inline.

gvn: Ah, makes sense

Issue #225, opened 2017-03-27
Fix layout on featured profiles


xmatthewx:
Currently, if a featured profile has a short quote and long list under more details, the layout of the quote looks broken ... card preserve the height of the hidden details.

alanmoo:
This is expected behavior, as the flipping animation necessitates the "sides" be equal height. The quote is positioned dynamically. @gvn and I feel like this requires a design fix (as well as specific card examples at screen sizes of where it's an issue)

xmatthewx:
We don't need the flip. We can fade. We went with full width cards so that a change in height wouldnt awkwardly bump layout. Details will almost always be longer than quote, making default state awkward.

gvn:
It's still going to be a similar issue with a fade, because the container will have a "jump" when it resizes itself for varying content heights. Making the box size based on the tallest content eliminates that "jump" and makes the transition feel less janky.

xmatthewx:
We need a different solution. We can't have awkward white space up and down the page by default. A jump is preferable to that. A transition to smoothly increase the height shouldn't be impossible. If you want design to prototype the transition, we can do that.

alanmoo:
Are you just talking about smaller screens? Things look good on desktop to me.

taisdesouzalessa:
I am not sure I understood this issue 100% so please correct me if I am wrong.

I think that we are talking about the mobile preview on staging, correct?
![people-cards-staging](https://cloud.githubusercontent.com/assets/22157921/24471623/8ea87c0a-1477-11e7-9758-aed3f72382b6.jpg)

We run into the issue of having same height from the back and front of the "cards" while designing and the way we solved that was to have the 1 column layout for bigger cards and a bigger typography for quotes on mobile (I got the example I found the most extreme here, in which the quote is very short).
![people-cards-design](https://cloud.githubusercontent.com/assets/22157921/24471774/068bcae2-1478-11e7-8126-77d943d89cd2.jpg)

If we use this model, we always align the quote vertically to the center, the spacing doesn't look weird and the front and back of the card can have the same height.

Do you think the designed solution proposed here works well @xmatthewx? Or you still feel that we should adjust the solution to avoid the spacing even more?

alanmoo:
Ah, somewhere along the line I missed the mobile design plan/saw an old one. This makes more sense and should clean it up; Let me try this out and see how it looks.

alanmoo:
@gvn I threw an estimate on this, let me know if you think it's too high/low.

[...]

PR #253, opened 2017-03-31
Improve featured person card layout

alanmoo:
Addresses #225. Not in exactly the same way as prototyped, but this feels like a really good improvement.

### DIFF
--- a/source/js/components/people/people.scss
+++ b/source/js/components/people/people.scss
@@ -32,23 +32,23 @@
   }
 
   .flippable-card {
-    transition: transform 1s ease;
+    transition: transform 1s ease, height 1s ease;
     transform-style: preserve-3d;
     position: relative;
 
     .bio-content {
-      backface-visibility: hidden;
-      width: 100%;
       transform: rotateY(0deg);
     }
 
     .quote-content {
       transform: rotateY(180deg);
+    }
+
+    .bio-content,
+    .quote-content {
       backface-visibility: hidden;
-      position: absolute;
-      top: 0;
-      height: 100%;
       width: 100%;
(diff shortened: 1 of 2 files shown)


source/js/components/people/people.scss
```diff
@@ -32,23 +32,23 @@
   }
 
   .flippable-card {
-    transition: transform 1s ease;
+    transition: transform 1s ease, height 1s ease;
```
gvn: smart

source/js/components/people/person.jsx
```diff
@@ -13,9 +13,26 @@ export default class Person extends React.Component {
...
+      this.setState({
+        'flippableStyle': {
+          height: this.refs.bioContent.clientHeight
+        }});
+    } else {
+      this.setState({
+        'flippableStyle': {
+          height: this.refs.quoteContent.clientHeight
+        }});
+    }
```
gvn: I think you can condense to just:

```
this.setState({
  flippableStyle: {
    height: this.state.flipped ? this.refs.bioContent.clientHeight : this.refs.quoteContent.clientHeight
  }
});
```

source/js/components/people/person.jsx
```diff
@@ -13,9 +13,26 @@ export default class Person extends React.Component {
...
     this.setState({'flipped': !this.state.flipped});
   }
 
+  componentDidMount() {
+    this.refs.quoteContent && this.setState({'flippableStyle': {
```
gvn: I wouldn't use this pattern outside JSX as it's slightly hacky despite their endorsement in that context.

Just use a standard `if` here.

PR #254, opened 2017-03-31
Moving partnership logo to featured card bio

### LINKED ISSUE #225: Fix layout on featured profiles

alanmoo:
Closes #225


source/js/components/people/person.jsx
```diff
@@ -50,7 +50,7 @@ export default class Person extends React.Component {
     }
 
     let socialLinks = Object.keys(filteredLinks).map((linkKey, index) => {
-      let classes = linkKey + ` mr-3`;
+      let classes = linkKey + ` mr-4`;
```
gvn: You shouldn't need to concatenate strings like this anymore:

Use this instead: `${linkKey} mr-4`

PR #257, opened 2017-04-04
adjusting type sizes

### LINKED ISSUE #169: Mobile Typography
**Issue:** The mobile typography [on staging](https://network.mofostaging.net/) is not following the mobile specs in [the mockup](https://invis.io/ZKAIKN1MN#/220168109_MOBILE-HomePage) - sizes, line height and spaces. 

NOTE: I am almost sure the mobile on staging is still a WIP so please let me know if it is too soon to file this ticket :).

gvn:
closes #169

### DIFF
--- a/source/sass/type.scss
+++ b/source/sass/type.scss
@@ -18,12 +18,13 @@ a {
   color: #3bb7ef;
 }
 
-@mixin scaleText($small) {
+@mixin scaleText($small, $large: null) {
   font-size: $small;
 
-  @media screen and (min-width: $bp-md) {
-    // "Mobile" is 75% the size of "Desktop"
-    font-size: $small * 1.333;
+  @if ($large) {
+    @media screen and (min-width: $bp-md) {
+      font-size: $large;
+    }
   }
 }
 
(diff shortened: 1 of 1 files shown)


source/sass/type.scss
```diff
@@ -48,58 +49,58 @@ a {
...
-  @include scaleText(54px);
+  @include scaleText(56px, 72px);
 }
 
 .h3-black {
   font-family: Arvo;
   font-weight: 100;
   letter-spacing: 0;
-  line-height: 48px;
+  line-height: 1.7;
```
alanmoo: I think this should be 1.2 so it's consistent with what the big size was previously

gvn: All the sizes have been tweaked, including the larger sizes, so I think design will have to take a look at the build and give me specific values.

gvn: These are based off what's in Sketch, but I don't know if that's what's desired.

alanmoo: Can we preview this somewhere for them to review without merging this? Because this is definitely going to need changes.

gvn: Not easily. I figure we should just land it since I expect there to be tweaks anyway once design can see the system in context.

gvn:
I just did another pass and tightened up the line heights.

PR #258, opened 2017-04-05
Compress JS on build

alanmoo:


gvn:
Good call, but let's leave compression off for faster local builds (it adds a few seconds).

Here's a PR to do that: https://github.com/mozilla/network/pull/272

Note you'll need to rebase this branch from master.

PR #271, opened 2017-04-06
fixing title case on home hero tagline

xmatthewx:
Capitalized "Are" looks awkward and doesn't seem correct


hannahkane:
I believe APA, Chicago, MLA, and AP would all capitalize "Are" in title case. See http://capitalizemytitle.com/

PR #281, opened 2017-04-07
refining mobile typography

gvn:
Based on latest Sketch revisions.

### DIFF
--- a/source/sass/type.scss
+++ b/source/sass/type.scss
@@ -128,7 +128,7 @@ a {
   font-weight: 400;
   color: #231f20;
   letter-spacing: 0.25px;
-  line-height: 1.3;
+  line-height: 1.333;
 
   @include scaleText(18px, 20px);
 }
@@ -137,7 +137,7 @@ a {
   font-weight: 400;
   color: #909090;
   letter-spacing: 0.5px;
-  line-height: 1.33;
+  line-height: 1.333;
   text-transform: uppercase;
 
   @include scaleText(12px);
@@ -146,7 +146,7 @@ a {
 .h6-white {
   letter-spacing: 0.5px;
   text-transform: uppercase;
-  line-height: 1.33;
+  line-height: 1.333;
 
   @include scaleText(12px);
   @include black-box-white-text;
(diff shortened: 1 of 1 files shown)


source/sass/type.scss
```diff
@@ -128,7 +128,7 @@ a {
   font-weight: 400;
   color: #231f20;
   letter-spacing: 0.25px;
-  line-height: 1.2;
+  line-height: 1.3;
```
mmmavis: It doesn't really matter for the final generated value but we use  `1.33` in all other places. For consistency let's change it to `1.33` here?

source/sass/type.scss
```diff
@@ -140,24 +140,24 @@ a {
...
-  @include scaleText(10px, 12px);
+  @include scaleText(12px);
   @include black-box-white-text;
 }
 
 @mixin p-lead {
   font-weight: 100;
   letter-spacing: 0;
-  line-height: 1.6;
+  line-height: 1.3;
```
mmmavis: `1.33`?

source/sass/type.scss
```diff
@@ -232,29 +232,29 @@ a {
...
   line-height: 1.5;
 
-  @include scaleText(14px, 16px);
+  @include scaleText(16px);
 }
 
 .link-small {
   letter-spacing: 0;
   line-height: 1.33;
   font-weight: 900;
```
mmmavis: Should it be `700` since it's `bold` in Sketch?

https://developer.mozilla.org/en/docs/Web/CSS/font-weight

source/sass/type.scss
```diff
@@ -232,29 +232,29 @@ a {
...
 
 .link-small-tags {
   letter-spacing: 0;
   line-height: 1.33;
 
-  @include scaleText(10px, 12px);
+  @include scaleText(12px);
 }
 
 .cta-link {
```
mmmavis: https://network.mofostaging.net/style-guide/ `.cta-link` seems to be too thin from what's in the mockup (`bold`). @sabrinang @taisdesouzalessa can you confirm that we do want it `bold`?

mmmavis: ah I got clarification from @taisdesouzalessa on Slack - not sure if you want to make this change here but we'd like to make some tweaks to cta-link, see https://github.com/mozilla/network/issues/273#issue-220022590 (near the bottom) for detail

source/sass/type.scss
gvn: Doesn't really make a tangible difference, but I can do that. `¯\_(ツ)_/¯`

PR #298, opened 2017-04-12
h1 bottom spacing tweaks

### LINKED ISSUE #269: QA: Global Issue - vertical space
Description: all pages that have the highlighted title on the white background should have the same vertical space.

Example - About Page (good vertical space - showed in the rectangle area)
image

The following pages should follow the same rule as About page:

- [ ] Opportunities pages ([page 1](https://network.mofostaging.net/opportunity/internet-health-report/), [page 2](https://network.mofostaging.net/opportunity/fellowships/), [page 3](https://network.mofostaging.net/opportunity/training/))

- [ ] [Sign up Page](https://network.mofostaging.net/sign-up/)

mmmavis:
Fixes #269

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -54,6 +54,7 @@ export default class JoinUs extends React.Component {
 
     let signupState = classNames({
       'row': true,
+      'py-3': true,
       'signup-success': this.state.signupSuccess && this.state.userSubmitted,
       'signup-fail': !this.state.signupSuccess && this.state.userSubmitted
     });
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -54,7 +54,6 @@ export default class JoinUs extends React.Component {
 
     let signupState = classNames({
       'row': true,
-      'py-3': true,
       'signup-success': this.state.signupSuccess && this.state.userSubmitted,
       'signup-fail': !this.state.signupSuccess && this.state.userSubmitted
     });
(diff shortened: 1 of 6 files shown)


source/pug/views/fellowships.pug
```diff
@@ -7,7 +7,7 @@ block centerNav
 
 block content
   .container
-    .row.py-5
+    .row.pt-5
```
alanmoo: Can we consistently use margins across these? This way we can take advantage of their collapsing nature and not worry about special markup for each case.

source/js/components/join/join.jsx
```diff
@@ -67,7 +66,7 @@ export default class JoinUs extends React.Component {
         <div className="col">
           <div className="row">
             <div className="col-12 join-content">
-              <div className="my-3 join-page-title">
+              <div className="mb-5 join-page-title">
```
alanmoo: I think there's a way to make this `my-5` and remove the `.my-5` from `sign-up.pug`. Can you try that out?

source/js/components/join/join.jsx
```diff
@@ -54,7 +54,6 @@ export default class JoinUs extends React.Component {
 
     let signupState = classNames({
       'row': true,
-      'py-3': true,
```
alanmoo: I think this is what removes padding from the "Join Us" component on opportunity pages, which is a regression. Keep in mind that component has 3 different versions.

mmmavis: @alanmoo what are the 3 versions? You meant form, submission failed, and submission succeed?

alanmoo: Full page, full width (12 column) and half width (6 column). There's a set of styles for each in the scss file. It may be easiest to move this padding into the appropriate selectors in that file.

source/js/components/join/join.scss
```diff
@@ -78,6 +78,9 @@
   }
 
   &.full-page {
+    margin-top: 2rem;
+    margin-bottom: 3rem;
+
```
mmmavis: Had to add custom css here since bootstrap doesn't have a class for margin `2rem`.  I can turn `margin-bottom: 3rem` into using `.mb-5` on the element if preferred.

alanmoo: This is fine; The utility classes get confusing the way these components are intertwined

PR #304, opened 2017-04-12
Add spacing and change some fonts

### LINKED ISSUE #266: QA: People Page - IMPORTANT
- [x] Add quote behind characters and quote seems to be cut on top (see red ellipse) **IMPORTANT**
quote

- [x] Increase space between name and role (quote side). Maybe add the class "mt-2"?
screen shot 2017-04-05 at 5 47 43 pm

- [x] Small gray: line height 1.5  - _I know it affects all the pages too and all it seems fine to me. Let me know if you find some weird case if that change happens._
screen shot 2017-04-05 at 5 10 21 pm
 
- [x] Add a bit more space between top or each card and first line of text (see red rectangles). It seems too tight.
cards

- [x]  in the quotes, let's use “Nunito Sans Light Italic” (instead of using “font-style:italic”)
italics

- [x] Add grey slant

[svg here](https://drive.google.com/open?id=0B3b1KWRwjxtoNWVYbXowajZ3eTQ)
screen shot 2017-04-05 at 5 26 34 pm

- [x] Reduce size of "more details" button (we don't want it to compete with main CTA). Changing the font-size and padding seems to do the trick.

screen shot 2017-04-05 at 4 55 27 pm

alanmoo:
Closes #266


source/sass/utilities.scss
```diff
@@ -10,6 +10,10 @@
   background-image: url(../_images/slanting-lines-pttn.svg);
   background-size: 150px;
   width: 100%;
+
+  &.gray {
```
gvn: Can we make this `slanty-background-gray` since it's a variant?

alanmoo: So the element uses `class= "slanty-background slanty-background-gray"`?

alanmoo:
Double check this is what you wanted?

gvn:
Yep LGTM

PR #312, opened 2017-04-13
Gh 265

gvn:


alanmoo:
Can you put the issue number in the comment so it's linkable? It really reduces friction to figuring out what's being solved (since it's not always in the commit title)

PR #319, opened 2017-04-17
Replace text value with variable - changed text color to black

### LINKED ISSUE #267: QA: Global Issue - text color
This issue affects many pages:

- [x] Copy is rendering dark gray in some pages
gray-copy
It seems the solution for this is in the css of body (currently is set to have a different color than black):
screen shot 2017-04-05 at 6 22 08 pm

- [ ] Other examples of gray copy (those ones are very close to black but not #000000)
image
image

taisdesouzalessa:
Fixes #267

### DIFF
--- a/source/sass/type.scss
+++ b/source/sass/type.scss
@@ -112,7 +112,7 @@ a {
 .h4-light-black {
   font-family: Arvo;
   font-weight: 100;
-  color: #231f20;
+  color: $black;
   letter-spacing: 0;
   line-height: 1.2;
 
@@ -121,7 +121,7 @@ a {
 
 .h4-headingcontract-black {
   font-weight: 400;
-  color: #231f20;
+  color: $black;
   letter-spacing: 0;
   line-height: 1.2;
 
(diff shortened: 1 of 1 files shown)


taisdesouzalessa:
cc: @mmmavis here are the latest changes

mmmavis:
Hey @taisdesouzalessa, woohooo another PR! 🍻 

I think we don't have to override the text color again since we already set `h1, h2, h3, h4, h5, h6` to be `$black`. See https://github.com/mozilla/network/blob/master/source/sass/type.scss#L1-L7

screen shot 2017-04-17 at 3 25 35 pm

Notice in the dev tool `color: #000` is crossed out and later being set again with the same value.

/cc @gvn in case I'm pointing @taisdesouzalessa to the wrong direction 😉

gvn:
That's right, you can probably just remove those color overrides entirely.

taisdesouzalessa:
@mmmavis I made the changes as requested. Please review :).

gvn:
If you add branches to this repo for PRs, please delete them after they're merged. 🎉

PR #324, opened 2017-04-18
Adding first visit notice

### LINKED ISSUE #288: Full page welcome on first-time visit
problem: people will visit a WIP site that requests/requires lots of engagement and feedback; notice is useful, but we need a space for more content

- full page at /welcome (not a modal)
- First time visit only
- This will include a long message Mark crafted
- Includes a couple of buttons or links to: "Preview site" (homepage) | "Help us build" (new page or gdoc) | "Name the network" (form or gdoc from 270) | "Join the network" (site's page)
- we will keep tiny notice at top of site (though probably suppressed on this page)
- Design should be simple! Like fellowship landing page. Nothing bespoke. 

Estimate is for both design and engineering.

gvn:
closes #288

### DIFF
--- a/source/js/components/takeover/takeover.jsx
+++ b/source/js/components/takeover/takeover.jsx
@@ -1,0 +1,66 @@
+import React from 'react';
+
+export default class Takeover extends React.Component {
+  show () {
+    this.setState({
+      isHidden: false
+    });
+  }
+
+  hide () {
+    this.setState({
+      isHidden: true
+    });
+  }
+
+  toggle () {
+    this.setState({
+      isHidden: !this.state.isHidden
+    });
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.hide = this.hide.bind(this);
+
+    this.state = {
+      isHidden: false
+    };
+  }
+
+  render() {
+    return (
+      <div hidden={this.state.isHidden}>
+        <header className="bar">
+          <div className="container py-3">
+            <img src="/_images/mozilla-on-black.svg" width="100"/>
+          </div>
+        </header>
+        <div className="container">
+          <div className="row">
+            <div className="col-9 py-5">
+              <div className="my-5">
+                <h1 className="h1-white">This is a Barn Raising</h1>
+              </div>
+
+              <p>There is a movement to keep the Internet healthy taking root around the world. Mozilla is a part of this movement – and wants to help it grow.</p>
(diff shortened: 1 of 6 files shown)


source/js/components/takeover/takeover.jsx
```diff
@@ -0,0 +1,68 @@
+import React from 'react';
+
+export default class Takeover extends React.Component {
+  show () {
```
alanmoo: Does this need to exist?

source/js/components/takeover/takeover.jsx
```diff
@@ -0,0 +1,68 @@
...
+  toggle () {
+    this.setState({
+      isHidden: !this.state.isHidden
+    });
+  }
+
+  constructor(props) {
+    super(props);
+
+    this.hide = this.hide.bind(this);
```
alanmoo: Can we get rid of this by making the original function definition
`hide = () => {`?

Because ES7/React voodoo?

source/js/components/takeover/takeover.jsx
```diff
@@ -0,0 +1,68 @@
...
+
+  hide () {
+    window.scrollTo(0, 0);
+
+    this.setState({
+      isHidden: true
+    });
+  }
+
+  toggle () {
```
alanmoo: Does this get used?

gvn: No, I can remove it.

source/js/components/takeover/takeover.jsx
gvn: To do that I believe we have to refactor our build system. If we do that then we can remove all the bindings as part of that...

source/js/components/takeover/takeover.jsx
gvn: Not really...I usually just include show and hide for this sort of component.

source/js/components/takeover/takeover.jsx
alanmoo: Ah, then not worth it; I thought it just worked automagically.

source/js/components/takeover/takeover.jsx
alanmoo: Usually makes sense, but easy enough to add back if we really need it.

source/js/components/takeover/takeover.jsx
gvn: I think we should probably do it, just in another ticket.

gvn:
@alanmoo – Made those tweaks...

PR #331, opened 2017-04-20
A header join button that really pops

### LINKED ISSUE #289: Hero sign-up button
- [x] Change Sign-up button in Hero to "Get Involved". 
 - [ ] Choose a new color for the button (maybe Orange?)
 - [ ] implement copy and color change
 
**Time boxing for design: 1 hr** (color only, quick test on other pages for conflict)

alanmoo:
Closes #289

### DIFF
--- a/source/sass/buttons.scss
+++ b/source/sass/buttons.scss
@@ -6,6 +6,29 @@
   &:hover {
     cursor: pointer;
   }
+
+  &.btn-pop {
+    background: #ffed00;
+    color: black;
+    box-shadow: 4px 4px black;
+    transition: all 0.25s ease;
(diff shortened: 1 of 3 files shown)


source/sass/buttons.scss
```diff
@@ -6,6 +6,29 @@
   &:hover {
     cursor: pointer;
   }
+
+  &.btn-pop {
```
gvn: Is there a reason for nesting this under `.btn` when the others aren't?

alanmoo: Not really. Fixed (but had to add `color:black` in the hover state

PR #337, opened 2017-04-24
indentation updated

vschs007:
indentation updated


gvn:
This isn't necessary, but thanks for your contribution. We should have tickets ready for contributors in the near future if you want to check back later. Cheers!

PR #339, opened 2017-04-24
measuring target margin for takeover – closes #338

### LINKED ISSUE #338: Bottom text cut off on Welcome page
Welcome page includes buttons in a container fixed to the bottom of the view. On mobile, more padding is need to avoid blocking some content. 

screen shot 2017-04-21 at 2 46 36 pm

gvn:
measuring target margin for takeover – closes #338

### DIFF
--- a/source/js/components/takeover/takeover.jsx
+++ b/source/js/components/takeover/takeover.jsx
@@ -31,7 +31,7 @@ export default class Takeover extends React.Component {
         </header>
         <div className="container">
           <div className="row">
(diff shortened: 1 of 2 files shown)


gvn:
Yeah, it's technically less supported, but we're using it already so sure.

alanmoo:
My thoughts exactly. Plus it degrades to a totally fine experience.

gvn:
Ok, sticky it is. R?

gvn:
Margin reduced...

PR #345, opened 2017-04-25
Adding image compression tasks

### LINKED ISSUE #292: Automate image optimization
It'll be good. Basically this will involve installing svgo as devDependency and setting up a `"build:svgo"` npm script that runs svgo on the `source/images` directory with in-place optimization.

gvn:
closes #292


package.json
```diff
@@ -14,6 +14,10 @@
     "build": "shx rm -rf dest && npm run fetch && run-p build:images build:js build:pug build:sass",
     "build-uncompressed": "shx rm -rf dest && npm run fetch && run-p build:images build:js-uncompressed build:pug build:sass",
     "fetch": "node scripts/fetch-json.js",
+    "optimize:svg": "find source/images -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle",
```
alanmoo: Is this cross platform?

gvn: Anywhere UNIX commands work...

alanmoo: But not Windows? @Pomax?

alanmoo: Ah since this command isn't essential to running the codebase, let's merge it in for now. If Windows support is required, we can refactor; This is already an improvement on what we had.

gvn: Yeah, we can file a follow-on to refactor as a Node script, but as-is this will only be run maybe once a month or so by Alan or myself.

PR #350, opened 2017-04-26
Allow for a configurable network api and network pulse domain when fe…

cadecairos:
…tching JSON


gvn:
Since there's another env methodology being used we should consolidate to just use this one.

PR #371, opened 2017-05-01
projects page revisions – closes #340

### LINKED ISSUE #340: QA - Projects Page
Summary by @xmatthewx: 
 - [ ] Project thumb and title should link to entry in Pulse (1)
 - [ ] Fix heading hierarchy as described in (2) below
 - [ ] Add "Discover much more..." as in mockup under (3) below. No need to add sign-up form.
 - [ ] Add space on mobile between each category and its 1st project (4)

In addition: 
 - [ ] Change categories to be our 5 issues. (drop games and mozfest). Tais and Hannah will provide copy. 

------- 
Original comment: 

Since the last changes on this page's structure we didn't have a QA. This ticket is to cover that.

- [ ] 1.  In the "Projects" page, the right side projects are not clickable. For instance, users can't click on "The Internet Health Report" or "The Glass Room". We have to add the links for all of them - could we add the links in both image and title? If not possible, in the title is the preference. The link should go to individual cards in Pulse. For instance, "The Internet health Report" should go to here: https://www.mozillapulse.org/entry/12. "The Glass Room" to here: https://www.mozillapulse.org/entry/15
image

- [ ] 2. Currently the projects for each section have the title with size h4 and we should have it as h5. It is important because the hierarchy is broken on mobile of we don't have it as h5

**STAGED VERSION:**
image

**DESIRED STATE:**
The solution is to implement the same classes as we have in the home page (People box) - "h5-black" and "my-2"
image

- [ ] 3. The staged solution doesn't have the closing text ("Discover much more...") 
image

- [ ] 4. MOBILE: Increase space between category links and the following image.
image

adding the class "mb-5" seems to work
image

gvn:
projects page revisions – closes #340


gvn:
Created follow-on for de-duping: #372

source/pug/views/projects.pug
```diff
@@ -13,11 +13,14 @@ block heroGuts
...
-    +projects("Games", "Building a healthy Internet isn't just about white papers. These games teach privacy best-practices, HTML, CSS and other digital skills.", "https://www.mozillapulse.org/search?keyword=game", data.pulseGames.results)
+    +projects("Digital Inclusion", "Projects, articles and tools that promote an open and inclusive internet.", "https://www.mozillapulse.org/issues/digital-inclusion", data.pulseInclusion.results)
+    +projects("Decentralization", "Projects that help ensure that the internet is controlled by many, rather than just a handful of global technology giants.", "https://www.mozillapulse.org/issues/decentralization", data.pulseDecentralization.results)
+    +projects("Privacy & Security", "Tools, tips, apps and more for staying safe online.", "https://www.mozillapulse.org/issues/online-privacy-and-security", data.pulsePrivacy.results)
+    +projects("Web Literacy", "Projects that empower people to shape their own internet experience.", "https://www.mozillapulse.org/issues/web-literacy", data.pulseLiteracy.results)
 
-    .row.pt-3.pb-5
-      .col.text-center
+    .row.my-5.pt-3.pb-5.d-flex.align-items-center
```
alanmoo: `.d-flex` is redundant, `.row` means it's flexed already

source/pug/views/projects.pug
```diff
@@ -13,11 +13,14 @@ block heroGuts
...
+      .col-6
```
alanmoo: This breaks at 320px wide. `.col-12.col-sm-6`? Same for below

source/pug/partials/pulse-projects.pug
```diff
@@ -12,8 +11,9 @@ mixin projects(name, description, url, projectData)
 
 mixin project(meta)
   .col-sm-12.col-md-6.mb-4.mb-md-0
-    img.project-image(src=meta.thumbnail)
-    h4.medium-black.my-2=meta.title
+    a(href=`https://www.mozillapulse.org/entry/${meta.id}`)
```
alanmoo: Pulse front end is hard coded here but the API is a variable. I can see that causing trouble. Can we set this as a variable in the same spot the others are declared, at the least?

gvn: Is there a reason you'd want to go to another instance of the Pulse site?

gvn: And are you suggesting we make the Pulse domain an environmental variable?

alanmoo: Only because the API is an env variable already. I suppose you could also remove that from being an env var and just point at production, then this would be fine. But leaving the API as an env var opens the door for a disconnect between the link and the entry you're looking at.

gvn:
@alanmoo R?

gvn:
I just pushed a change that will make this match Pulse (Option 3 as described in #372). I feel like this is at least our best interim solution.

mozilla network projects 20170503

PR #382, opened 2017-05-03
Changed line height of "@mixin black-box-white-text". 

taisdesouzalessa:
This change affected ".h1-white" and ".h1-home-white".

You know it worked if you see the gap between the 2 black boxes (in all headings) like in this example below:
image


taisdesouzalessa:
@gvn this is ready for review. Let me know if there is any change required :).

PR #392, opened 2017-05-04
home hero tweaks - closes #365

### LINKED ISSUE #365: Reduce font-weight and size of home page title
**Issue:** Secondary pages have light font-weight (nice!). Homepage has bold, which is too heavy. It also jumps up to 64px on wide screens, which feels too big.

screen shot 2017-04-27 at 1 25 07 pm

Matthew's comment: 

> Can we just make the large heading on all pages the same? Simplify.

gvn:
home hero tweaks - closes #365

### DIFF
--- a/source/pug/views/home.pug
+++ b/source/pug/views/home.pug
@@ -34,7 +34,7 @@ mixin newsItem(data, featured=false)
 
 block heroGuts
   .row
-    .col.p-4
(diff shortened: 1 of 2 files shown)


alanmoo:
I can't follow the original ticket to tell if it meets the needs, but technically this is fine.

Issue #397, opened 2017-05-04
Upcoming page - bug


taisdesouzalessa:
"Mozfest" is marked as "November 2017" when actually is "October 2017"

<img width="1068" alt="image" src="https://cloud.githubusercontent.com/assets/22157921/25729272/2ee1985e-30e9-11e7-8a3e-75188a843254.png">

PR #398, opened 2017-05-04
Update Mozfest date in Upcoming Page

taisdesouzalessa:
We should see "October 2017" now instead of "November 2017"

image

Fixes:  #397

### DIFF
--- a/source/json/upcoming.json
+++ b/source/json/upcoming.json
@@ -56,7 +56,7 @@
   },
   {
     "photo": "/_images/upcoming/mozfest.jpg",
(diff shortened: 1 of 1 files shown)


taisdesouzalessa:
@gvn @alanmoo - I manually changed the "upcoming.json" file to achieve that. However I am not sure this was the part I should be editing. Please advise if I did something wrong :).

alanmoo:
That seems to be how this section is implemented, thanks!

gvn:
Yes, Upcoming is still static JSON and hasn't been migrated to the CMS.

PR #401, opened 2017-05-05
QA - Projects Page - Pulse links styles - closes #387

### LINKED ISSUE #387: QA - Projects Page - little tweak
This ticket is related to number 2 issue from this ticket:  #340

- [x] One thing is missing: 
For the cards we have to implement the same classes as in the home page (People box) - "h5-black"

FROM:
image

TO:
image

- [x] And one thing is a little design bug: the hover state should be a black underline (like in the News item in the home page). Instead, he have it blue on this page:
image

taisdesouzalessa:
Projects Page - Changed the class of the titles of Pulse cards  and changed underline on hover from blue to black.

This is how should look like on hover:

image

### DIFF
--- a/source/sass/views/projects.scss
+++ b/source/sass/views/projects.scss
@@ -9,3 +9,8 @@
 #view-projects .btn {
   white-space: initial;
 }
+
+a.pulse-project
+ {
+	color: $black ;
+}
--- a/source/sass/views/projects.scss
+++ b/source/sass/views/projects.scss
@@ -10,7 +10,6 @@
   white-space: initial;
 }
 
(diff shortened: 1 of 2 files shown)


source/sass/views/projects.scss
```diff
@@ -9,3 +9,7 @@
 #view-projects .btn {
   white-space: initial;
 }
+
+a.pulse-project{
+  color: $black ;
```
alanmoo: This is one of those things where it feels like there should be a way to do this without having to write new CSS but I can't think of what that solution would be. So uh, carry on.

PR #433, opened 2017-05-16
Adding CSP

### LINKED ISSUE #390: Add CSP Meta tag
@cadecairos Any more details required?

gvn:
closes #390

### DIFF
--- a/defaults.env
+++ b/defaults.env
@@ -3,4 +3,4 @@ PULSE_DOMAIN=www.mozillapulse.org
 NETWORK_API_DOMAIN=network.mofoprod.net
 VIRTUAL_ROOT=/
 TARGET_DOMAIN=network.mofostaging.net
-CSP=default-src 'self'; script-src 'self' cdn.optimizely.com; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' *.mozilla.org; media-src 'self'; object-src 'self'; child-src 'self'; frame-src 'self';
+CSP=default-src 'self'; script-src 'self' 'sha256-iN7wpJdxHlpujRppkOA8N0+Mzp0ZqZr3lCtxM00Y63c=' cdn.optimizely.com; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' ws: *.mozilla.org; media-src 'self'; object-src 'self'; frame-src 'self';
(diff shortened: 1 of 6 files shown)


defaults.env
```diff
@@ -3,3 +3,4 @@ PULSE_DOMAIN=www.mozillapulse.org
 NETWORK_API_DOMAIN=network.mofoprod.net
 VIRTUAL_ROOT=/
 TARGET_DOMAIN=network.mofostaging.net
+CSP=default-src 'self'; script-src 'self' cdn.optimizely.com; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' *.mozilla.org; media-src 'self'; object-src 'self'; child-src 'self'; frame-src 'self';
```
cadecairos: instead of `frame-src` (deprecated), lets use `child-src`

gvn: I'm actually seeing `child-src` as deprecated after [running this policy through a validator](https://cspvalidator.org/#headerValue%5B%5D=default-src+'self'%3B+script-src+'self'+cdn.optimizely.com%3B+style-src+'self'+code.cdn.mozilla.net+fonts.googleapis.com%3B+img-src+*+data%3A%3B+font-src+'self'+fonts.gstatic.com+fonts.googleapis.com+code.cdn.mozilla.net%3B+connect-src+'self'+*.mozilla.org%3B+media-src+'self'%3B+object-src+'self'%3B+child-src+'self'%3B+frame-src+'self'%3B&strategy=intersection).

gvn: How about [this](https://cspvalidator.org/#headerValue%5B%5D=default-src+'self'%3B+script-src+'self'+cdn.optimizely.com%3B+style-src+'self'+code.cdn.mozilla.net+fonts.googleapis.com%3B+img-src+*+data%3A%3B+font-src+'self'+fonts.gstatic.com+fonts.googleapis.com+code.cdn.mozilla.net%3B+connect-src+'self'+*.mozilla.org%3B+media-src+'self'%3B+object-src+'self'%3B+frame-src+'self'%3B&strategy=intersection)?

cadecairos: if we add `'sha256-vcOkZrGWRfmgv+69QFqtI4SpPCnNHq6JIzLcQvzg5/8='` to `script-src` we will not have to disable live reload for development. that hash is tha SHA256 fingerprint of the live reload script.

cadecairos: Haha you're right about frame-src - it was deprecated in v2, undeprecated in v3 draft - but Firefox still complains that it's deprecated!

alanmoo:
Looks like the CSP isn't getting picked up- is this working right? https://observatory.mozilla.org/analyze.html?host=network.mofostaging.net

gvn:
My guess is that tool isn't pulling the meta tag, but just looking for HTTP headers.

cadecairos:
yup, you're right - it isn't aware of the meta tag.

We can just add 20 points in out minds :)

gvn:
+25 points! 🤣

PR #442, opened 2017-05-19
escaping data attributes in opportunity template

### LINKED ISSUE #351: Links in Join component do not work
When attempting to add a link to the description of a signup component (https://network.mofoprod.net/admin/landingpage/signup/1/change/), there seems to be some issues with encoding of quotation marks that causes the string to only partially render (the link and everything after it disappear).

Since the component is being generated by a data attribute that react then uses, I believe the problem is in the front end code. Perhaps something needs to be done when passing the props to react.

gvn:
closes #351


alanmoo:
Needs to be tested by creating a signup component that includes a link in the body/content/description (I can't remember) field of the component on Mezzanine, and seeing if that then gets react-ified properly.

Issue #260, opened 2017-04-05
QA: Opportunities Pages


taisdesouzalessa:
**------TO DO LIST - UPDATED - TO ENGINEERS------**

**[Internet Health](https://network.mofostaging.net/opportunity/internet-health-report/)**

- [x]  Link “your feedback” in the grey box. Link url: https://internethealthreport.org/v01/your-feedback/ **IMPORTANT**

- [x] Link "the report" in the grey box. Link: https://internethealthreport.org/v01/ 

**[Leadership Training](https://network.mofostaging.net/opportunity/training/)**
- [x]  Include start your training button and link to https://mozilla.teachable.com/p/open-leadership-101
<img width="327" alt="screen shot 2017-04-05 at 10 57 40 am" src="https://cloud.githubusercontent.com/assets/22157921/24719363/515f8380-19ee-11e7-8ee6-9d7f86e6b592.png">

- [x] Change class of the font of the first paragraph ("The skills you need...") 
<img width="796" alt="image" src="https://cloud.githubusercontent.com/assets/22157921/25765491/333e8290-31a2-11e7-9703-978b81701f11.png">

**[Fellowships](https://network.mofostaging.net/opportunity/fellowships/)**
- [x]   Link “open data” and “open web” **IMPORTANT**
Open data link: https://science.mozilla.org/programs/fellowships
Open web link: https://advocacy.mozilla.org/en-US/open-web-fellows/fellows2016
<img width="500" alt="screen shot 2017-04-05 at 10 44 14 am" src="https://cloud.githubusercontent.com/assets/22157921/24718860/6de5d57e-19ec-11e7-9034-5b5f2387f77c.png">

**All opportunity pages:**

- [x] Reduce space between Title and first paragraph (removing class "mt-5") of the paragraph works. _NEW REQUEST_
<img width="662" alt="image" src="https://cloud.githubusercontent.com/assets/22157921/25590794/b40dc1cc-2e66-11e7-9075-35dc82455b3c.png">

- [x] Add more padding top/bottom to the grey box (change class from "py-md-4" to " py-md-5":
<img width="1061" alt="image" src="https://cloud.githubusercontent.com/assets/22157921/25591040/d79e7734-2e67-11e7-9ea4-c79ca8c6b7ee.png">

- [x] MOBILE: Add more padding top and bottom in the grey box. Changing class from "py-3" to "py-5" seems to work. 
[...]

PR #454, opened 2017-05-22
Landing page template tweaks

gvn:
Related to #260

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -53,7 +53,7 @@ export default class JoinUs extends React.Component {
     });
 
     let signupState = classNames({
-      'row py-3 py-md-4 px-md-2': true,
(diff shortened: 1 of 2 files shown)


gvn:
Solves a few more of the items in #260. 

Finishing that ticket is blocked by promoting to API to prod.

source/js/components/join/join.jsx
```diff
@@ -53,7 +53,7 @@ export default class JoinUs extends React.Component {
     });
 
     let signupState = classNames({
-      'row py-3 py-md-4 px-md-2': true,
+      'row py-5 py-md-5 px-md-2': true,
```
alanmoo: Assuming this change is indeed supposed to affect mobile and medium+, the `py-md-5` can just go away.

Issue #446, opened 2017-05-19
Implement Analytics Event tracking


xmatthewx:
Building off of https://github.com/mozilla/network/issues/412, we'll wire up events listed in this [spreadsheet](https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=779058050)

xmatthewx:
@gvn @alanmoo @mmmavis – should we schedule a chat with @vojtechmzl on implementing GA Events? I know we have some experience, but I'm sure there are questions worth answering. We could look at how it's implemented in Pulse currently, so we can quickly dig into specifics. cc @taisdesouzalessa

gvn:
AFAIK anything that's just a normal anchor link to another page should already be tracked by stock GA (eg: "home feature tap"). Can we confirm we actually need to wire everything on the spreadsheet manually? I wouldn't want to be logging things twice in two different ways.

@vojtechmzl can you look at the spreadsheet and give some guidance on what isn't already tracked?

I believe we'll just need to wire up buttons that trigger actions facilitated by JS (eg: menu open/close).

vojtechmzl:
Hey @gvn, you're right about the home feature tap event, if it's pointing to another page on the site, then we don't need additional tagging. I think all of the other events make sense, including the clicks on the sign up button in the header, which we should track to help us understand the full conversion funnel. Let me know if there is anything I can help with. Thanks

vojtechmzl:
For outbound links that point to other Mozilla sites, let's use the following standard set of URL params:

?utm_source=mozillaNetwork&utm_medium=referral

xmatthewx:
For Homepage "Features", only some of those go to other pages. Some go out to external news (our Medium and 3rd party), some go to our other properties. While the destinations are different, the presentation to users is similar. However we track these, let's make sure we have a way to compare Features side-by-side so we can observe what drives people to dig deeper. Make sense?

vojtechmzl:
Hey @xmatthewx, the issue with tracking outbound links is that in order to log the event, we have to prevent the link from navigating to its destination until we create the callback to GA, which is not ideal for the UX. We can still track click-throughs to our sites by adding parameters to the URLs (e.g. utm_source=mozillaNetwork, etc.), but with other sites it is a bit of an issue. We can chat about this Monday, if that would be helpful.

gvn:
[...]

PR #460, opened 2017-05-23
Add tracking for dismissing splash page

alanmoo:
A little bit of #446

### DIFF
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "mofo-bootstrap": "3.2.0",
     "react": "15.4.2",
     "react-dom": "15.4.2",
+    "react-ga": "^2.2.0",
     "react-markdown": "2.4.6"
   }
 }
--- a/package.json
+++ b/package.json
@@ -80,7 +80,7 @@
     "mofo-bootstrap": "3.2.0",
     "react": "15.4.2",
     "react-dom": "15.4.2",
-    "react-ga": "^2.2.0",
+    "react-ga": "2.2.0",
     "react-markdown": "2.4.6"
   }
 }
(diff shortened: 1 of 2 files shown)


source/js/components/takeover/takeover.jsx
```diff
@@ -8,6 +9,12 @@ export default class Takeover extends React.Component {
       isHidden: true
     });
 
+    ReactGA.event({
+      category:`splash`,
+      action: `splash exit`,
+      label: `dismiss splash `
```
xmatthewx: For label, can distinguish between the two buttons? Using button string or add a name to element?

cadecairos:
Does react-ga just plug into existing GA scripts on the page? I though it needed to be i[nitialized before use](https://www.npmjs.com/package/react-ga#reactgainitializegatrackingid-options).

alanmoo:
When I tested this it seemed to fire the event upon dismissing the splash modal, so I thought it just hooked right in.

alanmoo:
If this needs changes (I'm guessing the `logJoinBtn` function isn't going to get called, but I can't confirm) can someone take this and run with it so we can get this in today?

alanmoo:
Even if we use the anchor link tracking, we should make sure we can identify a user as coming from the button on the splash vs the button in the header

package.json
```diff
@@ -80,6 +80,7 @@
     "mofo-bootstrap": "3.2.0",
     "react": "15.4.2",
     "react-dom": "15.4.2",
+    "react-ga": "^2.2.0",
```
gvn: Lock to `2.2.0`

PR #463, opened 2017-05-23
adding cms resets - closes #400

### LINKED ISSUE #400: Images in custom pages should automatically be responsive
Because users can embed images in custom pages, these images should be [responsive](https://v4-alpha.getbootstrap.com/content/images/#responsive-images) by default. What this means is we need to add `height: auto` to all images in custom pages via CSS targeting. (I solved this temporarily on the fellows page by editing the source of the content but this isn't an appropriate long-term solution.)

gvn:
adding cms resets - closes #400


gvn:
Yeah, I think some guidance from Design would be good.

PR #464, opened 2017-05-23
fix regresssion in nav states

gvn:


source/js/components/nav/nav.scss
```diff
@@ -60,9 +60,8 @@
   @at-root {
```
alanmoo: Oh you fancy

Issue #325, opened 2017-04-19
Prep analytics for first-visit Welcome view


xmatthewx:
During member preview, we'll have a Welcome message presented as a full site takeover. There's risk of losing visitors. 

We'll want to track a few things. Here's first stab. Needs refinement. 

 - [x] Source: if these people arrive via our preview engagement emails, we should know that. Add a source attribute to those links.  
 - [x] View: We should track Welcome views as [virtual page views](https://developers.google.com/analytics/devguides/collection/analyticsjs/single-page-applications#tracking_virtual_pageviews).
 - [x] Event: Welcome view offers 2 buttons and 2 links to proceed. We should track these as events.
 - [x] Bounce: There's high risk people will bounce. We need to track that, make sure it isn't happening too often for the 2500 people engaged via email.

xmatthewx:
more info in #412

alanmoo:
The buttons on the splash page now trigger events, this is live on production.

gvn:
I'm not sure who controls the welcome email, but they'll need to append a query string. @vojtechmzl can you provide what that should be?

gvn:
"Source: if these people arrive via our preview engagement emails, we should know that. Add a source attribute to those links."

^ Is blocked by query string and then delivering that query string to whoever controls the emails (@xmatthewx – can you assign that person?).

hannahkane:
Adding @stephaniemcv who I believe has the keys to the welcome email.

xmatthewx:
I believe this is all done! Including source attribute for visits via email. Closing. Reopen if needed.

PR #480, opened 2017-06-01
track welcome view as virtual page view

gvn:
Related to #325

### DIFF
--- a/source/js/main.js
+++ b/source/js/main.js
@@ -1,6 +1,7 @@
 /* eslint no-unused-vars: ["error", { "varsIgnorePattern": "React" }] */
 
 import React from 'react';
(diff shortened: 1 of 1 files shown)


alanmoo:
This was surprisingly difficult to test, but it looked right so I merged it in. Unfortunately, it doesn't seem to be working on staging. I filed a Devops ticket to create an analytics view so that we can more easily test analytics on our local builds. Once that's landed, we should be able to confirm it works before landing another PR.

gvn:
This has been really useful for me: https://chrome.google.com/webstore/detail/google-analytics-debugger/jnkmfdileelhofjcijamephohjechhna?hl=en

PR #494, opened 2017-06-06
allow mofoprod in csp

### LINKED ISSUE #502: Opportunity pages are broken on staging
(Due to CSP)

gvn:
This will fix broken landing pages on staging.

Closes #502

### DIFF
--- a/defaults.env
+++ b/defaults.env
@@ -3,5 +3,5 @@ PULSE_DOMAIN=www.mozillapulse.org
 NETWORK_API_DOMAIN=network.mofoprod.net
 VIRTUAL_ROOT=/
 TARGET_DOMAIN=network.mofostaging.net
-CSP=default-src 'self'; script-src 'self' 'sha256-iN7wpJdxHlpujRppkOA8N0+Mzp0ZqZr3lCtxM00Y63c=' cdn.optimizely.com https://www.google-analytics.com/analytics.js; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' ws: *.mozilla.org; media-src 'self'; object-src 'self'; frame-src 'self';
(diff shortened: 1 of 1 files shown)


gvn:
Yeah, that's a good point. If we change `ASSET_DOMAIN` to point to staging I believe that will work.

I'm wondering if there are other places `ASSET_DOMAIN` is used that might be problematic if we switch to staging, though.

@Pomax @cadecairos any thoughts?

defaults.env
```diff
@@ -3,5 +3,5 @@ PULSE_DOMAIN=www.mozillapulse.org
 NETWORK_API_DOMAIN=network.mofoprod.net
 VIRTUAL_ROOT=/
 TARGET_DOMAIN=network.mofostaging.net
-CSP=default-src 'self'; script-src 'self' 'sha256-iN7wpJdxHlpujRppkOA8N0+Mzp0ZqZr3lCtxM00Y63c=' cdn.optimizely.com https://www.google-analytics.com/analytics.js; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' ws: *.mozilla.org; media-src 'self'; object-src 'self'; frame-src 'self';
+CSP=default-src 'self'; script-src 'self' 'sha256-iN7wpJdxHlpujRppkOA8N0+Mzp0ZqZr3lCtxM00Y63c=' https://www.google-analytics.com/analytics.js network.mofoprod.net; style-src 'self' code.cdn.mozilla.net fonts.googleapis.com network.mofoprod.net; img-src * data:; font-src 'self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net; connect-src 'self' ws: *.mozilla.org; media-src 'self'; object-src 'self'; frame-src 'self';
```
cadecairos: I'm confused, `'self'` should allow `network.mofoprod.net` in both scripts-src and style-src

gvn: It won't on `network.mofostaging.net`, which is where the we're seeing CSS/JS not loading.

gvn: EG: https://network.mofostaging.net/opportunity/internet-health-report/

gvn:
Alternatively, we could tweak the `CSP` value just on staging to fix this issue. That doesn't solve the problem that Alan is raising though...

defaults.env
cadecairos: Ah okay, I understand the issue now

cadecairos:
> Alternatively, we could tweak the CSP value just on staging to fix this issue.

yeah, I think this is the right way to go


> That doesn't solve the problem that Alan is raising though...

Is ASSET_DOMAIN not set to the staging domain on staging? we should probably change that.

gvn:
Ok, I tweaked the `CSP` value for staging. I believe we also need a push of `network-api` to get the new template deployed.

`ASSET_DOMAIN` appears to be prod on network.mofostaging.net.

gvn:
Closing this in favor of overriding `CSP` on staging only.

Issue #203, opened 2017-03-20
Refactor featured Get Involved home content to use external JSON from Mezzanine


gvn:
- [x] Create schema and wire up static JSON in `network`
- [ ] Add new `highlights` model to Mezzanine https://github.com/mozilla/network-api/issues/220
- [ ] Add content to Mezzanine
- [ ] Switch `network` over to use Highlights JSON pulled from `network-api`

gvn:
Blocked by pending API work.

xmatthewx:
We should probably do a new estimate on this to include a means to select featured items outside of the entries themselves.

alanmoo:
Made an epic, this is the first item in it and it is complete, so I'm closing for accurate estimation records.

PR #512, opened 2017-06-12
Wire up homepage highlights to static JSON

gvn:
This sets us up for moving the "highlights" into Mezzanine.

Once this lands we can then add this model and switch to `network-api` as the source of the JSON instead of the flat static version.

Related to #203

### DIFF
--- a/source/json/highlights.schema.json
+++ b/source/json/highlights.schema.json
@@ -1,0 +1,43 @@
+{
+  "$schema": "http://json-schema.org/draft-04/schema#",
+  "definitions": {},
+  "id": "http://example.com/example.json",
+  "items": {
+    "id": "/items",
+    "properties": {
+      "description": {
+        "id": "/items/properties/description",
+        "type": "string"
+      },
+      "footer": {
+        "id": "/items/properties/footer",
+        "type": "string"
+      },
+      "image": {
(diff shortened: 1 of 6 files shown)


source/json/highlights.schema.json
```diff
@@ -0,0 +1,46 @@
...
+  "definitions": {},
+  "id": "http://example.com/example.json",
+  "items": {
+    "id": "/items",
+    "properties": {
+      "description": {
+        "id": "/items/properties/description",
+        "type": "string"
+      },
+      "footer": {
```
gvn: This is new since the schema we discussed as a group, but will allow adding the "See more projects" button (or other bits of HTML as needed).

alanmoo: What type of field would this be in the API admin?

gvn: Text/HTML

source/json/highlights.schema.json
```diff
@@ -0,0 +1,46 @@
...
+        "id": "/items/properties/footer",
+        "type": [
+          "string",
```
alanmoo: See if there's a way to indicate this is going to be a string that's rendered as HTML. If so, adjust it. If not that's fine too, just not ideal for clarity.

PR #544, opened 2017-06-23
Hero & Navbar revisions

### LINKED ISSUE #523: Implement new hero artwork
Design in #493

### LINKED ISSUE #524: Implement new navbar design
design in #492

gvn:
closes #523 
closes #524


source/js/main.js
```diff
@@ -26,6 +26,49 @@ let main = {
...
     });
+
+    // Track window scroll position and add/remove class to change sticky header appearance
+
+    let lastKnownScrollPosition = 0;
+    let ticking = false;
+    let elBurgerWrapper = document.querySelector(`.wrapper-burger`);
+
+    let adjustNavbar = (scrollPosition) => {
+      if (scrollPosition > 100) {
```
mmmavis: Can we put a comment here explaining where the hardcoded number `100` came from?

mmmavis: Is there a way to not use hardcoded number here? If not, can we put a comment here explaining where the hardcoded number `100` came from so that number doesn't look so magical?

gvn: Yep!

PR #554, opened 2017-07-10
adding Email buttons - closes #486

### LINKED ISSUE #486: Add email address to site
- [ ] 1. Add "EMAIL" as a line item under TWITTER on the "We'd love to hear from you" section of /about. It should be a mailto: pointing to network@mozillafoundation.org
 - [ ] 2. Add "Email" as the second item in the footer, to the right of Twitter 
 - [ ] 3. Add "Email" as the second item in nav menu footer, below Twitter

gvn:


gvn:
I actually did that re-ordering on purpose because I don't think the hierarchy makes sense as designed.

I think it should be email first and then have social media secondary since email is something everyone has, but Twitter and FB are not.

PR #556, opened 2017-07-11
Lint all scss files

arush0311:
Fixes https://github.com/mozilla/network/issues/543


gvn:
@Pomax can you confirm this works on Windows?

Pomax:
@gvn I don't get any errors on Windows during the test:scss task

arush0311:
Hey! Since we have installed shx why dont we use that to standardize every script for windows too?

gvn:
`shx` is being used to shim standard UNIX commands (like `rm`, `mkdir`, etc) for better portability, but it doesn't cover custom tools like `stylelint`.

PR #559, opened 2017-07-12
Hero update: title + blurb

taisdesouzalessa:
News Page: hero was updated with the new content.

This is the result that should be seen once it is implemented:
image


taisdesouzalessa:
@gvn could you review this code please? Thanks!

PR #564, opened 2017-07-13
Change source of "highlights" from hardcoded to network-api`

### LINKED ISSUE #517: Change source of "highlights" from hardcoded to network-api

gvn:
closes #517


gvn:
Note: I think the actual CMS content needs to be tweaked to match staging:

screen shot 2017-07-13 at 2 02 04 pm

I originally made the schema require images for all highlights, but I tweaked it (and the layout) to not break without them...

alanmoo:
Oh is the content on staging what I should have moved into the CMS? I copied the content from production.

gvn:
Yeah, it's slightly more up to date.

PR #571, opened 2017-07-17
switch to "auto" for scrollbars

Pomax:
fixes https://github.com/mozilla/network/issues/386

With "scroll", the menu always has horizontal and vertical scrollbars (particularly visible on windows, or on a mac with a 3rd party mouse plugged in), even when the content fits on the screen. This change effects scrollbars only when necessary.

### DIFF
--- a/source/js/components/nav/nav.scss
+++ b/source/js/components/nav/nav.scss
@@ -4,7 +4,7 @@
   height: 100%;
   width: 100%;
   z-index: 99999;
-  overflow: scroll;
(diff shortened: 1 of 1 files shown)


alanmoo:
I stumbled across this and recalled when this was added - let's make sure it doesn't reintroduce the issue solved by https://github.com/mozilla/network/pull/295

gvn:
Doesn't seem to cause a regression:

screen shot 2017-07-18 at 10 10 09 am

PR #572, opened 2017-07-17
switch to "auto" for scrollbars

Pomax:
With "scroll", the menu *always* has horizontal and vertical scrollbars (particularly visible on windows, or on a mac with a 3rd party mouse plugged in). This change effects scrollbars only when necessary based on screen dimensions.


Pomax:
...why did this open a second PR O_O

Issue #565, opened 2017-07-14
Implement revised hierarchy of people card content


sabrinang:
Based on design issue https://github.com/mozilla/network/issues/430#issuecomment-315158519, we want to implement a revised hierarchy for people cards that includes icons:
<img width="943" alt="image" src="https://user-images.githubusercontent.com/2492510/28194231-bc6414e6-67f7-11e7-8adf-4a317c7532be.png">

.svgs zipped here:
[icons.zip](https://github.com/mozilla/network/files/1147084/icons.zip)

.pngs:

Note: This issue is also related to https://github.com/mozilla/network/issues/521 refining member roles

![map-marker-icon](https://user-images.githubusercontent.com/2492510/28194521-b13c316e-67f9-11e7-8fa0-d1a6558f85ba.png)
![mozilla-icon](https://user-images.githubusercontent.com/2492510/28194522-b13d0530-67f9-11e7-83b2-62bc41dc3211.png)

xmatthewx:
R+ good to go. This is excellent. A small thing that has such a big impact. Nice work @sabrinang @gvn.

PR #626, opened 2017-08-01
adding icons to person metadata - #565

gvn:
adding icons to person metadata - #565


source/js/components/people/people.scss
```diff
@@ -22,9 +22,35 @@
     line-height: 1;
   }
 
-  .person-role-location {
-    font-weight: 600;
-    color: black;
+  .meta-block {
+    .meta-block-item {
+      display: flex;
+      align-items: center;
```
mmmavis: Should we just use `.d-flex` and `.align-items-center` on the element instead?

gvn: We could, but then we have to add those classes to every item.

mmmavis: k makes sense

PR #631, opened 2017-08-03
Add a heroku-prebuild hook so that the git submodule is checked out b…

cadecairos:
…efore building


cadecairos:
ignore this, it doesn't work.

PR #638, opened 2017-08-07
fix sort orders for Upcoming

gvn:


gvn:
@alanmoo - I guess I need another R+ since I rebased... 🤔

PR #655, opened 2017-08-09
remove build signal for jenkins

cadecairos:
This PR removes the code that wires up signals on models, and the debounce/throttle code for triggering Jenkins builds remotely.


gvn:
We'll still need the build to re-run when models' data changes. Is that what this is taking out?

cadecairos:
right, I had forgotten this was still statically generated. I'll keep the signals and integrate the signals with Heroku's build and deploy API

gvn:
Do you want to do that in this PR?

cadecairos:
yes, I'll work on this today

cadecairos:
@gvn I have force pushed a change to this branch. It updates the `build_static_site` method to use Heroku's Build API.

gvn:
This gets a "LGTM" R+ since I'm not really able to test...

PR #684, opened 2017-08-22
Add in env-controllable CSP settings

Pomax:
This PR adds CSP configuration to the standard Django settings.py file by using the django-csp middleware in combination with environment variables for each separate content category in the CSP specification.

Right now this PR only has the bare minimum in terms of default values for development, but this can be changed to the exact set we need either before landing, or as part of fast follow-up.

This PR closes https://github.com/mozilla/network/issues/679


cadecairos:
I've added CSP setting to staging. Merging this in.

PR #697, opened 2017-08-24
refine CSP

### LINKED ISSUE #695: CSP is now overly restrictive
Looks like CSP rules weren't ported over as defined here:

https://github.com/mozilla/network/blob/master/env.default#L5

Lots of breakages:

screen shot 2017-08-24 at 1 33 28 pm

Pomax:
- updated the network-api `env.default` with CSP values that will work for us
- removed meta-based CSP from template.
- removed the CSP env value from the network `env.default` file

This fixes #695


Pomax:
added `allow-scripts` as sandbox CSP policy so our own scripts are allowed to run =)

network-api/env.default
```diff
@@ -16,3 +16,18 @@ BUILD_THROTTLE_SECONDS=900
 SOCIAL_AUTH_GOOGLE_OAUTH2_KEY=
 SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET=
 SOCIAL_AUTH_LOGIN_REDIRECT_URL=http://localhost:5000/soc/complete/google-oauth2/
+CSP_DEFAULT_SRC='self'
```
cadecairos: any lines with just the `'self'` directive will need to also be surrounded by double quotes, otherwise the env loader will strip them out

network-api/env.default
```diff
@@ -16,3 +16,18 @@ BUILD_THROTTLE_SECONDS=900
...
+CSP_IMG_SRC=* data:
+CSP_FONT_SRC='self' fonts.gstatic.com fonts.googleapis.com code.cdn.mozilla.net
+CSP_CONNECT_SRC='self' ws: *.mozilla.org
+CSP_MEDIA_SRC='self'
+CSP_OBJECT_SRC='self'
+CSP_CHILD_SRC='self'
+CSP_FRAME_SRC=
+CSP_BASE_URI=
+CSP_FRAME_ANCESTORS=
+CSP_FORM_ACTION=
```
cadecairos: This should be set to `https://www.mozilla.org/en-US/newsletter/`

cadecairos: and 'self' so admin forms work 👍

network-api/env.default
```diff
@@ -16,3 +16,18 @@ BUILD_THROTTLE_SECONDS=900
...
+CSP_SANDBOX='allow-scripts'
```
cadecairos: sandboxing breaks code in the django admin that depends on localstorage - perhaps we should not enable sandboxing.

network-api/env.default
```diff
@@ -16,3 +16,18 @@ BUILD_THROTTLE_SECONDS=900
 SOCIAL_AUTH_GOOGLE_OAUTH2_KEY=
 SOCIAL_AUTH_GOOGLE_OAUTH2_SECRET=
 SOCIAL_AUTH_LOGIN_REDIRECT_URL=http://localhost:5000/soc/complete/google-oauth2/
+CSP_DEFAULT_SRC='self'
+CSP_SCRIPT_SRC='self' 'sha256-iN7wpJdxHlpujRppkOA8N0+Mzp0ZqZr3lCtxM00Y63c=' cdn.optimizely.com https://www.google-analytics.com/analytics.js
```
cadecairos: The Django admin interface relies heavily on inline scripts and CSS, so we'll have to add that here and to `script-src`. (remove the sha source or browsers will ignore unsafe-inline)

Pomax:
if it's important to land this soon, please edit the PR directly @gvn/@cade, as I won't have time to work on this today due to Advocacy work

cadecairos:
Not sure Cade really wants to jump in on this one, but hey, feel free to if you are in the mood!

cade:
I've missed y'all! Can you imagine a world without malicious hackers and the crazy security features we've had to layer on over the years like CSPs? Yeah... me neither.

I'm about to get into some lunch, so I'll let @cadecairos bring it home!

Issue #619, opened 2017-08-01
Update Opportunity templates to support subtree navigation, if present


alanmoo:
To close this ticket, multipage navigation should be implemented to the point of [Phase 1](https://redpen.io/wh681ba02d54215838).

Any page that's a child of "Opportunities" which has its own children (and those children pages themselves) should show the side navigation.

Note that the first item in the nav list is the parent, but it's styled to be identical to the children pages.

Also, pages that have the signup component will look a little strange due to the 2 columns next to the navigation, the guidance for now is simply "don't do that" so that these mini-site pages are single column + nav.

gvn:
@arush0311 (or maybe @alanmoo ) – What are the `page` object properties to access the subpages in the opportunity template?

I'm assuming its something along the lines of `page.landingpage.SOMETHING`?

arush0311:
@gvn I think its a `page_branch` variable. http://mezzanine.jupo.org/docs/content-architecture.html#page-menus

gvn:
@taisdesouzalessa Can you point me to the Sketch file for this?

taisdesouzalessa:
Hey @gvn I had a prototype for the latest version but it is broken now (my guess is that the css was updated and I have to relink now)
Here is the prototype - please see it at 320 [mobile] and 1280 [desktop]
https://taisslessa.github.io/simple-model-ui-v4/index.html

I could either:  fix the now broken prototype (it is the most accurate representation of the final state) or make updates to the Sketch file (some little changes happened that were only reflected in the prototype). Which one is more useful to you? I could do any of those by end of the day tomorrow if it works for you.

gvn:
It's definitely most useful for me to have a Sketch file. 👍

taisdesouzalessa:
@gvn here is the link to the Sketch file: https://drive.google.com/a/mozilla.com/file/d/0B3b1KWRwjxtoa093cVU5REhxdlU/view?usp=sharing

Prototype updated (you can also see it in mobile view - >376 - to check mobile behavior): https://taisslessa.github.io/simple-model-ui-v4/index.html

gvn:
I'm trying to figure out if there's a boolean I can access outside of the `page_menu` function that will indicate whether or not a page has children or not. I need this to adjust the template layout and hide the side navigation for standalone pages.

Any ideas @arush0311 @alanmoo ?

alanmoo:
`page.has_children` in a template tag should do it, no?
[...]

PR #702, opened 2017-08-28
Enable multipage Opportunity pages

gvn:
Unfortunately there is a small hack required for this to all work:

https://github.com/mozilla/network/compare/master...gvn:gh-619?expand=1#diff-ab5ab32e890ace9a68922aaf8f73db0cR23

Related: #619


network-api/app/networkapi/templates/pages/landingpage.html
```diff
@@ -53,22 +53,27 @@
...
-            <h1 class="h1-white">{{ page.landingpage.header }}</h1>
-          </div>{% endif %}
+          </div>
         </div>
-        <div class="row justify-content-around">{% if page.landingpage.signup %}
-          <div class="cms col-md-6 col-lg-5 mb-5">{{ page.landingpage.content | safe }}</div>
-          <div class="col-md-5 col-lg-5">
-            <div class="row">
-              <div class="join-us half-width col" data-cta-header="{{ page.landingpage.signup.header | escape }}" data-cta-description="{{ page.landingpage.signup.description | escape }}"></div>
+        <div class="row mb-5">{% if page.parent.title != "Opportunity" or page.children.all.count > 0 %}
```
alanmoo: it seems like `branch_level` works inside the child template, so can the check for level/"should we show the side menu in the first place" happen inside that?

gvn: I have tried a *ton* of approaches, including that one, and there are issues with it as well. This is really the best I could come up with after many hours of hammering on it.

source/sass/type.scss
```diff
@@ -278,3 +278,17 @@ a {
 
   @include scaleText(16px);
 }
+
+.opp-link {
+  font-family: Arvo;
+  font-size: 20px;
```
alanmoo: Why the hard coded font size?

gvn: Versus what?

alanmoo: The `scaleText()` mixin used on previous lines

gvn: Oh. Yeah I can add that.

PR #709, opened 2017-08-31
Fix #657 - Create homepage model

### LINKED ISSUE #657: Create Homepage model

gideonthomas:
This just creates the homepage model. Nothing super functional yet.

Since we want the page to look like what's described in https://github.com/mozilla/network/issues/659#issue-249371338, we use FKs in the related models instead of declaring the fields explicitly in the Homepage model. The homepage model itself is going to effectively function as a singleton, however we are not going to enforce that at the model tier. Instead, Mezzanine provides a way for the admin interface to treat the homepage as a singleton, per https://mezzanine.readthedocs.io/en/latest/admin-customization.html#singleton-admin.


gideonthomas:
@alanmoo that's why the homepage property is on each model vs. having the related model as a foreign key on the homepage.

Here's the explanation for this - Suppose we add the related models as properties on the homepage:
```python
class Homepage(models.Model):
    leader1 = models.ForeignKey(Person, related_name='homepages')
    leader2 = models.ForeignKey(Person, related_name='homepages')
    leader3 = models.ForeignKey(Person, related_name='homepages')
```
this approach has two big flaws in it:
1. It becomes hard to extend the properties on the homepage to beyond (or fewer than) 3 instances of the  related model (in this case `Person`) because we effectively have to add migrations to make it possible and deal with existing data and the null case.
2. The model in which the FK is applied implies that the related model is a many relation. That means that for `leader1`, it can only point to 1 `Person` object, but that `Person` object can be related to multiple `Homepage` models - which in this case doesn't make sense since we will always only have a single homepage. So in query terms in the above case, when we do `print(homepageObj.leader1)`, we get `<Person object>`. But when we ask the person object which homepage they belong to, it would be `print(personObj.homepages)` which will always return a list viz. `<List <Homepage>>` which is incorrect if we only have a single homepage.

Now we can probably circumvent the 2nd issue if we make it a 1:1 relationship so that each property in the `Homepage` model points to only 1 `Person` instance and vice-versa. However, the first issue would still remain.

To make sure we fix both issues (so that this won't be a problem later on esp. w.r.t. migrations), we flip the 1:Many relationship of `Homepage`:`Person` so that it becomes `Many:1` represents `Person`:`Homepage` with the implication that many people belong to a single homepage and 1 homepage contains many people. The code for that would be:
```python
class Homepage(models.Model):
    pass
class Person(models.Model):
    homepage = models.ForeignKey(Homepage, related_name='leaders')
```
so in query terms, `print(personObj.homepage)` would only print out a single homepage or null (which aligns with having only 1 homepage object) while `print(homepageObj.leaders)` will print out a list of person objects viz. the leaders that are featured on the homepage. The code above is what I have in my patch.
This also solves issue 1 because now, we can extend the number of people featured on the homepage as needed whenever we want without needing to create a migration. We can restrict the number on the admin interface if we want, but the good thing about that is that it only touches the UI and not the database.

alanmoo:
That makes a ton of sense, thanks. I’ll run this code when I get back to a computer.

PR #716, opened 2017-09-05
Make homepage related models Many-to-Many

gideonthomas:
After trying to implement the admin part of this, I had to change the way we think about the relationship between the Homepage and People, Highlights, and News.

tldr; it is easier to implement the relation as Many-to-Many at the db level while managing only one instance of the homepage ever (manifesting as a 1:Many) relationship at the admin/api level.

When we look at how the admin interface is structured, we see that what we let users change are the relationships between the homepage and the related model. We aren't changing the related model nor are we changing the homepage model itself. At the admin level, we are still asking the question - which people/news/highlights are related (or contained) in the homepage (note that we are focussing on the relation here). However, at the db level, since we want to deal with the relationship, the best way to expose the relationship and tinker with it is to completely disassociate the related models and have a separate model (which is equivalent to a bridge table in a many-to-many relationship) that handles the relationship so that we can always use that model to modify the relationship (which is all that it is meant for) vs. using the homepage or people/news/highlight models directly for managing the relationship (which is not what it is meant for). This allows us to treat the Homepage model effectively as a shell model that contains other models (which we control through the intermediary models named HomepageLeaders, HomepageNews, and HomepageHighlights). Not important, but this will also allow us to extend the homepage relations further later on if we need to (for e.g. adding more/removing leaders, tracking history of changes, etc.).


gideonthomas:
hmm, so, I don't understand the error that travis is failing on (seems to be on a healthcheck route?). @alanmoo, do you have any idea what this error is?

alanmoo:
I noticed this failure this morning when trying to redeploy master. I'm guessing something changed in testing and our code isn't specific enough?

gideonthomas:
where's the healthcheck route served from? The api or the front-end?

alanmoo:
Looks like it gets built as an output of an npm script.

The heroku build that failed earlier today suggested 
> - Dangerous semver range (>) in engines.node
>       https://devcenter.heroku.com/articles/nodejs-support#specifying-a-node-js-version

(https://dashboard.heroku.com/pipelines/49045b00-f8f2-4bee-9e5f-cfc684926fe5, which really I probably shouldn't have attempted to deploy, because immediately after triggering it I realized it would push untested changes from staging into production)

Issue #691, opened 2017-08-24
Allow sign-up widgets to pull from any newsletter


hannahkane:
We have one use case where we want to add a sign-up widget to a page, but the the user should not be signing up for the main Mozilla Network newsletter. Instead it's a TBD newsletter that's being created.

This ticket is for the technical work of allowing CMS admins to specify a particular newsletter per sign-up widget.

Pending product conversation with @xmatthewx

hannahkane:
Grooming notes: 
 - Add field to sign-up widget model for path for newsletter name
 - Limit the admins who can create a new sign-up form (we have an interest in keeping each newsletter-specific sign-up form consistent)
 - Don't need to validate that newsletter isn't used twice; will just trust people to follow the rules

hannahkane:
Just noting a UX challenge this opens up - every page has the "SIGN UP" button in the header that lets users sign up for the Mozilla Network newsletter. This is potentially quite confusing on pages with a different newsletter sign-up widget.

xmatthewx:
I'd like to pause this, push back on the need to service unique snowflakes while we're still trying to get the core services running optimally.

hannahkane:
@xmatthewx - OK. How do you want to address the immediate need for MITI?

xmatthewx:
I added this to our agenda to discuss with Jesse. We can loop back here after.

alanmoo:
Did we come to a decision on this last week?

xmatthewx:
I believe we said we'd allow this lone exception, with the expectation that this will become part of the main list in the future, distinguished as an interest. It can be engineered with that in mind.

xmatthewx:
@alanmoo - who can confirm this? should we hit up Steph?

alanmoo:
In my time off this managed to get onto production, but it's working as expected.

PR #720, opened 2017-09-06
Add support for custom newsletters

alanmoo:
For #691 

This allows (permissioned) users to type in the newsletter name when creating a signup component, and passes that name to the signup form when it gets used on opportunity pages.


network-api/app/networkapi/landingpage/models.py
```diff
@@ -54,4 +60,4 @@ def __str__(self):
 
     class Meta:
         verbose_name = 'Custom Page'
-        verbose_name_plural = 'Custom Pages'
+        verbose_name_plural = 'Custom Pages'
```
gideonthomas: undo this change?

network-api/app/networkapi/landingpage/models.py
```diff
@@ -19,6 +19,12 @@ class Signup(models.Model):
         help_text="Body of signup component"
     )
 
+    newsletter = models.CharField(
+        max_length=100,
+        help_text="The (pre-existing) SalesForce newsletter to sign up for",
+        default="mozilla-leadership-network"
```
gideonthomas: do you want to allow them to leave the field blank and then in the db it defaults to this? Or is this just a default for existing entries and they always have to put something in for this field in the admin interface?

network-api/app/networkapi/templates/pages/landingpage.html
```diff
@@ -69,7 +69,7 @@ <h1 class="h1-white">{{ page.landingpage.header }}</h1>
             <div class="row">{% if page.landingpage.signup %}
               <div class="cms col-md-6">{{ page.landingpage.content | safe }}</div>
               <div class="col-md-5 offset-md-1">
-                <div class="join-us half-width col" data-cta-header="{{ page.landingpage.signup.header | escape }}" data-cta-description="{{ page.landingpage.signup.description | escape }}"></div>
+                <div class="join-us half-width col" data-cta-header="{{ page.landingpage.signup.header | escape }}" data-cta-description="{{ page.landingpage.signup.description | escape }}" data-newsletter="{{ page.landingpage.signup.newsletter }}"></div>
```
gideonthomas: don't know if you want to `escape` this as well?

network-api/app/networkapi/landingpage/models.py
alanmoo: The latter. Also the fact that new entries get pre-populated with this in the admin is helpful too.

network-api/app/networkapi/templates/pages/landingpage.html
alanmoo: Nope, just a string here, not HTML.

alanmoo:
@gideonthomas I re-added the newline, think I need you to approve again.

PR #755, opened 2017-09-13
add a models2json management command that renders our API views to js…

cadecairos:
…on files for the front-end build

WIP.


alanmoo:
Is this still necessary with the JSX approach we're taking?

Issue #775, opened 2017-09-21
Reactify People


gvn:

alanmoo:
Looks good on staging to me!

PR #778, opened 2017-09-21
switching People over to XHRd React component

gvn:
related to #775

### DIFF
--- a/source/js/components/people/people.jsx
+++ b/source/js/components/people/people.jsx
@@ -1,14 +1,31 @@
 import React from 'react';
 import Person from './person.jsx';
-import peopleData from '../../../json/temp/people.json';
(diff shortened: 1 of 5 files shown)


source/js/components/people/people.jsx
```diff
@@ -1,14 +1,31 @@
...
+  componentDidMount() {
+    let xhr = new XMLHttpRequest();
+
+    xhr.addEventListener(`load`, () => {
+      this.setState({
+        people: JSON.parse(xhr.response)
+      });
+    });
+
+    xhr.open(`GET`, `https://${env.NETWORK_API_DOMAIN}/api/people/?format=json&featured=True&page=1`);
```
mmmavis: do we only need to get data from `page=1`?

PR #783, opened 2017-09-22
Add a management command that loads data from a fixture.

cadecairos:
The management command "heroku_release" will run automatically on heroku. If the "LOAD_FIXTURE" environment variable is set to true, it will not only migrate the database, but fill it with data resembling production. (no user data)

It has a default superuser:

testuser:networktest


Procfile
```diff
@@ -1,2 +1,2 @@
-release: ls network-api/app/networkapi && python network-api/app/manage.py migrate
+release: cd network-api && python network-api/app/manage.py migrate && python networkapi/app/manage.py checkenv
```
Pomax: should this be `migrate --no-input`?

network-api/app/networkapi/settings.py
```diff
@@ -113,6 +117,7 @@
     'adminsortable',
 
     # the network site
+    'networkapi',
```
Pomax: interesting... what does this do?

cadecairos: loads the management command. I felt it didn't belong in any other single app.

Procfile
cadecairos: Yup,  and the command is wrong too Lol

cadecairos:
@Pomax This is ready for a review now, when you've got the time.

I wanted to update the readme with details on how to use the command, but it seems we havent' merged the readme's for both repos yet?

cadecairos:
adding @alanmoo as a reviewer - since @Pomax will not be doing Moz stuff tomorrow

Procfile
```diff
@@ -1,2 +1,2 @@
-release: ls network-api/app/networkapi && python network-api/app/manage.py migrate
+release: cd network-api && python network-api/app/manage.py --no-input migrate && python networkapi/app/manage.py heroku_release
```
alanmoo: I believe this isn't gonna work. Double check it?

cadecairos: yup, don't know what my brain was doing.

app.json
```diff
@@ -0,0 +1,36 @@
+{
+  "name": "network",
+  "description": "The Mozilla Network site",
+  "addons": [
+    "heroku-postgresql:hobby-dev"
+  ],
+  "env": {
+    "HEROKU_APP_NAME": {
```
cadecairos: https://devcenter.heroku.com/articles/github-integration-review-apps#heroku_app_name-and-heroku_parent_app_name

This, plus some changes I made to a few jsx components and the main.js file should let review apps automagically work

README.md
```diff
@@ -9,15 +9,77 @@
...
-- `npm start`
+- `cp env.default .env`
 
-This will install all dependencies, build the code, start a server at [http://127.0.0.1:2017](http://127.0.0.1:2017), and launch it in your default browser.
+Install npm dependencies and build the static parts of the site by running the following commands:
+
+- `npm install`
+- `npm run build`
+
+Next, create a virtual environment using either `virtualenv` or `python3`'s virtual environment invocation. For the purposes of this README.md it is assumed you called this virtual environment `venv`.
```
alanmoo: Some npm scripts assume `venv` is the name as well

network-api/app/networkapi/management/commands/heroku_release.py
```diff
@@ -14,6 +14,4 @@ def handle(self, *args, **options):
 
         if settings.LOAD_FIXTURE:
             call_command('loaddata', 'app/networkapi/fixtures/test_data.json')
-            site = Site.objects.get(id=1)
-            site.domain = 'localhost:8000'
-            site.save()
+            call_command('update_site_domain')
```
alanmoo: Why do we need this for Heroku? Shouldn't it be a different url?

source/js/components/news/news.jsx
```diff
@@ -21,7 +28,7 @@ export default class News extends React.Component {
       });
     });
 
-    xhr.open(`GET`, `https://${env.NETWORK_API_DOMAIN}/api/news/?format=json&featured=True&page=1`);
+    xhr.open(`GET`, `https://${networkApiDomain}/api/news/?format=json&featured=True&page=1`);
```
alanmoo: I think @gvn recently refactored this out with the JSX migration. You may not need these changes if that's landed in master. If we do still need the current domain in the front-end, can we make it not baked in at build-time so that we don't run into the "heroku promotion" bug we've got on Learning.m.o and other sites?

(This comment applies to the other JSX components too)

network-api/app/networkapi/management/commands/heroku_release.py
cadecairos: The fixture data relates pages to sites via domain, strangely, and is currently set to use `localhost:8000`

This will make sure the pages loaded from the fixture are visible in the admin interface, and shouldn't affect anything beyond that. If it does, we can try using the HEROKU_APP_NAME variable to update the site and pages in the release command.

source/js/components/news/news.jsx
cadecairos: > not baked in at build-time
👍

This branch appears to be up to date, but we might consider holding off until @gvn's work lands.

gvn: I'm done with migrating the CMS driven components to JSX. 

Is there a reason we'd want to pull from the Heroku app's API instead of prod (or whatever NETWORK_API_DOMAIN is potentially overridden to)? Most of the time we're going to want to see the site with prod's data.

cadecairos:
@alanmoo / @Pomax okay, fixed up lint problems

source/js/components/news/news.jsx
```diff
@@ -3,6 +3,13 @@ import PropTypes from 'prop-types';
 import moment from 'moment';
 import env from '../../../../env.json';
 
+let networkApiDomain = env.NETWORK_API_DOMAIN
+
+// HEROKU_APP_DOMAIN is used by review apps
+if (!networkApiDomain && env.HEROKU_APP_NAME) {
```
alanmoo: If Heroku is cloning the environmental variables from staging, can review apps unset a single one automatically?

cadecairos: I've actually elected in the app.json to not inherit from staging. all that needs to be set should be set automatically. In the case of `HEROKU_APP_NAME` we have no control over what variable this value is assigned to and must check for it manually.

alanmoo:
Can we land this once CI finishes? (I just rebased the branch)

gvn:
Getting errors on `/people`

screen shot 2017-10-02 at 9 39 14 am

cadecairos:
@gvn is `NETWORK_API_DOMAIN` set in your .env?

gvn:
@cadecairos Yeah, I did a fresh copy of `env.default`. Other subpages/XHR'd components work, so I think it's a bug with that page.

cadecairos:
@gvn I found&fixed the issue, along with a couple others in the latest commit.

gvn:
I'm now hitting a bunch of CORS issues, eg:

`Failed to load https://network.mozilla.org/api/homepage: Redirect from 'https://network.mozilla.org/api/homepage' to 'https://network.mozilla.org/api/homepage/' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource. Origin 'null' is therefore not allowed access.`

cadecairos:
This appears to be a bug in Webkit/Blink that sets the Origin header to `null` on XHR requests that get 302's (even if they're not cross-origin redirects)

In Firefox, the Origin header is properly set...

gvn:
Putting the protocol back in the URLs solves [the issue](https://github.com/mozilla/network/pull/783#issuecomment-333673519). 

Can we do that?

alanmoo:
Yes, but as @cadecairos pointed out to me in a call we just had, we have to remember to update the env var with the protocol right before we deploy this code.

gvn:
The env var `NETWORK_API_DOMAIN ` is just the domain. Protocol should be able to stay `https://` unless we really feel it needs to vary. It is currently hardcoded to `https` and I haven't had any need to change it...

alanmoo:
I believe @gvn was saying in a call that since the change from protocol to protocol-less never happened on staging/production, we should just be able to change this back and land this PR. Sound right @cadecairos?

cadecairos:
@gvn @alanmoo @Pomax rebased, and updated so that protocol can now be specified in the network site variable that the front-end depends on.

R?

alanmoo:
When I hit http://localhost:8000/api/homepage/ I get a 500 triggered by `/network/network-api/app/networkapi/news/serializers.py in get_glyph, line 27`

Anyone else?

gvn:
I get a 200 on that URL with default env and the route fully populated.

gvn:
This is working for me with the new refactored env.

cadecairos:
> When I hit http://localhost:8000/api/homepage/ I get a 500 triggered by /network/network-api/app/networkapi/news/serializers.py in get_glyph, line 27
> 
> Anyone else?

yup, I see this too. might be some strangeness with the fixture data. I'll investigate.

cadecairos:
Oh this is fun. the data assumes images are stored locally, but the fixture data has images stored on S3.

cadecairos:
@alanmoo @Pomax and I successfully located the source of this bug, and it's fixed in https://github.com/mozilla/network/pull/783/commits/101f27f617db40f1fda7e0406aebf24ced27b76f

we should clean up this code in a new bug: #846

Pomax:
Yep, this works quite well. Good find about the serializers, we should definitely make sure indirection can't break things (always tricky to get right)

network-api/LICENSE
```diff
@@ -1,21 +0,0 @@
-MIT License
```
Pomax: can we actually remove this file here?

network-api/app/networkapi/homepage/serializers.py
```diff
@@ -43,7 +44,7 @@ def serialize_relation(
     for related_field in relation:
         related_objs.append(getattr(related_field, related_field_name))
 
-    return related_serializer(related_objs, many=True).data
+    return related_serializer(related_objs, many=True, context=context).data
```
Pomax: 🥇

network-api/LICENSE
cadecairos: Isn't that what this does?

cadecairos:
followup: #847

PR #785, opened 2017-09-22
reactifying News

gvn:
related to #776


gvn:
Yeah, the glyphs are broken on prod right now, but will be fixed next time we do a prod push! 👍

PR #789, opened 2017-09-25
updated svg - error image

taisdesouzalessa:
This is a fix for the inconsistent illustration we were seeing in Firefox:
image

Expected result:
image

The text and the pattern (polka dots) were transformed to paths (pattern was not visible in Firefox) - although we have more shapes now the final size of the file was reduced: from ~186KB to ~84KB


taisdesouzalessa:
@gvn transforming the pattern into paths created a lot of new lines of code. The final size of the file is smaller, though. If the amount of new code seems too much, please let me know and I could upload a PNG instead, as you suggested.

Issue #793, opened 2017-09-26
Move Network Leaders component down on the homepage


xmatthewx:
@xmatthewx commented on [Tue Sep 26 2017](https://github.com/mozilla/network-pulse/issues/705)

Move the Network leaders block to lower on the homepage. Below News/Latest and above Get Involved.

taisdesouzalessa:
Mockups posted here: https://github.com/mozilla/network/issues/791#issuecomment-333688967

gvn:
@taisdesouzalessa I'm unclear on what the final direction is based on #791. I see several variations.

Can you point me to the final comp?

taisdesouzalessa:
@gvn based on the discussion we had [here](https://github.com/mozilla/network/issues/791#issuecomment-333699733) we are not changing the hero image for now. So those mockups don't apply.

This ticket's request is to move the Network leaders block to lower on the homepage (below News/Latest and above Get Involved). No other changes required. Is that clear or still confusing? I can do a quick mockup if it helps.

gvn:
Ok that makes sense. I was confused by your link above to mockups...

taisdesouzalessa:
Yep, I can see how it is confusing - I should have updated this page as well with the decisions we made on that thread...

xmatthewx:
looks good on staging!

PR #836, opened 2017-10-10
Move Network Leaders component down on the homepage - #793

gvn:
#793


mmmavis:
so @gvn , where's the final comps 😉 ?can you point me there? 👍

mmmavis:
@gvn nvm, Tais reached out to me on this just now.

gvn:
There aren't comps after all...just reordering stuff. "Move the Network leaders block to lower on the homepage. Below News/Latest and above Get Involved."

PR #839, opened 2017-10-12
Limit news excerpt length on backend

alanmoo:
Pending approval of https://github.com/mozilla/network/issues/821#issuecomment-336137037

@gideonthomas Limiting the `excerpt` field after-the-fact will just truncate the existing content in the db if anything is there, right? (Not that I think anything is there currently, or that it matters)


alanmoo:
Update: This is good to land from a functionality standpoint once reviewed.

network-api/app/networkapi/news/models.py
```diff
@@ -57,8 +57,8 @@ class News(models.Model):
         help_text='URL link to the article/media clip',
     )
     excerpt = models.TextField(
-        max_length=1000,
-        help_text='A small extract from the article',
+        max_length=200,
```
gideonthomas: since this is a `TextField`, no max length is enforced at the db level...the `200` will only be enforced at the model level so it's relatively safe to change.

gvn: Limit per design is 146.

alanmoo: See comment in original thread. 200 is approved. We all know how being super strict about web content length is unrealistic with regards to layout, so there's some flex space here.

gvn: K

PR #849, opened 2017-10-12
default to using prod for NETWORK_SITE_URL

gvn:
IMO this makes a more sensible default...thoughts?


mmmavis:
Leaving this for you to review @cadecairos ... I don't know much about how Network site is set up. 🙈

PR #853, opened 2017-10-13
Turn HEROKU_APP_NAME (if it exists) into a hostname …

cadecairos:
and add it to the ALLOWED_HOSTNAMES list


cadecairos:
This PR is intended to support enabling review apps.

Issue #822, opened 2017-10-05
Implement section divider ui alignment


kristinashu:
From alignment work in mozilla/network#582 and to align better with new brand, mozilla.org, Network site, and IRH sites.

Comp https://redpen.io/ygc29beb3e61e39afb

<img width="1021" alt="screen shot 2017-10-05 at 10 46 23 am" src="https://user-images.githubusercontent.com/1068783/31242194-1c1e3aa6-a9bb-11e7-949e-4ff70ac3d6ce.png">

Update dashed divider to gradient + highlight text. Note that sometimes the divider is gradient + highlight text and sometimes it is only the highlight text (no gradient)([example](https://redpen.io/tw35abb62b45df306b)).

- gradient colors are same as header banner (from #3DFEFD to #B7D43F), 4px in height
- 30px vertical space between gradient and text
- text is Zilla Slab Highlight 25px with 40px of black before text starts

Related to Pulse nav update https://github.com/mozilla/network-pulse/issues/731

cc @gvn

kristinashu:
Sweet! Looks good except the dividers at top of the Landing page and also the Upcoming page look a little too close to the banner image.

Example:
![screen-shot-2017-10-17-at-8 40 56-am](https://user-images.githubusercontent.com/1068783/31674924-044d766a-b318-11e7-86e4-ec9d40397f21.png)

Please increase the the gap from what I think is 30px to 50px.

Padding on all the other dividers (example: 'Network Leaders' and 'Get Involved' lower down on the Landing page) look good.

kristinashu:
Sorry, one more thing... can you add the selected state? So if I'm on the projects page then "Projects" should be underlined (like the hover state). It seems to be there on narrow screens but not desktop.

gvn:
Changes are up on staging to verify: https://network.mofostaging.net/

kristinashu:
Awesome!

Issue #820, opened 2017-10-05
 Implement nav ui alignment


kristinashu:
From alignment work in mozilla/network#582 and to align better with new brand, mozilla.org, Network site, and IRH sites.

https://redpen.io/ygc29beb3e61e39afb

<img width="1022" alt="screen shot 2017-10-05 at 10 13 42 am" src="https://user-images.githubusercontent.com/1068783/31240638-2ee66e1a-a9b6-11e7-992e-a81308078c21.png">

### Updates include:
- keep logo the same
- show menu items (pull out from hamburger)
- update menu items Open Sans Bold 16px (can use Nunito instead of Open Sans)
- keep signup CTA the same
- hover state for nav items and Username is black underline 2px high
- keep total nav bar height the same (88px)

### Mobile:
- similar to now
- move hamburger icon and close icon to the right
- reversed menu icon [Menu.svg.zip](https://github.com/mozilla/network/files/1364119/Menu.svg.zip)

<img width="970" alt="screen shot 2017-10-06 at 4 37 15 pm" src="https://user-images.githubusercontent.com/1068783/31302197-acadc308-aab4-11e7-957b-1d6bed88fda0.png">

cc @gvn

kristinashu:
The menu items look just a little too tight next to each other. Could you please add an extra couple px between each menu item? 

Example:
![screen-shot-2017-10-17-at-8 31 24-am](https://user-images.githubusercontent.com/1068783/31674391-90acfe66-b316-11e7-972e-a5f09680c656.png)

Otherwise, everything else looks great!

gvn:
What would you like the spacing to be exactly?

kristinashu:
If it's 18px now, then please increase it to 28px (that might be @2x).

gvn:
Changes are up on staging to verify: https://network.mofostaging.net/

kristinashu:
The space between nav items looks good but please make the space between the logo and the first nav item the same as between individual items (ie make the blue box the same as the pink ones).

![screen-shot-2017-10-19-at-4 32 48-pm](https://user-images.githubusercontent.com/1068783/31798952-e2ff859a-b4eb-11e7-9b3c-81c57f37c7e7.png)

gvn:
Changes are up on staging to verify: https://network.mofostaging.net/

kristinashu:
Perfecto!

PR #859, opened 2017-10-17
Ui align fixups

gvn:
Addressing design feedback from: #822 #820


cadecairos:
uhhh, @gvn I tried to rebase this so that I could use it to test out Review Apps, and it seems to have pulled in the `app.json` commit, even though it's already landed on master... :|

alanmoo:
Scratch that, it was a CORS issue.

PR #862, opened 2017-10-18
Fix domain issues with fixture loading

cadecairos:
1. Change all example.com and network.mozilla.org references in the fixtures to localhost:8000
2. Change the default site domain before loading the fixure data. This will ensure that future heroku "releases" dont error out, since the domain will always be as expected.


cadecairos:
to test:

on current master branch, `python app\manage.py heroku_release` will error on subsequent runs.
on this branch it will work every time.

when testing be sure to flush the db `python app\manage.py flush`

PR #863, opened 2017-10-19
Add a new route '/environment.json' to expose whitelisted env var as JSON

### LINKED ISSUE #843: Create route that exposes whitelisted env vars as JSON object
1. A route handler, maybe `/environment`?
2. We can have it serve the same variables we output in `build-env.js`
3. accepts GET requests, response type "application/json"
4. Ensure it only renders whitelisted vars.

patjouk:
Fixes #843

### DIFF
--- a/network-api/app/networkapi/views.py
+++ b/network-api/app/networkapi/views.py
@@ -1,0 +1,13 @@
+from django.http import JsonResponse
+from django.views import View
+from mezzanine.conf import settings
+
+
+class EnvVariablesView(View):
+    """
+    A view that permits a GET to expose whitelisted environment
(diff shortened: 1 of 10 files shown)


network-api/app/networkapi/views.py
```diff
@@ -0,0 +1,13 @@
...
+from mezzanine.conf import settings
+
+
+class EnvVariablesView(View):
+    """
+    A view that permits a GET to expose whitelisted environment
+    variables in JSON.
+    """
+
+    def get(self):
```
cadecairos: I get an error here when running this: `get() takes 1 positional argument but 2 were given`

PR #864, opened 2017-10-19
ALLOWED_HOSTS does not require the protocol!

cadecairos:
this should enable the admin site to work on review apps.


alanmoo:
Which env var was the one with a protocol issue recently? Not this one?

cadecairos:
> Which env var was the one with a protocol issue recently? Not this one?

That was a frontend variable, this is different

PR #865, opened 2017-10-19
Pull env via XHR

### LINKED ISSUE #844: Consume env vars via route during bootstraping of app

gvn:
closes #844


gvn:
Note: I left the `build:env` script intact because the generated JSON is being used in the build system. I'd be open to removing that in a follow-on ticket, but it's going to take more refactoring.

cadecairos:
> I'd be open to removing that in a follow-on ticket, but it's going to take more refactoring.

yeah, let's do that in a follow-up: #866

Issue #425, opened 2017-05-15
QA > Hover state on email signup button is barely noticeable.


flukeout:
I'd increase the contrast between the normal and hovered states...

![image](https://cloud.githubusercontent.com/assets/25212/26079554/06b76574-3978-11e7-9bcd-d777ffb2af8e.png)

hannahkane:
estimate is just design

gvn:
@taisdesouzalessa If you want to come up with a darker color value for hover I can tweak this pretty easily.

taisdesouzalessa:
@gvn here is the hex code for the button hover state: `#e7c121` - I connected with Kristina and Sabrina to make sure we are all aligned on it.

Here are the demos of it in action:
![signup-yellow3](https://user-images.githubusercontent.com/22157921/31636553-c640a182-b27f-11e7-8bc6-025f48ebd3cb.gif)
![signup-yellow4](https://user-images.githubusercontent.com/22157921/31636555-c83ba856-b27f-11e7-9d47-9d0ed7a35d3c.gif)

Tks!

taisdesouzalessa:
It looks great @gvn! Closing this issue.

PR #873, opened 2017-10-24
darker hover state for buttons - #425

gvn:
#425


alanmoo:
Plz open from the same repo for review app?

gvn:
I don't think review apps are working, but: https://github.com/mozilla/network/pull/874

PR #874, opened 2017-10-24
darker hover state for buttons - #425

gvn:


alanmoo:
Thanks! I turned on automatic review apps for this, one minor issue filed in #875.

PR #880, opened 2017-11-01
Fix s3 assets on review apps

### LINKED ISSUE #875: Review apps don't show s3 assets
I think we're still missing an env var or two.

Images with a host of `assets.mofostaging.net` in the staging app instead get a host of
`network-site-staging-pr-874.herokuapp.com`, for example

patjouk:
fix #875 

I've added a few s3 env variables that will inherit their values from the staging stack. It should fix the issue around assets.

### DIFF
--- a/app.json
+++ b/app.json
@@ -11,6 +11,24 @@
     "DJANGO_SECRET_KEY": {
         "generator": "secret"
     },
+    "ASSET_DOMAIN": {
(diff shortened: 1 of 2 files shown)


patjouk:
Images are not broken anymore 🎉 
This PR is now ready for review @cadecairos

PR #882, opened 2017-11-02
adding og:image - closes #602

### LINKED ISSUE #602: Add og:image
See: https://developers.facebook.com/docs/sharing/best-practices#images

gvn:
#602


gvn:
Since we're not (or shouldn't be, anyway) sharing staging URLs to Facebook, it should be safe to hardcode the production domain for OG tags.

I tweaked the template to do that...

gvn:
I realize it's not perfect, but it should really be fine. When do we ever need non-production URLs in OG tags?

alanmoo:
Highly unlikely. "Bugs" probably isn't the right description, but if we're trying to change some OG content there might be a "why isn't this working on staging?! Oh it's hardcoded to production values" type of minor time-sink. Not enough to be worth redoing much work right now, just something to keep in mind.

PR #825, opened 2017-10-05
Fix #798: Removed signal related files and depentant methods

jayyyin:
Signals removed,
tester changed,
build and decorator files removed
fixes https://github.com/mozilla/network/issues/798

PR #920, opened 2017-11-22
Fix #798: Removed signal related files and depentant methods + rebased master, rebranched and redone changes.

### LINKED ISSUE #798: Remove Signals for building using the Heroku Build API
as per #749, and with #656 landed, the home page now pulls in content via XHR, not during the npm build step.

It should be safe to pull out this functionality now.

jayyyin:
most up to date changes for removing the signal related files


jayyyin:
original pull request here https://github.com/mozilla/foundation.mozilla.org/pull/825

alanmoo:
Thanks for the update! @cadecairos or @patjouk can you take a look at this over the next few days and merge it in if it's all set?

jayyyin:
@cadecairos yep just removed them, must have missed them when I removed the other ones

cadecairos:
I apologize that I didn't catch this before, but we also need to remove all the related setting from [settings.py](https://github.com/mozilla/foundation.mozilla.org/blob/bb8a32b50adf7eca14cd8deaba5c639705892f71/network-api/networkapi/settings.py)

Specifically these lines:

1. [here](https://github.com/mozilla/foundation.mozilla.org/blob/bb8a32b50adf7eca14cd8deaba5c639705892f71/network-api/networkapi/settings.py#L28-L29)
2. [here](https://github.com/mozilla/foundation.mozilla.org/blob/bb8a32b50adf7eca14cd8deaba5c639705892f71/network-api/networkapi/settings.py#L31-L33)
3. [here](https://github.com/mozilla/foundation.mozilla.org/blob/bb8a32b50adf7eca14cd8deaba5c639705892f71/network-api/networkapi/settings.py#L419-L424)

jayyyin:
@cadecairos np, I just removed the specified settings, should the rest of the HEROKU api stuff I see in there be touched? is that the name of the old api?

cadecairos:
> should the rest of the HEROKU api stuff I see in there be touched? is that the name of the old api?

Thanks for bringing this up, we need to keep [this block](https://github.com/jayyyin/network/blob/c9aa8729f61ef6a8757be0571840a52f84fa9d1d/network-api/networkapi/settings.py#L84-L86), because it enables review apps to work properly. This means we should add back in [this line](https://github.com/mozilla/foundation.mozilla.org/pull/920/files#diff-2b1ca8a00cb2ce5d77f49ee88010d26fL31), which served two purposes before, but is still needed.

jayyyin:
@cadecairos alright thanks for the info, and I've added the line back in, should be good to go now from what I can tell.

cadecairos:
Thanks so much for the PR @jayyyin!

jayyyin:
@cadecairos Thank you!, it's been a pleasure.

Issue #950, opened 2018-01-12
Build fellowships homepage template


alanmoo:
Build a static homepage based on the wireframes at https://projects.invisionapp.com/share/UQF78CYKM#/screens/271941410_It-4-Overview.

UI design to come, but this should be enough for a start of a scaffolding. No carousel.

mmmavis:
I guess this is a good ticket to start. Assigning it to myself. 

@alanmoo is `/network-api/networkapi/fellows/templates/fellows_home.html` where I'm supposed to add the majority of HTML/template code to?

alanmoo:
Yup!

mmmavis:
Done and merged into `fellowships` branch.

PR #965, opened 2018-01-17
fellowships homepage wireframe

mmmavis:
Related to #950

WIP Comps: https://projects.invisionapp.com/share/ZJ9YTAGEA#/screens/273312987

This PR adds elements to the fellowships homepage. I didn't wanna bother adding CSS since the comps haven't been finalized and Sketch file is not ready yet. We'll fine tune stylings at a later stage.

### DIFF
--- a/network-api/networkapi/fellows/templates/fellows_home.html
+++ b/network-api/networkapi/fellows/templates/fellows_home.html
@@ -8,4 +8,45 @@
   </div>
   <div class="col-md-6"></div>
 </div>
+<div class="row mb-5">
+  <div class="col-md-12">
+    <h2 class="h3-black">Meet the Latest Cohort</h2>
+  </div>
+  <div class="col-md-6 fellow-card">
+    <div class="row">
+      <div class="col-md-4">pic</div>
+      <div class="col-md-8">
+        <h5>Firstname Surname</h5>
+        <p>This copy is short bio ofeach Fellow and a brief descripiton of the project they foceused on during their Fellowship.</p>
+      </div>
+    </div>
+  </div>
+  <div class="col-md-6 fellow-card mb-4">
+    <div class="row">
+      <div class="col-md-4">pic</div>
+      <div class="col-md-8">
+        <h5>Firstname Surname</h5>
+        <p>This copy is short bio ofeach Fellow and a brief descripiton of the project they foceused on during their Fellowship.</p>
+      </div>
+    </div>
+  </div>
+  <div class="col-md-6 fellow-card mb-4">
+    <div class="row">
+      <div class="col-md-4">pic</div>
+      <div class="col-md-8">
+        <h5>Firstname Surname</h5>
+        <p>This copy is short bio ofeach Fellow and a brief descripiton of the project they foceused on during their Fellowship.</p>
+      </div>
+    </div>
+  </div>
+  <div class="col-md-6 fellow-card mb-4">
+    <div class="row">
+      <div class="col-md-4">pic</div>
+      <div class="col-md-8">
+        <h5>Firstname Surname</h5>
+        <p>This copy is short bio ofeach Fellow and a brief descripiton of the project they foceused on during their Fellowship.</p>
+      </div>
+    </div>
+  </div>
+</div>
 {% endblock %}
--- a/network-api/networkapi/fellows/templates/fellows_home.html
+++ b/network-api/networkapi/fellows/templates/fellows_home.html
@@ -1,5 +1,16 @@
 {% extends "foundation-base.html" %}
 
+{% block hero_banner %}
+<div class="row">
+  <div class="col-md-12 col-lg-6 py-5">
+    <div class="mb-4">
+      <h1 class="h1-white">Fellowships</h1>
+    </div>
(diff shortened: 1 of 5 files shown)


network-api/networkapi/templates/foundation-base.html
```diff
@@ -72,6 +72,34 @@
         </div>
       </div>
     </div>
+    {% if showHeroBanner %}
```
ScottDowne: you can change this if to use blocks.

Like this:

```
{% block hero_banner %}{% endblock %}
```

You can also fill in that block with a default banner unrelated to fellows, but I don't think we need to in our case.

Then in fellows_home.html somewhere you can do:

```
{% block hero_banner %}
    <div id="hero">
      <div class="container">
        <div class="row">
          <div class="col-md-12 col-lg-6 py-5">
            <div class="mb-4">
              <h1 class="h1-white">Fellowships</h1>
            </div>
            <p class="lead-black">Our Fellowships empower bright minds in the fields of public policy, science and journalism.</p>
          </div>
        </div>
      </div>
    </div>
{% endblock %}
```

and then remove the `, { 'showHeroBanner': True }` in fellows/views.py

ScottDowne: I think. I've  not tested this...

mmmavis: Sounds good. Will fix!

network-api/networkapi/templates/foundation-base.html
```diff
@@ -72,6 +72,34 @@
...
+            <div class="mb-4">
+              <h1 class="h1-white">Fellowships</h1>
+            </div>
+            <p class="lead-black">Our Fellowships empower bright minds in the fields of public policy, science and journalism.</p>
+          </div>
+        </div>
+      </div>
+    </div>
+    {% endif %}
+    <div id="fellowships-nav" class="mb-5">
```
ScottDowne: Is the nav supposed to be available to all pages extending foundation-base.html?

mmmavis: From what I see on https://projects.invisionapp.com/d/main#/console/10016110/263404839/preview ... yea I think the fellowship nav is always there.

It seems like the homepage is the only page that has the hero banner though. Did I understand this correctly, @taisdesouzalessa ?

network-api/networkapi/fellows/templates/fellows_home.html
```diff
@@ -1,5 +1,161 @@
...
+    <a href="" class="cta-link">Learn More</a>
+  </div>
+</div>
+
+<div class="row mb-5">
+  <div class="col-md-12 my-5">
+    <hr class="hr-gradient mt-0">
+    <h4 class="h2-headings-white">Meet Some of the Mozilla Fellows</h4>
+  </div>
+  <div class="featured-person col-6 col-sm-4 col-md-3 mb-4 mb-sm-0" key={index}>
```
ScottDowne: What's the index from?

network-api/networkapi/fellows/templates/fellows_home.html
```diff
@@ -1,5 +1,161 @@
...
+      <li>Sugar plum caramels jelly.</li>
+      <li>Marshmallow jelly beans sesame snaps fruitcake jelly-o icing bear claw dessert.</li>
+    </ul>
+  </div>
+  <div class="col-md-6">img</div>
+</div>
+
+<div class="row py-5 px-md-2">
+  <div class="col-md-2">img</div>
+  <div class="col-md-10">
```
ScottDowne: A lot of chunks in the body block can probably be rendered from an inclusion tag (kinda like setting up a component)

Although these last two chunks of html are almost the same cept the top one (block this comment in line attached to) has `col-md-2` then `col-md-10` and the last one just under this comment has `col-md-1` and `col-md-11` which is the type of minor variance that makes things like an inclusion tag a bit annoying.

ScottDowne: I'm also not sure if any of these inclusion tags would be used outside of this page or not, or just this page, and if so, does jdnago have a way to do it in a light way way.

network-api/networkapi/fellows/templates/fellows_home.html
mmmavis: ha this was some copy and paste mistake

network-api/networkapi/utility/templatetags/active_nav.py
```diff
@@ -0,0 +1,15 @@
+from django import template
```
mmmavis: copied from https://maorleger.wordpress.com/2015/11/20/active-page-styling-for-bootstrap-navbar-and-django/

should I leave a comment here and cite that URL?

mmmavis:
@alanmoo I think all other fellowships pages will inherit from `foundation-base.html` so I left fellowships nav code there and not `fellows_home.html`.

network-api/networkapi/utility/templatetags/active_nav.py
alanmoo: Probably useful for future reference.

mmmavis:
🎉

PR #970, opened 2018-01-18
Django 1.11

cadecairos:
1. Use the lastest master branch commit of Mezzanine (Django 1.11 support)
2. Update Django to 1.11
3. Fix custom adminsortable template tags that were passing Context into the render method
4. Remove deprecated Middleware

### DIFF
--- a/network-api/networkapi/utility/templatetags/adminsortable_tags_custom.py
+++ b/network-api/networkapi/utility/templatetags/adminsortable_tags_custom.py
@@ -10,7 +10,6 @@ def render_sortable_objects(
     objects,
     sortable_objects_template='adminsortable/shared/objects.html'
 ):
-    context.update({'objects': objects})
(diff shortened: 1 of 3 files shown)


network-api/networkapi/utility/templatetags/adminsortable_tags_custom.py
```diff
@@ -44,10 +51,13 @@ def render_object_rep(
     obj,
     forloop
 ):
-    context.update({'object': obj, 'forloop': forloop})
     tmpl = template.loader.get_template(
         '{app_name}/adminsortable_objects_custom.html'.format(
             app_name=context.get('opts').app_label
```
gideonthomas: this is still working?

cadecairos: The trick was to pass `opts` along from context in every level of the template render.

PR #971, opened 2018-01-18
Adding some stub urls for the fellowships site.

ScottDowne:


network-api/networkapi/fellows/urls.py
```diff
@@ -2,5 +2,8 @@
 from networkapi.fellows import views
 
 urlpatterns = [
-    url('', views.fellows_home),
+    url(r'^$', views.fellows_home),
```
mmmavis: @alanmoo are these all the routes we are going to have?

mmmavis: @alanmoo 

Ah from the page structure section on https://docs.google.com/document/d/1dVRxKDmSIkrJdKhPqxHzhrByJ6fFkf0eCsQgm12sckI/edit# I guess this PR just didn't cover all the routes. And we need to change `/support` to be `/supportthefellows`?

alanmoo: `/support` is nicer

Issue #981, opened 2018-01-22
swap guetzli for imagemin-guetzli


Pomax:
as far as I can tell, based on [https://github.com/imagemin/guetzli-bin]'s README.md this is the same command line utility, but doesn't break when running `npm install` on windows.

PR #982, opened 2018-01-22
replaces guetzli with imagemin-guetzli

Pomax:
same functionality, but doesn't break windows installs.

closes https://github.com/mozilla/foundation.mozilla.org/issues/981


gvn:
Works for me (on OSX).

PR #996, opened 2018-01-24
(DO NOT MERGE YET) Fellowships work

mmmavis:
I'm opening up a PR to see if review app for the `fellowships` branch works.


network-api/networkapi/urls.py
```diff
@@ -30,6 +30,9 @@
 
 urlpatterns = list(filter(None, [
     url(r'^admin/', include(admin.site.urls)),
+    url(r'^soc/', include('social_django.urls', namespace='social'))
+    if settings.SOCIAL_SIGNIN else '',
+    url(r'^fellowships/', include('networkapi.fellows.urls')),
```
cadecairos: on line 47 below, there's this code:

```python
url(r'^soc/', include('social_django.urls', namespace='social')),
 
# Don't remove this redirect until Fellowships pages are live
url(r'^fellowships/$', RedirectView.as_view(
             url='https://advocacy.mozilla.org/open-web-fellows'
         ))
if settings.SOCIAL_SIGNIN else None,
```

There are two things we need to look at here:
Firstly, we'll want to update the route pattern to something like this
```python
url(r'^fellowship/(?P<path>.*)', RedirectView.as_view(
    url='/fellowships/%(path)s',
    query_string=True
)),
```
Secondly the line with 
```python
if settings.SOCIAL_SIGNIN else None,
``` 
should actually follow 
```python
url(r'^soc/', include('social_django.urls', namespace='social'))
```
... but there are two of the above url configs in the file, so lets de-dupe that!

source/js/fellowships.js
```diff
@@ -0,0 +1,183 @@
...
+
+import React from 'react';
+import ReactDOM from 'react-dom';
+
+import Person from './components/people/person.jsx';
+import LoadingIndicator from './components/loading-indicator/loading-indicator.jsx';
+
+let pulseApiDomain = ``;
+let pulseDomain = ``;
+const DIRECOTRY_PAGE_FILTER_OPTIONS = {'program_year': `2017`};
```
cadecairos: `DIRECOTRY_PAGE_FILTER_OPTIONS` => `DIRECTORY_PAGE_FILTER_OPTIONS `

source/js/fellowships.js
```diff
@@ -0,0 +1,183 @@
...
+const DIRECTORY_PAGE_TYPE_ORDER = [ `science`, `open web`, `tech policy`, `media`];
+
+function getFellows(params, callback) {
+  Object.assign(params, {'profile_type': `fellow`});
+
+  let queryString = Object.entries(params).map(pair => pair.map(encodeURIComponent).join(`=`)).join(`&`);
+  let req = new XMLHttpRequest();
+
+  req.addEventListener(`load`, () => {
```
cadecairos: This could be a one-liner:

`req.addEventListener(`load`, () => callback.call(this, JSON.parse(req.response)));`

Do we really need to bind the callback to `this`?

PR #1007, opened 2018-01-26
Campaign pages and petition

alanmoo:
ohboy


network-api/networkapi/campaign/views.py
```diff
@@ -0,0 +1,85 @@
...
+    try:
+        petition = Petition.objects.get(id=pk)
+    except ObjectDoesNotExist:
+        return Response(
+            {'error': 'Invalid petition id'},
+            status=status.HTTP_400_BAD_REQUEST,
+        )
+
+    data = {
+        'entry.1017670790': request.data['givenNames'],
```
alanmoo: @gideonthomas This will need to be updated when we want to do this for other petitions, right?

cadecairos: If this is going to be a long-ish term solution - i.e. other petitions, we need to abstract these field key names to  fields in the petition model.

Issue #954, opened 2018-01-12
Build "fellowship applications closed" page


alanmoo:
https://projects.invisionapp.com/share/DQF7R82UA#/screens/272095069

mmmavis:
@taisdesouzalessa what happens after user has successfully signed up the newsletter?

alanmoo:
Is it clear where that form submits to as well? Is it the general `mozilla-foundation` newsletter?

mmmavis:
Closing. PR just got merged into `fellowships` branch!

PR #1018, opened 2018-01-30
Fellowships application closed page & CSS cleanup for 'join-us' component

mmmavis:
Related to #954 

Sketch file can be downloaded [here](https://drive.google.com/open?id=1ZqIVpHwzx9AWk9Ef4l22VD82FK4yYWoi)

### fellowships application closed page
- can be visited by going to `/fellowships/apply` or by clicking the "Apply to be a Fellow" link on the fellowships nav
- there's no "application open" state - clicking on the "Apply to be a Fellow" link will link out to an external page during the application open phase (we'll figure out how we do this switch on a followup PR)

### .join-us Join component
- cleaned up styling for the `full-page` variation. This affects the existing Join component used on `/sign-up` - it's now 10-col wide (got okay from Tais already for this change)
- added the 3 variations to `/style-guide`

### DIFF
--- a/network-api/networkapi/fellows/templates/fellows_apply.html
+++ b/network-api/networkapi/fellows/templates/fellows_apply.html
@@ -1,0 +1,21 @@
+{% extends "fellowships-base.html" %}
+
+{% block template_id %}fellows-apply{% endblock %}
+
+{% block body %}
+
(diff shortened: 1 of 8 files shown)


mmmavis:
🎉

Issue #1019, opened 2018-01-30
Responsive fellowships nav


mmmavis:
Make sure the nav we have for fellowships site works and is responsive. 

@taisdesouzalessa can you give me a ping once the comps are finalized? Thanks! 😄

taisdesouzalessa:
Hey @mmmavis they are ready to go: https://drive.google.com/file/d/1ZqIVpHwzx9AWk9Ef4l22VD82FK4yYWoi/view?usp=sharing

This to consider: content still not 100% yet, we may have some few feedback pieces here and there - but styling and structure seems ok, I'll let you know if anything changes as soon as I know.

Next step: working on assets

mmmavis:
For more details about how the nav should work see

- https://github.com/mozilla/foundation.mozilla.org/issues/977#issuecomment-360321769
- https://invis.io/Z4FGLVDEP3W#/275358045_TEST2-MOBILE_-_FellowshipsUI_-_Overview_Copy

mmmavis:
Closing!
Work has been merged into the main `fellowships` branch.

Issue #977, opened 2018-01-19
Opportunity pages > multi page-menu behaviour


taisdesouzalessa:
I was wondering if we could fix a behaviour we are currently having in the multi-page menu, since this model of menu will also be used on the Fellowship Pages. The issue is that, once user select a category on the menu, the category doesn't stay visible

Current behaviour (once I click "Video" I still see "Net Neutrality Groups"):
![current-behaviour](https://user-images.githubusercontent.com/22157921/35163036-35753256-fcfa-11e7-960a-0c41b0a4c4fe.gif)

It was designed to have the category replaced on top once users click on that.
![desired-behaviour-opt1](https://user-images.githubusercontent.com/22157921/35163107-7d1c8b22-fcfa-11e7-8988-8bdcdce0388f.gif)

If this is too complicated to do, we could just have the word "menu" on top, always visible and as users select their option, it is still highlighted, but hidden under menu. 
<img width="344" alt="image" src="https://user-images.githubusercontent.com/22157921/35163141-9961a826-fcfa-11e7-80ae-b15f64268a3f.png">

@alanmoo @xmatthewx I assigned this ticket to you both to : 1. get your opinion on this issue (if you think there is another solution I am not seeing) and 2. triage this ticket

alanmoo:
As far as I can see, you're absolutely right, this is a bug and should show the current page. I think "Menu" would be too confusing.

Let's figure out #830 first, and then we can tackle this.

xmatthewx:
I agree totally. 1) Let's attempt to show current selection. If that's impossible for some weird reason, Menu is our best fallback. 2) This is less important than 830, but still pretty important. Thanks @taisdesouzalessa for catching this.

alanmoo:
I just poked at the code, I don't think it's too difficult to show, we just need the right mezzanine tag swapped out in one file.

taisdesouzalessa:
@alanmoo I did some studies in how this top menu could behave and talked to @mmmavis about what could be the best behaviour for the users. We both agree that something like the example below would be a better user experience. Wondering if you think it costs lots of time to implement something like that. If so, we can go with original plan and do that later on, after MVP.

![mobile-menu-behaviour-small](https://user-images.githubusercontent.com/22157921/35364461-219820a8-0124-11e8-841c-fb3d437be68b.gif)

alanmoo:
Seems reasonable, at first glance anyway.

anweshknayak:
[...]

PR #1020, opened 2018-01-31
Responsive fellowships nav & better multi-page nav on mobile

mmmavis:
Related to 

#1019 - responsive fellowships nav
#977 - better multi-page nav on mobile

Proposed change: https://github.com/mozilla/foundation.mozilla.org/issues/977#issuecomment-360321769

### DIFF
--- a/source/js/components/multipage-nav/multipage-nav.jsx
+++ b/source/js/components/multipage-nav/multipage-nav.jsx
@@ -33,8 +33,6 @@ export default class MultipageNav extends React.Component {
       );
     });
 
-    console.log(activeLinkLabel);
-
     links.unshift(<div key="current-active-link">
       <button className="expander" onClick={this.toggle}>
         <div className="d-flex justify-content-between">
--- a/source/js/components/multipage-nav/multipage-nav.jsx
+++ b/source/js/components/multipage-nav/multipage-nav.jsx
@@ -18,18 +18,31 @@ export default class MultipageNav extends React.Component {
   }
 
   render() {
+    let activeLinkLabel;
     let links = this.props.links.map((link, index) => {
       let className = `opp-link${ link.isActive ? ` opp-link-active` : `` }`;
+
+      if (link.isActive) {
+        activeLinkLabel = <a className={`active-link-label d-inline-block ${className}`}>{link.label}</a>;
+      }
 
       return (
         <div key={`link-${index}`}>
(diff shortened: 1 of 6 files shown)


source/js/components/multipage-nav/multipage-nav.jsx
```diff
@@ -18,19 +18,30 @@ export default class MultipageNav extends React.Component {
   }
 
   render() {
+    let activeLinkLabel;
     let links = this.props.links.map((link, index) => {
       let className = `opp-link${ link.isActive ? ` opp-link-active` : `` }`;
```
mmmavis: We should probably change all `opp-link` and `opp-link-active` class names to something more general like `multipage-link` and `multipage-link-active`. If this sounds sane let's tackle it in a follow-up ticket.

source/sass/type.scss
```diff
@@ -41,7 +41,7 @@ a {
   line-height: 1.5;
 }
 
-@mixin yellow-underline-nav-link($active-class-name, $font-family: "Zilla Slab") {
+@mixin yellow-underline-nav-link($active-class-name, $font-family: "Nunito Sans") {
```
mmmavis: @taisdesouzalessa 
This code change will make **all** multi-page nav links use Nunito Sans (which includes the links on the opportunity page). Sounds good? If not, I revert opportunity page links back to Zilla Slab.

taisdesouzalessa: That sounds good, Mavis. Since we will be using this style across pages, could we use Nunito Sans Semi Bold instead or the regular weight? So the element on the sidebar looks more like a menu and less like a body text. For all nav items - desktop and mobile. I'll alter the mockups to accommodate that change. Sounds good?
image

mmmavis: Let's discuss this in a new ticket. From the [style guide](https://foundation.mozilla.org/style-guide/) it seems like all the anchor styles are in bold (`font-weight: 700`). I hesitate to throw an exception there (at least not in this PR).

Will create a ticket and tag you @taisdesouzalessa and @gvn

source/js/components/multipage-nav/multipage-nav.jsx
gvn: Yeah that makes sense

mmmavis:
🎉 🎉 🎉

PR #1026, opened 2018-02-06
Create campaigns with fake data

alanmoo:
This probably needs some fine tuning (and needs to add petitions), but it's better than what we've currently got, and this code doesn't get run on Prod so I'd like to get this in for others to use today

### DIFF
--- a/network-api/networkapi/management/commands/flush_models.py
+++ b/network-api/networkapi/management/commands/flush_models.py
@@ -3,6 +3,7 @@ from django.core.management.base import BaseCommand
 # Models
 from networkapi.highlights.models import Highlight
 from networkapi.landingpage.models import LandingPage, Signup
+from networkapi.campaign.models import Campaign
 from networkapi.milestones.models import Milestone
 from networkapi.news.models import News
 from networkapi.people.models import (
@@ -28,6 +29,9 @@ class Command(BaseCommand):
 
         self.stdout.write('Dropping Signup objects...')
         Signup.objects.all().delete()
+
+        self.stdout.write('Dropping CampaignPage objects...')
+        Campaign.objects.all().delete()
 
         self.stdout.write('Dropping Highlight objects...')
         Highlight.objects.all().delete()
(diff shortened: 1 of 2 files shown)


network-api/networkapi/management/commands/load_fake_data.py
```diff
@@ -65,12 +66,16 @@ def handle(self, *args, **options):
...
         self.stdout.write('Generating LandingPage objects')
-        opportunity = LandingPageFactory.create(title='opportunity', content='This is placeholder')
+        opportunity = LandingPageFactory.create(title='Opportunity', content='This is placeholder')
         [LandingPageFactory.create(parent=opportunity) for i in range(5)]
 
         self.stdout.write('Generating LandingPage objects with Signup CTAs')
         [LandingPageFactory.create(parent=opportunity, signup=SignupFactory.create()) for i in range(5)]
 
+        self.stdout.write('Generating CampaignPage objects')
+        campaigns = LandingPageFactory.create(title='Campaigns', content='This is placeholder')
```
cadecairos: The bad grammar in this content has bothered me since I typo'd it originally. can you update that :)

alanmoo: How's this?

cadecairos: much better!

network-api/networkapi/management/commands/flush_models.py
```diff
@@ -3,6 +3,7 @@
 # Models
 from networkapi.highlights.models import Highlight
 from networkapi.landingpage.models import LandingPage, Signup
+from networkapi.campaign.models import CampaignPage
```
cadecairos:
```
    from networkapi.campaign.models import CampaignPage
ImportError: cannot import name 'CampaignPage'

```

alanmoo: This is what I get for trying to multitask

PR #1031, opened 2018-02-06
Add fields for petition component

alanmoo:


network-api/networkapi/campaign/models.py
```diff
@@ -82,6 +82,26 @@ class Petition(models.Model):
...
+        help_text="Text content of the share button",
+        default="Share this",
+        blank=True
+    )
+
+    thank_you = models.CharField(
+        max_length=140,
+        help_text="Message to show after thanking people for signing",
+        default="Thank you for signing too!",
+        blank=False
```
gideonthomas: we can remove this, it's the default

alanmoo: This prevents someone from deleting that and saving though, correct?

PR #1034, opened 2018-02-06
place petition in-page or below-page based on child count

Pomax:
- if no children, place the petition in the page, below the regular content
- if there are children, place the petition at the end of the page, above the footer.


alanmoo:
I'm gonna file a more clear ticket for this to be worked on next week

PR #1038, opened 2018-02-07
WIP share buttons for campaigns.

ScottDowne:


source/js/components/petition/petition.jsx
```diff
@@ -292,12 +312,23 @@ export default class Petition extends React.Component {
    * @returns {jsx} What users see when their petition sign up succeeded.
    */
   renderThankYou() {
-    return (
-      <div>
-        <p>{this.props.thankYou}</p>
-        <a href={this.props.shareLink} className="btn btn-info">{this.props.shareText}</a>
-      </div>
-    );
+    if (this.props.shareLink) {
```
alanmoo: This means existing petitions that have a share link and no button links will continue to work as-is?

ScottDowne: That's the idea, but I contemplated having an if for each type of value, and if non exist, we just don't see any buttons.

env.default
```diff
@@ -33,7 +33,7 @@ AWS_SQS_REGION=
 # CSP config
 
 CSP_DEFAULT_SRC="'none'"
-CSP_SCRIPT_SRC='self' 'unsafe-inline' cdn.optimizely.com https://www.google-analytics.com/analytics.js
+CSP_SCRIPT_SRC='self' 'unsafe-inline' cdn.optimizely.com https://www.google-analytics.com/analytics.js https://c.shpg.org/352/sp.js
```
gideonthomas: I know it's not related to your PR but can you a follow-up bug to make sure we figure out how to remove `unsafe-inline` and still have it work, especially since we're now collecting user data

network-api/networkapi/templates/pages/campaign.html
```diff
@@ -17,20 +17,32 @@
...
-          data-cta-description="{{ page.campaign.petition.description | escape }}"
-          data-newsletter="{{ page.campaign.petition.newsletter }}"
-          data-checkbox1="{{ page.campaign.petition.checkbox_1 | escape }}"
-          data-checkbox2="{{ page.campaign.petition.checkbox_2 | escape }}"
-          data-share-link="{{ page.campaign.petition.share_link }}"
-          data-share-text="{{ page.campaign.petition.share_link_text }}"
-          data-thank-you="{{ page.campaign.petition.thank_you }}"
-        ></div>{% endwith %}
+        {% with page.campaign.petition as petition %}
+          {% if page.campaign.petition.share_twitter %}
```
gideonthomas: again, not completely related to you pr but everything from line 21 to 41, can you change `page.campaign.petition` to just `petition`

source/js/components/petition/petition.jsx
gideonthomas: let's file a follow-up issue to disable this field on the admin.

ScottDowne: I think the idea is we can use either or. This doesn't replace the previous functionality, just allows us to pick.

Or are you thinking a way to toggle (disable) the field on and off via the admin page?

env.default
ScottDowne: I think unsafe-inline is there for optimizely, and not share progress.

And there isn't a way currently to make optimizely work without unsafe-inline unless we find a replacement to optimizely.

network-api/networkapi/templates/pages/campaign.html
ScottDowne: Is that going to just work, or do I need to do a sort of
```
var petition = page.campaign.petition
```
type thing? Don't know off hand what the syntax for that would be so I used js to make it easier.

env.default
alanmoo: Are we using optimizely on the foundation site? (We probably need to in the future, but I just don't think that's what the `unsafe-inline` is for at this point)

PR #1053, opened 2018-02-13
Share base template for opportunity and campaign pages

gvn:
This achieves a few things:

- Opportunity and Campaign templates now share a base (More DRY)
- The "guts" of the Opportunity template are no longer generated from Pug. You can just edit `landingpage.html` directly now that it's no longer compiled.
- All Pug and Django pages will share common header/footer code (`source/pug/templates/master.pug`)


source/pug/templates/base-for-django.pug
```diff
@@ -0,0 +1,18 @@
+//-
+  This template is used to generate HTML for usage in Mezzanine.
+  There are {{ TOKENS }} in place for Mezzanine to target with CMS data.
+  `npm build` will compile this to `/dest/opportunity/template/index.html`
```
mmmavis: is this line true for campaign pages too?

gvn: Ah I need to update this...not quite accurate

source/pug/templates/base-for-django.pug
```diff
@@ -0,0 +1,18 @@
...
+
+extends ../templates/master.pug
+
+block masterParams
+  - let isMezzanine = true
+  - let pageTitle = `{{ page.meta_title }}`;
+
+include ../partials/sign-up-box.pug
```
alanmoo: What's this doing?

gvn: Ah, I think actually nothing. That was an artifact from the previous opportunity template, but it's an unused mixin apparently.

PR #1060, opened 2018-02-14
Fixes to petition layout.

ScottDowne:


mmmavis:
The code change here will also increase the input field width on 1-column petition page from

screenshot-2018-2-14 mozilla foundation some water rate 1

to

screenshot-2018-2-14 mozilla foundation some water rate

I'd like to get designer's okay on making this styling changes here. /cc @kristinashu

kristinashu:
Hmm the short field would be better here but if there's no way around it then the long fields are ok.

ScottDowne:
Looking into it.

xmatthewx:
good catch @mmmavis!!! those are some long input fields!! i tested on mobile, but didn't think of 1-column layout. i guess you can only do bootstrap relative to screen-size, not container size.

ScottDowne:
I'm trying a fix where I make the container a bit smaller on the 1 column layout, to roughly match what it was agfter my change. This has the side effect of also making the checkbox labels line wrap if they are really long (but part of me thinks long checkbox input text is actually kinda not great either, so maybe two wins here?)

ScottDowne:
My above solution makes 1 column layout terrible on mobile though.

I think I need help from Mavis to get this working in a bootstrapy way.

ScottDowne:
Think I figured it out.

mmmavis:
Since DOM elements in this codebase are super nested... I think a more future proof solution would be making the input fields as wide as their container. If we want fields look narrower, make the container narrower.

My suggestion would be changing the `col-md-6` class we applied to petition `<input>` to `w-100`.

screenshot-2018-2-14 mozilla foundation some water rate
screenshot-2018-2-14 mozilla foundation choose industry republican

PR #1064, opened 2018-02-15
End the CODEOWNERS madness

alanmoo:


gvn:
Why not just delete it?

alanmoo:
Because I didn't see that as an option in the web admin and was so annoyed with it I wanted a fast solution 😅

PR #1071, opened 2018-02-16
Initial wagtail setup

Pomax:
This PR is for adding the basicmost wagtail setup into our existing codebase, so we can work on moving small bits and pieces over one by one, without having the wagtail settings and files getting in the way of a clean merge into master.

There's not a lot that can go wrong here from what I can tell, but let's get everyone's eyes on this for the initial merge, while it's still small and easy to spot problems specific to wagtail, rather than any work we do around code/file transitions etc.

This PR contains some work by Alan that turns several of our "just data" models into something that wagtail understands as a snippet, adding admin UI for creating arbitrary new instances, as well as code that allows you to directly log into the wagtail cms with the same oauth2 moz creds we currently allow for mezzanine.


alanmoo:
The google auth works fine on wagtail.mofostaging.net, no need to use the review app link

network-api/networkapi/urls.py
```diff
@@ -27,6 +30,9 @@
 
 urlpatterns = list(filter(None, [
     url(r'^admin/', include(admin.site.urls)),
+    url(r'^cms/', include(wagtailadmin_urls)),
+    url(r'^documents/', include(wagtaildocs_urls)),
+    url(r'^wagtail/', include(wagtail_urls)),
```
cadecairos: Can/should we feature flag all the things that wagtail does in `urls.py` and `settings.py`? Then we can make sure we only enable it on staging and prod when we're ready to have people log in there and start porting content.

Pomax: I'm fine with that, @alanmoo ?

alanmoo: Will the migrations still be able to happen on the production database if we feature flag settings?

alanmoo: They'll run, but successfully? They require certain apps to be installed properly, don't they?

cadecairos: The migrations should be agnostic to installed apps, since they're working directly with the ORM

alanmoo: Ok, let’s do the feature flag but default to “disabled” and let it go to staging that way so we confirm.

network-api/networkapi/settings.py
```diff
@@ -94,6 +95,8 @@
 
 SITE_ID = 1
 
+ENABLE_WAGTAIL = env.ENABLE_WAGTAIL
```
alanmoo: I think this isn't quite right. When I pull this branch I get an error: `https://github.com/mozilla/foundation.mozilla.org/pull/1071AttributeError: 'Env' object has no attribute 'ENABLE_WAGTAIL'`

Issue #984, opened 2018-01-23
Add Notice on display of Draft pages


xmatthewx:
Currently, all Mezzanine authors can view pages with Draft status on the frontend as if the page was Published. This is great for a preview, but dangerous since many staff can assume a page is live and visible to the public when it is not.  

Steps to Repeat: 
 - Create a new Opportunity Page via Mezz
 - Save as draft
 - Navigate to page 
 - Expected: page is visible as preview, with notice at top to Publish
 - Actual: page is visible with no notice

Requirements
 - The notice can be simple. Should display at top of all CMS editable pages that are not current Published.
 - [ ] Yellow bar (`#faf2cc`) across the top with text that says: 
   - [ ] "Draft preview. You must save as Published to make it public."
 - [ ] Use notice component styles previously developed by Gavin
 - [ ] A Page can be `Published`, but not live because of a future date or expiration date. We should show notice on ALL pages that are not visible. So, in pseudo code `if (status = draft || now != published ) { display_draft_notice = true }. We could get fancy with conditional notice copy, but that feels like extra complexity. Nice to have but probably not worth it. 

cc @alanmoo to triage

xmatthewx:
![image](https://user-images.githubusercontent.com/53411/35293958-55bfd5e2-0043-11e8-8840-3a565f82cda7.png)

alanmoo:
To implement this requires some extra Django/Mezzanine template code in `opportunity.pug`, which will get generated and added to `landingpage.html`.

I'm not sure what notice component styles you're referring to though. @gvn can you point them out?

gvn:
I'm not sure either. We didn't do a custom alert style for Foundation AFAIK.

xmatthewx:
I was thinking of the barn raising notice last spring. Yellow bar.

alanmoo:
Looks like we need a wrapper
```
{% if not page.published %}
DANGER WILL ROBINSON
{% else %}
```

PR #1085, opened 2018-02-22
Add Notice on display of Draft pages #984

gvn:


gvn:
I think that's the first time I've been asked to make something uglier. 😆

PR #1086, opened 2018-02-22
Show opportunity/campaign children on prmary pages

alanmoo:
For easier review apps. Still not perfect, but an improvement on what we have.

### DIFF
--- a/network-api/networkapi/templates/pages/menus/side-menu-items.html
+++ b/network-api/networkapi/templates/pages/menus/side-menu-items.html
@@ -1,11 +1,11 @@
 {% load pages_tags %}
 {% for page in page_branch %}
   {% if page.in_menu %}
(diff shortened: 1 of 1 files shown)


alanmoo:
@xmatthewx It's going to be way too much effort to show the children pages, is it useful to have this do something though? https://foundation-mofostaging-pr-1086.herokuapp.com/opportunity/ for example shows first level pages.

@cadecairos @patjouk This is related to the issue of not being able to easily access "fake" pages on review apps.

xmatthewx:
I'm unsure.  Is there a way we could do this on staging, but not production?

alanmoo:
I don't think so. I'm not sure it's worth putting more effort in this until we switch to Wagtail, so I'm gonna just close it.

PR #1089, opened 2018-02-23
[WIP] Opportunity pages

Pomax:
This is a WIP continuation of the `wagtail` branch that's been filed as PR mostly so we have a review app to do some testing on. There is enough in place for seeing how we like building opportunity pages atm using "stream field" templating (similar to Wix or Squarespace page building) with enough basic blocks that things like the Aadhaar page can be rebuilt (sans actual signup form right now).

(Please ignore the million or so migrations, as a work in progress every little model change requires a new migration to keep things chugging along)


alanmoo:
This is going to need `unsafe-eval` in the `script-src`. Do we add that to `env.default` or try to find a way to not need it?

Pomax:
We're going to have to add this to `env.default` until we actually fix wagtail itself - it uses jQuery which in turn uses an unsafe script load mechanism for some of its work, and we need to figure out a way to make it stop that (at the very least as a codepatch for us, but ideally as a PR for wagtail to merge upstream as well)

Pomax:
@cadecairos it looks like appveyor isn't building the `.env` off of the `env.default`, is that something that's easy to fix?

cadecairos:
We should add any configuration we need to the appveyor.yml file

Pomax:
Also currently an issue: admin view "LIVE" links point to nonexistent resources, rather than the actual page URLs, so a page might be published as `.../wagtail/info` but the admin will insist the live view URL is `.../cms/pages/1/None`, which doesn't even make sense.

I've asked about this over on https://stackoverflow.com/questions/48957300/wagtail-admins-live-button-is-pointing-to-a-nonexisting-url

alanmoo:
We should retarget this to master now, correct?

Pomax:
Temporarily stood up https://github.com/mozilla/foundation.mozilla.org/pull/1107 in order to get around the fact that `git rebase -i master` with 40 commits is a week's worth of merge conflict resolution, and a good old "copy files, checkout master, checkout new branch, copy files back, hey look, a single commit without merge conflicts!" takes a normal amount of time.

Pomax:
closing in favour of #1107

PR #1107, opened 2018-02-27
[WIP] minisites in wagtail

Pomax:
# Wagtail 2.0 + auto-upgrade changes

models for:
- generic modular page
- minisite subclass
- campaign/opportunity pages
- generic CTA
- signup subclass
- petition subclass

block definitions for:
- link button
- image + text block
- bio block
- vertical spacing

also included:
- minisite template tag for sidebar menu
- minisite template tag for CTA block generation

# Testing:

Grab the branch, update your `.env` to have `ENABLE_WAGTAIL=True`, and then please download the following .zip file and unpack it to the root - it has a `network-api` dir that contains a sqlite database file as well as a media directory filled with images that together mirror the opportunity and campaigns that we currently have on Mezzanine. The admin login for this db is `admin/1Mozilla` and is accessed through [http://localhost:8000/cms](http://localhost:8000/cms).

said zip file : [network-api-db-and-images.zip](http://pomax.nihongoresources.com/downloads/network-api-db-and-images.zip) (28MB, updated 2018-03-20, 10:19 PST)

# Known issues so far:

- ~leadership training: video element does not seem to render~
- ihr: headings inside articles are not getting the appropriate class (should header have a class field?)
- ~ihr: ghost button doesn't have the ghost class (link button should have a button type selector)~
- ~network feedback: there is no iframe field it looks like~
- network 50: fragment-id hash linking seems impossible because "that's not a URL"...
  - open question on https://stackoverflow.com/questions/49415788/fragment-id-linking-in-wagtails-rich-text-content atm
- tech policy fellows: as blocks, images do not have a size indicator
- miti: table layout - do we want to allow tables in the richtext editor, or do we want a special block for this (did not copy icons)
- ~NSF: missing its video embed~
~- design: crude copy-paste atm, so buttons etc don't look right~
- ~project brief: iframe~
- global sprint overview: no table for griddin'
- ~global sprint tools and working open: no youtube embeds...~
- weblit: image/text does not do multiple paragraphs, spacing looks off too
- Any of the → pages: the header styling is super off
	- Should these be "models" that are registered as snippets, with a snippet block picker instead?
- log in, log back out, try to log back in, wrong redirect URL pointing to nonexistent profile. Fun

# Documentation

There is documentation! It's found over on [https://github.com/mozilla/foundation.mozilla.org/blob/wagtail-minisites/network-api/networkapi/minisites/README.md](https://github.com/mozilla/foundation.mozilla.org/blob/wagtail-minisites/network-api/networkapi/minisites/README.md)

# Known follow-up:

- vspace: offer only the set of boostrap values, as a pulldown menu.
- image grid: flex grid with up to X (x=3?) images+captions


PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,4 @@
+Related issue:
```
Pomax: I have no idea why this file keeps popping in and out of existence between branches. It's maddening.

env.default
```diff
@@ -34,7 +34,7 @@ XSS_PROTECTION=True
 # CSP config
 
 CSP_DEFAULT_SRC="'none'"
-CSP_SCRIPT_SRC='self' 'unsafe-inline' cdn.optimizely.com https://www.google-analytics.com/analytics.js https://c.shpg.org/352/sp.js
+CSP_SCRIPT_SRC='self' 'unsafe-eval' 'unsafe-inline' cdn.optimizely.com https://www.google-analytics.com/analytics.js
```
Pomax: if anyone knows what `https://c.shpg.org/352/sp.js` is, do let me know (we should probably have that explained in dev docs somewhere because just looking at the url's content isn't super indicative that this is a safe script at all)

Pomax: As secondary comment: it looks like unsafe-eval is necessary for the a jquery-ui datepicker. If we can fix that dependency we can hopefully get rid of `unsafe-eval` again. I'm not super keen on landing this PR until that's sorted.

network-api/networkapi/settings.py
```diff
@@ -131,16 +131,17 @@
 
     'wagtail.wagtailforms' if ENABLE_WAGTAIL else None,
     'wagtail.wagtailredirects' if ENABLE_WAGTAIL else None,
-    'wagtail.wagtailembeds' if ENABLE_WAGTAIL else None,
+    'wagtail.wagtailembeds',
```
Pomax: `manage <anything>` can't run without this allowed to load

network-api/networkapi/settings.py
```diff
@@ -131,16 +131,17 @@
...
     'wagtail.wagtailsites' if ENABLE_WAGTAIL else None,
     'wagtail.wagtailusers',
     'wagtail.wagtailsnippets',
     'wagtail.wagtaildocs',
-    'wagtail.wagtailimages' if ENABLE_WAGTAIL else None,
+    'wagtail.wagtailimages',
```
Pomax: same as above

network-api/networkapi/settings.py
```diff
@@ -131,16 +131,17 @@
...
     'wagtail.wagtailsearch' if ENABLE_WAGTAIL else None,
     'wagtail.wagtailadmin' if ENABLE_WAGTAIL else None,
     'wagtail.wagtailcore',
     'wagtail.contrib.modeladmin',
+    'wagtail.contrib.wagtailstyleguide' if ENABLE_WAGTAIL and DEBUG else None,
```
Pomax: a convenience app for devs, with some information which icon shorthand codes exist, etc.

env.default
alanmoo: It's share progress. @ScottDowne Just put it in.

network-api/networkapi/minisites/README.md
```diff
@@ -0,0 +1,142 @@
+# Minisites in Wagtail
+
+This document explains how the wagtail `minisites` app works with respect to the "networkapi" Django/Mezzanine instance.
+
+The base app has the following directory structure:
+```
+- minisites
```
gvn: Maybe call this `wagtail` instead?

network-api/networkapi/minisites/templates/minisites/blocks/image_text_block.html
```diff
@@ -0,0 +1,8 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<div>
+	{% block prefix %}{% endblock %}
```
gvn: We typically use 2 space indentation...should we add linting for that?

network-api/networkapi/minisites/README.md
Pomax: I'd rather keep the name related to what the app adds "into" django/wagtail, which right now is all the code necessary for minisites - open to better names based on the content it allows for, but "wagtail" is already the name of the CMS framework so I'd rather not call it that.

network-api/networkapi/minisites/templates/minisites/blocks/image_text_block.html
Pomax: It's not caught by `flake8` at the moment, so we might be able to do that if we can run an indent linting on `*.html` inside the `/network-api/networkapi/` dir. I guess the question would be what we're linting for: as code "on the python side" tabs are pretty PEP8-conformant, so it might actually be more worth it to see if we can make sure that final page HTML as generated by wagtail is nicely indented without changing the markup semantics?

network-api/networkapi/minisites/README.md
alanmoo: When we eventually move the entire site into the wagtail tree, will that be done in this app as well? If so, perhaps we should call this something like `wagtailPages`?

Pomax: Ideally we find a name that we don't then change a second time once mezzanine is out of the picture.

alanmoo: I'm inclined to agree, but I think "wagtail" here is more of a useful adjective, as "pages" isn't the most useful app name.

Pomax: true, but down the line we're removing that "wagtail" because it's all wagtail, and then we're left with "pages", which is a universally bad name. I'm perfectly happy punting on that and just using `wagtail-pages` but eventually we'll need a better name

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,363 @@
...
+class iFrameBlock(blocks.StructBlock):
+    url = blocks.CharBlock(
+        help_text='Please note that only URLs from white-listed domains will work.'
+    )
+    width = blocks.IntegerBlock(default=800)
+    height = blocks.IntegerBlock(default=450)
+
+    class Meta:
+        template = 'minisites/blocks/iframe_block.html'
+
```
alanmoo: This file's pretty big from a developer-understanding-at-a-glance-of-what-it-does standpoint. Can the above blocks be moved into their own file?

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,363 @@
...
+
+class MiniSiteNameSpace(ModularPage):
+    subpage_types = [
+        'CampaignPage',
+        'OpportunityPage',
+    ]
+
+    """
+    This is basically an abstract page type for setting up
```
alanmoo: Should this model be specifically defined as abstract? https://docs.djangoproject.com/en/2.0/topics/db/models/#abstract-base-classes

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,363 @@
...
+    minisite namespaces such as "campaign", "opportunity", etc.
+    """
+    def get_context(self, request):
+        """
+        Extend the context so that mini-site pages know what kind of tree
+        they live in, and what some of their local aspects are:
```
alanmoo: This is a nice function

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,363 @@
...
+        blank=True
+    )
+
+    newsletter = models.CharField(
+        max_length=100,
+        help_text='The (pre-existing) SalesForce newsletter to sign up for',
+        default='mozilla-foundation'
+    )
+
+    order = models.PositiveIntegerField(
```
alanmoo: What's this for?

network-api/networkapi/minisites/models.py
Pomax: yes

network-api/networkapi/minisites/models.py
Pomax: no, because it's "basically" an abstract page, except it gets used to make concrete pages for the CMS to have top level "opportunity" / "campaign" / (whatever) roots.

network-api/networkapi/minisites/models.py
Pomax: A required field for `SortableMixin` snippets. Without an ordering field, you cannot reorder them.

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,363 @@
...
+        they live in, and what some of their local aspects are:
+        """
+        context = super(MiniSiteNameSpace, self).get_context(request)
+        ancestors = self.get_ancestors()
+        root = next((n for n in ancestors if n.specific_class == self.specific_class), self)
+        context['mini_site_title'] = root.title
+
+        children = self.get_children().live()
+        has_children = len(children) > 0
+        ancestors = self.get_ancestors()
```
Pomax: duplicate lines 168/169

network-api/networkapi/minisites/templates/minisites/blocks/iframe_block.html
```diff
@@ -0,0 +1,9 @@
+{% load wagtailcore_tags %}
+
+<iframe
+    width="{{ block.value.width }}"
```
alanmoo: Is this going to make it easy to break responsiveness on pages?

Pomax: iframes the bane of responsiveness to begin with: they don't "fit" their content like a div or the like does, they *must* have a specified width and height attribute. We could set those in percentages but that's about the max in terms of responsiveness I know of.

network-api/networkapi/wagtailcustomization/static/wagtailadmin/js/modal-workflow.js
```diff
@@ -0,0 +1,130 @@
+/* A framework for modal popups that are loaded via AJAX, allowing navigation to other
```
alanmoo: Why's this here?

network-api/networkapi/minisites/models.py
alanmoo: Is sorting something we already have for these/why do we need it?

network-api/networkapi/minisites/templates/minisites/blocks/iframe_block.html
alanmoo: Right, I think a default of 100% would be useful. @gvn, thoughts?

network-api/networkapi/minisites/models.py
```diff
@@ -0,0 +1,360 @@
...
+        btn_class = self['styling']
+        if self['outline'] is True:
+            btn_class = btn_class.replace('btn-', 'btn-outline-')
+        return btn_class
+
+
+class LinkButtonBlock(blocks.StructBlock):
+    label = blocks.CharBlock()
+
+    # We use a char block because UrlBlock does not
```
alanmoo: Is UrlBlock the component that results in this UI? 
image

network-api/networkapi/minisites/models.py
Pomax: not sure, and excellent question. It's in there right now mostly "to keep our options open" but it's easy to take out before this PR is given the all clear if we don't see a usecase for having sorting available on CTA snippets.

network-api/networkapi/minisites/models.py
Pomax: No, it's just a model property block for "this string counts as a URL" which is rendered on pages using an `<a...>` element. It does not have anything to do with admin functionality.

network-api/networkapi/minisites/templates/minisites/blocks/iframe_block.html
gvn: Agree. 💯

Pomax: What minimal height should they be?

PR #1111, opened 2018-02-27
Adding updated style and icons to share buttons.

ScottDowne:
fixes a small part of https://github.com/MozillaFoundation/Design/issues/266


ScottDowne:
@gvn These are the updated button styles and icons.

PR #1126, opened 2018-03-01
[WIP] wagtail 2.0 + auto-upgrade changes

Pomax:
upgrading wagtail to 2.0


network-api/networkapi/management/commands/heroku_release.py
```diff
@@ -1,16 +1,16 @@
-from django.core.management.base import BaseCommand
```
Pomax: I think this is because I did a global CR+LF => LF line ending replacement... were CR+LF crucial here?

network-api/networkapi/management/commands/update_site_domain.py
```diff
@@ -1,11 +1,11 @@
...
-
-
-class Command(BaseCommand):
-    help = 'update the default django site domain to localhost:8000'
-
-    def handle(self, *args, **options):
-        site = Site.objects.get(id=1)
-        site.domain = 'localhost:8000'
-        site.save()
+from django.core.management.base import BaseCommand
```
Pomax: I think this is because I did a global CR+LF => LF line ending replacement... were CR+LF crucial here?

network-api/networkapi/utility/middleware.py
```diff
@@ -1,30 +1,30 @@
-from django.http.response import HttpResponseRedirectBase
```
Pomax: I think this is because I did a global CR+LF => LF line ending replacement... were CR+LF crucial here?

network-api/networkapi/management/commands/heroku_release.py
cadecairos: probably not

PR #1149, opened 2018-03-05
redirected /fellowship to /fellowships

mmmavis:
As suggested in https://github.com/mozilla/foundation.mozilla.org/pull/996#pullrequestreview-100604752

### DIFF
--- a/network-api/networkapi/urls.py
+++ b/network-api/networkapi/urls.py
@@ -41,13 +41,11 @@ urlpatterns = list(filter(None, [
     url(r'^wagtail/', include(wagtail_urls))
     if settings.ENABLE_WAGTAIL else None,
 
+    url(r'^fellowship/(?P<path>.*)', RedirectView.as_view(
+        url='/fellowships/%(path)s',
+        query_string=True
+    )),
     url(r'^soc/', include('social_django.urls', namespace='social')),
-
-    # Don't remove this redirect until Fellowships pages are live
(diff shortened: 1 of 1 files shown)


network-api/networkapi/urls.py
```diff
@@ -41,14 +41,12 @@
     url(r'^wagtail/', include(wagtail_urls))
     if settings.ENABLE_WAGTAIL else None,
 
+    url(r'^fellowship/(?P<path>.*)', RedirectView.as_view(
+        url='/fellowships/%(path)s',
+        query_string=True
+    )),
```
mmmavis: not sure if I need to add an extra line in between these 2 lines?

network-api/networkapi/urls.py
```diff
@@ -41,14 +41,12 @@
...
     url(r'^soc/', include('social_django.urls', namespace='social')),
```
cadecairos: please drop this line too, it's a dupe of line 33

network-api/networkapi/urls.py
cadecairos: nope, not unless pep8 says we need it

mmmavis: weird, thought i deleted it 👀

network-api/networkapi/urls.py
mmmavis: weird, thought i deleted it 👀

PR #1157, opened 2018-03-06
Fellowships improvements

mmmavis:


source/js/fellowships.js
```diff
@@ -60,12 +58,14 @@ function groupFellowsByAttr(attribute, fellows) {
   let fellowsGroup = {};
 
   fellows.forEach(fellow => {
-    let attr = fellow[attribute].toLowerCase();
+    let attr = fellow[attribute];
```
mmmavis: i wonder if i should `try.. catch` here instead

cadecairos: nah, I think the early return does the trick well

Issue #1199, opened 2018-03-16
[Fellowship Mini Site QA] Fellowship Directory


taisdesouzalessa:
### Issue
Currently, in our [Directory Page](https://foundation.mozilla.org/fellowships/directory/), once users click on the button "See Fellows for Previous Years" they are taken to [a page](https://foundation.mozilla.org/fellowships/directory/archive/) that has all Fellows, from all years, including 2018. The problem is that this behavior doesn't match the expectation of the title "See Fellows from Previous Years"

### Solution
Change the [archive page](https://foundation.mozilla.org/fellowships/directory/archive/) to display Fellows from 2017 and before that, as proposed in [the comp](https://invis.io/ZJ9YTAGEA#/276734822_v2-FellowshipsUI_-_Fellows_Directory_Specific_Program_Page).

cc: @alanmoo @community-impact for visibility

gvn:
This is on prod now if you want to verify, @taisdesouzalessa !

taisdesouzalessa:
It looks great!! Yay, thank you so much @gvn :)

PR #1200, opened 2018-03-16
Change the archive page to display Fellows from 2017 and before that …

gvn:
Related issue: #1199

### DIFF
--- a/source/js/fellowships.js
+++ b/source/js/fellowships.js
@@ -6,9 +6,11 @@ import ReactDOM from 'react-dom';
 import Person from './components/people/person.jsx';
 import LoadingIndicator from './components/loading-indicator/loading-indicator.jsx';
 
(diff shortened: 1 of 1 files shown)


source/js/fellowships.js
```diff
@@ -136,7 +138,7 @@ function renderFellowsOnDirectoryByTypePage() {
   getFellows({}, fellows => {
     let fellowsByYear = groupFellowsByAttr(`program_year`, fellows);
 
-    let sections = Object.keys(fellowsByYear).sort().reverse().map(year => {
+    let sections = Object.keys(fellowsByYear).filter((year) => parseInt(year, 10) !== CURRENT_YEAR).sort().reverse().map(year => {
```
alanmoo: Can you make this `< CURRENT_YEAR`?

gvn: That is effectively the same thing though?

Issue #1202, opened 2018-03-19
[Fellowship Mini Site] Change button copy on lauch + helper text


taisdesouzalessa:
Once the applications launch, do the following changes on this page:
https://foundation.mozilla.org/fellowships/apply/

- [x] change button copy to "Apply to be a fellow today" 
- [x] button linking to Fluxx: https://mozilla.fluxx.io/user_sessions/new)  
- [x] add this helper copy: "Applications are due by 5 pm Eastern Time on April 20th, 2018. `<strong>` Make sure to register by April 18th `</strong>`."

@alanmoo adding you here to triage this issue. If needed, I can do those changes. I have the bandwidth.

@community-impact and @auremoser could you let us know here when this copy should change? 

cc: @xmatthewx adding you to the loop

auremoser:
Looks great, thank you @taisdesouzalessa , if we can bold the April 18th date that would be great. I just don't want people to miss it!

taisdesouzalessa:
good point, Aurelia :). Added this note to the original ticket :)

alanmoo:
@taisdesouzalessa If you want to open a PR labeled "Do not merge yet" you're welcome to take it on. If you can't assign it to Gavin.

taisdesouzalessa:
Sure, I can do it @alanmoo between today and tomorrow.

taisdesouzalessa:
@auremoser @community_impact here is how the button with the helper text will look like. I think having 2 dates there may be a bit confusing from the user's perspective. With the current copy, I made the whole "April 18th" line bold to try to alleviate that. Do we need to have those 2 dates?

<img width="822" alt="image" src="https://user-images.githubusercontent.com/22157921/37673783-5c4a0cfe-2c2e-11e8-9ac4-eed3a5ccb049.png">

annmarie123:
@taisdesouzalessa Yes we do as applicants who are registering for the first time, need to be approved by staff, thus we do not want last minute "help" emails about their application at 11pm Friday April 20th.

taisdesouzalessa:
Thanks for the context @annmarie123 - leaving it as is then :).

taisdesouzalessa:
@alanmoo PR is ready, assigned Gavin for review.

auremoser:
+1 to Ann Marie, it is confusing, but the reality is that if they don’t register by the 18th we probably won’t invite them to Fluxx in time to submit by the 20th. I’ve prepped a canned email to respond to any confusion!

[...]

PR #1205, opened 2018-03-20
Preparation of button for when applications open

taisdesouzalessa:
Related issue:  #1202
Summary: button copy changes + added helper text when we launch Fellowships application.

This is how this button should look like:
image


taisdesouzalessa:
This is ready to merge now and be tested on staging - could we do that @gvn @alanmoo ?

gvn:
It's merged and should be on staging in 5-10 minutes...

PR #1231, opened 2018-03-29
Deficiency fixes

Pomax:
Fixes template mismatching between current foundation template and what we had when wagtail got stood up.

- fixes side nav positioning
- fixes signup placement
- fixes two-column for pages like equifax
- made vertical spacer set up a div with bootstrap spacing classes instead of some rem values

Also adds in the django-debug-toolbar when `DEBUG=True` and `DISABLE_DEBUG_TOOLBAR=False` in the env


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/bootstrap_spacer_block.html
```diff
@@ -0,0 +1 @@
+<div class="mt-{{value.size}}"></div>
```
Pomax: this missing newline is intentional, and one of those times where github's insistence on it is plain wrong.

network-api/networkapi/settings.py
```diff
@@ -33,6 +33,7 @@
     CORS_WHITELIST=(tuple, ()),
     DATABASE_URL=(str, None),
     DEBUG=(bool, False),
+    DISABLE_DEBUG_TOOLBAR=(bool, False),
```
alanmoo: Not a blocker, but I wonder if we're approaching a point where we should have an "environment" variable that sets a few of these automatically (set to "development" or "production")

Pomax: That would be `DEBUG`. if you turn that on and don't set anything else, the debug toolbar will work just fine. This env var is for explicitly *disabling* the debug toolbar, despite DEBUG being true (for instance, it's getting in the way of some particular design/testing work)

PR #1237, opened 2018-04-02
New Homepage

gvn:
# WIP PLZ NO MERGE 😰


source/js/components/home-news/home-news.jsx
```diff
@@ -20,8 +20,7 @@ export default class HomeNews extends React.Component {
           <h5 className="mb-2">
             <a href={item.link} className={featured ? `h4-medium-black` : `h5-black`}>{item.headline}</a>
           </h5>
-          { item.author && <p className="italic-black mb-1">by {item.author}</p> }
-          <p className="small-gray">{moment(item.date, `YYYY-MM-DD`).format(`MMMM YYYY`)}</p>
+          <p className="small-gray">by { item.author } on {moment(item.date, `YYYY-MM-DD`).format(`MMMM YYYY`)}</p>
```
alanmoo: Author isn't a required field, so we should still check for its existence.

source/sass/views/home.scss
```diff
@@ -40,3 +40,9 @@
     width: 50%;
   }
 }
+
+#view-home {
+  .h1-white {
+    font-size: 48px;
```
alanmoo: This feels like not the right solution, especially since this text isn't white. Is it the right class to begin with?

source/pug/views/home.pug
```diff
@@ -5,36 +5,46 @@ block masterParams
...
+      .col.mb-2
+        hr.hr-gradient
+    .row
+      .col-md-6
+        h4.h2-headings-white Trending projects
+      .col-md-6.d-flex.justify-content-end
+        a.cta-link(href=`https://www.mozillapulse.org/add`) Add your project
+    .row
+      .col
+        .pulse-project-list(for="", size="6", checked="true")
```
alanmoo: Let's use the same cropping behavior that the Pulse client uses so we don't get images that are too tall. See https://www.mozillapulse.org/entry/588 vs what shows up on the review app.

source/sass/views/home.scss
gvn: This is the correct class for the end result...we're going to change the class names post-Wagtail migration so that they are color agnostic.

source/sass/global.scss
```diff
@@ -14,13 +14,13 @@ body {
...
 
   // Homepage
-  &#view-home .wrapper-2 {
-    background:
-      linear-gradient(to left, rgba(61, 254, 253, 1) 0%, rgba(183, 212, 63, 1) 100%),
-      url(../_images/hero-image-crop@2x.jpg) center top / auto 100% no-repeat,
-      url(../_images/bg-repeat-crop@2x.jpg) center top / 2px 100% repeat-x;
-    background-blend-mode: overlay, normal, normal;
-  }
+  // &#view-home .wrapper-2 {
```
gvn: Commenting out for use in upcoming work...

PR #1242, opened 2018-04-02
add figure + figure grid components

Pomax:
This will probably require a CSS update to make things in `.figure-grid` properly spaced etc.

The flexboxing is hardcoded atm and that's pretty bad!


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/figure_grid_block.html
```diff
@@ -0,0 +1,9 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<div class="figure-grid" style="display: flex; flex-direction: row; ">
```
Pomax: this is pretty terrible, but in order to "see it work" I added a hardcoded style here, rather than add in a rule for `.figure-grid` and descendants to our CSS file and probably getting the "where to put that" wrong.

Issue #1234, opened 2018-04-02
Build updated home page


kristinashu:
Based on design in https://github.com/mozilla/foundation.mozilla.org/issues/1138

### High-level list of things to implement [these designs](https://redpen.io/p/mz472b7821ed66aed4):

**Nav**
- gradient bar background
- remove Signup button (maybe replace with [Donate](http://donate.mozilla.org/))

**Banner**
- remove background
- reduced width of text
- add CTA button
- option for in body text links
- add image

**News**
- Update title copy
- Simplify News date and byline #1217

**Network Leaders**
- remove entire section (title, images, text, button)

**Happening now**
- update title copy
- 2 up feature boxes (image, white box over top (same as overlap as featured news) section title in small grey, title, body copy)
- 3 down feature list (image, section title in small grey, title, body, dividing line)

**Projects
- add section title
- pull Featured Pulse projects

**Quote**
- add quote style - might want a different quote at the bottom of different pages

**Email Signup**
- remove "Get Connected" title
- background image is full width and reduced height
- remove icon

**Mobile**
- note full width images and some content is hidden

kristinashu:
I believe this is the only asset you need (other ones can use images from current news, Get Involved featured, pulled from Pulse):
![aadhaar](https://user-images.githubusercontent.com/1068783/38208677-bc47aa70-3666-11e8-8497-093193a7964a.jpg)

PR #1254, opened 2018-04-03
happening now component updates

gvn:
Related issue: #1234

### DIFF
--- a/source/js/components/highlights/highlights.jsx
+++ b/source/js/components/highlights/highlights.jsx
@@ -9,37 +9,44 @@ export default class Highlights extends React.Component {
   }
 
   render() {
-    // Limit to 2 additional highlights
-    let highlights = this.props.data.slice(1, 3).map((item, index, array) => {
+    let superHighlights = this.props.data.slice(0, 2).map((item, index) => {
       return (
-        <div key={index}>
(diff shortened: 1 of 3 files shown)


source/js/components/highlights/highlights.jsx
```diff
@@ -9,37 +9,44 @@ export default class Highlights extends React.Component {
   }
 
   render() {
-    // Limit to 2 additional highlights
-    let highlights = this.props.data.slice(1, 3).map((item, index, array) => {
+    let superHighlights = this.props.data.slice(0, 2).map((item, index) => {
```
mmmavis: should we create a `const` for `2`?

source/js/components/highlights/highlights.scss
```diff
@@ -0,0 +1,7 @@
+.highlights {
+  .placeholder {
+    height: 0;
+    padding-top: 57%;
+    background: linear-gradient(to right, #3dfefd, #b7d43f);
```
mmmavis: Can we create vars for the 2 colours here? They are used elsewhere as well.

source/js/components/highlights/highlights.jsx
gvn: I mean...we could, but it's always going to be 2.

PR #1264, opened 2018-04-04
Adding quote component

gvn:
Related issue: #1234


gvn:
(Approved by Kristina IRL)

gvn:
Note: There will be different copy coming in, but code is good to review...

PR #1285, opened 2018-04-05
Build a better image block

Pomax:
This replaces the default wagtail image component with one that lets users pick a size (icon, small, medium, large, mapping to 100, 200, 400, and 800px constrained w/h respectively), with the same alignment options (left/right/full) but then straight in the admin view rather than as part of the picker.

screen shot 2018-04-05 at 2 35 18 pm

This affects:
- image block
- figure block (+figure grid block)

Not affected:
- image/text combo block


Pomax:
Fair point, let's have a talk about that. At least we should be able to force that issue entirely through CSS, so no model changes should be necessary in any follow-up we decide on.

kristinashu:
Here's how I see images being used on the site. The four different sizes seem to work as long as text is next to the two smaller sizes. 

ih copy 2x

Pomax:
the (image+text) block is a not affected by this patch right now, we only built that to accommodate a single image size  (due to using bootstrap classes to size the image vs. text column)

kristinashu:
Ok, I think it's ok to have several images sizes.

PR #1290, opened 2018-04-06
Make homepage have the right body class

alanmoo:


network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
```diff
@@ -1,5 +1,6 @@
-{% extends "pages/base-compiled.html" %} {% load wagtailcore_tags wagtailimages_tags %} {% block content %}
-
+{% extends "pages/base-compiled.html" %} {% load wagtailcore_tags wagtailimages_tags %} 
+{% block bodyID %}home{% endblock %}
```
Pomax: let's put some whitespace around this, so it's clear that it's implementing one of the blocks pre-specified by base-compiled

Issue #1284, opened 2018-04-05
Nav naming and ordering


kristinashu:
### Nave updates:
- Remove "Project" page
- Rename "Get Involved" to "Initiatives"
- Rename "Upcoming" to "Participate"
- Rename "About" to "About Us"
- Add "Internet Health" and link to https://internethealthreport.org/ (open in new tab)

### Reorder to:
- Initiatives
- Participate      
- Internet Health
- People      
- About Us

kristinashu:
w00t

PR #1291, opened 2018-04-06
Nav naming and ordering #1284

gvn:
Related issue: #1284


source/js/components/nav/nav.jsx
```diff
@@ -61,11 +61,11 @@ export default class PrimaryNav extends React.Component {
             <div className="col-sm-auto align-self-end">
               <ul className="px-5 menu-nav">
                 <li><a className="home" href="/">Home</a></li>
+                <li><a className="get-involved" href="/get-involved">Initiatives</a></li>
+                <li><a className="upcoming" href="/programs/upcoming">Participate</a></li>
+                <li><a className="upcoming" href="https://internethealthreport.org" target="_blank" rel="noopener noreferrer">Internet&nbsp;Health</a></li>
```
alanmoo: Should this "upcoming" class be removed?

gvn:
Without the wrap it breaks the layout because the line length gets too long.

PR #1292, opened 2018-04-06
Add basics of a primarypage

alanmoo:
The template needs more work but once it's up, the About page work can start.


network-api/networkapi/wagtailpages/models.py
```diff
@@ -310,6 +310,14 @@ def __str__(self):
...
 
 
+class PrimaryPage(Page):
+    body = StreamField(base_fields)
+
+    content_panels = Page.content_panels + [
+        StreamFieldPanel('body'),
+    ]
+
+
```
Pomax: let's also add a `parent_page_types = ['Homepage']` so that people can't spuriously add these under other page types

PR #1303, opened 2018-04-06
Easy way out for building primary pages

alanmoo:


Pomax:
I've updated this PR in a way that I think we can move forward with. The static content's been moved into its own dir (handy!) and the primary page template extends off of modular_page (the class, however, does not extend from the ModularPage class, because of migration complications I don't want to solve right now.. it's a django bug apparently?)

nested primary pages spawn a horizontal menu (not mobile-friendly yet) similar to opportunities/campaigns, with the top level document always being called `Overview` in the menu now. 

This PR also fixes the "highlight page you're on" bug (simple typo on the menu tag code, so fixed for both).

IMPORTANT FOR TESTING: use [this handy prepared database](https://github.com/mozilla/foundation.mozilla.org/files/1900345/db.zip) (1MB .zip file), which has admin account `admin/1Mozilla` and consists of a homepage with the static pages and an about page that has stubs for all its sub pages, as well as a single opportunity page with stub children to verify that the menu works correctly both on primary pages as well as minisite pages.

network-api/networkapi/wagtailpages/templates/wagtailpages/primary_page.html
```diff
@@ -1,13 +1,19 @@
-{% extends "pages/base-compiled.html" %} {% load wagtailcore_tags wagtailimages_tags %}
+{% extends "./modular_page.html" %}
+{% load wagtailcore_tags wagtailimages_tags primary_page_tags %}
 
-{% block bodyID %}primary{% endblock%}
+{% block bodyID %}{{ page.title }}{% endblock%}
```
alanmoo: The main goal of this tag is for CSS, so I think it's better to keep this to the template name rather than a piece of CMS data. Was there a specific reason for this change?

network-api/networkapi/wagtailpages/templates/wagtailpages/static/initiatives_page.html
```diff
@@ -1,8 +1,6 @@
-{% extends "./primary_page.html" %}
+{% extends "../primary_page.html" %}
 {% load wagtailcore_tags wagtailimages_tags %}
 
-{% block bodyID %}initiatives{% endblock%}
-
```
alanmoo: Similar to above; This is for CSS that might exist now to be applied to initiatives, but as we move forward we should only target CSS by page template, not page title.

PR #1307, opened 2018-04-09
Allow PrimaryPage type as child of PrimaryPage

alanmoo:
For sections like “About”


Pomax:
superceded by the bigger primary page PR

PR #1320, opened 2018-04-12
Introducing Pipenv

patjouk:
In answer to points raised in the developers' survey I ran during the work week, I propose we use [pipenv](https://docs.pipenv.org) instead of pip and virtualenv.

## Major changes

- `requirements.txt` are replaced by `pipfile` and `pipfile.lock`. You will need to commit both of them after installing or updating dependencies.
- You won't be using pip but pipenv instead.
- You won't have to manually create virtualenvs anymore.
- Except for a few limited ones, dependencies won't be pinned anymore. Updating dependencies just got easier :)

## What you need to install

*Pipenv:*
MacOs:
`$ brew install pipenv`

Others:
`$ pip install pipenv`

*Inkove*

`$ pip install invoke`

### Pipenv
- `$ pipenv install --dev`: Creates your project's virtualenv and installs both required and dev dependencies.

- `$ pipenv install <package>`: Installs the latest version of that package and adds it to the pipfile. It automatically triggers a `$ pipenv lock` and update the lock file.

- `$ pipenv run python [command]`: Automatically run that python command in your virtualenv. Ex: `$ pipenv run python network-api/manage.py runserver`.

- `$ pipenv update`: update unpinned dependencies. If you want to check which dependencies needs to be updated before installing those, run `$ pipenv update --outdated`

- `$ pipenv graph`: Shows a graph of your installed dependencies

- `$ pipenv shell`: Activates your virtualenv. Small gotcha: pipenv automatically loads your `.env` file, but only once. If you update your `.env` file, you will have to exit your virtualenv and start it again.

The full documentation is [available here](https://docs.pipenv.org/).

### Invoke

Pipenv commands might seem long. I added a few invoke tasks to make everyone life easier:

- `$ inv -l`: list available invoke tasks
- `$ inv makemigrations`: Creates new migration(s) for apps
- `$ inv migrate`: Updates database schema
- `$ inv runserver`: Start a web server
- `$ inv setup`: Prepare your dev environment after a fresh git clone
- `$ inv test`: Run tests

Check the readme for more details on both pipenv and invoke.

### Does the pipfile have a script section like in a `package.json` file?

No, as [explained here](https://github.com/pypa/pipenv/issues/826). That's also why I wrote some invoke tasks.

### Why do some dependencies are pinned?

- Django is pinned to the latest patch (1.11.12) because of Mezzannine. When we move to Django 2, we should unpin it. Meanwhile, let's keep an eye on the latest security patches and manually update.
- Mezzannine stuff: unpinning it creates an issue with Django 1.11. I tried to figure out what was going on and realized that we will remove it at some point and stop digging my nice rabbit hole ^^
- Psycopg2: "it's complicated". The short answer is I want to see what happens when they land 2.8. The long answer [is here](http://initd.org/psycopg/articles/2018/02/08/psycopg-274-released/).
- whitenoise: we're using the beta version. If I use the wildcard, it will go back to the latest stable version (3.3.1) [because of pip](https://pip.pypa.io/en/stable/reference/pip_install/#pre-release-versions). When the v4 is out, we can unpin it.


requirements.txt
```diff
@@ -13,7 +13,7 @@ filebrowser_s3==1.0.0
 wagtail==2.0
 
 # Social auth is useful
-social-auth-core==1.1.0
+social-auth-core
```
cadecairos: How come pipenv removed the version here, but not elsewhere?

patjouk: Probably my fault, not pipenv >.<

gvn:
This seems pretty useful. We should update the venv/python-y tasks in `package.json` if we switch to this.

See:

https://github.com/mozilla/foundation.mozilla.org/blob/master/package.json#L20
https://github.com/mozilla/foundation.mozilla.org/blob/master/package.json#L25

gvn:
Oh, also we should update the README in this PR... ✨

patjouk:
Of course :D I can't do it now because it depends on the workflow we adopt.

PR #1323, opened 2018-04-12
Fix homepage banner in wagtail

### LINKED ISSUE #1322: Wagtail homepage header text styling is incorrect
The description text on the header of the homepage is using the wrong sizes- looks like it's due to some extra markup being added by Wagtail.

https://foundation.mozilla.org/
vs 
https://foundation.mozilla.org/wagtail

My guess is we need to change `p` elements to inherit font sizes, though that might have global implications. In the interest of getting over to wagtail, maybe we should target that specific `p` to say `font-size: inherit; line-height: inherit`.

alanmoo:
Closes #1322 and ups the resolution on that image so it looks better on retina displays.

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
@@ -5,14 +5,14 @@
 
 {% block content %}
 <div class="container">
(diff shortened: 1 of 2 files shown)


alanmoo:
I'm not in love with this solution but it seems the least risky for impacting other pages right now.

Issue #1321, opened 2018-04-12
Update heading class names on primary wagtail pages


kristinashu:
This is a blocker for launch.

Need to update heading class names on primary wagtail pages. 

Headings should match our [style guide](https://foundation.mozilla.org/style-guide/).

Example: wagtail looks like https://foundation.mozilla.org/wagtail/participate/
<img width="1205" alt="screen shot 2018-04-12 at 9 12 14 am" src="https://user-images.githubusercontent.com/1068783/38693158-f72caeac-3e3a-11e8-9039-6e855c86d8dd.png">

Should look like original pages https://foundation.mozilla.org/participate/
<img width="1265" alt="screen shot 2018-04-12 at 10 19 38 am" src="https://user-images.githubusercontent.com/1068783/38693183-0e3fe032-3e3b-11e8-9809-227d00e34f71.png">

cc @alanmoo @taisdesouzalessa

alanmoo:
I believe I just addressed this in #1346

PR #1324, opened 2018-04-12
switching to new classnames - #1321

gvn:
Related issue: #1321


gvn:
This won't show up on the staged PR, so for design review see this screenshot:

screen shot 2018-04-12 at 11 45 08 am

kristinashu:
Looks good as long as that is reflected on all the primary pages, not just Participate.

alanmoo:
So which pages need to be compared? I see a few more on "Initiatives" but I think that's it.

kristinashu:
@alanmoo probably all the wagtail pages: Initiatives, Participate, People, About Us, (News looks ok I think). Not sure what's happening with all the Opportunity and Campaign pages.

alanmoo:
Opportunity and Campaign will likely need updates in the CMS editor, not in the code.

PR #1332, opened 2018-04-12
button class fixes

### LINKED ISSUE #1314: Fix class for ghost buttons
Seems like the markup is emitting `btn-outline-ghost` but the CSS class name is `btn-ghost`.

@Pomax @gvn

Pomax:
This changes the `error` class (which doesn't exist) to `danger` (which does exist), and ignores the outline checkbox for the `ghost` class, which is already an outline-styled button.

Fixes #1314 

image


gvn:
Some of these button styles are coming from mofo-bootstrap and not our current Foundation style guide, which only has these 3:

screen shot 2018-04-12 at 3 08 13 pm

alanmoo:
Yeah, we need to clarify what we're using mofo-bootstrap for these days; but I think that's an issue for another ticket.

PR #1333, opened 2018-04-12
fix the paragraph margin CSS in rich-text content

### LINKED ISSUE #1331: Wagtail QA: Bullet points
From @kristinashu  in #1325 

- Spacing after bullet points should be the same or larger than before them
Screen cap from production looks confusing because bullets look grouped with content bellow, not above
image

Pomax:
Fixes #1331 

image


kristinashu:
Much better!

PR #1334, opened 2018-04-12
Ensure unclassed headings have the proper fonts

### LINKED ISSUE #1329: Wagtail QA: Heading Styles
From @kristinashu  in #1325 

- will [this](https://github.com/mozilla/foundation.mozilla.org/pull/1324#issuecomment-380906497
) fix the heading styles here too?

alanmoo:
I think this fixes #1329

### DIFF
--- a/source/sass/type.scss
+++ b/source/sass/type.scss
@@ -80,6 +80,7 @@ a {
   }
 }
 
+h1,
 .h1-heading {
   font-family: "Zilla Slab";
   font-weight: 400;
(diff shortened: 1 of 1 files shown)


gvn:
Fine with me as long as these are the defaults that Design wants to roll with...

alanmoo:
@kristinashu @taisdesouzalessa Does this make sense? The richtext field outputs raw `h2, h3, h4` etc so these are the styles that will be applied to those elements when used in those fields.

kristinashu:
@alanmoo sorry, what do you mean? What does raw `h2, h3, h4` end up looking like on the site?

kristinashu:
Ideally a user would be able to chose a heading or paragraph style in the editor and then it would look like one of these on the site:

image

The names don't have to align. For example it could just be `h3` in the editor but on the site it should look like `.h3-heading` style.

PR #1338, opened 2018-04-12
Narrow content when needed

### LINKED ISSUE #1336: Wagtail QA: copy width
All copy heavy pages (i.e ones that aren't bespoke) should use 8 out of the 12 bootstrap columns. 

Example: copy width on [Overview](https://foundation.mozilla.org/wagtail/about/), [Trademarks](https://foundation.mozilla.org/wagtail/about/trademarks/), [Public Records](https://foundation.mozilla.org/wagtail/about/public-records/), and any future pages that are paragraphs of copy should be narrower. Should look like https://foundation.mozilla.org/fellowships/apply/)

Pomax:
This adds a "narrowed page content" checkbox to the "Settings" tab for all PrimaryPage and ModularPage page (sub)types, defaulting to `False`.

image

When checked, content on the page is set using `col-8` rather than `col-12`.

With the setting unchecked:

image

With the setting checked:

image

Fixes #1336


kristinashu:
Amazing!

alanmoo:
Updated to support the refreshed sub-nav. @kristinashu please confirm this is right: 
image

kristinashu:
@alanmoo looks good with nav! Did you update mobile? I can't actually preview it, right?

PR #1340, opened 2018-04-13
Streamfield quote

### LINKED ISSUE #1326: Create "quote" streamfield component
Like the one on the homepage, but so that we can insert it in a streamfield. This is a request per @jessevondoom before launching the about page.

alanmoo:
Closes #1326 
image

Single quote for now, but the admin accepts multiple, with the disclaimer that they won't be used yet.


kristinashu:
I believe we decided to use spacers and to not build the margin in here so ignore my above comment.

Issue #1316, opened 2018-04-11
update wagtail Petition class


Pomax:
it has diverged from what's in use in Mezzanine at the moment, and needs the twitter/facebook/email share options.

PR #1383, opened 2018-04-17
update petition cta model to reflect current mezzanine model

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/1316


network-api/networkapi/wagtailpages/models.py
```diff
@@ -242,6 +242,24 @@ class Petition(CTA):
         blank=True
     )
 
+    share_twitter = models.CharField(
```
gideonthomas: might also want to add `editable=False` to `share_link` and `share_link_text` (and also remove it from the admin) so that we aren't allowed to use that field anymore but at the same time maintaining backwards-compatibility for existing data (+ add a note as to why we're doing this)

Issue #1382, opened 2018-04-17
[QA Opportunity Pages] Add page anchors


kristinashu:
Works well on Mezzanine https://foundation.mozilla.org/opportunity/network-50/ but is missing on Wagtail.

kristinashu:
Decided to not bother with anchors!

PR #1397, opened 2018-04-17
Add an AnchorBlock for fragmentID linking

Pomax:
This PR adds an "anchor" component to the streamfield set, allowing content editors to add explicit fragmentids to pages for navigation purposes. The block effects an `<a>` element with an `id` attribute corresponding to the user specified fragment id, which can then be used for `#fragmentid` linking. This element has been given CSS to effect zero on-page width and height.

This solves the problem of wagtail not having a way to set up fragmentids _at all_ but does mean that any page with a fragment id needs to split its content at that point. I'm okay with that, but that also in itself is a hint that there might be something structural going on with the on-page data (for instance, lists of bios, some kind of picture/image block, etc) that might be better solved by adding some kind of automated anchor generation to specific elements.

This component feels "necessary to overcome wagtail limitations right now", but not a long term solution.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/1382


network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -2,6 +2,16 @@
 from wagtail.images.blocks import ImageChooserBlock
 
 
+class AnchorBlock(blocks.StructBlock):
+    fragmentid = blocks.CharBlock(
+        help_text='Use this to place an explicit fragmentid on the page for linking purposes.'
```
alanmoo: This text is too technical. Use something like `The text entered here will enable direct linking to this part of the page, using a # before it. For example, "foo" allows you to use <a href="#foo">`

Pomax:
@alanmoo We need more precise terminology here because I don't know we're talking about when we're saying "users". Let me suggest:

- users = visitors of the website
- staff = people who generate content on our site (ignoring employment status)
- devs = the people who maintain wagtail

alanmoo:
I'm referring to staff

source/sass/wagtail.scss
```diff
@@ -1,4 +1,14 @@
 .cms {
+  a.anchor-block {
+    /*
+      special purpose element for fragmentid anchoring,
+      something wagtail does not support natively.
+    */
+    display: inline-block;
+    width: 0;
+    height: 0;
```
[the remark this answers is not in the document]
Pomax: gah, that should have been block, not inline-block. Good catch O_o

network-api/networkapi/wagtailpages/customblocks.py
Pomax: that's a very long text for a tiny, tiny block. Can we condense that? (maybe refer to somewhere in the "how do I wagtail"?)

alanmoo: Sure. Not a "use this" just a "like this" suggestion

Pomax:
I think in this case we're running into a limitation of wagtail. Some of our pages are "oodles of bio blurbs" which we then want an auto-generated ToC for. @kristina are there other usecases where explicit `#...` linking would be necessary?

Pomax:
we are not going down this long, dark road. This is not a tribute.

PR #1428, opened 2018-04-19
Create a pulse entry listing component

Pomax:
This adds a pulse entry listing component for pages to use in stream content, with user fields for:

- search terms
- max number of results
- reverse chronological order
- from-featured-entries-only

image


source/js/main.js
```diff
@@ -258,37 +258,37 @@ let main = {
...
-          <SingleFilterFellowList
-            env={env}
-            filterType={target.dataset.filterType}
-            filterOptions={target.dataset.filterOptions.split(`,`)}
-            selectedOption={target.dataset.selectedOption}
-          />, target
-        );
-      });
-    }
+    let singleFilterFellowList = Array.from(
```
Pomax: updated because it used the exact same pattern as `pulseProjectList` and changing one but not the other would have just made the code inconsistent.

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -209,3 +211,49 @@ class Meta:
...
+    max_number_of_results = blocks.IntegerBlock(
+        min_value=0,
+        default=0,
+        required=False,
+        help_text='The maximum number of results to fetch (use 0 for "no maximum")',
+    )
+
+    reverse_chronological_listing = blocks.BooleanBlock(
+        default=False,
+        help_text='Check this box to select oldest entries first.',
```
alanmoo: This is confusing. If a box labeled "reverse chronological listing" is checked, I'd expect to see newest first.

PR #1435, opened 2018-04-20
Make wagtail redirects work

alanmoo:


network-api/networkapi/urls.py
```diff
@@ -77,7 +77,6 @@
 ]))
 
 
-handler404 = RedirectView.as_view(url='/errors/404')
 handler500 = 'mezzanine.core.views.server_error'
```
Pomax: let's file a follow-up to remove this line too. Mezzanine should not be handling 500 errors anymore, especially if they come from Wagtail =D

Issue #1407, opened 2018-04-18
[QA Opportunity] Figure Grid isn't working


kristinashu:
Figure Grid component isn't working on Wagtail:
https://foundation.mozilla.org/wagtail/opportunity/global-sprint/
<img width="431" alt="image" src="https://user-images.githubusercontent.com/1068783/38907068-d8d78278-426e-11e8-9cd6-c5b8611d5809.png">

It should look like it does on Mezzanine:
https://foundation.mozilla.org/opportunity/global-sprint/
<img width="882" alt="image" src="https://user-images.githubusercontent.com/1068783/38907076-e4ce9fc6-426e-11e8-9da1-6d0863fcbd11.png">

Slack convo:
P:
that's actually a good case of "We have mutually exclusive needs"
you landed a PR that allows the flexbox to wrap, but that means with images set to original size, it wraps immediately
so it might be wise to undo that wrap, and say "don't use more than x figures. If you need more, use another grid"
A:
Mmm. Yeah I was landing for about page, recognized that need may be different elsewhere
P:
or use more if you need like 10 icons next to each other of course
is the about page use-case fixable by "using another flexgrid" instead?
A:
This is something gvn should run
Note that specific use of flex grid on the leadership page will eventually go away as we move to better profile cards. Not sure about that need for flexgrid elsewhere

kristinashu:
Updated it to 200x200. It's ok but not great ...

<img width="934" alt="image" src="https://user-images.githubusercontent.com/1068783/38963940-3766f840-4329-11e8-91cd-ad1989edbd09.png">

gvn:
This should be good to go now, but we should close this after it's switched over to the new component on prod.

kristinashu:
It's live and it looks great! https://foundation.mozilla.org/opportunity/global-sprint/

PR #1449, opened 2018-04-23
Figure Block 2: Electric Boogaloo

gvn:
Related issue: #1407


gvn:
For Design review:

localhost_8000_cms_pages_6_edit_preview_

kristinashu:
This looks great for Opportunity pages like https://foundation.mozilla.org/opportunity/global-sprint/

Just confirm that the [About Us Leadership page](https://foundation.mozilla.org/about/leadership/) will not use this component but will use a separate 'custom' one instead?

gvn:
I think both pages could use the same component.

We can either add an option to the CMS to toggle between layout options (default/fill-space or all-the-same-size-regardless-of-quantity) or hardcode an override at the template level.

I'm always inclined toward presenting the fewest options to users per [Flexibility–usability tradeoff](https://en.wikipedia.org/wiki/Flexibility%E2%80%93usability_tradeoff), but this may be a case where an in-CMS toggle would be an acceptable compromise. I'm torn...

gvn:
Could we at least evaluate how the Leadership page looks when it shares the same layout style?

kristinashu:
Unfortunately, pics on the Leadership page should really be all the same size. People under "Emeritus Board Members" and "Management" can't look bigger than people in "Board of Directors". Because Leadership is a primary page, I think it's ok for some hard coding/overrides to happen. Agree that a toggle could get too complicated.

kristinashu:
Yes perfect!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/figure_block2.html
```diff
@@ -0,0 +1,17 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+{% with url=value.url %}
+
+<div class="col-12 {% if count > 3 %}col-sm-3{% elif count == 3 %}col-sm-4{% elif count == 2 %}col-sm-6{% endif %}">
+	{% if url %}<a href="{{ url }}">{% endif %}
+    {% image value.image fill-890x890-c100 format-jpeg class="w-100" %}
+  {% if url %}</a>{% endif %}
+
+	<p>
```
alanmoo: Should this be inside a `{% if value.caption %}`?

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -131,6 +143,15 @@ class Meta:
         template = 'wagtailpages/blocks/figure_grid_block.html'
```
alanmoo: Can you add `group = 'Deprecated'` here so it shows up in the streamfield grouped as such? And for the figureBlock if that's the case?

alanmoo:
Also rebase off master once #1461 lands; your migrations are already out of sync and that PR will just make it happen again.

PR #1451, opened 2018-04-24
pip3 instead of just pip

patjouk:
Someone ran into an issue of invoke using python 2.7 on MacOs instead of python 3.

From [Python Homebrew page](https://docs.brew.sh/Homebrew-and-Python):

```
pip3 points to Homebrew’s Python 3.x’s pip (if installed)
pip and pip2 point to Homebrew’s Python 2.7.x’s pip (if installed)
```


gvn:
Let's get a R+ from someone with OSX. Cade, you use Linux, right?

Issue #1474, opened 2018-04-28
[Components] Spacings Implementation ticket


taisdesouzalessa:
Based on discussion on #1404, here is the implementation ticket.

- [x] Add margin top and bottom of 16px by default on all components (expected behavior is to have a margin collapse as per [Gavin's comment](https://github.com/mozilla/foundation.mozilla.org/issues/1404#issuecomment-383751450)). Here is a [list of the components](https://docs.google.com/document/d/1zpw8Kn9NQEkUCAlzeh_VIvUFXEKdWRc8oskoVrd0a1Y/edit?usp=sharing) that need to be updated  (I don't have full knowledge of the items that are also components, so please correct me if something seems off/is missing there @gvn) 

- [x] Keep the line break available to users <img width="30" alt="image" src="https://user-images.githubusercontent.com/22157921/39158149-3e0ccb50-4713-11e8-9648-1a69e0acf032.png"> 

- [x] "Healthy Internet" component should have a bigger margin-top. Margin bottom is 8px+48px so margin-top should be the same (8px+48px)
<img width="581" alt="image" src="https://user-images.githubusercontent.com/22157921/39388940-b55398e2-4a38-11e8-9723-927ea878f399.png">

cc: @kristinashu @xmatthewx @gvn let me know if this makes sense to you. Assigning this task to Gavin

gvn:
I think we should also change the "Spacer" component to be optionless and just be 16px tall.

@taisdesouzalessa @kristinashu thoughts?

taisdesouzalessa:
I think the "Spacer" component can be useful because there are scenarios (like "Example B" [here](https://mozilla.invisionapp.com/share/H4HRH4BKMQW#/293086382_Desktop)) in which the layout would benefit if different blocks of text have more spacing between them instead of the constant 16px.

That being said, I do think our spacer has an overwhelming amount of options (@sabrinang and I talked about it in the past and we have some opinions about it - feel free to add your comments here, Sabrina):
<img width="973" alt="image" src="https://user-images.githubusercontent.com/22157921/39443154-8d9ec632-4c68-11e8-986c-b422b335d213.png">

Maybe just 3 would be enough? And named: **small** (equivalent to my-2 which is equal to 8px)/**medium** (equivalent to my-4 which is equal 24px)  / **large** (equivalent to my-5 which is equal 48px)

I even wonder if the "medium" is necessary since it is so close to 16px...what do you think?

gvn:
[...]

PR #1475, opened 2018-04-30
adding default vertical margins to components - #1474

gvn:
Related issue: #1474

### DIFF
--- a/source/sass/utilities.scss
+++ b/source/sass/utilities.scss
@@ -1,3 +1,11 @@
+@import '../../node_modules/bootstrap/scss/_variables';
+
+// default for vertical margins ($spacer = 1rem = 16px)
+.my-default {
+  margin-top: $spacer;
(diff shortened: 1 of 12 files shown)


gvn:
This is going to be hard to stage for a design review. @taisdesouzalessa is there anything in particular you'd want to see?

source/sass/utilities.scss
```diff
@@ -1,3 +1,11 @@
+@import '../../node_modules/bootstrap/scss/_variables';
+
+// default for vertical margins ($spacer = 1rem = 16px)
+.my-default {
```
mmmavis: lolz i was like "what's MY default" until I saw these lines...

gvn:
I'm going to merge this since it's very minor visual tweaks. @taisdesouzalessa you can play around with it once it lands on staging and file followups if there are any issues. Should be very consistent though...

Issue #1480, opened 2018-05-01
[Wagtail Component] FigureGrid2: move caption closer to image


kristinashu:
New FigureGrid2 is looking good! https://foundation.mozilla.org/about/leadership/

Only issue is when there are two rows and the caption is right in between the two images. Please move the caption up closer to the image.

Screencap on Prod:
<img width="1110" alt="image" src="https://user-images.githubusercontent.com/1068783/39493499-ec5f8ec4-4d47-11e8-81d4-677a900d5994.png">

Based on https://github.com/mozilla/foundation.mozilla.org/issues/1474#issuecomment-385525302 @taisdesouzalessa recommends 8px.

<img width="253" alt="image" src="https://user-images.githubusercontent.com/1068783/39493605-52446552-4d48-11e8-8971-0111516a3ec7.png">

@taisdesouzalessa how much space should there be between the caption and then next row of images? Maybe 24px? I don't think 16px will be enough.

cc @gvn

taisdesouzalessa:
> @taisdesouzalessa how much space should there be between the caption and then next row of images? Maybe 24px? I don't think 16px will be enough.

@kristinashu I would go even further: 32px so it is clear on mobile which copy belongs to which picture since they will be stacked. Below you can see screenshots of the copy with margin-top 8px and margin-bottom 16px:

**Desktop:**
<img width="842" alt="image" src="https://user-images.githubusercontent.com/22157921/39552624-e6d177c4-4e1e-11e8-8a86-1f84345c828f.png">
 **Mobile:**
<img width="319" alt="image" src="https://user-images.githubusercontent.com/22157921/39552741-5055038c-4e1f-11e8-8f13-5ef6f445de23.png">

kristinashu:
Sounds good Taís! Yes to white space!

gvn:
For design review:

<img width="916" alt="screen shot 2018-05-03 at 2 16 03 pm" src="https://user-images.githubusercontent.com/206379/39603271-9df45882-4edc-11e8-874a-6cb260b2be38.png">
<img width="390" alt="screen shot 2018-05-03 at 2 16 22 pm" src="https://user-images.githubusercontent.com/206379/39603272-9e0ce316-4edc-11e8-9fda-c2fcae845f85.png">

gvn:
This should be fixed.

PR #1496, opened 2018-05-03
adjusting margins of figure grid 2 - #1480

gvn:


gvn:
For design review:

screen shot 2018-05-03 at 2 16 03 pm
screen shot 2018-05-03 at 2 16 22 pm

PR #1505, opened 2018-05-07
unifying external link behavior - closes #1504

### LINKED ISSUE #1504: Make all external links target blank
Also: https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/

- Use JS/Regex to determine if a link is external and add necessary attributes for external target and safety.

gvn:


source/js/main.js
```diff
@@ -40,10 +40,25 @@ let main = {
...
+        this.decorateExternalLinks();
       });
     });
   },
 
+  decorateExternalLinks() {
+    Array.from( document.querySelectorAll(`a`) ).forEach((link) => {
+      let href = link.getAttribute(`href`);
+
+      // Define an external link as any URL with `//` in it
```
mmmavis: Should we also check if the domain in `href` is the same as `TARGET_DOMAIN`? I'm not very familiar with how things are on Wagtail and if there's any field on Wagtail that requires admins to enter full URL instead of relative URL.

gvn: Hmm...that's a good point. I'll work on an update for that

gvn:
Ok, now if someone puts a full link in to a page on `foundation.mozilla.org` it won't open a new tab for it.

PR #1507, opened 2018-05-08
linking internet health in nav - closes #1489

### LINKED ISSUE #1489: Link Internet Health page in nav
Follow up from #1298

Internet Health main nav item can now link to https://foundation.mozilla.org/internet-health/ instead of the redirect.

gvn:


gvn:
Not really anything to design review here since it's just a link change.

ScottDowne:
I think it's a bug that I missed to catch when I did the review. It's only updated on the mobile nav.

ScottDowne:
Quick fix. https://github.com/mozilla/foundation.mozilla.org/pull/1518

Not sure in the long run how to handle this in a better way. Django templating still escapes me for reuse of little blocks of html like this.

Issue #1506, opened 2018-05-08
Global Sprint 2018 Featured Projects


acabunoc:
The website is only fetching the most recent 48 entries

On the site (48 projects):
https://foundation.mozilla.org/opportunity/global-sprint/2018-projects/

In Pulse (58 projects):
https://www.mozillapulse.org/search?keyword=mozsprint%202018

It *should* fetch all the entries with the tags `mozsprint` and `2018`

acabunoc:
If I edit 'Max number of results:', I still get the same result

alanmoo:
This is by design from a performance standpoint. The Pulse project component in Wagtail tops out at 48 items; We're not trying to replicate Pulse inside the Foundation site. I'd recommend updating the page itself to be clearer that this is a portion of projects. You could try adding a button to the bottom of the list if you want, or you could reduce the list by only showing "featured" Pulse entries (right now that's a total of 3 matching projects).

acabunoc:
People are freaking out over not being featured anymore & this was not explained when we decided to use the CMS. Can you make an exception for this page?

alanmoo:
Are "featured" projects simply all Global Sprint Projects?

acabunoc:
yes!

acabunoc:
all the projects on Pulse that is. We don't approve projects that aren't 'featured'.

alanmoo:
Ok. Looks like the problem is that the Pulse API returns pages of 48 by default, yet the component doesn't handle pagination.

Let's make the React component take the "Maximum" and pass that in to the API call for now, as a query parameter of "page_size". That way in the admin, you can set a maximum higher than the number of featured projects. This will slow the page down as the number of featured projects grows, but it's the simplest fix for now.

(For implementation, let's make sure we still default to 48 projects if "0" is set as the maximum)

acabunoc:
Thanks @gvn!

The Sprint starts in ~ 6 hours. Any word when this will be live?

alanmoo:
It's live

PR #1510, opened 2018-05-08
allow arbitrary maximums for pulse project lists - #1506

gvn:


gvn:
Kinda weird...but `¯\_(ツ)_/¯`

For returning every result I'd suggest putting in a very large max, like 1000.

source/js/components/pulse-project-list/pulse-project-list.jsx
```diff
@@ -29,7 +29,7 @@ export default class PulseProjectList extends React.Component {
       });
     });
 
-    projectXHR.open(`GET`, `https://${this.props.env.PULSE_API_DOMAIN}/api/pulse/v2/entries/?format=json&search=${query}${this.props.featured ? `&featured=True` : ``}`);
+    projectXHR.open(`GET`, `https://${this.props.env.PULSE_API_DOMAIN}/api/pulse/v2/entries/?format=json&page_size=${this.props.max ? this.props.max : 48}&search=${query}${this.props.featured ? `&featured=True` : ``}`);
```
alanmoo: So 0 gets evaluated as false in the ternary and thus fetches 48?

gvn: Correctamundo.

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -267,7 +267,7 @@ class PulseProjectList(blocks.StructBlock):
         min_value=0,
         default=0,
         required=False,
-        help_text='The maximum number of results to fetch (use 0 for "no maximum")',
+        help_text='The maximum number of results to fetch (use 0 for default maximum of 48)',
```
alanmoo: Believe it or not, this needs a migration, thus the CI failure

gvn: Just noticed that...hang on

Issue #1372, opened 2018-04-17
Redirects on About Us subpages


kristinashu:
Based on Jesse's [endpoints doc](https://docs.google.com/document/d/1zhOwOET0yH3xQC83tcKmlwGT9s2lLv05wdUqRJsm5MI/edit?usp=sharing), subnav menu items that we don't have copy for yet should redirect. 

#### Redirects should be:
- /about/impact (https://advocacy.mozilla.org/)
- /about/strategy (https://mzl.la/foundation-strategy)
- /about/careers (https://careers.mozilla.org/)

#### Note that order in nav should be:
- /about
- /about/impact
- /about/leadership
- /about/trademarks
- /about/public-records
- /about/strategy
- /about/support
- /about/careers

cc @alanmoo for triage

kristinashu:
Maybe related to other redirect tickets: https://github.com/mozilla/foundation.mozilla.org/issues/937 https://github.com/mozilla/foundation.mozilla.org/issues/1216

kristinashu:
@alanmoo is this possible now?

alanmoo:
Not without some code still- Wagtail redirects do work, but if the page exists (a 200 http response), the redirect won't happen. So we need a way for the pages to be added to the menu, but for the redirect to still happen and I have a feeling that may need to happen with some hard coding embedded into our URLs.py file.

alanmoo:
To be clear, here's what I tried: 
1. Create new child pages under`About` called `impact`, etc. and make sure "show page in navigation" box is checked. 
2. Add a wagtail redirect for those routes
3. Observe that as long as the page exists (even in draft mode), the redirect doesn't work. But if the page doesn't exist, we can't get it to show up in the nav without hacking something in.

kristinashu:
I don't quite get it but don't have permissions anyways so can you add these links and redirects?

alanmoo:
Talked to @Pomax now that he's back, this is his first priority.

Summary of approach for posterity: We're probably going to need to do this kind of thing again in the future, so let's not hard code anything in. Secondly, let's not fight Wagtail. The most likely solution involves creating some type of pages for the desired links so they show up in the navigation, but adjusting something (be it a new "redirect" template or Wagtail's routing) so that those pages can redirect as needed. Pomax is going to ask the Wagtail folks what they'd suggest and formulate a plan.

Pomax:
[...]

PR #1531, opened 2018-05-15
redirect handling using a RedirectingPage and serve()

Pomax:
This closes https://github.com/mozilla/foundation.mozilla.org/issues/1372 by adding a RedirectingPage page type that can be used in place of a real page, with the title showing up in menus but the page itself effecting an immediate redirect to a custom URL when clicked by the user.

In terms of use, you just create a new page in the admin and pick this page type, then fill in the URL field, and that's pretty much it.


network-api/networkapi/wagtailpages/utils.py
```diff
@@ -31,7 +31,9 @@ def get_descendants(node, list, depth=0, max_depth=2):
     discovery pass of all menu-listable children to some root node.
     '''
     if (depth <= max_depth):
-        title = node.specific.header if node.specific.header else node.title
+        title = node.title
```
Pomax: this change was necessary due to `RedirectingPage` not having a header attribute, so it would literally crash Django when trying to build the menu.

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -31,7 +31,9 @@ def get_descendants(node, list, depth=0, max_depth=2):
...
+        if hasattr(node.specific, 'heading') and node.specific.heading:
```
Pomax: "if exists, and non-empty, use it"

network-api/networkapi/wagtailpages/models.py
```diff
@@ -475,6 +486,7 @@ class Homepage(MetadataPageMixin, Page):
         'NewsPage',
         'ParticipatePage',
         'MiniSiteNameSpace',
+        'RedirectingPage',
```
Pomax: Not sure about this one: should the root page be allow to have redirecting pages @alanmoo? It kind of feels like none of the homepage children should ever be "a different website", but that's not my call.

alanmoo: Feels weird, but let it. The Internet Health Report was already like that for a while, except that it was hard coded into the primary navigation.

alanmoo:
Important: These are 302 redirects, not 301, right?

Pomax:
according to https://docs.djangoproject.com/en/2.0/ref/request-response/#django.http.HttpResponseRedirect, yes

network-api/networkapi/wagtailpages/utils.py
Pomax: hm, it should be header.  As for the simplification, would `getattr(node.specific, 'header', node.title)` then work?

Issue #1311, opened 2018-04-10
Implement Campaign Landing Page 


beccaklam:
### Checklist:
- [x] Desktop: Form is sticky and always present
- [x] Mobile: CTA button is always present at bottom 
- [x] Form filed inputs are similar to http://bradfrost.com/blog/post/float-label-pattern/ 
- [x] Include form error states [[bug](https://github.com/mozilla/foundation.mozilla.org/issues/1311#issuecomment-387580009)]
- [x] Form title (ex "Add your name", Empty form field text, and checkbox text are from our [Style Guide ](https://foundation.mozilla.org/style-guide/))
- [x] Content on left side will reflect Wagtail rich text editor (so image is optional and font styles are based on the editor)
- [x] Follow bootstrap grid

### Comp
Initial "basic" template page https://redpen.io/gf9574a99f7765a29a and grid

<img width="983" alt="image" src="https://user-images.githubusercontent.com/1068783/39936954-189a2400-5503-11e8-819a-c380165ccebb.png">

### Mobile
<img width="319" alt="image" src="https://user-images.githubusercontent.com/1068783/40207689-b2aa8738-59ea-11e8-9bbe-53f7b2983093.png">

kristinashu:
### Comp
- starting with our initial "basic" template page https://redpen.io/gf9574a99f7765a29a

![red-pen-57207e7a-c287-4135-8c53-3fce523b51d7](https://user-images.githubusercontent.com/1068783/38707391-0ac31f16-3e66-11e8-811b-e59b9356d7cc.jpg)

### Existing components
- Use nav that is currently on production
- Callout above the CTA already exists https://foundation.mozilla.org/campaigns/equifax-breach/

### Content
- Content on left side will reflect Wagtail rich text editor (so image is optional and font styles are based on the editor)

### Form 
- Updated our form style to something like http://bradfrost.com/blog/post/float-label-pattern/ 
- Form title (ex "Add your name", Empty form field text, and checkbox text are from our [Style Guide ](https://foundation.mozilla.org/style-guide/))
- Orange #FF5C00 button and text is a new color not in our style guide
- @ScottDowne and @beccaklam to work out details for the best form UX!

### Desktop bootstrap grid
<img width="970" alt="screen shot 2018-04-12 at 3 30 55 pm" src="https://user-images.githubusercontent.com/1068783/38707539-8ef8f814-3e66-11e8-93a5-2c05e7d6ae66.png">

### Mobile
[...]

PR #1555, opened 2018-05-23
Updates to campaing page to match layout grid.

ScottDowne:

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -4,7 +4,7 @@
 {% block subcontent %}
 <div class="row">
     <div class="cms{% if page.cta and uses_menu == False %}
-        col-md-6
+        col-lg-7
     {% else %}
         col-12
     {% endif %}{% if uses_menu and page.narrowed_page_content %} col-md-9{% endif %}">
@@ -21,7 +21,7 @@
     </div>
 
     {% if page.cta != None %}
-    <div class="cta col-md-6">
+    <div class="cta col-lg-4 offset-lg-1">
         <div class="narrow-sticky-button-container hidden">
           <a href="#cta-anchor" class="narrow-sticky-button btn">TAKE ACTION</a>
         </div>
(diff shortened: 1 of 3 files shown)


ScottDowne:
@kristinashu can you drop the ticket in here for this when you do the review so gvn has some context? (I lost the ticket)

kristinashu:
Yup, here's the ticket https://github.com/mozilla/foundation.mozilla.org/issues/1311

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
```diff
@@ -21,7 +21,7 @@ <h1>{{ block.value }}</h1>
     </div>
 
     {% if page.cta != None %}
-    <div class="cta col-md-6">
+    <div class="cta col-lg-4 offset-lg-1">
```
alanmoo: This looks like it breaks multipage campaigns. https://foundation-mofostaging-pr-1555.herokuapp.com/campaigns/multi-page/

source/js/components/petition/petition.scss
```diff
@@ -30,7 +30,7 @@
       transition: margin-bottom 0.3s ease-in-out, visibility 0s ease-in-out 0.3s;
     }
 
-    @media (min-width: $bp-md) {
+    @media (min-width: $bp-lg) {
```
alanmoo: How do I get the sticky button to show up?

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
ScottDowne: Cool cool, I'll look into that.

source/js/components/petition/petition.scss
ScottDowne: just make the page narrow enough.

ScottDowne: if you load the page while wide, and then make it narrow, you'll need to refresh.

This is because the button goes away if the form is visible, and if the page is wide on load, the button hides.

kristinashu: I think it wasn't showing up because there was no body content. I've added some and now it seems works again https://foundation-mofostaging-pr-1555.herokuapp.com/campaigns/single-page/ (note: just resizing the window doesn't work, it only shows up if you load the page at a narrow width).

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
ScottDowne: Pushed an updated, should fix it.

PR #1558, opened 2018-05-23
Only hide title if we're using a menu, not when we have children page…

ScottDowne:
…s, because if a parent page has children that are all told to not be displayed in the menu, we end up with no menu, thus no title.

Related issue:
Summary:

If you don't know who to ask for a review, try @alanmoo or @xmatthewx. If you do, please delete this line.

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/mini_site_name_space.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/mini_site_name_space.html
@@ -38,7 +38,7 @@
 
 
 {% block h1 %}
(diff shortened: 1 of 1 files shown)


ScottDowne:
What's happening is we hide the title on narrow view if the page has children pages, and rely on the narrow view nav (drop down) to show the page title.

This breaks when a page has children, but all children are set to not be displayed in the menu, making a case where we have children but no menu (this is what happened in the localized pages).

All my fix does is change the check from "do I have children" to "do I have a menu" which is much more accurate for that the css class is trying to achieve.

Issue #1476, opened 2018-04-30
[Wagtail] Narrow grid width when no side nav or petition form by default


kristinashu:
CMS pages have a "Narrowed page content" checkbox. 

<img width="1327" alt="image" src="https://user-images.githubusercontent.com/1068783/39449431-dc97e0e8-4c7c-11e8-844b-118575e7338f.png">

Problem is that users won't know when to check or uncheck this box. User generated pages should always be narrow when there is no side nav or side petition. And should be otherwise always be full-width. 

All our Opportunity/Campaign pages without nav or petition should be narrow:
- https://foundation.mozilla.org/opportunity/leadership-training/
- https://foundation.mozilla.org/opportunity/network-50/
- https://foundation.mozilla.org/opportunity/fellowships-faq/
- https://foundation.mozilla.org/opportunity/fellowships-research/
- we don't have any Campaign pages like this yet, but we will

All our Primary pages that are CMS generated should be narrow:
- https://foundation.mozilla.org/about/
- https://foundation.mozilla.org/about/leadership/ 
- https://foundation.mozilla.org/about/public-records/
- https://foundation.mozilla.org/about/trademarks/

kristinashu:
cc @Pomax (and @ScottDowne)

This https://github.com/mozilla/foundation.mozilla.org/pull/1568 was a short term fix to make sure petition pages look good regardless if "make narrow" check box is selected or not. 

This ticket is a more long term solution to put the onus on the template rather than the user to determine the page width.

kristinashu:
From #1405 

https://redpen.io/gnce740629099e94c7

![template 2x](https://user-images.githubusercontent.com/1068783/40739611-835da756-63fb-11e8-92d8-4d0115740623.jpg)

Pomax:
this got implemented quite a while back.

PR #1562, opened 2018-05-24
Fixes so a narrow campaign page doesn't look horrible if it ever happ…

ScottDowne:
Fixes https://github.com/mozilla/foundation.mozilla.org/issues/1564

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -3,11 +3,14 @@
 
 {% block subcontent %}
 <div class="row">
-    <div class="cms{% if page.cta and uses_menu == False %}
+    <div class="cms{% if page.cta and uses_menu == False and page.narrowed_page_content == False %}
         col-lg-7
-    {% else %}
-        col-12
-    {% endif %}{% if uses_menu and page.narrowed_page_content %} col-md-9{% endif %}">
+      {% elif page.narrowed_page_content == True %}
+        col-lg-8
+      {% else %}
+        col-lg-12
+      {% endif %}{% if uses_menu and page.narrowed_page_content %} col-md-9{% endif %}"
+    >
 
       {% block html_content %}
         {% for block in page.body %}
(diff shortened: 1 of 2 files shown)


kristinashu:
I'm not sure how to check this? It looks good on https://foundation-mofostaging-pr-1562.herokuapp.com/campaigns/single-page/ but it looked good last time too. I need a way to see it as though the "Narrowed page content" is checked off. @alanmoo any suggestions?

ScottDowne:
@kristinashu I logged into it and made that page narrow, now looks like this https://foundation-mofostaging-pr-1562.herokuapp.com/campaigns/single-page/

kristinashu:
Thanks! https://foundation-mofostaging-pr-1562.herokuapp.com/campaigns/single-page/ and https://foundation-mofostaging-pr-1562.herokuapp.com/campaigns/multi-page/ both look less broken now. It it isn't ideal but definitely a better fallback than what we had. 

Here's the longer-term solution https://github.com/mozilla/foundation.mozilla.org/issues/1476

alanmoo:
@cadecairos @patjouk There may be a bit more need for variable campaign pages.

@Pomax Handing this to you for review because you know the structure here. Might there be a need for simplification of the template complexities?

alanmoo:
@gvn too for visibility because you may have thoughts on templates

kristinashu:
Taís made these a little while ago to show the different template rules:

[Campaign Page (2 templates)](https://mozilla.invisionapp.com/share/6PH6YDCKSUT#/292486894_Campaign_Template)
[Opportunity Page (2 templates)](https://mozilla.invisionapp.com/share/SEH6YI72GKW#/292486895_Opportunitiy_Page_Template)

What isn't represented here is that /campaign should also be able to look like Opportunity pages (aka the url says `campaign` but it looks like an opportunity page - no petition form).

ScottDowne:
I noticed there might have been a bit of a bug in some of the offset usage on narrow pages, which once fixed, allowed me to remove some of my changes. Testing now but I did push up the changes already.

ScottDowne:
No wait that probably won't work, or isn't what I thought it looked like.

ScottDowne:
@kristinashu question, in narrow content set pages, what should happen with the screen is 1200px+ wide? Should it stay narrow or just go large? I ask because the code seems to imply both, and it's kinda conflicting from what I can tell.

ScottDowne:
I'm going to drop my changes, it's clear this is more of a mess than just applying some css.

ScottDowne:
For any devs reading this, the conflict I'm seeing is here: `offset-md-2 col-lg-12` in mini_site_name_space.html offset-md-2 looks to apply a left margin offset on medium and above widths, which we then also say on lg and above make everything 12 wide. Meaning on large and above we have a total 14 units usage wide, which seems kinda busted from my understanding of bootstrap?

So I guess we can either remove the col-lg-12 and replace it with something that ends up as 8, meaning on large widths we're still narrow. Or stop enforcing narrow on 1200px wide screens by adding offset-lg-0.

If that's not making sense, think of it this way, col-lg-12 is implying we don't want to be narrow on large and up, and offset-md-2 is implying we do want to be narrow on medium and up. Am I reading this correctly?

PR #1563, opened 2018-05-24
SQS petitions

Pomax:
This adds functionality to the petition handling codepaths so that we can get the schema described on page 8 in https://docs.google.com/document/d/1LeJ3l9OSNADQWPyVOUXHUlDKxRZe_YABXyRDBIW7NEY/edit# to a new SQS location for easier consumption.

This is still work in progress, and requires a Salesforce-approved picker for the `country/state` data (currently a freeform text field in petition.jsx), and might need a more elaborate codepath separation in the `campaign/views.py:petition_submission_view()` function, based on whether what will now be "legacy" data to the now "old" google sheet SQS location, or whether we're sending new data to the new SQS location.

## Things to make sure of for testing:

update your .env so that you're sending petitions to yourself:
```
NETWORK_SITE_URL=http://localhost:8000
...
TARGET_DOMAIN=localhost:8000
```
make sure you edit a petition for whichever campaign page in http://localhost:8000/cms you want to be testing with to make sure its `requires country` and `requires postal code` checkboxes are set, so that the petition will ask you for those.


network-api/networkapi/wagtailpages/models.py
```diff
@@ -181,6 +181,23 @@ class OpportunityPage(MiniSiteNameSpace):
 
 @register_snippet
 class Petition(CTA):
+    # This field, if set, overrides the default SQS queue location so
+    # that the legacy location is used by the server, instead.
+    legacy_petition = models.BooleanField(
```
ScottDowne: How is this going to work, is the idea we go in and check this for all old petitions when we ship?

ScottDowne:
@Pomax we should be getting @cadecairos to look at some of this too.

Otherwise, I'm trying to come up with something for the country dropdown to manage the list of countries.

network-api/networkapi/wagtailpages/models.py
cadecairos: could we make it a data migration?

Pomax: can do.

network-api/networkapi/campaign/views.py
```diff
@@ -19,6 +19,7 @@
         aws_secret_access_key=settings.AWS_SQS_SECRET_ACCESS_KEY,
     )
 queue_url = settings.PETITION_SQS_QUEUE_URL
+legacy_queue_url = settings.LEGACY_PETITION_SQS_QUEUE_URL
```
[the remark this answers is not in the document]
Pomax: the problem is basically the same - if we say this:
```
queue_url = settings.NEW_PETITION_SQS_QUEUE_URL
legacy_queue_url = settings.PETITION_SQS_QUEUE_URL
```
then it's just as weird, really. As a dev my first though would be "why is the legacy url using the current url value?". 

As a one-time deal I'd prefer we do a one time switch-over (we only have a handful of petitions) when we deploy, but let's get @alanmoo's opinion here, too.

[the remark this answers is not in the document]
alanmoo: Why not just call it something like `crm_queue_url`? Eventually we could remove the prefix, but there's no real need to do so.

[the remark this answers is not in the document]
cadecairos: > crm_queue_url

This could work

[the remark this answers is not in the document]
Pomax: renamed to crm_queue_url (with new var) and gs_queue_url (with current var).

Pomax:
@ScottDowne, if you can look this over in terms of the petition itself, and see where there's still todo's for follow up, that would be great. Anything else you pick out in the code is of course super helpful too =)

@cadecairos, @gideonthomas, as a fairly significant change to the way petitions are sent (new ones will go to a new place, with a new payload schema), I would like to ask you twoto give this a thorough review, with an eye on "if I had to implement this change, does the code I'm looking at now make sense".

network-api/networkapi/settings.py
```diff
@@ -387,6 +388,7 @@
 
 # The SQS endpoint where we will send petition signups to
```
cadecairos: minor nit: endpoints!

ScottDowne:
@Pomax is there anything setup in the review app I can see to test with?

network-api/networkapi/wagtailpages/migrations/0026_auto_20180524_1225.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.11.13 on 2018-05-24 19:25
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+
+
+def mark_existing_petitions_as_legacy(apps, schema_editor):
+    petitions = apps.get_model("wagtailpages", "Petition")
+    for i in petitions.objects.all():
```
cadecairos: I think you should be able to do `petition.objects.all().update(legacy_petition=True)` to have Django make the change in one SQL statement.

source/js/components/petition/salesforce-country-list.js
```diff
@@ -0,0 +1,254 @@
...
+  "VU": "Vanuatu",
+  "VE": "Venezuela, Bolivarian Republic of",
+  "VN": "Viet Nam",
+  "VG": "Virgin Islands, British",
+  "WF": "Wallis and Futuna",
+  "EH": "Western Sahara",
+  "YE": "Yemen",
+  "ZM": "Zambia",
+  "ZW": "Zimbabwe"
+}
```
cadecairos: semicolon :)

network-api/networkapi/wagtailpages/migrations/0026_auto_20180524_1225.py
Pomax: handy!

network-api/networkapi/settings.py
Pomax: not sure I understand the nit, it should be singular in this sentence, and then also a singular in the next one?

network-api/networkapi/campaign/views.py
```diff
@@ -58,6 +72,64 @@ def petition_submission_view(request, pk):
...
+            {'error': 'Server is missing campaign for petition'},
+            status=status.HTTP_500_INTERNAL_SERVER_ERROR,
+        )
+
+    data = {
+        "campaign_id": cid,
+        "first_name": request.data['givenNames'],
+        "last_name": request.data['surname'],
+        "email": request.data['email'],
+        # We already submit an email subscription separately
```
pmac: @Pomax in a meeting today it was suggested that I'd need to have basket send a subscription when it saw this key come in as `True` (I assume it's boolean?), but this comment suggest that this would be redundant. Do you know if I should or should not fire off a subscription for the mozilla newsletter for the email address of the signer when this flag is true?

alanmoo: I also thought this was going to/should happen Basket-side, but seem to recall Pomax had a reason for it happening on the Foundation site server side.

Pomax: We've had this in place effectively "since forever" so when that was brought up during the first or second meeting we decided it wasn't worth changing that. However, we're not married to doing things client-side in the slightest so if the PM decision is to do this in on the basket side we can remove this code pretty much whenever is convenient

PR #1566, opened 2018-05-25
Adding wagtail-modeltranslation to get at least some sort of localization.

Pomax:
## The details

This adds https://github.com/infoportugal/wagtail-modeltranslation in addition to the basic Django changes necessary for i18n outlined in http://docs.wagtail.io/en/latest/advanced_topics/i18n/index.html, in order to effect an l10n/i18n'd version of wagtail.

This adds new fields for any field that is marked as "should be localizable" (indicated in a `translations.py` file) for each language that should be supported site-wide.

For streamfield content, copy buttons are provided that allow users to "copy over" the content from any locale into another locale-specific version.

Content now also auto-routes to `localhost:8000/en/...` if your browser thinks you're in an English locale, and you can manually explicitly navigate to any other supported locale without the CMS going "no I'm pretty sure you're in an `en` browser, I will force you back to `en` now."

So this all "works", but it's not ideal. We're landing this now mostly because the proper solution is going to require quite a bit of work on the wagtail side, and we don't have _any_ localization options right now other than "make a new page, mess with the slug to pretend it's a localized version, but really it's just a completely different page".

### Getting the database set up

This needs to be tested "properly", with a copy of the production database, using postgresql. This means you'll need to have that installed (`brew install postgresql` on mac; download and run the official installer for windowsl; if you use linux/unix, you know how to install things for your favourite flavour, so just do that for postgresql)

1) grab a copy of the production database by using `pg_dump DATABASE_URL > foundation.psql`, where the database url is the value associated with the `DATABASE_URL` environment variable for our production app on heroku.

once that finishes, open it in your favourite editor and take note of who the owner is:

```
SET search_path = public, pg_catalog;

--
-- Name: clean_user_data(); Type: FUNCTION; Schema: public; Owner: ...... <= we want to know this string
--
```

2) run `createdb foundation` so that you have a postgresql database to work with.

3) run `psql foundation` to connect to that database.

4) run `CREATE ROLE theownernamefromthedbdump WITH SUPERUSER;`, making sure to have that semi-colon at the end, and making sure NOT to quote the `owner` string.

5) run `\i foundation.psql`, which should now import the entire database for you.

You will now also need to update your `.env` file to make sure you're using this database, setting `DATABASE_URL=postgres://localhost:5432/foundation`.

### Getting the database uplifted to localized form

This is less work than the previous steps. Get yourself in the foundation.mozilla.org dir if you're not there already, and run:

```
$ inv makemigrations
$ inv migrate
$ inv l10n-sync
$ inv l10n-update
```
Your database should now be all fixed up, and you can run everything with `npm start` or `inv runserver` (the latter is probably faster since we're not actually concerned with any changes to the page-injected react code)

### What's my username?

You copied the production database, so it's your normal username. However, if you don't have google authentication set up for local access (which is a bit of a chore, tbh) you can also run `pipenv run python network-api/manage.py createsuperuser` and just create yourself a superuser for testing so you don't have to log in through google. 

## Known issues and quirks

- page URLs are based on `slug` values, so if you want a page with the same URL save for the language code, you need to double-check that only the `[en]` slug has a value in the PROMOTE tab when you publish. This is an odd bug and I've filed https://github.com/infoportugal/wagtail-modeltranslation/issues/213 to see if there's a solution that can be implemented in the library instead of us trying to document it as "you need to publish the page, then edit it again, go to the promote tab, and manually delete all slugs that aren't EN"
- plain Django models (say, snippets like Petitions and Signups) get localised just fine but unlike `Page`  objects they don't get their base fields hidden in the admin. I'm not sure why, I filed https://github.com/infoportugal/wagtail-modeltranslation/issues/210 to see whether that's something that can be fixed.
- Localizers need to get CMS access. That feels both obvious, and really wrong, so let's not train staff to consider this "the way to localize". This is a hack. It gets us where we need to be as far our users can tell, but it's not okay and we should probably make sure _everyone_ understands this is a temporary fix.
- multi-column fields that contain 2 or 3 fields before localisation suddenly contain 12 or 18 (or more) fields after localization depending on how many languages cause duplication. Obviously, this completely wrecks Wagtail's ability to determine what the CSS class should be for regulating column width, so I hacked it up. Since it's impossible to "reverse-guess" what the original number in the css class was, I've added an override that just strips the class from multi-column fields. It's not pretty, but "not pretty" is better than "impossible to make sense of".


alanmoo:
@TheoChevalier Do you have thoughts on this as a next step for l10n? I know it's not perfect, but we're trying to move within the framework of what Wagtail offers us.

gvn:
So, to clarify, does this depend on editors logging into Wagtail to translate, or is there a way to leverage something like Transifex?

Pomax:
I don't know if there's a clean Transifex integration using this library yet.

Pomax:
Hm, see what `pipenv install wagtail-modeltranslation` says when you run it manually. Did you run through the "new" foundation setup to make sure pipenv etc. are all initialised correctly?

network-api/networkapi/settings.py
```diff
@@ -345,14 +349,21 @@
...
+LANGUAGES = (
+    ('en', _('English')),
+    ('de', _('German')),
+    ('pt', _('Portuguese')),
+    ('es', _('Spanish')),
+    ('fr', _('French')),
+    ('pl', _('Polish')),
+)
 
+# We do not define an explicit LANGUAGES listing,
```
gideonthomas: this seems contrary to what's defined above?

Pomax:
Only the fields that we indicate in translations.py will get the special sauce treatment, I don't think any of those are non-nullable, but obviously a good idea to see what happens. If no one pre-empts me today, I'll have a look on Monday.

Pomax:
Based on https://github.com/infoportugal/wagtail-modeltranslation/issues/207#issuecomment-392533693 it sounds like Transifex is not going to help us with the CMS-hosted dynamic content.

Pomax:
One genuine problem is that, for some reason, tests now crash python. Not just "fail":
```
c:\Users\Mike\Documents\git\collaborations\mozilla\foundation.mozilla.org>inv test
Running flake8
Loading .env environment variables…
Running tests
Loading .env environment variables…
Creating test database for alias 'default'...
Got an error creating the test database: database "test_foundation" already exists

Type 'yes' if you would like to try deleting the test database 'test_foundation', or 'no' to cancel: yes
yes
Destroying old test database for alias 'default'...
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
psycopg2.ProgrammingError: column wagtailcore_page.title_en does not exist
LINE 1: ...ore_page"."numchild", "wagtailcore_page"."title", "wagtailco...
                                                             ^
HINT:  Perhaps you meant to reference the column "wagtailcore_page.title".


The above exception was the direct cause of the following exception:

  File "network-api/manage.py", line 22, in <module>
    execute_from_command_line(sys.argv)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\__init__.py", line 364, in execute_from_command_line
    utility.execute()
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\__init__.py", line 356, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\commands\test.py", line 29, in run_from_argv
    super(Command, self).run_from_argv(argv)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\base.py", line 283, in run_from_argv
    self.execute(*args, **cmd_options)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\base.py", line 330, in execute
    output = self.handle(*args, **options)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\management\commands\test.py", line 62, in handle
    failures = test_runner.run_tests(test_labels)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\test\runner.py", line 601, in run_tests
    old_config = self.setup_databases()
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\test\runner.py", line 546, in setup_databases
    self.parallel, **kwargs
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\test\utils.py", line 187, in setup_databases
    serialize=connection.settings_dict.get('TEST', {}).get('SERIALIZE', True),
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\base\creation.py", line 77, in create_test_db
    self.connection._test_serialized_contents = self.serialize_db_to_string()
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\base\creation.py", line 121, in serialize_db_to_string
    serializers.serialize("json", get_objects(), indent=None, stream=out)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\serializers\__init__.py", line 129, in serialize
    s.serialize(queryset, **options)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\core\serializers\base.py", line 80, in serialize
    for count, obj in enumerate(queryset, start=1):
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\base\creation.py", line 117, in get_objects
    for obj in queryset.iterator():
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\models\query.py", line 53, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\models\sql\compiler.py", line 899, in execute_sql
    raise original_exception
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\models\sql\compiler.py", line 889, in execute_sql
    cursor.execute(sql, params)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\utils.py", line 94, in __exit__
    six.reraise(dj_exc_type, dj_exc_value, traceback)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\utils\six.py", line 685, in reraise
    raise value.with_traceback(tb)
  File "C:\Users\Mike\.virtualenvs\foundation.mozilla.org-X7v0FXSr\lib\site-packages\django\db\backends\utils.py", line 64, in execute
    return self.cursor.execute(sql, params)
django.db.utils.ProgrammingError: column wagtailcore_page.title_en does not exist
LINE 1: ...ore_page"."numchild", "wagtailcore_page"."title", "wagtailco...
                                                             ^
HINT:  Perhaps you meant to reference the column "wagtailcore_page.title".
```

@cadecairos @patjouk any idea what that might be? (other than "yay, psycog~")

patjouk:
TL;DR: it might be an issue around test database, with a few solutions to test [described here](https://github.com/infoportugal/wagtail-modeltranslation/issues/175#issuecomment-368046055).

I took a look this morning and found an issue around migrations: `migrate` says we're missing one while `makemigrations` says no change happened. My guess was that something was interefering with Django migrations: I couldn't find the field "title" in any migrations file but, for some reason, psycopg was expecting it.

I don't know why I was looking at the `settings.py` but I saw `'wagtail_modeltranslation.makemigrations'` [on L174](https://github.com/mozilla/foundation.mozilla.org/pull/1566/files#diff-2b1ca8a00cb2ce5d77f49ee88010d26fR174). As a test, I removed that line: Django found the missing migration, I applied it and was able to run the tests.

But, reading the wagtail docs, I realized that removing that line is [not something you want to do](https://wagtail-modeltranslation.readthedocs.io/en/latest/caveats.html?highlight=makemigrations#wagtail-s-page-patch) so I looked at their `migrate` command. It took me to [this issue](https://github.com/infoportugal/wagtail-modeltranslation/issues/175#issuecomment-368046055) where someone describe the same problem you're running into, with different solutions to try.

I hope it helps :)

Pomax:
Things to check up on:

- [ ] enabling /country-LOCALE/ codes rather than just /country/
- [ ] check if we can track page changes (maybe tap into the revision system?) so that we know when strings that have localized versions need to be updated
- [ ] check how we can get strings into the JSX component side of things (add them as `data-` attributes on the CMS side?)

chadsansing:
@chadsansing

Pomax:
The floor recognizes @chadsansing

Pomax:
Seriously, what the heck is this `npm ERR! nan not accessible from fsevents` error...

network-api/networkapi/settings.py
```diff
@@ -187,6 +194,7 @@
 
     'whitenoise.middleware.WhiteNoiseMiddleware',
     'django.contrib.sessions.middleware.SessionMiddleware',
+    'django.middleware.locale.LocaleMiddleware',  # should be after SessionMiddleware and before CommonMiddleware
```
alanmoo: _This_ is useful commenting. Thank you.

network-api/networkapi/wagtail-l10n-customization/static/css/language_toggles.css
```diff
@@ -0,0 +1,51 @@
+form .l10n-hidden {
```
alanmoo: Is this a copy from another file? Using the element type for CSS targeting seems overly specific.

Pomax: we want to be overly specific. The most specific rule wins.

Pomax:
What's weird about it? (Where's the rest of the window screenshot?)

PR #1570, opened 2018-05-28
updating header on petitions

ScottDowne:


ScottDowne:
Fixes https://github.com/mozilla/foundation.mozilla.org/issues/1311#issuecomment-392572772

kristinashu:
This is so small that I don't think it requires code review. Can we just commit this to staging?

Pomax:
agreed =)

PR #1571, opened 2018-05-29
Header align

ScottDowne:
@kristinashu this is the patch that aligns the petition with the h1 better. It was trickier than it looks.

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -11,13 +11,7 @@
       {% endif %}"
     >
 
-      <h1 class="h1-heading {% if uses_menu %}hidden-sm-down{% endif %}">
-      {% if page.header %}
-        {{ page.header }}
-      {% else %}
-        {{ page.title }}
-      {% endif %}
-      </h1>
+      {% include "./h1-heading.html" %}
       {% block html_content %}
         {% for block in page.body %}
 	        {% if block.block_type == 'heading' %}
@@ -26,7 +20,7 @@
 	            {% include_block block %}
 	        {% endif %}
         {% endfor %}
-    {% endblock %}
+      {% endblock %}
     </div>
 
     {% if page.cta != None %}
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -11,6 +11,13 @@
       {% endif %}"
     >
 
+      <h1 class="h1-heading {% if uses_menu %}hidden-sm-down{% endif %}">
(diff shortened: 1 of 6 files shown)


kristinashu:
@Pomax does this need code review?

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
```diff
@@ -1,6 +1,8 @@
 {% extends "./mini_site_name_space.html" %}
 {% load wagtailcore_tags mini_site_tags %}
 
+{% block h1 %}{% endblock %}
```
Pomax: If the block needs to be set empty, does that mean this was only an issue for campaign pages, or also opportunity pages? (it feels like if we're overriding it here to be empty, and then hardcode it in the subcontent, that block may be in the wrong spot in the parent template).

ScottDowne: Technically it was only for campaign pages, because it was related to the petition snippet.

That said, I'm pretty sure if this changed happened on opportunity pages, it wouldn't show a visible difference.

If the h1 block is only being used for those two page types, I think I can clean this up.

ScottDowne: So this isn't an override on the minisitenamespace, it's an override on the modular page, which right now is only used on minisitenamespace, but these comments here concern me: https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/models.py#L313-L318

I don't want to drift too far from the initial plan here. If modular page was meant to be for primary page too, but there was an issue, I don't want to be making campaign/opportunity based only changes in it.

ScottDowne: Hitting an issue where I need to know how to reuse small bits of html in django templates. Blocks won't work, and making a whole new file and including it seems super silly. (but might be the expected way)

ScottDowne: If you're still curious, the fix basically pastes the code in the two places we need it, and removes the block... Like, nothing fancy here.

ScottDowne:
Forgot to mention, this has updates.

kristinashu:
Design still looks good!

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
```diff
@@ -11,6 +11,13 @@
       {% endif %}"
     >
 
+      <h1 class="h1-heading {% if uses_menu %}hidden-sm-down{% endif %}">
```
Pomax: this is the exact same code as in mini_site_name_space.html, can we turn this into a fragment and `{% include "h1-heading.html" %}` that in both places?

Issue #1560, opened 2018-05-24
[Campaigns] Multi-page petition page layout


kristinashu:
We need our [multi-page petition pages](http://foundation.mozilla.org/campaigns/aadhaar/) to align with our single-page petition pages where the form is nice and prominent.

#### Steps include:
- [x] Move sub-nav from side to top below the title and style (see below for style details)
- [x] Use the same layout grid as single-page petitions (7 body content | 1 gap | 4 form)
- [x] Mobile: sub-nav turns into a dropdown like we have now on [see Aadhaar](http://foundation.mozilla.org/campaigns/aadhaar/)

#### Follow up tickets:
- Add "next section" wayfinding #1561
- Style all our dropdowns [need ticket]

#### Sub-nav styling:
- 15px 
- #000
- 24px gap between menu items
- Inactive
  - Nunito Sans Regular 
  - Line: 1px | #CCCCCC
- Selected/Hover
  - Nunito Sans Bold
  - Line: 3px | #000 | length of the word

#### [Extra Large:](https://redpen.io/pg11648642fd877912)
![multipage copy 2x](https://user-images.githubusercontent.com/1068783/40510255-f0d8fb78-5f50-11e8-93a2-294a671054f9.jpg)

#### [Medium:](https://redpen.io/zy0a3e69d2aaef09dc)
<img width="654" alt="image" src="https://user-images.githubusercontent.com/1068783/40499082-b28515f2-5f35-11e8-9940-2856e145e41a.png">

#### [Small and Extra Small:](https://redpen.io/hx831a776ac7d00a3f)
<img width="318" alt="image" src="https://user-images.githubusercontent.com/1068783/40499062-a9569e42-5f35-11e8-908a-5ff9bb62dda1.png">

beccaklam:
@kristinashu Looks good to me! Only thing is I think stylistically Nat's subnav on the nav UI's she's developing uses a thicker line for selected items that's aligned in the middle of a thinner grey line ... should we make them the same? I'm fine with adapting Nat's style for the campaign pages if so. CC: @natalieworth

kristinashu:
@beccaklam @natalieworth I would like to get them as similar as possible. I've mocked up a quick comparison https://redpen.io/mz3c1d83d188065adf

kristinashu:
I went with something in between, specs [above](https://github.com/mozilla/foundation.mozilla.org/issues/1560#issue-326180443) have been updated. Thanks!

[...]

PR #1580, opened 2018-05-31
Horizontal nav

ScottDowne:
Building the nav as shown in these comps https://redpen.io/pg11648642fd877912

I'm actually not aware of a ticket specifically for this. @kristinashu is there a ticket for just the nav?


ScottDowne:
Actually, probably should send this through design review first (forgot)

kristinashu:
Comp
image

PR
image

- [ ] Here's the ticket https://github.com/mozilla/foundation.mozilla.org/issues/1560 it has some specifics about the font styling (what you have is close but the sub nav items should be a thinner font weight and slightly smaller)

- [ ] Like in the mobile view, we don't need to show "Key Challenges" since people see it in the menu

- [ ] Mobile: nav items seem to have disappeared from dropdown
screen shot 2018-06-04 at 12 57 07 pm

ScottDowne:
@kristinashu updated.

I think given that most of the header changes are already like they are in my PR and not relatedto the nav, to keep this pr smaller and easier to review, let's do header changes in another ticket if possible.

kristinashu:
Cool, sounds good! 

Sorry but I realize now that making the font change weights shifts all the text over on hover. My bad! Please make the font bold all the time.

nav

Issue #1587, opened 2018-06-05
[Campaign QA] Hide title on multipage petitions


kristinashu:
Follow up from https://github.com/mozilla/foundation.mozilla.org/pull/1580

### Comp:
<img width="1217" alt="screen shot 2018-06-05 at 1 27 15 pm" src="https://user-images.githubusercontent.com/1068783/41001573-2e6faf94-68c6-11e8-98d9-0ecc3fd8cfed.png">

### Screen cap from staging:
![navtitle](https://user-images.githubusercontent.com/1068783/41001586-35fa4814-68c6-11e8-89a1-d9f759a8f5cd.jpg)

ScottDowne:
This got done.

PR #1592, opened 2018-06-06
campaign page header changes.

ScottDowne:


kristinashu:
Ticket for reference https://github.com/mozilla/foundation.mozilla.org/issues/1587

Pomax:
would have been nice to keep this unified, but if that hampers implementing the design, r+

Issue #1408, opened 2018-04-18
[Wagtail UX] Simplify button options


kristinashu:
There are too many button style options and I'm not sure what the Outline checkbox does.

<img width="962" alt="image" src="https://user-images.githubusercontent.com/1068783/38907242-c9030362-426f-11e8-9aaa-9ab8255aa029.png">
 

### To do:
- [x] Remove Outline checkbox if it is not needed
- [x] Reduce number of button styles to two .btn-ghost and .btn-normal from the [style guide](https://foundation.mozilla.org/style-guide/).

<img width="579" alt="image" src="https://user-images.githubusercontent.com/1068783/38907249-d71e0258-426f-11e8-9185-f2c1d998b205.png">

kristinashu:
Looks like this is live! Closing.

PR #1593, opened 2018-06-07
 simplified LinkButton - #1408

gvn:


gvn:
For design review:

screen shot 2018-06-07 at 9 54 23 am

Pomax:
@gvn github rebase failed, so I spun a new migration the "traditional" way, and CI passes again. Far as I can tell there are no problems, so merge should be safe, but if you want to double check there's no harm in playing it safe.

PR #1599, opened 2018-06-07
show draft pages in menus if the viewing user is CMS-authenticated

Pomax:
This PR changes the menu building logic so that pages that are in draft are _not_ skipped over when the viewing user has an authenticated session with the CMS (i.e. they are logged into the CMS when viewing pages).

This PR also dries out the tags for showing menus, since all three tags used the exact same data objects


ScottDowne:
Just cleaning up my backlog, and found this. (sorry about missing it)

If this is still valid, and you rebase it, I can take a look asap.

Pomax:
still relevant, so: rebased.

PR #1585, opened 2018-06-04
Revised ImageTextBlock - #1514

gvn:

Issue #1514, opened 2018-05-08
[Wagtail Component] Image + Text


taisdesouzalessa:
### Image + Text

IMPORTANT: this component has 2 variations. If I am not mistaken, @gvn suggested we build one component with a checkbox if users want the small icon scenario. Correct me if I am wrong. We can discuss implementation in the CMS if necessary.

**Examples** 
_Medium size image_
https://foundation.mozilla.org/opportunity/tech-policy-fellows/
https://foundation.mozilla.org/opportunity/web-literacy/core-curriculum/frame/
https://foundation.mozilla.org/opportunity/global-sprint/ (one logo)
_Small icon size image_
https://foundation.mozilla.org/opportunity/mozilla-information-trust-initiative/
https://foundation.mozilla.org/opportunity/web-literacy/core-curriculum/frame/

**Recommendation - General**

- [ ] Image always on the left
- [ ] Paragraph component styling (h2-h6, p and links)
- [ ] 1:1 aspect ratio
- [ ] Image top aligned with top of text

**Medium size image recommendations:**

- [ ] 3 columns
- [ ] Mobile: images is full width
- [ ] Images for partner organizations should allow links (clickable images)
- [ ] IMPORTANT: some images may need design support - see “Internet Health” pages example, in which a white space had to be added to icons so the page looks more balanced

**Medium size image examples:**
[Comp here](http://mozilla.invisionapp.com/share/Z7I9P86BET8)
Logo example:
<img width="223" alt="image" src="https://user-images.githubusercontent.com/22157921/39787774-e6ed6fd0-52db-11e8-8b7f-ce7176b8881b.png">

**Small icon size image recommendations:**

- [ ] Instead of making the icon occupy 1 column (which can make a very tiny icon on narrower grids), we should use the same model as we use for the “healthy internet” mini banner. Although it breaks the grid, it makes more sense on different screen sizes.

**Small icon size image examples:**
[Comp here](https://mozilla.invisionapp.com/share/STHUH4DF8UB#/294659434_icon___Text)
[Comp here - another example](https://mozilla.invisionapp.com/share/3RHUH34WFX8#/294659431_Image___Text_-scenario_2-)

**CMS Fields:**

- [ ] Upload image
- [ ] Checkbox if it is an icon
- [ ] Paragraph component styling (h2-h6, p and links)
- [ ] For medium images, caption and url (both optional)

**This will replace our current component:**
Image text

gvn:
[...]

PR #1608, opened 2018-06-19
new ImageTextBlock2 block - #1514

gvn:
Recreating the work done in #1585 as a new block since refactoring the existing one resulted in an incompatible migration.

Design has already been reviewed and approved.


Pomax:
Yeah, this looks good. We should probably see if we can help the wagtail folks along in implementing schema diffing (it's just JSON, "_how hard could it be(tm)_") so we can write normal migrations against streamfield changes =/

PR #1611, opened 2018-06-20
Update CI

patjouk:
Ref: MozillaFoundation/mofo-devops#597
Removing npm update since Travis now support `npm ci` out of the box.


patjouk:
Ok, nevermind, it works only with npm 5.8 (cf [Travis announcement](https://changelog.travis-ci.com/npm-ci-support-62786)) so we still need to update npm...

PR #1615, opened 2018-06-22
Delete accounts that are not staff

patjouk:
close MozillaFoundation/mofo-devops-private#140

Add a scheduled task that will run every week to remove any accounts that are not staff or with a '@mozillafoundation.org' email address.

@alanmoo I am kicking people out that I shouldn't with this rule?

Post-merge todo:
- [ ] Add Heroku Scheduler add-on with `python network-api/manage.py delete_non_staff` to prod


alanmoo:
As long as people who can log in and edit continue to be able to, even if they're not `@mozillafoundation.org`, we should be good. Test case is Luca's account.

patjouk:
I decided to remove apscheduler: I realized that the `clock` dyno would always be running, which is definitely overkill for a 10s script that would run every Sunday 😅 Instead, we should use Heroku Scheduler add-on that will spin a dyno for the duration of the script execution only.

@alanmoo the current setup would kick Luca's account. I'll check if people are part of a group or not too.

network-api/delete_non_staff.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def delete_non_staff():
+    from django.contrib.auth.models import User, Group
+    from django.db.models import Q
+
+    print("Deleting non staff users")
+    group_q = Group.objects.all()
+    non_staff = User.objects.exclude(
+        Q(email__contains='mozillafoundation.org') |
```
cadecairos: this should probably be `Q(email__endswith='@mozillafoundation.org') |`, which will ensure sneaky people don't get by with something like "trollolol+mozillafoundation.org@example.com" - which seems unlikely, but lets just protect against it anyways.

Pomax: I am totally going to start using that now.

patjouk: You're totally right, let's fix that XD

network-api/networkapi/tests.py
```diff
@@ -28,3 +29,64 @@ def test_no_migrations_missing(self):
...
+
+        self.assertEqual(User.objects.count(), 1)
+
+
+class MozillaFoundationUsersNotDeletedTest(TestCase):
+
+    def setUp(self):
+        User.objects.create(username='Alex', email='alex@mozillafoundation.org')
+
+    def test_mozilla_Foundation_users_not_deleted(self):
```
cadecairos: can you fix the capitalization of this test name?

patjouk: done!

PR #1624, opened 2018-06-27
Remove most of mezzanine

Pomax:
Let me start by saying this: 

image

This is going to be a multi-step process, where we first "empty out" all the apps that are mezzanine only, or were relevant in mezzanine context only, and then once everything looks good, do an update that removes migrations for emptied out apps, so we get rid of dumb things like one the migration files in the campaign app relying on a mezzanine.page object, and despite not having its own data anymore (because they're now done through wagtail), preventing `mezzanine.pages` from getting removed.

Also note that all the things that were accessible on the old `/admin` have been removed (but not the things living on `/api`) because the `/admin` route no longer exists. There is no access to the Django admin anymore, only the wagtail CMS admin. As such, JS and template overrides we had in place for making the Django admin look/interact nicer are now wholly irr-🐘


alanmoo:
>  irr-🐘

ಠ_ಠ

Anyway, given the multi-step-ness, are the warnings as part of the build log (visible on heroku when searching for the review app password) expected?

Pomax:
ooooh yeah okay this needs fixing - we call the /api/homepage route to kick off the actual homepage data react parsing run. So..... putting that back in for now.

Pomax:
Turned out that this was `main.js` was still set to ask about the old homepage before it did the react swapins, so I removed that check and just jump straight to react swapins, and took out the swapins for the old homepage elements, since those are generated by the CMS now.

network-api/networkapi/static/README.md
```diff
@@ -0,0 +1 @@
+why is this directory required?
```
gvn: LOL

alanmoo:
It seems like this might be a good PR for which to migrate a fully formed database for a bit more QA before we land it. Thoughts?

network-api/networkapi/middleware.py
```diff
@@ -1,13 +1,13 @@
...
-
-if settings.REFERRER_HEADER_VALUE:
-    referrer_value = settings.REFERRER_HEADER_VALUE
-
-
-class ReferrerMiddleware(MiddlewareMixin):
-    def process_response(self, request, response):
-        response['Referrer-Policy'] = referrer_value
-        return response
+from django.conf import settings
```
alanmoo: Was mezzanine.conf simply importing django.conf under the hood, and then adding some stuff?

Pomax:
So the one thing that's genuinely preventing removing mezzanine fully is the highlights model, which has a `footer` field that is of a type `mezzanine.core.fields.RichTextField`. Looking at the actual highlights we have in the system atm, none of them seem to actually be using rich text, rather they use plain text. Oddly enough, some are links, but in text form rather than genuine links, so.... maybe we can just turn this into a basic Django text field first, and then see if it needs to be rich text instead?

Pomax:
I need to un-remove campaigns: we're using that app for petition posting, which is **fairly critical**

alanmoo:
Highlights: we might be able to drop the footer field entirely. I think it was used when we had highlights on the homepage in a specific way, but isn't necessary anymore. Take a look at the templates that use highlights; If they don't use the footer, gut it.

Does campaigns use Mezzanine? What for?

network-api/networkapi/highlights/migrations/0007_nullify_homepage.py
```diff
@@ -0,0 +1,15 @@
...
+import wagtail.core.fields
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('highlights', '0006_auto_20180208_2205'),
+    ]
+
+    operations = []
```
gideonthomas: what is the purpose of this migration?

network-api/networkapi/news/migrations/0012_nullify_homepage.py
```diff
@@ -0,0 +1,15 @@
...
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('news', '0011_auto_20171012_1326'),
+    ]
+
+    operations = [
```
gideonthomas: what does this migration do?

network-api/networkapi/static/README.md
gideonthomas: it won't work without this directory? Probably because we set the `STATICFILES_DIR` setting. If we remove that, we won't need this folder anymore.

network-api/networkapi/urls.py
```diff
@@ -46,12 +49,11 @@
 
     # network API routes:
 
-    url(r'^api/people/', include('networkapi.people.urls')),
+    url(r'^api/campaign/', include('networkapi.campaign.urls')),
```
gideonthomas: why do we still need this url set?

network-api/networkapi/urls.py
```diff
@@ -46,12 +49,11 @@
...
+    url(r'^api/highlights/', include('networkapi.highlights.urls')),
```
gideonthomas: same question as above

network-api/networkapi/settings.py
```diff
@@ -529,3 +483,6 @@ def show_toolbar(request):
 # Review apps' slack bot
 GITHUB_TOKEN = env('GITHUB_TOKEN')
 SLACK_WEBHOOK_RA = env('SLACK_WEBHOOK_RA')
+
+# REQUIRED FOR AS LONG AS MEZZANINE HAS NOT BEEN FULLY PURGED
+TESTING = False
```
gideonthomas: where is this used?

network-api/networkapi/settings.py
```diff
@@ -258,18 +258,9 @@
                 'django.template.context_processors.static',
                 'django.contrib.auth.context_processors.auth',
                 'django.contrib.messages.context_processors.messages',
-                'mezzanine.conf.context_processors.settings',
-                'mezzanine.pages.context_processors.page',
-                'networkapi.context_processor.review_app',
```
gideonthomas: why is this removed?

network-api/networkapi/highlights/migrations/0007_nullify_homepage.py
Pomax: it allows safe modification of 006 to remove mention of mezzanine

network-api/networkapi/urls.py
Pomax: because the campaign app _does_ still do work for us. It's the POST handler for petition data to be forwarded to gsheets/salesforce via SQS

network-api/networkapi/urls.py
Pomax: we use highlights in wagtail.

network-api/networkapi/settings.py
Pomax: no idea: it's required by mezzanine for some purpose that we don't care about, since it'll get removed again when we do the second part of this work that takes out mezzanine entirely. Django/Wagtail don't need this value

network-api/networkapi/settings.py
Pomax: it should not have been. Restoring.

Issue #1627, opened 2018-06-27
Remove the client-side newsletter signup for campaign petitions.


Pomax:
For the new petitions this will be taken care of by Salesforce already.

Pomax:
This seems to still be in the codebase, in `source/js/components/join.jsx`.

PR #1628, opened 2018-06-27
basket cleanup for petitions

Pomax:
This removes the client-side newsletter signup for petitions, and closes https://github.com/mozilla/foundation.mozilla.org/issues/1627

### DIFF
--- a/source/js/components/petition/petition.jsx
+++ b/source/js/components/petition/petition.jsx
@@ -1,7 +1,6 @@
 import React from 'react';
 import ReactGA from 'react-ga';
 import classNames from 'classnames';
-import basketSignup from '../../basket-signup.js';
 import SALESFORCE_COUNTRY_LIST from './salesforce-country-list.js';
 
 class FloatingLabelInput extends React.Component {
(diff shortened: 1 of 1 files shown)


Pomax:
@alanmoo this simply removes the basket signup for new petitions, but also turns off client-side signup for old (now obsolete?) petitions. Are we okay with that are there strong reasons to keep newsletter signup for no longer relevant petitions working?

alanmoo:
No, I'm not comfortable with turning those off until we've migrated the existing petitions to CRM campaigns.

source/js/components/petition/petition.jsx
```diff
@@ -261,31 +259,6 @@ export default class Petition extends React.Component {
     });
   }
 
-  /**
-   * sign the user up for the mozilla newsletter.
-   *
-   * @returns {promise} the result the XHR post attempt.
-   */
-  signUpToBasket() {
-    this.setState({ basketSubmitted: true });
```
ScottDowne: I think `basketSubmitted` can be removed from other parts of the code too?

ScottDowne:
@Pomax just one potential comment?

Other than that the code looks fine, I'll let you and Alan worry about his comment.

source/js/components/petition/petition.jsx
Pomax: good point

Pomax:
actually landed in revised form as part of the petition refactor already

Issue #1629, opened 2018-06-27
add email validation to the petition email field


Pomax:
Right now, "cake" is an acceptable email address. And that doesn't even make sense!

desiree-mofo:
Has this fix been applied to staging? I tested this today (Jul 01) and it's still accepting invalid email addresses

desiree-mofo:
@alanmoo  - do you know what the hold up is on getting this issue resolved? I believe the code fix was being worked on by @Pomax as of Tues, but he's out today and I need some progress on this as it's the final issue holding up Flash at the moment.

PR #1630, opened 2018-06-27
added simple email validation

Pomax:
Fixes https://github.com/mozilla/foundation.mozilla.org/issues/1629

It'd be nice to cache the `...GroupClass` conditionals and use those in place of the same conditionals used in `render` to show the warning text, but that's a refactor that we should probably do outside this PR


gvn:
Is the input for email not type `email`? We should be able to leverage more robust built in native validation:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/email

PR #1640, opened 2018-06-29
More official wagtail inventory

alanmoo:
The main wagtail-inventory repo hasn't landed a proper release yet, but it works locally in our project.

There is a command that needs to be run on the server once this hits- `./manage.py block_inventory`. @patjouk Do you have thoughts on where that needs to happen?


patjouk:
That management command needs to happen at the release step of the [Procfile](https://github.com/mozilla/foundation.mozilla.org/blob/master/Procfile#L1), after the migrate step.

Pomax:
It looks like stuff landed since we opened this: I just installed the wagtail-inventory master branch, and added their app to settings and then ran `migrate` + `block_inventory`, and things seem to work perfectly fine.

alanmoo:
So does this need an update?

Pomax:
We do need an update: I asked if a new version could be published, so we should be able to set the version to ==0.5 rather than using a git url, and everything should work great!

Pomax:
@alanmoo can I lob this back at you for testing? This works for me, if it works for you as well I say we merge this.

alanmoo:
Sure, but is the deploy failing to build because of Github issues or something else?

Pomax:
```
       Your Pipfile.lock (87fee9) is out of date. Expected: (6ea965).
       Aborting deploy.
```
let me push a lock update

Pomax:
One interesting problem here is that our fake data doesn't have streamfields, so there's no way to truly verify this works except on staging =)

(beyond verifying it gets installed properly and adds the settings -> block inventory option)

Pomax:
@alanmoo gtg as far as I'm concerned but you probably want to counter-gtg that before we hit merge.

alanmoo:
Review app works for me!

Pomax:
For unclear reasons, this ran into the following errors on staging:
```
usage: manage.py block_inventory [-h] [--version] [-v {0,1,2,3}]
                                 [--settings SETTINGS]
                                 [--pythonpath PYTHONPATH] [--traceback]
                                 [--no-color]
manage.py block_inventory: error: unrecognized arguments: --no-input
```

I've filed https://github.com/cfpb/wagtail-inventory/issues/15 to see if the maintainer of the library is interested in adding support for `--no-input`, and hopefully that results in a new version that lets us add this plugin as part of the automated deployment.

Issue #1588, opened 2018-06-05
Hide "subscribe" option on petitions if user came from an email


alanmoo:
MVP (as there are probably a bunch of ways this will not be effective)
If the URL contains a query parameter `subscribed=1`, we should not show the "sign up for the newsletter" checkbox on petitions

taisdesouzalessa:
Really like this idea, such a good way to remove unnecessary info for specific users.

desiree-mofo:
@Pomax  and @ScottDowne  - I think this is what we were discussing last week in our petition meeting? Wanted to make sure y'all were aware of this issue. I was going to enter something similar, but think we can work from this one...

ScottDowne:
@desiree-mofo Question on this, if the signup is removed, does that mean the country list should go first?

I guess it comes down to if the country is used for the signature or the signup or both. If it's only used for signup, we should probably remove it too.

ScottDowne:
It looks like the only thing we're removing is the checkbox that says "Yes, I want to receive email updates about Mozilla’s campaigns." and everything else stays the same.

PR #1652, opened 2018-07-04
subscribed query string for petitions.

ScottDowne:
Fixes https://github.com/mozilla/foundation.mozilla.org/issues/1588

### DIFF
--- a/source/js/components/petition/petition.jsx
+++ b/source/js/components/petition/petition.jsx
@@ -575,12 +575,14 @@ export default class Petition extends React.Component {
                 {this.state.userTriedSubmitting && !this.refs.privacy.checked && <small className="has-danger">Please check this box if you want to proceed</small>}
               </label>
             </div>
-            <div className="my-2">
-              <label className="form-check-label">
-                <input disabled={disableFields} type="checkbox" className="form-check-input" id="PrivacyCheckbox" ref="newsletterSignup" />
-                <span className="h6-heading form-text">Yes, I want to receive email updates about Mozilla’s campaigns.</span>
-              </label>
-            </div>
+            { this.props.subscribed ? null :
+              <div className="my-2">
+                <label className="form-check-label">
+                  <input disabled={disableFields} type="checkbox" className="form-check-input" id="PrivacyCheckbox" ref="newsletterSignup" />
+                  <span className="h6-heading form-text">Yes, I want to receive email updates about Mozilla’s campaigns.</span>
+                </label>
+              </div>
+            }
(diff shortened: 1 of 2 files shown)


source/js/components/petition/petition.jsx
```diff
@@ -575,12 +575,14 @@ export default class Petition extends React.Component {
...
-            <div className="my-2">
-              <label className="form-check-label">
-                <input disabled={disableFields} type="checkbox" className="form-check-input" id="PrivacyCheckbox" ref="newsletterSignup" />
-                <span className="h6-heading form-text">Yes, I want to receive email updates about Mozilla’s campaigns.</span>
-              </label>
-            </div>
+            { this.props.subscribed ? null :
+              <div className="my-2">
+                <label className="form-check-label">
+                  <input disabled={disableFields} type="checkbox" className="form-check-input" id="PrivacyCheckbox" ref="newsletterSignup" />
```
Pomax: let's just change this ID at the same time. It should be NewsletterSignup, not PrivacyCheckbox

source/js/components/petition/petition.jsx
```diff
@@ -575,12 +575,14 @@ export default class Petition extends React.Component {
                 {this.state.userTriedSubmitting && !this.refs.privacy.checked && <small className="has-danger">Please check this box if you want to proceed</small>}
               </label>
             </div>
-            <div className="my-2">
-              <label className="form-check-label">
-                <input disabled={disableFields} type="checkbox" className="form-check-input" id="PrivacyCheckbox" ref="newsletterSignup" />
-                <span className="h6-heading form-text">Yes, I want to receive email updates about Mozilla’s campaigns.</span>
-              </label>
-            </div>
+            { this.props.subscribed ? null :
```
Pomax: This also needs a change to L233 now, which is currently `newsletterSignup: !!(this.refs.newsletterSignup.checked),` - this will throw when we don't have that element on the page, because `this.refs.newsletterSignup` won't exist, and so trying to access `checked` will cause a TypeError.

`newsletterSignup: this.props.subscribed || !!(this.refs.newsletterSignup.checked),` should do the trick.

ScottDowne: Good catch, but I might fix it in a slightly different way.

This would attempt to sign them up again (I think), probably wouldn't fail because they email they use is probably already in the system (and if it's not, would be signed up unexpectedly(I have two emails and don't always pay attention to which I use for what)) which is at best redundant, but at worst unexpected.

Pomax:
lgtm!

Issue #1639, opened 2018-06-29
[Pulse component on Wagtail] Implementation ticket - for shipping 📦


taisdesouzalessa:
Based on conversation on: #1617

### Here is the comp:
![animation-pulse-component](https://user-images.githubusercontent.com/22157921/42108724-2ba91872-7b90-11e8-9eff-34bd23a98024.gif)

### InVision:
https://mozilla.invisionapp.com/share/QCM7H95P2BK#/306182735_CMS_-_Projects_-_V6_-v4_Updated-

### Checklist:

- [ ] Search section: add label `Search:`, placeholder copy `Search across title, description, and tags`, helper text `Try your search first at mozillapulse.org/search.`

- [ ] Number of results section: add label `Max number of results:`, default number `6` and helper text `Choose 1-12. If you want visitors to see more, link to a search or tag on Pulse. `

- [ ] Featured section: add label `Featured:`, checkbox (default empty)`Display only featured entries` and helper text `Featured items are selected by Pulse moderators.`

- [ ] Horizontal hairline

- [ ] Advanced filters section: Label `ADVANCED FILTERS:` Copy: `Options to display fewer, more targeted results`

- [ ] `Issues:` Dropdown box with `All` chosen by default (if users don't choose an option, all will apply). Once the dropdown arrow is clicked, the full list is displayed in alphabetical order: `All` (in light grey, since it is the selected option), `Decentralization`, `Digital Inclusion`, `Online Privacy & Security`, `Open Innovation` and `Web Literacy` - If users choose another option, the list remains the same and the chosen option is displayed in the box and it is light grey in the expanded view of the list.

- [ ] `Type of help needed:`  Dropdown box with `All` chosen by default (if users don't choose an option, all will apply). Once the dropdown arrow is clicked, the full list is displayed in alphabetical order: `All` (in light grey, since it is the selected option), `Attend`, `Create content`, `Code`, `Design`, `Fundraise`, `Join community`, `Localize & translate`, `Mentor`, `Plan & organize`, `Promote`, `Take Action`, `Test & feedback`, `Write documentation`- If users choose another option, the list remains the same and the chosen option is displayed in the box and it is light grey in the expanded view of the list.

- [ ] `Sort:`  Dropdown box with `Find and show the most recent` (selected by default). Once the dropdown arrow is clicked, 2 options are displayed: `Find and show the most recent` (in light grey, since it is the selected option) and `Find and show the oldest`

[...]

PR #1668, opened 2018-07-10
Revised Pulse Project Component

gvn:
Related issue: #1639


gvn:
@taisdesouzalessa if you want to try this out in the CMS you can go to https://foundation-mofostaging-pr-1668.herokuapp.com/cms/

gvn:
Made changes where possible.

taisdesouzalessa:
Design review: 💯

gvn:
@Pomax I'm trying to determine if this component schema change is safe or going to blow things up again. It's mostly additive FWIW...

Worst case I can leave the existing `PulseProjectList` intact and move these tweaks to a new `PulseProjectList2`.

Pomax:
Styling looks a bit odd in my FF, but codewise this should be safe. We're redeclaring newest_first but the data in it stays "the same" so pages that had it set to `True` or `False` stay set to that as far as I can tell.

FF screenshot:

image

PR #1669, opened 2018-07-10
updated wagtail to 2.1.1

Pomax:
This also removes the need for our own jquery datepicker copy, which we had put in so that we wouldn't have to allow `unsafe-eval` in our Content Security Policy.


Pomax:
Why does installing wagtail 2.1.1 remove all those `markers` from the lock file...? O_o

Would you know, @patjouk?

patjouk:
@Pomax I don't have time to investigate (stuff to finish before going on holidays!) but you should check this PEP about markers: https://www.python.org/dev/peps/pep-0496/

Pomax:
hm, that PEP is marked as rejected, raising some more questions about why they were there in the first place. I guess if we don't see any difference, their removal probably doesn't matter.

PR #1673, opened 2018-07-16
package json bump

gvn:


gvn:
(Whoever gets to the review first...)

PR #1678, opened 2018-07-18
Add nav indicator for private pages

alanmoo:
STR:

Open campaign or opportunity multi-page minisite in editor.
Change visibility of a page so it’s only available to logged in users
View live version of that page (and siblings)
🙈

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/mini_site_horizontal_nav.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/mini_site_horizontal_nav.html
@@ -4,7 +4,7 @@
 <div class="mb-4 hidden-sm-down">
   <div class="pb-2 multipage-nav horizontal-nav" id="multipage-nav">
     {% for entry in menu_pages %}
-    <a class="pb-2 multipage-link horizontal-link mr-4 {% if current.title == entry.page.title %}multipage-link-active{% endif %} {% if entry.restriction != None %}multipage-link-restricted{% endif %}" href="{% pageurl entry.page %}">{% if entry.depth == 2 %}→ {% endif %}{{ entry.menu_title }}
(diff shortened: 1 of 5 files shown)


network-api/networkapi/wagtailpages/templates/wagtailpages/tags/mini_site_horizontal_nav.html
```diff
@@ -4,7 +4,7 @@
 <div class="mb-4 hidden-sm-down">
   <div class="pb-2 multipage-nav horizontal-nav" id="multipage-nav">
     {% for entry in menu_pages %}
-    <a class="pb-2 multipage-link horizontal-link mr-4 {% if current.title == entry.page.title %}multipage-link-active{% endif %}" href="{% pageurl entry.page %}">{% if entry.depth == 2 %}→ {% endif %}{{ entry.menu_title }}
+    <a class="pb-2 multipage-link horizontal-link mr-4 {% if current.title == entry.page.title %}multipage-link-active{% endif %} {% if entry.restriction != None %}multipage-link-restricted{% endif %}" href="{% pageurl entry.page %}">{% if entry.depth == 2 %}→ {% endif %}{{ entry.menu_title }}
```
Pomax: it feels like we should tap into the string that's in `restriction`, too. rather than setting `multipage-link-restricted`, can we  set something like `multipage-{{restriction}}-restricted` and then define three CSS rules for the three options? (having a 🙈  for private, 🔒 for password locked, and 👥 for groups or something? (aaaand on that note, why does github only offer "family_man_...")

source/sass/components/nav.scss
```diff
@@ -234,6 +234,22 @@
...
+  .multipage-groups-restricted::after {
+    content: '👥';
+  }
+
+  .multipage-password-restricted::after {
+    content: '🔒';
+  }
+
+  .multipage-draft::after {
+    content: '🐣';
```
Pomax: not quite seeing the connection to "draft", but frankly I also don't care enough to go "no! fix this!" =D

alanmoo: It’s “partially hatched” 😂

It was Matthew's suggestion 🙂

PR #1680, opened 2018-07-18
donation modal for campaign petitions

### LINKED ISSUE #1589: Add fundraising modal after petition signing
Goal: CMS authors have the ability to decide if a modal asking for donations should appear after users sign a petition.

Requirements:
* The copy of the modal (including CTA button) can be changed on a per instance basis using the CMS.
* When users click the CTA, a new tab is opened to the donate site
    * The link to the donate site passes along the campaign ID and information that identifies which modal version the user came from (for A/B testing purposes), as well as the user's email subscription status if they came from an email (#1588) or signed up during the petition.
* When the user comes back to the campaign page, the modal will be closed with the "share petition" links visible where the petition originally was located.

Suggestions:
Based on the first requirement, my instinct is that the modal content should be its own snippet model, so that CMS authors can simply attach it to a campaign so it happens, or not attach it.

Design:
https://redpen.io/xf5d9f47242418277e
https://redpen.io/sc98112fcc645c8449

Placeholder content:
(headline:) Thanks for signing! While you're here, we need your help.

Mozilla is a nonprofit organization fighting for a healthy internet, where privacy is included by design and you have more control over your personal information. We depend on contributions from people like you to carry out this work. Can you donate today?

(Button:) Yes, I'll chip in
No, I'll share instead 

This is synthesized from the conversation in https://github.com/mozilla/donate.mozilla.org/issues/1999

Pomax:
Closes #1589

This adds a donation modal to any campaign petition that is marked as having a modal attached, by:

- adding a Donation snippet 
- adding the ability to tie a Donation snippet to a campaign page
- having the presence of a donation snippet trigger donation modals on the campaign page when a petition is successfully submitted

## page example

The following is a screenshot of the in-CMS part of this work, giving campaign pages a "Donation modal" section where one or more donation-requesting modals can be bound to the campaign.

image

Campaigns can have more than one, so that we can do A/B testing around things like copy.

## modal snippet

The donation modal itself is defined by data that we can specify as a snippet, similar to petitions themselves. One modal can be used by multiple campaign pages (in the same way that a single petition can be used by multiple campaign pages).

image

## styling

The styling is bootstrap, with tweaks to get it as close to the comps as I can.

image

The only open issue styling wise that I can see is that the modal body text is not using the right font, but it doesn't look "off" if you don't know what the font is supposed to be, so this would probably not be something to block the PR on if we want to get a modal up before @kristinashu is back from PTO


gvn:
Has this had a design review?

network-api/networkapi/campaign/views.py
```diff
@@ -11,6 +11,8 @@
 
 from networkapi.wagtailpages.models import Petition
 
+gs_sqs = False
```
Pomax: changes to this file are mostly because I needed to do localhost testing and these changes are critical for that. I can file this as separate PR if desired, though.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -321,9 +324,19 @@ class CampaignPage(MiniSiteNameSpace):
         help_text='Choose existing or create new sign-up form'
     )
 
+    def get_donation_modal_json(self):
+        modals = self.donation_modals.all()
```
Pomax: This is the fun part, from a campaign perspective.

source/js/components/petition/donation-modal.jsx
```diff
@@ -0,0 +1,70 @@
...
+          <h3 className={ classNames(`h3-heading`, `text-center`) }>
+            { this.props.heading }
+          </h3>
+          <p className={ classNames(`body-large`, `text-center`) }>
+            {this.props.bodyText}
+          </p>
+        </div>
+
+        <div className="text-center">
+          <a className="btn btn-normal" href={this.donateURL} target="_blank" onClick={e => this.userElectedToDonate(e)}>
```
Pomax: I've made this an `<a>` because it means people get to see where this will lead them with a mouseover, but we can make this a `<button>`, too. As an actual navigating element, though, `<a>` feels way more semantically correct.

source/js/components/petition/petition.jsx
```diff
@@ -1,78 +1,15 @@
...
-
 export default class Petition extends React.Component {
   constructor(props) {
     super(props);
     // Default props defined at end of file
 
-    // TODO: this needs a better solution, because explicitly binding
-    //       ten+ functions is kind of silly. The code should already
-    //       guarantee the correct `this` will be used.
-    this.submitDataToApi = this.submitDataToApi.bind(this);
```
Pomax: using arrow functions in event handling means all these this binders can go, which is super pleasant.

source/js/components/petition/petition.jsx
```diff
@@ -348,6 +319,14 @@ export default class Petition extends React.Component {
     return input.match(/[^@]+@[^.@]+(\.[^.@]+)+$/) !== null;
   }
 
+  userElectedToDonate() {
+    this.setState({ showDonationModal: false });
```
Pomax: I assume we want analytics here, but I don't know which events and what payload those should have.

source/js/components/petition/petition.jsx
```diff
@@ -348,6 +319,14 @@ export default class Petition extends React.Component {
...
+  }
+
+  userElectedToShare() {
+    this.setState({ showDonationModal: false });
```
Pomax: I assume we want analytics here, but I don't know which events and what payload those should have.

source/js/components/petition/petition.jsx
```diff
@@ -404,10 +383,33 @@ export default class Petition extends React.Component {
...
       </div>
     );
   }
 
+  /**
+   * This renders a donation modal on the page as full-page-overlay,
+   * provided the petition HTML specifies that as a thing that should happen.
+   */
+  renderDonationModal() {
+    // This is where can do client-side A/B testing
```
Pomax: More fun to be had, from a campaign perspective

source/js/components/petition/petition.jsx
```diff
@@ -1,78 +1,14 @@
 import React from 'react';
 import ReactGA from 'react-ga';
 import classNames from 'classnames';
+import DonationModal from './donation-modal.jsx';
+import FloatingLabelInput from './floating-label-input.jsx';
+import CountrySelect from './country-select.jsx';
 import basketSignup from '../../basket-signup.js';
-import SALESFORCE_COUNTRY_LIST from './salesforce-country-list.js';
-
-class FloatingLabelInput extends React.Component {
```
Pomax: this class and the next kept being incredibly in the way so I put them in their own files.

source/js/components/petition/petition.scss
```diff
@@ -6,7 +8,8 @@
 .cta {
   .sticky-cta {
     position: sticky;
-    top: 57px;
+    top: 5rem;
```
Pomax: the header does not appear to have a fixed height, rather being defined by the text in it, so I made this top offset 5rem to match better - without it, the CTA would end up scrolling on top of/underneath the header, which looked really janky.

gvn:
Generally for front end PRs our process is now:

1. Flag a designer to review the staged build and address any feedback.
2. Once design gives a R+ then assign a dev to review as normal.

Pomax:
@gvn our designer is deep in the woods though, so in this case getting the code reviewed made the most sense. There is plenty of not-styling code that needs to work and can be reviewed independent of design review.

Pomax:
@alanmoo that is.. odd, The review app does not look the way it's supposed to. I'll have to dig into the why.

Pomax:
The review app had desynced somehow, and was many commits behind what it should have been. I've recreated the review app and things should a lot more normal @alanmoo @gvn @gideonthomas

alanmoo:
I did a little bit of digging- It seems like it's `position: -webkit-sticky` that's causing problems for the post-sign modal. But it's used to keep the petition on the page with longer content, as I added [here](https://foundation-mofostaging-pr-1680.herokuapp.com/campaigns/multi-page/) to simulate a real petition. Maybe the div that kicks off the modal needs to be moved out of the sticky container?

source/js/components/petition/petition.jsx
```diff
@@ -1,78 +1,14 @@
...
-        </select>
-      </div>
-    );
-  }
-}
 
 export default class Petition extends React.Component {
   constructor(props) {
     super(props);
-    // Default props defined at end of file
```
alanmoo: Why does this work now?

Pomax: What do you mean "now"?

Pomax: using proper arrow functions in the handle calls.

PR #1681, opened 2018-07-19
update share progress required classes

Pomax:
Fixes shareprogress not actually generating share links for us


network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
```diff
@@ -55,7 +55,6 @@ <h1>{{ block.value }}</h1>
             {% cta page %}
           </div>
         </div>
-        <script src="https://c.shpg.org/352/sp.js"></script>
```
Pomax: we already have this in the petition.html template code. Loading it twice makes no sense.

alanmoo:
Code looks reasonable

PR #1687, opened 2018-07-20
Wagtail inventory

Pomax:
Reinstated PR for https://github.com/mozilla/foundation.mozilla.org/pull/1640, which had to be backed out over a runtime flag incompatibility on staging.

/cc @alanmoo


Procfile
```diff
@@ -1,2 +1,2 @@
-release: cd network-api && python ./manage.py migrate --no-input
+release: cd network-api && python ./manage.py migrate --no-input && python ./manage.py block_inventory --no-input
```
Pomax: This `--no-input` flag breaks the staging deploy. I've filed cfpb/wagtail-inventory#15 to see if the maintainer of the library is interested in adding support for this flag, and hopefully that results in a new version that lets us add this plugin as part of the automated deployment.

alanmoo: So this can't be reviewed until then?

alanmoo:
Wait, I thought we were waiting for a change upstream. Why would this work if #1640 didn't?

patjouk:
From what I understood, we don't need that "no-input" flag anymore. Can you confirm @Pomax ?

Pomax:
indeed, after conferring with the library author, we should be able to run without `--no-input`

PR #1691, opened 2018-07-23
Actually remove mezzanine

Pomax:
The prep work for this PR landed last week, which means this PR can finally be filed: this PR squashes the migrations for apps that used to rely on mezzanine, erasing all mention of it, removes mezzanine from our settings.py, and removes it from our pipfile.

This still leaves the `networkapi/homepage` and `networkapi/landingpage` directories, which will be the last piece of follow-up (you can't delete the directory _and_ have their squashed migration run, so... )

> One thing that we **really** need to test is what happens with S3 files now. Can we safely rip out mezzanine's `filebrowser_safe`, with our own `s3_filebrowser` on top, or will that lead to massive problems?

we removed `glyph` from the news model for now.

We still need `filebrowser_safe`, `filebrowser_s3`, and `future` in the dependency list for now thanks to old migrations, but after this lands, we can delete the old migrations, leave the squash, and remove `filebrowser_safe` and `future` again.


alanmoo:
This does not seem to want to deploy. Related to #1711?

Pomax:
@alanmoo it was the pipfile.lock being out of sync. Running into this quite a bit.

PR #1692, opened 2018-07-23
Remove mezzanine apps/files after prod has been updated with #1691

Pomax:
https://github.com/mozilla/foundation.mozilla.org/pull/1691 removes mezzanine and blanks several migrations that used to involve it - once that has landed into prod, we can land this PR, which removes files that no longer serve any purpose, but still had to exist in order to have Django pick up on the fact that they got migrated to "empty", etc.


Pomax:
CI passes, heroku build succeeds, everything looking good.

PR #1699, opened 2018-07-30
Buyers guide MVP

alanmoo:
Add model support for Buyer's Guide as a separate app. This way content teams can add content while we build out templates for this year/migrate previous templates. It's restricted to access by logged in users.


network-api/networkapi/buyersguide/models.py
```diff
@@ -0,0 +1,90 @@
...
+        help_text='Price range',
+        blank="True",
+    )
+    image = models.FileField(
+        max_length=2048,
+        help_text='Image representing this prodct',
+        upload_to=get_product_image_upload_path,
+        blank=True,
+    )
+    camera = models.NullBooleanField(
```
Pomax: why do we need true, false, or null?

network-api/networkapi/buyersguide/views.py
```diff
@@ -0,0 +1,16 @@
+from django.shortcuts import render
+from networkapi.buyersguide.models import Product
+from django.contrib.auth.decorators import login_required
+
+
+# Create your views here.
+@login_required
```
Pomax: is this "for now" or are these pages supposed to only be visible to ourselves?

network-api/networkapi/buyersguide/models.py
alanmoo: Because the answer to some of these is "unknown". This allows editors to leave the field unfilled.

network-api/networkapi/buyersguide/views.py
alanmoo: For now- basically a feature flag so we can merge this in without risk of it being stumbled upon.

Pomax: Can we put a comment in here with a refernce to a github issue along the lines of "the login requirement for buyersguide"?

Issue #1648, opened 2018-07-03
Implement Initiatives page


kristinashu:
Based on design in https://github.com/mozilla/foundation.mozilla.org/issues/1605
Comps https://redpen.io/p/er1b89aec9ac7facb2

Hero banner
- image, title, body
- rarely edited

Sub nav
- use [current style](https://foundation.mozilla.org/about/) for now (update can be done later https://github.com/mozilla/foundation.mozilla.org/issues/1156)
- add external link icon
- links: [MozFest](https://mozillafestival.org/), [Fellowship](http://foundation.mozilla.org/fellowships/), [Internet Health Report](https://mzl.la/ihr), [Advocacy](https://advocacy.mozilla.org/)

4 themes
- each has image, title, body, button with external link icon 
- rarely edited
- [animated transition similar to this example](http://werkstatt.fuelthemes.net/werkstatt-fullscreen-split/)

Featured programs
- title + intro text (sometime edited)
- 3-6 cards (sometimes edited) (same as home page Happening Now cards but 3-up instead of 2-up)
- 3-6 mini cards (sometimes edited)

### Comps
![initiatives ui 2x](https://user-images.githubusercontent.com/20977376/42648876-127d5606-85d6-11e8-821e-213d45b7eef7.jpg)

kristinashu:
Gavin has done all the P1's for Initiatives.

Opened up a new ticket for P2's https://github.com/mozilla/foundation.mozilla.org/issues/1718 and a separate ticket for Participate page https://github.com/mozilla/foundation.mozilla.org/issues/1717

kristinashu:
@gvn please close this ticket if that all sounds good to you

gvn:
I'm going to leave it open until we get it landed, but yeah...should be soon 🤞

gvn:
Landed! https://foundation.mozilla.org/initiatives/

![jenga](https://user-images.githubusercontent.com/206379/43728817-16c9e826-995b-11e8-9c94-27a21b336d4d.gif)

PR #1712, opened 2018-08-02
New Initiatives Page (Take 2)

alanmoo:
Related: #1648

@gvn we had to roll this back yesterday. Opening this up again to roll forward, but there's a bug that needs fixing

- [x] #1714 Secondary Nav


beccaklam:
@gvn Also ... wondering if we could change the initiatives paragraph div styling to:
<div class="col-12 col-md-5 px-4 py-5 px-md-0 px-lg-3 align-self-center></div>
and the image div to:
<div class="col-12 col-md-5 offset-md-1"></div>
We can also chat about this over vidyo just to make sure it looks right before pushing the commit.

alanmoo:
@beccaklam Please file that in another ticket to make it easier to address individually, and edit the original comment in this ticket to add that number if it's a launch blocker.

gvn:
@alanmoo A fix for #1714 should be in here now

gvn:
AFAIK this is good to merge but I'll hold off until we get the missing content, which is:

- [ ] What highlights are featured?
- [ ] Copy for "Sub H3", which is lorem ipsum in the comp. We can also leave this blank.

alanmoo:
They can be created on the live site if they don't already exist there, as the highlights model already exists there. That way, once the Initiatves page makes it live, someone just has to point to those highlights to hook them up.

https://foundation.mozilla.org/cms/highlights/highlight/

beccaklam:
Sydette has provided all the content we need in the aforementioned doc!

gvn:
Awesome. I'm merging this and I'll also put the initial content in the CMS this time around.

We'll plan to do a prod push Monday. 🎉

PR #1713, opened 2018-08-02
doubling outputted image size

alanmoo:
I'm not sure why a PR into the Initiatives branch isn't looking good, but this should end up back in there after #1712


gvn:
manually recreating in #1712

Issue #1750, opened 2018-08-20
Add comment field to petitions


alanmoo:
Petitions should have an optional comment field. Three states:

* Comments off
* Comments on, but optional
* Comments required

This passes to basket, with a key that @pmac will confirm (I believe "comments").

Pomax:
field name for sending to basket documented over on https://github.com/mozmeao/basket/issues/87

PR #1751, opened 2018-08-20
add optional comment field to petition

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/1750

This PR adds a `comment_requirements` selector to petitions with options `none`, `optional`, and `required`, defaulting to `none`. These values have the following effect:

- `none`: no comment field is presented in the petition at all
- `optional`: a comment field is presented in the petition, but it need not be filled in
- `required`: a comment field is presented in the petition, and the petition cannot be submitted unless it has content.

This value is sent to the campaign data handler in `networkapi/campaigns/view.py` where it is copied over as `comment` attribute as per discussions in slack

Test pages:

- any of the predefined campaigns (no comment set)
- https://foundation-mofostaging-pr-1751.herokuapp.com/cms/pages/38/edit/
  - using https://foundation-mofostaging-pr-1751.herokuapp.com/cms/snippets/wagtailpages/petition/15/ (optional comment)
- https://foundation-mofostaging-pr-1751.herokuapp.com/cms/pages/39/edit/
  - using https://foundation-mofostaging-pr-1751.herokuapp.com/cms/snippets/wagtailpages/petition/16/ (required comment)

Live versions:

- https://foundation-mofostaging-pr-1751.herokuapp.com/en/campaigns/comments-optional/
- https://foundation-mofostaging-pr-1751.herokuapp.com/en/campaigns/comments-required/


source/js/components/petition/floating-label-textarea.jsx
```diff
@@ -0,0 +1,22 @@
+import React from 'react';
```
Pomax: It would be nice to unify this with `floating-label-input.jsx` as the only difference is really the jsx element type, which we could just throw into the constructor. That said, it does make importing a little more muddled.

Pomax:
good catch. I'll have to see why it's doing that (and also what I messed up in this rebase because the migrate is not kicking in correctly)

gvn:
Looks like the migration needs to be redone since I just landed one.

Good opportunity to give feedback on [this documentation](https://github.com/mozilla/foundation.mozilla.org/pull/1763)?

PR #1753, opened 2018-08-21
Add 2019 fellows directory option by default

alanmoo:
@annmarie123 I assume we want this?

cc @Sydette


annmarie123:
@alanmoo Yes please! All new fellows should have pulse profiles by now. I was planning to go into their accounts and set them to 2019 and fellows so they will automatically appear in the directory, but if this can be done automatically, let me know.

alanmoo:
@annmarie123 you do need to flip their accounts individually. I did the first one yesterday to test out this PR. (You may need to check the "Extended info" checkbox if just setting the program/year doesn't work)

PR #1763, opened 2018-08-23
probably bad instructions on resolving migration conflicts

gvn:


Pomax:
This looks about right to me. If you need to go nuclear you can also empty the entire migrations dir and then (with an up to date master) say `git checkout master migrations`, after which you can run `makemigrations` but that probably doesn't need to be in the readme?

README.md
```diff
@@ -126,6 +126,19 @@ You will now also need to update your `.env` file to make sure you're using this
 
 If you need to reset this database, rerun step 2 (with `dropdb foundation` as first command) through 5 to get back to a clean copy of the production database.
 
+
+### Resolving conflicting Django migrations
+
+**AKA: What to do when a migration lands before yours**
+
+- Create a new, separate local instance of `foundation` per "Setup steps" above.
```
Pomax: Hmm, this doesn't quite feel required. You should be to do this in your branch without setting up a new copy.

Pomax: meh, let's just merge anyway =D

Issue #1749, opened 2018-08-20
Participate page: design review P2's


kristinashu:
Follow up from https://github.com/mozilla/foundation.mozilla.org/pull/1748

### Time commitment tags 
- [x] Hand should be transparent but it looks like it has white fill (let me know if you need a different svg?)
<img width="198" alt="image" src="https://user-images.githubusercontent.com/1068783/44356435-32288000-a464-11e8-970b-2ac949c4fa5d.png">

### Top hero banner ([see comp](https://redpen.io/ctffa61f18d45a7825) or Abstract)
- [x] Paragraph text should be .body-large

### Top hero banner Mobile
- [x] Please stack the image above the text  (see comp below)
![screen-shot-2018-08-20-at-10 38 12-am](https://user-images.githubusercontent.com/1068783/44358291-6bafba00-a469-11e8-9bc6-619aeec0c481.jpg)

### Intro text on Tablet
- [x] Make text full width (see comp below)
<img width="768" alt="image" src="https://user-images.githubusercontent.com/1068783/44363562-6195b780-a479-11e8-8398-8b2020f9232b.png">

### Donate & Share CTA
- [x] Paragraph text should be .body-large
- [x] Less padding above and more below the gray box
<img width="1187" alt="image" src="https://user-images.githubusercontent.com/1068783/44357746-b8929100-a467-11e8-9625-99d179647de9.png">

### Donate & Share CTA Mobile
- [x] Put image above the text (see comp below)
<img width="319" alt="image" src="https://user-images.githubusercontent.com/1068783/44357902-3a82ba00-a468-11e8-814a-773cc1f1b7c0.png">

### Bottom Evergreen Cards
- [x] Paragraph text should be .body-large
- [x] Is it possible to vertically center the right side content? (see comp below) 
<img width="715" alt="image" src="https://user-images.githubusercontent.com/1068783/44363375-e7fdc980-a478-11e8-88a8-c310bb0407a1.png">

### Bottom Evergreen Cards Tablet
- [ ] If possible, make image square (see comp below) 
<img width="572" alt="image" src="https://user-images.githubusercontent.com/1068783/44363957-92c2b780-a47a-11e8-8380-389df3382bb9.png">

### Bottom Evergreen Cards Mobile
- [x] image should be full width and if possible look like regular card (see comp below)
[...]

PR #1785, opened 2018-08-27
Participate P2 Revisions

gvn:
Related issue: #1749


gvn:
@kristinashu We're having problems uploading images to staged PRs right now, so I took screenshots of 3 breakpoints for your design review:

localhost_8000_cms_pages_38_edit_preview_320
localhost_8000_cms_pages_38_edit_preview_780
localhost_8000_cms_pages_38_edit_preview_1200

kristinashu:
Thanks for the screen caps! The only thing that looks slightly off is the Donate/Share cta on mobile. I like that the image is on top of the text but the gap between the two (below the image and above the text "If you can, will you give today?") seems a little too big so it takes up more vertical space than needed. Is it possible to reduce that gap? Otherwise, everything else is looking good!

Screen cap of the screen cap:

image

gvn:
How's this?

localhost_8000_cms_pages_38_edit_preview_

kristinashu:
🚢 it!

PR #1804, opened 2018-09-06
PNI Scaffolding

gvn:
@alanmoo – This is obviously a work in progress, but I figure it's better to land in chunks (and safe since it's non-visible to the public and fairly sandboxed).


gvn:
Whoops forgot @alanmoo is out. 

@Pomax can you review?

Pomax:
Interesting bit of scrollbar on windows: 

image

network-api/networkapi/buyersguide/templates/bg_base.html
```diff
@@ -0,0 +1,65 @@
...
+        </ul>
+        <div class="org-info mozilla">
+          <div class="logo-container"><a class="logo" href="https://mozilla.org"></a></div>
+          <p>Mozilla is a global non-profit dedicated to putting you in control of your online experience and shaping the future of the web for the public good. Visit us at <a href="https://mozilla.org">mozilla.org</a>.</p>
+        </div>
+      </div>
+    </footer>
+  </body>
+</html>
+<script src="/_js/bg-main.compiled.js"></script>
```
Pomax: technically this is invalid HTML since nothing's allowed to be between `</body></html>` or after `</html>` - given the current state of browsers, could we just move it to before `</body>` and mark it as `defer` so that it doesn't actually load until after the DOM has been fully parsed?

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -1,19 +1,31 @@
...
+
+{% block guts %}
+
+<div class="container">
+  <h2>{{product.name}}</h2>
+
+  <div class="creepometer mb-5"></div>
+
+  <img src="{{mediaUrl}}{{product.image}}" width="500"/>
+  <div>
```
Pomax: It feels like we've hit upon one of those few legitimate uses for `<table>` here.

network-api/networkapi/buyersguide/views.py
```diff
@@ -16,3 +16,8 @@ def buyersguide_home(request):
 def product_view(request, productname):
     product = Product.objects.get(name__iexact=productname)
     return render(request, 'product_page.html', {'product': product, 'mediaUrl': settings.MEDIA_URL})
+
+
+@login_required
```
Pomax: 👍

network-api/networkapi/buyersguide/templates/product_page.html
gvn: Possibly!

network-api/networkapi/buyersguide/templates/bg_base.html
gvn: Shouldn't matter, but sure.

PR #1806, opened 2018-09-07
Streamfield provider

cadecairos:
Summary:

This PR adds a custom Faker provider that can output streamfield JSON. As I write this, it currently supports the following components: header, paragraph, image, spacer, image_text2, quote.

Right now the CMSPageFactory class is the only factory using this provider, but we can use it anywhere we want. It's also very easy to add more components as we develop them.

There's an additional step in the load_fake_data command that generates 20 images with random colors, this is done so that a variety of images can be associated with streamfield image components.


.travis.yml
```diff
@@ -1,8 +1,8 @@
 language: python
 python:
 - '3.6'
-services:
-- postgresql
+addons:
+  postgresql: "9.6"
```
patjouk: Why the change to a specific version of postgresql?

.travis.yml
```diff
@@ -12,6 +12,7 @@ install:
 - pipenv run python network-api/manage.py collectstatic --no-input
 before_script:
 - psql -c 'create database network;' -U postgres
+- pipenv run python network-api/manage.py migrate --no-input
```
patjouk: It's not necessary: migrations will be applied when the test database is created (run it with verbose level 2 if you're curious)

cadecairos: And that's what I thought, but my test runs on travis were failing because none of the tables were getting created. I'll remove this line and cross my fingers that the issues I saw don't return

.travis.yml
cadecairos: I was having some Database issues so I wanted to pin the version we test with to the same one we use in production. Travis defaults to 9.2 but we run 9.6 in prod.

Pomax:
Is this in an easily rebasable form, or will this need to (effectively) be redone?

cadecairos:
I'll rebase and test it, then finish the last change request that @patjouk made in her review.

cadecairos:
Rebased and updated. ready for a final review.

PR #1842, opened 2018-09-20
Bump webpack from v3 to v4, with additional fixes to make that work.

Pomax:
supersedes https://github.com/mozilla/foundation.mozilla.org/pull/1835

updates webpack from v3 to v4, with a change to the webpack config because `module.loaders` was renamed to `module.rules`. Small change, but dependabot has no way of knowing that. The CLI invocation was also taken out of webpack itself, so `webpack-cli` has been added to make sure we can still actually _use_ webpack =)


Pomax:
Big mystery: compile runs fine!

*but it does not actually write the output file*

PR #1845, opened 2018-09-20
Bump all the babel things.

Pomax:
This builds on https://github.com/mozilla/foundation.mozilla.org/pull/1842 and updates babel to a shinier, newer, "everything is done differently now" version.

One upside: instead of `es2015` we can now finally say "use whatever you need to support the same browsers we target for our CSS", based on browserlist. So that's pretty great?

I found one file that broke the bundle but it only needed the smallest of fixes. That said, probably good to see if main.js and bg-main properly load on all the pages?


Pomax:
@gvn rebased, R+?

PR #1863, opened 2018-09-26
Remove ajv-cli as dependency

Pomax:
Dependabot had a PR to update this, but digging through the code shows we don't actually use `ajv-cli` anywhere (nor anything `ajv`), so rather than bump the version, this PR just removes it from the dependency list.


gvn:
Yeah should be fine...it was for validating pulled JSON against now unused JSON schemas.

Issue #1802, opened 2018-09-06
Define Creepiness Service


gvn:
### API Outline:

**POST /privacynotincluded/vote**
```
{
	attribute: "creepiness" (or "confidence")
	rating: 0-99, // can be boolean by sending "true" or "false"
	id: 235325
}
```

**Requires CSRF**

Response codes:
201 is created.
400 is error (will return error text).

**GET /privacynotincluded/product/235325**
```
{
    id: 235325,
    name: “Smart Thermostat”,
	votes: [{
		attribute: "creepiness",
		rating: 45, // average of all votes
		votes: 3245,
		vote_breakdown: { // This can be an array
			"1": 34, // 0-19 vote
			"2": 355, // 20-39 vote
			"3": 34, // etc...
			"4": 355,
			"5": 23
		}
	}, {
		attribute: "confidence",
		rating: 1,
		votes: 345,
		vote_breakdown: {
			"true": 34, // "would buy"
			"false": 355
		}		
	}]
}
```

**GET /products**

Returns ranked product list array (most to least creepy?). Use same schema-per-item as individual `/product` above.

General Questions: 

- What level of rate limiting should we enforce? (via @cadecairos)

CC: @alanmoo

gvn:
@kristinashu We're trying to get a better sense of how the vote tallying works.

Is it total or per product?
Is it votes on creepiness and/or confidence?

kristinashu:
It's the total votes for that specific product on that specific question. 

cc @taisdesouzalessa

gvn:
Ok, I updated the outline above to reflect that.

cadecairos:
@kristinashu Is there a design mockup I could comment on and reference to help finalize this proposed API before we start coding?

kristinashu:
We have [comps for the Product detail page](https://redpen.io/p/qz6bdb2811e058ffb1). This is where users can vote on a product and see how other people voted on for that product. Here are the steps:
![red-pen-0cc70402-4983-4162-c519-94b00acce0b4](https://user-images.githubusercontent.com/1068783/45182254-99209580-b1d5-11e8-806b-109f73ecfcab.jpg)

We are still exploring home page ideas but this is where people will be able to see an overview of all the results. For example, all the products listed in order of creepiest to least creepy.

Happy to hop on vidyo if you need more clarification.

gvn:
For the "found it creepy"/"would buy" percentages...how is that determined?

[...]

PR #1864, opened 2018-09-26
Caas

cadecairos:
Related issue: https://github.com/mozilla/foundation.mozilla.org/issues/1802
Summary: Implement Product voting for PNI, implement a script to periodically summarize votes for each product, implement a serializer that outputs product vote statistics.


cadecairos:
@gvn I wrote a Serializer that generates the JSON as we described in #1802, and use it to generate the objects passed to the views here: https://github.com/mozilla/foundation.mozilla.org/pull/1864/files#diff-fdf5c15be366077e370c1d3b6d42def7R24

We also needed ways to query/filter products, right? Do we have details on what exactly is required?

patjouk:
Cool stuff! Can you add tests around the management command to check if the tally of the vote breakdown looks like we expect it to be? Thanks!

network-api/networkapi/buyersguide/models.py
```diff
@@ -148,3 +154,125 @@ class Product(models.Model):
 
     def __str__(self):
         return str(self.name)
+
+
+class ProductVote(models.Model):
+    votes = models.IntegerField(
+        null=False,
```
patjouk: It's at [False by default](https://github.com/django/django/blob/stable/1.11.x/django/db/models/fields/__init__.py#L145-L151), you don't need to specify it.

network-api/networkapi/buyersguide/models.py
```diff
@@ -148,3 +154,125 @@ class Product(models.Model):
...
+    )
+
+    class Meta:
+        abstract = True
+
+
+class RangeProductVote(ProductVote):
+    attribute = models.CharField(
+        max_length=100,
+        null=False,
```
patjouk: same as before

network-api/networkapi/buyersguide/models.py
```diff
@@ -148,3 +154,125 @@ class Product(models.Model):
...
+        validators=[
+            ValueListValidator(valid_values=['creepiness'])
+        ]
+    )
+    average = models.IntegerField(
+        null=False,
```
patjouk: I won't point all of them, but you get the idea ;)

network-api/networkapi/buyersguide/serializers.py
```diff
@@ -0,0 +1,56 @@
...
+                'attribute': 'confidence',
+                'vote_breakdown': {}
+            }
+
+            for vote_breakdown in boolean_product_vote.booleanvotebreakdown_set.all():
+                breakdown['vote_breakdown'][str(vote_breakdown.bucket)] = vote_breakdown.count
+
+            votes.append(breakdown)
+
+        return {
```
gvn: The schema has shifted a little bit for Product in my latest PR:

https://github.com/mozilla/foundation.mozilla.org/blob/c704458880858e4da1478bc1f1172868be9463c0/network-api/networkapi/buyersguide/models.py#L42

cadecairos: Should I base my PR on yours?

network-api/networkapi/buyersguide/serializers.py
```diff
@@ -0,0 +1,56 @@
...
+            'name': instance.name,
```
gideonthomas: nit: use the `model_to_dict` helper to generate the dict: https://docs.djangoproject.com/en/2.1/_modules/django/forms/models/

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,84 @@
...
+
+        # persist the vote
+        vote.save()
+
+        return Response('Vote recorded', status=201, content_type='text/plain')
+    except ObjectDoesNotExist:
+        return Response('Invalid product', status=400, content_type='text/plain')
+    except ValidationError as ex:
+        return Response(f'Payload validation failed: {ex}', status=400, content_type='text/plain')
+    except Error:
```
gideonthomas: let's at least log this

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,84 @@
...
+vote_throttle_class = UserVoteRateThrottle if not settings.TESTING else TestUserVoteRateThrottle
 
 
 # Login required so we can continue to develop this and merge into master without the public seeing it.
 # Remove this when we're ready to launch.
 @login_required
 def buyersguide_home(request):
     products = Product.objects.all()
-    return render(request, 'buyersguide_home.html', {'products': products})
+    serialized_products = [ProductSerializer(product) for product in products]
```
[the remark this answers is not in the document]
cadecairos: well that's handy!

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,84 @@
...
+    return render(request, 'buyersguide_home.html', {'products': serialized_products})
```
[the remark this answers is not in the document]
cadecairos: In this case the dict being passed here is the context when rendering the template. (I think)

network-api/networkapi/buyersguide/serializers.py
gvn: Yeah, it's actually landed now.

network-api/networkapi/buyersguide/views.py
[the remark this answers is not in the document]
gideonthomas: is there a reason we're serializing it then? As opposed to just passing in the model instances directly?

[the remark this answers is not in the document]
cadecairos: We're using the serializer to generate a `votes` attribute that presents the aggregate vote data in a nice format.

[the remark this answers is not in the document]
gideonthomas: serializer seems to be extra overhead for just that one thing though. Are you able to just do it in this function itself since you need the rest of the model data anyway?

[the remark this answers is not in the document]
cadecairos: Maybe, I'll look into an alternative implementation

kristinashu:
cc @Pomax

cadecairos:
@patjouk I've added a test that checks that the management script generates the right vote counts

@gideonthomas I've removed that Serializer and set up the `votes` attribute on the model directly.

@gvn I'm still not sure what API routes we might need here. I'm passing the new attribute to the existing templates but do we need to expose products with an API endpoint?

gvn:
Yeah, we'll need to expose products with an route as well. I wasn't sure if you wanted to do that in this PR or later, but feel free to do either approach.

cadecairos:
Let's land this and get a DRF route in another ticket

network-api/networkapi/buyersguide/models.py
```diff
@@ -1,6 +1,14 @@
 from django.db import models
+from django.core.validators import MaxValueValidator, MinValueValidator
+from django.forms import model_to_dict
+
+from networkapi.buyersguide.validators import ValueListValidator
 from networkapi.utility.images import get_image_upload_path
-# from wagtail.snippets.models import register_snippet
+
+VALID_VOTE_ATTRIBUTES = [
```
Pomax: crufty?

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,84 @@
...
     products = Product.objects.all()
+    products = [p.to_dict() for p in products]
     return render(request, 'buyersguide_home.html', {'products': products})
 
 
 @login_required
 def product_view(request, productname):
     product = Product.objects.get(name__iexact=productname)
-    return render(request, 'product_page.html', {'product': product, 'mediaUrl': settings.MEDIA_URL})
+    return render(request, 'product_page.html', {'product': product.to_dict()})
```
Pomax: we probably want to keep that mediaURL, and then figure out of we still use it in a new issue

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,84 @@
...
+    except KeyError as ex:
+        return Response(f'Missing attribute in payload: {ex}', status=400, content_type='text/plain')
+
+    # validate the data types passed in the request payload
+    if not (isinstance(product_id, int) and isinstance(value, (int, bool))):
+        return Response('Invalid payload - check data types', status=400, content_type='text/plain')
+
+    try:
+        product = Product.objects.get(id=product_id)
+        vote_class = RangeVote
```
Pomax: can we capitalise this as a class? (VoteClass maybe?)

network-api/networkapi/settings.py
```diff
@@ -487,3 +489,9 @@ def show_toolbar(request):
 
 # Used by load_fake_data to ensure we have petitions that actually work
 PETITION_TEST_CAMPAIGN_ID = env('PETITION_TEST_CAMPAIGN_ID')
+
+# Buyers Guide Rate Limit Setting
+BUYERS_GUIDE_VOTE_RATE_LIMIT = env('BUYERS_GUIDE_VOTE_RATE_LIMIT')
+
+# Detect if we're testing
+TESTING = sys.argv[1:2] == ['test']
```
Pomax: this feels like it's position based, probably a good idea to harden this by doing an `in` check, like `TESTING = 'test' in sys.argv`?

network-api/networkapi/buyersguide/management/commands/aggregate_product_votes.py
```diff
@@ -0,0 +1,118 @@
...
+            # Calculate the average total creepiness score for this product, default to 50 if there aren't votes.
+            creepiness_avg = creepiness_query_set.aggregate(
+                Avg('value', output_field=IntegerField())
+            )['value__avg'] if creepiness_vote_count > 0 else 50
+
+            # define an object for recording this product's creepiness bucket totals
+            creepiness_bucket_totals = {}
+
+            # For each bucket group, filter votes on the low and high values, and record the count of each
+            for creepiness_bucket in self.range_filter_values.keys():
```
Pomax: what is `self` here?

cadecairos: OH. `self` is the Test instance. it's referring to this variable: https://github.com/mozilla/foundation.mozilla.org/pull/1864/files/1f4ffa989c910c3d2cbbce049f48bc895a65e57e#diff-faeee7b1bd84dcf1f1ecd9d0661bb9b9R21

Pomax: 404 on that link?

(we might want to add a comment so we keep that knowledge local to the code =D)

network-api/networkapi/buyersguide/tests.py
```diff
@@ -1,3 +1,272 @@
...
-# Create your tests here.
+from networkapi.buyersguide.factory import ProductFactory
+from networkapi.buyersguide.models import RangeVote, BooleanVote
+from django.core.management import call_command
+
+
+class ManagementCommandTest(APITestCase):
+
+    def setUp(self):
+        self.vote_url = reverse('product-vote')
```
Pomax: nit, but: this doesn't need to be here, we can get the `reverse('product-vote') value outside the class, as a constant, and refer to that in both `ManagementCommandTest` and `BuyersGuideVoteTest`

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,88 @@
...
+
+vote_throttle_class = UserVoteRateThrottle if not settings.TESTING else TestUserVoteRateThrottle
 
 
 # Login required so we can continue to develop this and merge into master without the public seeing it.
 # Remove this when we're ready to launch.
 @login_required
 def buyersguide_home(request):
     products = Product.objects.all()
+    products = [p.to_dict() for p in products]
```
Pomax: just to prevent weirdness, `products = [p.to_dict() for p in Product.objects.all()]` and then just remove L22 above it?

network-api/networkapi/buyersguide/views.py
```diff
@@ -1,23 +1,88 @@
...
     return render(request, 'buyersguide_home.html', {'products': products})
 
 
 @login_required
 def product_view(request, productname):
+    print(productname)
```
Pomax: the dreaded print statement!

cadecairos: just testing you.

network-api/networkapi/buyersguide/management/commands/aggregate_product_votes.py
cadecairos: yeah I rebased the branch so the hashes changed. I think it's clear if you look at parameters for the method this code is in (something I missed while skimming through the code.)

cadecairos:
@Pomax this is ready for a re-review

Issue #1860, opened 2018-09-25
Buyers Guide: QA product detail page


kristinashu:
From ticket https://github.com/mozilla/foundation.mozilla.org/issues/1819 and [doc](https://docs.google.com/document/d/14wTXrEmYmQFWYlmkwFXJtHJZPiseXTgAokJq7_BHx28/edit#heading=h.725dh6j4j01a).

### Images
- [x] hero image is broken **_(high priority)_**
- [x] Related Product images are broken **_(high priority)_**

### Categories Nav
- [ ] should be sticky _(low priority)_
- [ ] font weight 700 instead of 900 _(low priority)_
- [ ] font size from 17px to 15px (new update so that it fits better on all screen widths)  _(medium priority)_
- [ ] Nav container should be consistent height (right now it changes depending if selected state bar is there or not) _(low priority)_

### Grid
- [x] grid is generally too narrow (update from 8 to 10 columns) _(medium priority)_

### Intro
- [ ] copy link functionality isn't working in Safari on iPhone _(medium priority)_
- [ ] needs some sort of feedback so the user knows that the link has been copied **_[need to mock up]_** _(medium priority)_
- [ ] narrow layout: copy link should be moved lower down _(medium priority)_

### Criteria
- [x] need to add link to Privacy Policy **_(high priority)_**
- [ ] accordions work well but should [style text in italics and 14px](https://redpen.io/we4fedfbca381f4804) _(medium priority)_
- [ ] undetermined face has been updated from green to gray _(low priority)_
- [x] grades should be individual and add "can't determine :|" **_(high priority)_**
- [x] update password question and answer: "If a password is required, you must change the default password"  Yes :) or No :( or Not required :|  **_(high priority)_**
- [ ] add "Customer support" text above list of ways to contact them _(medium priority)_ 
- [x] remove "Customer support" field from CMS (we don't show the answer) **_(high priority)_**
- [ ] medium layout: phone icon disappears _(medium priority)_
- [ ] medium layout for both 3 icons sections: list should stack and go into mobile view layout sooner _(medium priority)_
- [ ] narrow layout for both 3 icons sections: icon centering and spacing is off _(medium priority)_
- [ ] narrow layout: smile/frown faces shouldn't change size _(medium priority)_
- [x] add Twitter as a contact **_(high priority)_** (this is a new feature, sorry)

gvn:
That sounds do-able to me.

gvn:
[...]

PR #1875, opened 2018-09-28
PNI: High priority detail page items

gvn:
Related issue: #1860


gvn:
@kristinashu This contains the high pri items from #1860 

https://foundation-mofostaging-pr-1875.herokuapp.com/privacynotincluded/

kristinashu:
Thank you for all these, it's looking good! There are a few things but these could be done as follow ups:

- [ ] You removed the Customer Help field which is good but should also remove the Helptext field for it.
screen shot 2018-09-28 at 3 07 58 pm

- [ ] Grades are great! But can you remove the required field * (not sure why this one has it and others don't but would rather none have it)?
screen shot 2018-09-28 at 3 08 28 pm

- [ ] Please update copy in this one to "If a password is required, you must change the default password" and update "Can't be determined" to "Not required".
screen shot 2018-09-28 at 3 09 57 pm

Issue #1805, opened 2018-09-07
Petition signees already subscribed to newsletter don't see share buttons


gideonthomas:
STR:
1. Go to https://foundation.mozilla.org/en/campaigns/venmo-put-our-privacy-first/?subscribed=1
2. Sign the petition and click "no I'll share instead" in the modal.
3. You won't see the share buttons but will instead see the form as if an error occurred.

Pretty sure it's because we conditionally render the signup checkbox based on the `subscribed=1` query param out [here](https://github.com/mozilla/foundation.mozilla.org/blob/8ad24d5192a16a57857b53a4cc374cbfb6b7cc32/source/js/components/petition/petition.jsx#L634-L641). This makes [this ref](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/petition/petition.jsx#L252) `undefined` which throws an error in the promise chain thus breaking it.

PR #1896, opened 2018-10-03
make sure to test we have an argument to work with

Pomax:
Fixes https://github.com/mozilla/foundation.mozilla.org/issues/1805 by making sure we test to see if the function was even given an argument before we do an `instanceof` check.


Pomax:
https://foundation-mofostaging-pr-1896.herokuapp.com/en/campaigns/single-page/?subscribed=1 should allow you to confirm now.

PR #1897, opened 2018-10-03
remove pillow dependency

Pomax:
While wagtail relies on pillow, we don't explicitly use it ourselves, and so we don't need it in our our pipfile


cadecairos:
@Pomax can you rebase and fix the conflict so we can merge this?

Pomax:
I can't really get the pip lock file to only update the pillow part and the master hash, a new venv ends up bumping some things automatically.

Pomax:
Aparently, no rebasing/respinning I can do on my windows machine seems to fix this piplock issue.

Pomax:
Redid the PR on current master.

Pomax:
@cadecairos @patjouk just reset you for review, just to make sure we had eyes on it before merge =)

PR #1898, opened 2018-10-03
missing migrations?

Pomax:
Apparently there were missing migrations in master. This is the result of `inv makemigrations`.


patjouk:
It was linked to a change in wagatail-metadata. I've merged that missing migration and the dependency update in #1911.

PR #1907, opened 2018-10-08
Pin pip to 18.0 for now

patjouk:
Related issue: CI is broken because of an issue around pipenv/pip.
Summary: I'm pinning pip to 18.0 for now but the next step would be to unify Heroku and CI config: they should run the same version of pipenv and pip. This will be a second PR because I need to discuss it with @cadecairos 

Feel free to merge it if I'm out of the office by the time it's reviewed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ python:
 services:
 - postgresql
 install:
-- pip install --upgrade pip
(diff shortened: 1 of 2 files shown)


patjouk:
Not needed anymore since pipenv devs released a new version.

PR #1914, opened 2018-10-09
Update Django, boto3, django-admin-sortable

patjouk:
Update Django to 1.11.16 + a bunch of dependencies.

close #1912, close #1910, ref MozillaFoundation/mofo-devops-private#170


Pomax:
I can't remember, what was keeping us on Django 1.11 rather than 2?

patjouk:
The previous CMS was not compatible with Django 2.

Issue #1821, opened 2018-09-13
Buyers Guide: categories page


kristinashu:
Based on design in #1739

Comps https://redpen.io/p/qz6bdb2811e058ffb1

Same functionality as [live site](https://advocacy.mozilla.org/en-US/privacynotincluded/category/toys)

![category 2x](https://user-images.githubusercontent.com/1068783/46311952-8ac95d80-c578-11e8-93ff-99c52f583ead.jpg)

PR #1920, opened 2018-10-10
Category page

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/1821

Adds a product category as snippet-administratable thing, with default categories based on the comp, and which are used to generate the nav menu now, linking to category pages. Each product can have multiple categories - when on a category page the relevate cat. will be menu-highlighted, and when clicking through to a product no cat is highlighted.

This PR also updates the fake data generator so that it generates 20 products, each assigned to at least one, but possibly more categories.


network-api/networkapi/buyersguide/views.py
```diff
@@ -20,21 +20,39 @@
...
+
+@login_required
+def category_view(request, categoryname):
+    category = BuyersGuideProductCategory.objects.get(name__iexact=categoryname)
+    products = [p.to_dict() for p in Product.objects.filter(product_category__in=[category]).distinct()]
+    return render(request, 'category_page.html', {
+        'categories': BuyersGuideProductCategory.objects.all(),
+        'category': category,
+        'products': products,
+        'mediaUrl': settings.MEDIA_URL,
```
Pomax: this is not enough to get images to work properly.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -911,7 +911,6 @@ def get_context(self, request):
         # We need to expose MEDIA_URL so that the s3 images will show up properly
         # due to our custom image upload approach pre-wagtail
         context = super(Homepage, self).get_context(request)
-        print(settings.MEDIA_URL)
```
Pomax: spotted this while figuring out why MEDIA_URL doesn't include the AWS directory, so... remove'd

kristinashu:
@Pomax what url can I see this at? I'll do a design review too.

kristinashu:
This is good to push but I will open a follow up ticket to update the font styles:
- Category title .h3-heading
- Disclaimer .body with italic 
- Disclaimer should be moved to the bottom of the page
- Product company .h6-heading-uppercase
- Product name .h5-heading

category 2x

source/sass/buyers-guide/views/category.scss
```diff
@@ -0,0 +1,89 @@
+.category-listing {
```
gvn: Most of this page layout should be achievable with Bootstrap and other existing type classes (see https://foundation.mozilla.org/en/style-guide/), but I'll leave it up to you if you want to refactor given time constraints.

network-api/networkapi/buyersguide/urls.py
```diff
@@ -5,6 +5,7 @@
 urlpatterns = [
     url(r'^$', views.buyersguide_home, name='buyersguide-home'),
     url(r'^about/', views.about_view, name='about-view'),
+    url(r'^category/(?P<categoryname>[\w\W]+)/', views.category_view, name='category-view'),
```
gvn: Might want to use `categories/` for consistency with the recently renamed `products/` route fragment. @cadecairos may have thoughts.

cadecairos: yes, lets keep the resource names in our routes pluralized

kristinashu:
Any chance we can push this live today (even though Pomax is off today) so we don't have to wait until Monday? Planning to do a bunch of content entry tomorrow.

source/sass/buyers-guide/views/category.scss
Pomax: it's  good idea, although I will need to learn myself some bootstrap before I can do this, so I'll file it as a followup: https://github.com/mozilla/foundation.mozilla.org/issues/1924

network-api/networkapi/buyersguide/urls.py
Pomax: :+1:

PR #1929, opened 2018-10-13
Fixes #1643 made petition form work properly

### LINKED ISSUE #1643: Do not display FB, Email and Twitter buttons if not configured in Snippet
For petition pages, please do not display the FB, Email and Tweet buttons if the snippet related to the petition doesn't have these options configured

techac:
I made a change in petition.jsx to encounter an error some development purpose but the page /campaigns/single-page got rendered and form was submitted succesfully which should not happen .. Can you explain me why?

Thanks....

[ ] Migrations squashed

### DIFF
--- a/network-api/networkapi/buyersguide/templates/product_page.html
+++ b/network-api/networkapi/buyersguide/templates/product_page.html
@@ -25,7 +25,7 @@
       </div>
       <div class="body-large mb-5">{{product.blurb}}</div>
 
-      <div class="creep-vote-target my-5">
+      <div class="creep-vote-target my-5" data-product-name="{{product.name}}">
           {% csrf_token %}
           <input type="hidden" name="productID" value="{{ product.id }}">
           <input type="hidden" name="votes" value="{{ product.votes | safe }}">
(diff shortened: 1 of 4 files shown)


techac:
@alanmoo I intended to generate a error on '/campaigns/single-page/' and made these changes but the form was successfuly submitted ... Can you please help me out ..

alanmoo:
That makes sense- changing that wouldn't change if the form was successfully submitted or not, just the reaction to it. But that's not what needs to change for this issue- you need to add some if statements that check for the existence of share links, and then conditionally render the share buttons.

techac:
@alanmoo in petition.jsx I there is a method renderThankYou() and I tried removing onclick attribute for facebook SHARE but on submiiting the form an clicking on share, the console shows error "No attribute 'click' of null" .... Why the petition.jsx is not getting compiled and I am getting the same output

techac:
@alanmoo I there was a error showing up in console, which is removed which I thik is the call fr this ticket... If not suggest me what to add .. Thanks

Issue #1888, opened 2018-10-02
Buyers Guide: build creep-o-meter


kristinashu:
Based on design in https://github.com/mozilla/foundation.mozilla.org/issues/1739

Comps https://redpen.io/p/qz6bdb2811e058ffb1

Follow up from larger product page https://github.com/mozilla/foundation.mozilla.org/issues/1819

### Creep-O-Meter
- [x] back-end ticket https://github.com/mozilla/foundation.mozilla.org/issues/1802
- [x] animation - will be about 80 svg faces
  - Nat's animation design ticket https://github.com/mozilla/foundation.mozilla.org/issues/1859

Q1: 
- [x] text with number of votes and majority of results 
- [x] slider with animated face

Q2: 
- [x] text with number of votes and majority of results 
- [x] binary selector 
- [x] submit button (for both Qs)

Results:
- [x] simplified slider results
- [x] shows Your Vote indicator
- [x] binary results
- [ ] share icons (links should be done in https://github.com/mozilla/foundation.mozilla.org/issues/1899#issuecomment-429068810)
- [x] comments link scrolls user down the page to comment section

![red-pen-7fe5b11b-a9af-418c-891b-01f35caf7e46](https://user-images.githubusercontent.com/1068783/46363525-236ae680-c628-11e8-8619-3e4289cef842.jpg)

kristinashu:
From what I gather, @Pomax will work on linking up the voting submissions and results (as he works on the home page results) and then @alanmoo or @mmmavis will finish styling the content.

kristinashu:
🙌

PR #1946, opened 2018-10-16
Make creep-o-meter touch friendly

alanmoo:
Continues #1888 

It's still not perfect (the page can scroll while moving slider left/right, for example) but I'd say it's MVP.


source/js/buyers-guide/components/creepometer/creepometer.scss
```diff
@@ -25,11 +25,12 @@
   }
 
   .handle {
+    border-radius: 50%;
```
alanmoo: Just putting these in alphabetical order when adding a property.

PR #1954, opened 2018-10-16
use get_object_or_404

### LINKED ISSUE #1953: Buyers Guide: ensure 404 on nonexistent category
https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/views.py#L31 uses `get` rather than `get_or_404` so if it fails to find the correct category it will generate a 500 error, rather than a 404 error.

Pomax:
Closes #1953 

[x] Generates new fake data, if necessary
[ ] ~Wagtail documentation has been updated, if necessary~
[x] Includes tests
[ ] ~Migrations squashed~

### DIFF
--- a/network-api/networkapi/buyersguide/tests.py
+++ b/network-api/networkapi/buyersguide/tests.py
@@ -317,3 +317,11 @@ class BuyersGuideViewTest(TestCase):
         request = self.factory.get('/privacynotincluded/products/this is not a product')
         request.user = self.user
         self.assertRaises(Http404, product_view, request, 'this is not a product')
+
+    def test_category_view_404(self):
+        """
+        Test that the category view raises an Http404 if the category name doesn't exist
+        """
+        request = self.factory.get('/privacynotincluded/categories/this is not a category')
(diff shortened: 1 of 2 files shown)


network-api/networkapi/buyersguide/tests.py
```diff
@@ -317,3 +317,11 @@ def test_product_view_404(self):
...
         request.user = self.user
         self.assertRaises(Http404, product_view, request, 'this is not a product')
+
+    def test_category_view_404(self):
+        """
+        Test that the category view raises an Http404 if the category name doesn't exist
+        """
+        request = self.factory.get('/privacynotincluded/categories/this is not a category')
+        request.user = self.user
+        self.assertRaises(Http404, product_view, request, 'this is not a category')
```
cadecairos: you should import the `category_view` from `networkapi.buyersguide.views` and pass it here instead of the product view.

Pomax: good catch

Issue #1892, opened 2018-10-02
Buyers Guide QA: P1 product detail page


kristinashu:
Follow up from https://github.com/mozilla/foundation.mozilla.org/issues/1860
Comps https://redpen.io/p/qz6bdb2811e058ffb1

### Images
- [x] hero: banner is taking up too much vertical space, update image size from 500x500 to 400x400

### Intro
- [x] narrow layouts: copy link should be moved lower down ([see comps](https://redpen.io/fk5813dda329fda302))

### Criteria
- [x] P1: make product title link through to buy product (url is already in cms)
- [x] P1: undetermined face has been updated from green to gray [icon-face-neutral.svg.zip](https://github.com/mozilla/foundation.mozilla.org/files/2480810/icon-face-neutral.svg.zip)
- [x] P1: update "Shares your information with third parties" to "Shares your information with 3rd parties for unexpected reasons" (from Abigail)
- [x] add "Customer support" text above list of ways to contact them ([see comp](https://redpen.io/we4fedfbca381f4804))
- [x] P1: remove mailto for email (usually just a url to contact page)
- [x] **P1: make the Grade text linkable (most products will link to a unique external url)**
- [x] **P1: completely remove "Requires an account" question**

### Related Products
- [x] P1: mobile: image needs vertical spacing
- [x] mobile: images should be two across, not just one large one https://redpen.io/fk5813dda329fda302
Screen cap from staging: 
<img width="507" alt="image" src="https://user-images.githubusercontent.com/1068783/46364552-fc61e400-c62a-11e8-9d6a-e580b8d7c64d.png">

### Updates
- [x] **P1: if no updates are in this section, then the entire section should be hidden from users**

mmmavis:
@kristinashu 

> P1: make the Grade text linkable (most products will link to a unique external url)

Where can I find the external url you mentioned?

kristinashu:
Each product will link to a different url, so this should be added as a field in the cms.

Does that make sense? Happy to hop on a call and chat through all of these if you like.

kristinashu:
This is a big ticket so I've moved all those P2's the a new ticket. 

The only remaining things for this one is - P1: completely remove "Requires an account" question. 

After that, please close!

kristinashu:
🎉 yay thank you @mmmavis!

PR #1985, opened 2018-10-19
Some updates and fixes to Buyer's Guide's product page

### LINKED ISSUE #1982: Buyers Guide QA: Posts from updates aren't showing up
I have selected an update for PS4 but it isn't showing up on the site.

image

image

mmmavis:
Related to #1892 & fixes #1982

This PR only covers the following

### Images
- [x] hero: banner is taking up too much vertical space, update image size from 500x500 to 400x400

### Intro
- [x] narrow layouts: copy link should be moved lower down ([see comps](https://redpen.io/fk5813dda329fda302))

### Criteria
- [x] **P1: make product title link through to buy product (url is already in cms)**
- [x] **P1: undetermined face has been updated from green to gray [icon-face-neutral.svg.zip](https://github.com/mozilla/foundation.mozilla.org/files/2480810/icon-face-neutral.svg.zip)**
- [x] **P1: update Shares your information with third parties" to "Shares your information with 3rd parties for unexpected reasons" (from Abigail)**
- [x] add "Customer support" text above list of ways to contact them (see comp)
- [x] **P1: remove mailto for email (usually just a url to contact page)**

### Related Products
- [x] **P1: mobile: image needs vertical spacing**
- [x] mobile: images should be two across, not just one large one https://redpen.io/fk5813dda329fda302

### Updates
- [x] **P1: if no updates are in this section, then the entire section should be hidden from users**


network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -72,7 +73,7 @@ <h3 class="h3-heading h3-heading-small">What does it know about me?</h3>
         {% else %}
           {% criterion "privacy_policy_reading_level" "Privacy policy reading level" product.privacy_policy_reading_level product.privacy_policy_helptext %}
         {% endif %}
-        {% criterion "share_data" "Shares your information with third parties" product.share_data product.share_data_helptext %}
+        {% criterion "share_data" "Shares your information with 3rd parties for unexpected reasons" product.share_data product.share_data_helptext %}
```
Pomax: Why this change? Shareprogress text length limitation? (as a number below 20, normally this should stay "third").

source/images/buyers-guide/icon-face-neutral.svg
```diff
@@ -1 +1,9 @@
-<svg version="1.1" id="Layer_1" xmlns="http://www.w3.org/2000/svg" x="0" y="0" viewBox="0 0 24 24" xml:space="preserve"><style>.st25{fill:#4cbfa9}</style><path class="st25" d="M12 0C5.4 0 0 5.4 0 12s5.4 12 12 12 12-5.4 12-12S18.6 0 12 0zm0 22.5C6.2 22.5 1.5 17.8 1.5 12S6.2 1.5 12 1.5 22.5 6.2 22.5 12 17.8 22.5 12 22.5z"/><path class="st25" d="M7.9 11.1c.7 0 1.2-.5 1.2-1.2s-.6-1.1-1.2-1.1-1.2.5-1.2 1.1.5 1.2 1.2 1.2zm0-1.2v.1s-.1 0 0-.1zM16.1 11.1c.7 0 1.2-.5 1.2-1.2s-.5-1.2-1.2-1.2-1.2.5-1.2 1.2.6 1.2 1.2 1.2zm0-1.2v.1s-.1 0 0-.1zM17.2 15.5H6.8c-.3 0-.6.3-.6.6s.3.6.6.6h10.3c.3 0 .6-.3.6-.6s-.2-.6-.5-.6z"/></svg>
\ No newline at end of file
+<svg id="Layer_1" data-name="Layer 1" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 24 24">
```
Pomax: this file looks like it needs an `svgo` treatment.

source/sass/type.scss
```diff
@@ -70,7 +72,7 @@ h1,
   font-weight: 400;
   letter-spacing: 0;
 
-  @include scaleText(36px, 40px, 56px, 64px);
+  @include scaleText(36px, 40px, $h1-font-size, 64px);
```
Pomax: Why only the  third scale? Should the other values be based on `$h1-font-size`, too? (like x2/3,  x3/4, x1, x1.15?)

source/sass/buyers-guide/views/product.scss
```diff
@@ -4,6 +4,26 @@
 
 .product-detail {
   margin-top: -88px;
+
+  @media (min-width: $bp-md) {
```
Pomax: where is `$bp-md` coming from? (it's not imported anywhere, so if we ever refactor this might suddenly go very wrong)

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -11,19 +11,20 @@
...
 
   <div class="row">
     <div class="product-detail bg-white p-5 col-12 col-md-10 offset-md-1">
       <div class="product-price mb-3">{{product.company}} ${{product.price}}</div>
-
-      <div class="d-flex justify-content-between align-items-center">
-        <h2 class="h1-heading">{{product.name}}</h2>
+      <div class="intro d-flex flex-column flex-md-row justify-content-between">
+        <div>
+          <h2 class="h1-heading"><a href="{{product.url}}" href="_blank">{{product.name}}</a></h2>
```
Pomax:
```suggestion
          <h2 class="h1-heading"><a href="{{product.url}}" target="_blank">{{product.name}}</a></h2>
```

source/sass/buyers-guide/views/product.scss
```diff
@@ -4,6 +4,26 @@
...
+    .intro {
+      position: relative;
+
+      .h1-heading {
+        margin-right: 8rem;
```
Pomax: Hm, could we not achieve sufficient margin with bootstrap?

source/images/buyers-guide/icon-face-neutral.svg
mmmavis: Ummm... I did... using the same configs as what we have in `package.json` 

> `svgo --multipass --pretty --enable=removeTitle`

source/sass/buyers-guide/views/product.scss
mmmavis: Nope, Bootstrap margin classes only go up to `5` e.g,. `mb-5` which adds 3rem to margin bottom and this is the max margin we can get from Bootstrap

source/sass/type.scss
mmmavis: Turned 56px which is the font size we set for desktop into a variable so I can use that variable to position the `.copy-link` properly in `product.scss`

mmmavis: If we don't want to touch this `type.scss` I can just hardcode 

```
.copy-link {
  position: absolute;
  top: 28px; /* instead of $h1-font-size/2 */
  right: 0;
}
```

source/sass/buyers-guide/views/product.scss
mmmavis: these breakpoint variables are set in `/source/sass/buyers-guide/bg-main.scss`

network-api/networkapi/buyersguide/templates/product_page.html
mmmavis: not sure but this text is what we have in the ticket /cc @kristinashu

mmmavis:
@Pomax I fixed the wrong attribute name from `_blank` from `href` to `target`. Thanks for catching it.

Please also see my comments above.

network-api/networkapi/buyersguide/templates/product_page.html
kristinashu: This is not related to Shareprogess. This is one of the criteria. Please use "3rd" to help shorten the text. This has been approved by Jen and Abigail the lawyer.

source/sass/buyers-guide/views/product.scss
alanmoo: Should that element just be in its own column then?

source/images/buyers-guide/icon-face-neutral.svg
Pomax: ah, I've been running it without arguments, didn't really we have --pretty turned on in the package.json

kristinashu:
@mmmavis things are looking good, I've check off all the point that are done in the first comment.

Couple things, sorry if they were unclear before:
- [ ] For "remove mailto for email (usually just a url to contact page)": we still need that field, it just doesn't need the "mailto" automatically added to it (because we are often linking to a contact page url, not an email address).
- [ ] For "if no updates are in this section, then the entire section should be hidden from users": "updates" and "related products" should be separate. Seems like with this update that now "related products" only shows up if there is also an "update". Related products should show up even if there are no Updates.

source/sass/type.scss
Pomax: It feels a bit bug-introducing if we only change one of the four values from number to variable, so if we can turn the others into variables as well, that'd be nicer.

source/sass/buyers-guide/views/product.scss
Pomax: guess we'll have to refactor this a little after launch.

source/sass/buyers-guide/views/product.scss
Pomax: that's probably worth doing, since that suggests it's not on the right bootstrap grid if we need that much margin.

mmmavis: I'm not sure how we can achieve our mobile and desktop design solely by using bootstrap.

(pay attention to the order of "copy link")
image

I was thinking to rely on flexbox's `order` but since the 3 elements title, copy link, description aren't even on the same row on desktop, flexbox's `order` won't help. Suggestions?

mmmavis: @alanmoo @Pomax ^

mmmavis:
addressed the 2 items @kristinashu mentioned above

source/sass/buyers-guide/views/product.scss
alanmoo: I would think you should be able to use markup like this:

```
<div class="row">
<div class="title col-9">title</div>
<div class="copy col-3">copy link</div>
<div class="blurb col-12">paragraph</div>
</div>
```

And then add `order` to the unique classnames using CSS and some bootstrap media queries. (Those aren't actually the column classes you'd need, they're just for example.)

source/sass/buyers-guide/views/product.scss
mmmavis: @Pomax Do you mind filing a ticket for this? I think we can clean this bit by moving

```scss
$bp-xs: #{map-get($grid-breakpoints, xs)}; // < 576px
$bp-sm: #{map-get($grid-breakpoints, sm)}; // >= 576px
$bp-md: #{map-get($grid-breakpoints, md)}; // >= 768px
$bp-lg: #{map-get($grid-breakpoints, lg)}; // >= 992px
$bp-xl: #{map-get($grid-breakpoints, xl)}; // >= 1200px
```

to its own `.scss` file and `@import` this file into `bg-main.scss` but I'm not sure if this is what you meant. If we do decide to tackle this we'll need to do it for the main foundation site `main.scss` as well. I think filing a ticket to address both cases would be helpful.

source/sass/buyers-guide/views/product.scss
mmmavis: awesome! i like this clean approach

mmmavis:
@kristinashu @Pomax PR updated. Please review.

PR #2019, opened 2018-10-23
Cache sorted products dict in memory for one day

cadecairos:
This PR adds a local memory cache that I use to store the sorted products dict for one day.

Locally I saw the homepage load in 450ms before the cache, and in 50ms after - a huge improvement, and one I expect to be even more noticeable on production where the request is taking around 5000ms


cadecairos:
fyi: If we need to bust the cache, we can do so with a dyno reboot.

Issue #2029, opened 2018-10-23
Buyers Guide: audit the model for and `addslashes` need


Pomax:
Give https://github.com/mozilla/foundation.mozilla.org/pull/2010, which added `addslashes`, and then https://github.com/mozilla/foundation.mozilla.org/pull/2026, which selectively takes them away depending on a property name, we should audit the code to see:

1.  why we need `addslashes` at all (can we be better with the code we're generating?), and
2. whether we can cleanly set up something that'll yield the correct HTML without using nested template logic.

PR #2032, opened 2018-10-24
Un-break answers in buyer’s guide

### LINKED ISSUE #2030: Inconsistency between prod and local dev: bugs on deletes_data/child_rules
/cc @cadecairos @alanmoo @kristinashu 

### what's wrong?

https://foundation.mofostaging.net/privacynotincluded/products/apple-watch-4/ shows off a bug in product page rendering, where its model looks like this:

image

but its associated product page looks like this:

image

Where React has created the following DOM (note the `criterion-undefined` entries):

image

Off of the following Django HTML:

```
<div class="mb-5">
 <div class="criterion-target" data-meta='{"class": "null","id": "must_change_default_password","question": "If a password is required, you must change the default password","helptext": "","answer": "Not required"}'></div>
 <div class="criterion-target" data-meta='{"class": "true","id": "security_updates","question": "Automatic security updates","helptext": "","answer": "Yes"}'></div>
 <div class="criterion-target" data-meta='{"class": "true","id": "delete_data","question": "Deletes data it stores on you","helptext": "","answer": "Yes"}'></div>
 <div class="criterion-target" data-meta='{"class": "null","id": "child_rules","question": "Has parental controls","helptext": "","answer": "Can\&#39;t determine"}'></div>
 </div>
```

### trying this locally

Copying the production database and loading it locally, I see the same model content:

image

But no errors on the associated product page:

image

Because React generated this DOM:

image

off of the following Django HTML:

```
<div class="mb-5">
 <div class="criterion-target" data-meta='{"class": "null","id": "must_change_default_password","question": "If a password is required, you must change the default password","helptext": "","answer": "Not required"}'></div>
 <div class="criterion-target" data-meta='{"class": "true","id": "security_updates","question": "Automatic security updates","helptext": "","answer": "Yes"}'></div>
 <div class="criterion-target" data-meta='{"class": "false","id": "delete_data","question": "Deletes data it stores on you","helptext": "","answer": "No"}'></div>
 <div class="criterion-target" data-meta='{"class": "true","id": "child_rules","question": "Has parental controls","helptext": "","answer": "Yes"}'></div>
 </div>
```

So the same codebase, on the same database, on two different computers, yielding two very different HTML sources. I ... don't know how this is possible.

(yet?)

alanmoo:
I was using the wrong filter

I believe this closes #2030 and #2029

### DIFF
--- a/network-api/networkapi/buyersguide/templates/tags/criterion.html
+++ b/network-api/networkapi/buyersguide/templates/tags/criterion.html
@@ -1,1 +1,1 @@
-<div class="criterion-target" data-meta='{"class": "{{ class }}","id": "{{ id }}","question": "{{ question | escapejs }}","helptext": "{{ helptext | escapejs }}","answer": "{{ answer | escapejs }}"} '></div>
+<div class="criterion-target" data-meta='{"class": "{{ class }}","id": "{{ id }}","question": "{{ question }}","helptext": "{{ helptext | escapejs }}","answer": "{{ answer }}"} '></div>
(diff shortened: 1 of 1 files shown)


kristinashu:
But now it's adding https://foundation-mofostaging-pr-2032.herokuapp.com/ to all the links which breaks the Privacy Policy link, Reading Level link, and the Customer Support email.

Example: https://foundation-mofostaging-pr-2032.herokuapp.com/privacynotincluded/products/source-strategy/

Will that be fixed separately?

alanmoo:
Good catch. The fake data generator on my local machine didn't generate those links so I didn't think to check for them. Let's try this commit, I don't think there's free text answers or questions, so I think this should still work- I was being overzealous and applying the filter to everything.

Pomax:
this criterion code keeps biting us, so I've filed https://github.com/mozilla/foundation.mozilla.org/pull/2036 to use normal templating rather than trying to generate JSON inside a data attribute, superceding this PR.

PR #2036, opened 2018-10-24
Server-side criterium rendering.

### LINKED ISSUE #2029: Buyers Guide: audit the model for and `addslashes` need

### LINKED ISSUE #2030: Inconsistency between prod and local dev: bugs on deletes_data/child_rules

### LINKED ISSUE #2031: Buyers Guide QA: if Reading level Can't be determined then it should say that
Example: https://foundation.mozilla.org/privacynotincluded/products/fredi-baby-monitor/ and https://foundation.mozilla.org/privacynotincluded/products/tractive-gps-3g-pet-tracker/

If the Reading Level in the CMS is set to "Can't Determine" then the page should say "Can't determine". But on Prod is says "Level 0" (this is a new issue, used to be ok).

CMS
screen shot 2018-10-23 at 5 02 18 pm

Prod
screen shot 2018-10-23 at 5 02 45 pm

Pomax:
Supercededs #2032

Closes #2029, closes #2031


source/sass/global.scss
```diff
@@ -91,3 +91,7 @@ button:hover {
     color: transparentize($white, 0.5);
   }
 }
+
+.hidden {
```
alanmoo: Bootstrap should have a class for this already

Pomax: it does, `d-none`, which is a terrible name. I'll use it, but let the record show that the only right name for this CSS is `hidden`.

source/js/buyers-guide/components/criterion/criterion.jsx
```diff
@@ -1,12 +1,37 @@
 import React from 'react';
 
+// Helper function to get the text content from
+// a DOM element. Returns false if the query
+// selector fails to find an element.
+function getValueFrom(e, qs) {
+  let v = e.querySelector(qs);
+  return v ? v.textContent : false;
```
Pomax: not quite right yet, we want that HTML to actually do things.

network-api/networkapi/buyersguide/templates/fragments/product-criterion.html
```diff
@@ -0,0 +1,20 @@
+<div class="criterion criterion-{{ value|lower }} criterion-{% if value == bad %}bad{% else %}good{% endif %}">
```
alanmoo: This seems like you're just outputting `criterion-bad criterion-bad` or `criterion-good criterion-good`. The existing template outputs the criterion name?

alanmoo: Ah, I think I see now. This has to do with the bad attribute being passed in to the fragment?

source/sass/buyers-guide/views/product.scss
```diff
@@ -107,20 +107,7 @@
...
-.criterion-privacy_policy_reading_level.criterion-17,
-.criterion-privacy_policy_reading_level.criterion-18,
-.criterion-privacy_policy_reading_level.criterion-19,
-.criterion-share_data.criterion-true,
-.criterion-must_change_default_password.criterion-false,
-.criterion-security_updates.criterion-false,
-.criterion-delete_data.criterion-false,
-.criterion-child_rules.criterion-false,
-.criterion-manage_security.criterion-false {
+.criterion-bad {
```
alanmoo: Looks like the mapping of "reading levels >=13 are bad" didn't get translated into the refactor.

alanmoo:
@kristinashu I combed through this refactor/improvement, but could probably use a second set of eyes to make sure I'm not missing any functionality. Basically the "criteria" items have been refactored-  I edited [this item](https://foundation-mofostaging-pr-2036.herokuapp.com/privacynotincluded/products/wife-computer/) to set every item to Yes, No, and then Unknown, comparing it to [this staging item](https://foundation.mofostaging.net/privacynotincluded/products/petchatz-hd/). Yes and No looked fine, but "Unknown" items have different words and the reading levels need "bad" styling. If there's any other functionality I'm not aware of to check for, please note it here?

Issue #1968, opened 2018-10-18
Buyers Guide: Styling QA for creep-o-meter


kristinashu:
Low priority styling updates:
- [x] Before voting: likely to buy icons the same width
- [x] Before + After voting: vote text (in both locations) shoudl be `h6-heading-uppercase` from style guide
- [x] After voting: remove "so far:" text from title
- [x] After voting: all bars should be #e1e1e1
- [x] After voting: Likely to buy text in the bars is odd
- [x] After voting at medium width: the two results should stack sooner, they are too squished right now. The before voting screens seem to be nicely responsive and stack sooner.

![screen-shot-2018-10-18-at-12 23 57-pm](https://user-images.githubusercontent.com/1068783/47180701-07c53880-d2d6-11e8-916d-c7b8a9075e45.jpg)
 
<img width="525" alt="screen shot 2018-10-18 at 10 22 24 pm" src="https://user-images.githubusercontent.com/1068783/47199110-c1e39100-d324-11e8-9428-92598589afed.png">

Pomax:
Note that the PR does not introduce a gradient bar in the post-vote screen, for the bars-with-faces. This is not implemented right now, if we still want that: let's file a separate issue for that.

PR #2041, opened 2018-10-24
styling fixes

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/1968

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/components/**/*.jsx\" webpack.config.js",
     "test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/components/**/*.scss\" --syntax scss",
     "test": "npm run build && run-s test:**",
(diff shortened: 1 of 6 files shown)


kristinashu:
> After voting: remove "so far:" text from title

Please also remove the `:` at the end.


> After voting at medium width: the two results should stack sooner, they are too squished right now. The before voting screens seem to be nicely responsive and stack sooner.

Does seem like this has been done but I actually think this should be done as a separate and bigger update to improve styling at break point between around 980-770px.

package.json
```diff
@@ -26,7 +26,7 @@
     "test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/components/**/*.jsx\" webpack.config.js",
     "test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/components/**/*.scss\" --syntax scss",
     "test": "npm run build && run-s test:**",
-    "watch:images": "chokidar 'source/images/**/*' -c 'npm run build:images'",
+    "watch:images": "chokidar \"source/images/**/*\" -c \"npm run build:images\"",
```
alanmoo: I see you're making it consistent with the rest of the file, but...what is going on here?

PR #2049, opened 2018-10-25
PNI homepage cache refresh from CMS

cadecairos:
This PR adds:

1. An admin only view+URL that refreshes the homepage admin cache when pressed
2. An override template for the list view that adds a button to the top of the page


Pomax:
Duplicates the lambda, but that's kind of fine in this case.

Issue #1969, opened 2018-10-18
Buyers Guide: Creep-O-Meter tappable  


kristinashu:
This is a bonus! Would be if a user could also click on the gradient (instead of the face/circle) and that would move the face over there. Definitely a bonus tho. Just wonder if some user might try to do that. But could test and see at mozfest first.

PR #2053, opened 2018-10-25
allow people to click somewhere along the creepometer

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/1969 because bonuses are the fun bits of code.


alanmoo:
This was a great call, I saw lots of people using it this way during Mozfest!

Issue #1823, opened 2018-09-13
Buyers Guide: donate popup


kristinashu:
Based on design in #1739

Comps https://redpen.io/p/qz6bdb2811e058ffb1

### Functionality
- slides up from the bottom of the page
- user must be on the site for 10 second [need to discuss] before it slides up
- clicking close icon, donate, or outside of the popup will close it

### Copy
- `We made this guide with support from people like you`
   Zilla Slab Light 34px with 42px line height
- `Our supporters told us they are uncertain about how to be safer online. We listened. This guide is a result.` 
  .normal paragraph but white
- `Help us keep this work going` 
  Zilla Slab Regular 22px with 28px line height
- `Support Mozilla`

### Styling
- donate button similar to header but bigger 48px height
- background color of #333333 
- earth image

![earth 2x](https://user-images.githubusercontent.com/1068783/47503952-348fc780-d863-11e8-8df2-6c45aca69920.jpg)

### Comps

![donate copy 2x](https://user-images.githubusercontent.com/1068783/47504431-268e7680-d864-11e8-91a8-eb41b98717b8.jpg)

<img width="319" alt="image" src="https://user-images.githubusercontent.com/1068783/47504704-b2080780-d864-11e8-99f5-376d2c6d1b47.png">

Pomax:
Do we have a higher quality version of that earth image? (there's a lot of jpg artifacting in the one linked)

WillatMozFdn:
Copy & overall design look fine to me.  Thanks!!!

PR #2054, opened 2018-10-25
Buyer's Guide - Donate modal

Pomax:
Adds a new `donate-modal.jsx` component that slides up from below at (eventually) 10 seconds

Closes https://github.com/mozilla/foundation.mozilla.org/issues/1823 when actually done.


source/js/buyers-guide/components/donate-modal/donate-modal.jsx
```diff
@@ -0,0 +1,64 @@
...
+    }
+  }
+
+  render() {
+    if (this.state.dismissed) {
+      return null;
+    }
+
+    return (
+      <div className={`donate-modal ${this.state.visible ? 'show' : ''}`}>
```
Pomax: to be honest, we could even hard-code this in the base template, but that makes the interaction a bit more work in bg-main.js. Thoughts, @alanmoo ?

alanmoo: No specific opinions here, one way or the other. I'd say optimize for browser performance; I noticed pages choking up a bit before the user could scroll when loading on older phones.

alanmoo:
This is looking pretty good. What's remaining other than a few more adjustments/cleanups like console.log and timing?

kristinashu:
(mental note: we should add this to the main foundation site too)

kristinashu:
@alanmoo this is ready for you to review!

source/js/buyers-guide/components/donate-modal/donate-modal.scss
```diff
@@ -0,0 +1,98 @@
...
+      transform: rotate(45deg);
+    }
+
+    &::after {
+      transform: rotate(-45deg);
+    }
+  }
+
+  .container {
+    padding-top: 60px;
```
alanmoo: I'm going to land this because it could be useful to have before the email blast goes out. But things like this can be added using Bootstrap utility classes in markup, rather than targeting existing bootstrap containers. See https://getbootstrap.com/docs/4.1/utilities/spacing/

PR #2077, opened 2018-10-29
Vote timestamp

cadecairos:
This PR adds a `created_at` field to vote records that is set upon creation of the record. The migration automatically sets existing vote records to `timzone.now()` during the migration.


alanmoo:
Sweet, thank you!

PR #2078, opened 2018-10-30
Fix Cyclical redirect

### LINKED ISSUE #2074: Buyers Guide: cyclical redirect?
I'm seeing `Failed to load resource: net::ERR_TOO_MANY_REDIRECTS` for pointing to `/privacynotincluded/products/apple-watch-4/static/embed.js:1`, not sure yet why that's getting generated, but it's there...

cadecairos:
Closes #2074 

We created this cyclical redirect because any actual 404 url under the `/privacynotincluded/` path would eventually hit the redirect created by the i18n routing. This would append a `/en/` to the path, and redirect it back into the route we added to strip that locale prefix, thus creating a loop.

The only real way to prevent this issue is to stop trying to "un-i18n-ify" the `privacynotincluded` namespace with redirect views. The site is in english, so it can safely live at `/en/` and we can safely rely on the i18n code to redirect non-en prefixed urls in the wild to the prefixed version.

### DIFF
--- a/network-api/networkapi/buyersguide/tests.py
+++ b/network-api/networkapi/buyersguide/tests.py
@@ -7,7 +7,7 @@ from django.test import TestCase, RequestFactory
 
 from networkapi.buyersguide.factory import ProductFactory
 from networkapi.buyersguide.models import RangeVote, BooleanVote, Product
-from networkapi.buyersguide.views import buyersguide_home, product_view, category_view
+from networkapi.buyersguide.views import product_view, category_view
 from django.core.management import call_command
 
 VOTE_URL = reverse('product-vote')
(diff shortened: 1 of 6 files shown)


cadecairos:
@Pomax I think I broke something in the i18n management commands?

network-api/networkapi/buyersguide/tests.py
```diff
@@ -318,20 +318,17 @@ def test_homepage(self):
         """
         Test that the homepage works.
         """
-        request = self.factory.get('/privacynotincluded/')
-        response = buyersguide_home(request)
+        response = self.client.get('/en/privacynotincluded/')
+        # response = buyersguide_home(request)
```
Pomax: rogue comment

Pomax:
Removed debug toolbar to see if that either fixes the problem, or gives us a _real_ clue as to what's going wrong.

Pomax:
🐻 👨 🚀 💥

PR #2093, opened 2018-11-01
Fetch Comment counts from Talk and upsert to stats DB

cadecairos:
Uses Talk's GraphQL endpoint to comment counts for all the comment streams known to Talk, and upserts the data to the same database as the product vote stats.


cadecairos:
testing can be tricky, if you'd like I can help you get the code running tomorrow during our meeting

Issue #1984, opened 2018-10-19
Improve Buyer's guide homepage performance


alanmoo:
- [ ] Add caching (in Django settings?) for product images
- [ ] Reduce image load impact on homepage in some way (template resize filter, lazy load, etc. Not picky)

xmatthewx:
@cadecairos @alanmoo - have either of you had a chance to look at this? feels important in general, and especially considering the unique bandwidth issues at MozFest.

cadecairos:
I can write a script to add Cache-Control metadata to images uploaded for the buyer's guide.

cadecairos:
Just to update this issue, We've applied caching to the product data fetched from the database, which has reduced [TTFB](https://en.wikipedia.org/wiki/Time_to_first_byte) times from roughly 4.5 seconds to 0.2-0.4 seconds

@patjouk has an in-progress branch implementing Cloudinary as an image storage and delivery service, so we can specify in our templates the dimensions and compression level of images, and let cloudinary handle the work.

xmatthewx:
After this is done, should we look at cloudinary for all foundation site and pulse stuff? pulse is littered with oversized images.

patjouk:
I think that's the idea. I started exploring Cloudinary with pulse-api but switch to buyers guide because priorities :D

Pomax:
There's two merged PRs, but still unchecked boxes in the original comment, is this issue still open or did we forget to tick off the boxes and close the issue?

PR #2116, opened 2018-11-05
[WIP] Add Cloudinary support for the Buyers Guide

patjouk:
Ref #1984

This is a WIP PR on the work I'm doing to add Cloudinary support for the Buers Guide. It won't work on a review app for now because it's missing cloudinary credentials (cf todos).

### Todos:
- [x] Update image sizes in template,
- [x] Remove Cloudinary support when in local dev mode (in views, templates, and hide the cloudinary_image/image in admin dependending on what we're using).
- [x] Update docs,
- [x] Update app.json + add env vars to staging,
- [x] Adapt factories to use images on Cloudinary for review apps
- [x] Prepare images migration from S3 to Cloudinary
- [ ] Seperate staging and prod images: separation is done, now I need to find a way to avoid duplicate images (copydb script)

### Devops todos
- [x] Add cloudindary to status gator or subscribe to status notifications,
- [x] Check what happens when we reach the limits of our current plan: can we get a notifications before/check how much they charge for usage outside the plan (they probably won't cut their service but hey, you never know)

@kristinashu What kind of default sizes should I use for the products' picture. I have 300px width for the homepage and category page, and 400px for the product page (this one is the one actually used). We will replace hardcoded image sizes by breakpoints in the future, this is temporary and just to get this ready for launch.  

## Images Migration
I added a new management command called `image_migration`. The way it works is that it get images from s3, upload a copy of it on Cloudinary and update the `cloudinary_image` field of the product.

Deploy process (for staging and prod):
- Remove images on cloudinary: everything in `foundationsite` can be deleted. For foundationsite-staging, keep the `images/placeholders` content: it's used for review apps.
- Deploy code and wait until it's done,
- Set the new env vars but keep `USE_CLOUDINARY` at False for now,
- run `heroku run python network-api/manage.py image_migration -a [ProdApp or StagingApp]`,
- Once it's done, change `USE_CLOUDINARY` to True. Wait for the website to be redeployed by Heroku and check if the images are using a cloudinary url instead of a S3 assets url.

**What to do if images are broken?**
- If it's only one or two images, something failed during the migration: update the missing images in admin and it should fix the issue. You might need to refresh the homepage cache.
- If it's every images: check in admin if products have images. If yes, try reloading the page again. If it's still broken, set `USE_CLOUDINARY` to False to use images on S3 instead while you're debugging the issue.


kristinashu:
@patjouk yes, 300px and 400px sound right to me.

kristinashu:
@patjouk you mentioned you set the image at 85% quality. Does that mean it's reducing all the images everywhere? Or just on the home page? Or just mobile? Ideally, the product detail pages would show the full resolution. 

I'm also testing to see if 85% is too much or not but noticing something... I uploaded an image to faker product "time similar" which is the first product on the homepage. The image is updated on the [product detail page](https://foundation-mofostaging-pr-2116.herokuapp.com/en/privacynotincluded/products/time-similar/) and the [category page](https://foundation-mofostaging-pr-2116.herokuapp.com/en/privacynotincluded/categories/Entertainment/) but I don't see it on the home. It's not my browser caching but it is something else caching? Totally fine if it's intentional, just want to make sure it isn't a bug.

patjouk:
Right now, I've set up 85% for every template: we can change it for the
product page only if we want, etc.
For the homepage, it's a cache issue: I can't invalidate it for the moment
because of a bug but Chris is aware of it.

I'll sync with Alan today to see what other kinds of optimizations we could
squeeze in.

patjouk:
After changing the sizes and quality, we're now at 1.30MB for the homepage. I have a few issues to fix so I don't think we gonna support srcset for now but it's definitely something we can revisit in the future.

network-api/networkapi/buyersguide/models.py
```diff
@@ -347,6 +448,13 @@ def __str__(self):
         return str(self.name)
 
 
+# We want to delete the product image when the product is removed
+@receiver(pre_delete, sender=Product)
```
cadecairos: This is nice.

PR #2117, opened 2018-11-05
Buyersguide en only

cadecairos:
In buyers guide views, it will redirect if the path is not prefixed with `/en/`

Includes tests.


network-api/networkapi/buyersguide/tests.py
```diff
@@ -329,18 +329,42 @@ def test_localised_homepage(self):
         response = self.client.get('/privacynotincluded/')
         self.assertEqual(response.status_code, 302, 'simple locale gets redirected')
 
+    def test_only_en(self):
+        """
+        Test that the homepage redirects away from the other locales the site supports
+        """
+        response = self.client.get('/fr/privacynotincluded', follow=True)
```
cadecairos: Not gonna lie, something weird was going on here. when running the site, I saw 302's in the browser, but when using the test client, it was returning 301. Decided to get the client to follow redirects and check the end result by inspecting the redirect_chain.

network-api/networkapi/buyersguide/views.py
```diff
@@ -14,6 +15,8 @@
 
 vote_throttle_class = UserVoteRateThrottle if not settings.TESTING else TestUserVoteRateThrottle
 
+path_regex = re.compile(r"^/\w\w/")
```
Pomax: do we just want `\w\w` or do we want `[a-z]{2}(-[A-Z]{2})?`

cadecairos: This project only supports [these locales](https://github.com/mozilla/foundation.mozilla.org/blob/8be51a581f3696173dd729e9634a1cec1099e5f6/network-api/networkapi/settings.py#L333-L340) at this time, so we don't need to worry about anthing more complicated than a two-character code. I can make them `[a-z]{2}` but it would accomplish the same thing

alanmoo: Go for the more verbose one now so it doesn't trip us up in the future, for the same reason Pomax describes.

network-api/networkapi/buyersguide/views.py
```diff
@@ -28,7 +31,19 @@ def get_average_creepiness(product):
...
+    return path.startswith('/en/')
+
+
+def get_en_redirect(path):
+    redirect_path = re.sub(path_regex, '/en/', path, count=1)
+    return redirect(redirect_path, permanent=False)
+
+
 def buyersguide_home(request):
+    if not path_is_en_prefixed(request.path):
```
Pomax: This feels like it wants to be a decorator, given how many places we repeat this code in.

cadecairos:
@Pomax this is ready for re-review

Issue #1843, opened 2018-09-20
Add referral source to Particpate Donate button


xmatthewx:
We have an awesome new CTA to donate in the middle of the participate page (see below).

It doesn't look like this has UTM parameters in the link to track donation source. We should add that. Also probably good to add an analytics event if it isn't already there. This will give us a baseline we can use later to test alternate CTAs. 

cc @alanmoo @kristinashu to triage

<img width="1115" alt="image" src="https://user-images.githubusercontent.com/53411/45845311-cbacb100-bcf2-11e8-955e-38d5fcdb0477.png">

mmmavis:
@xmatthewx do you know where I can look for values for UTM and GA events that we wanna use?

xmatthewx:
For GA, we track events for both pulse and site in [this spreadsheet](https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=779058050). I added line 23 for this. 

category | action | label
:-- | :-- | :--
donate | donate button tap | [page title]

@alanmoo - can you lend a hand with UTM? Should we use the same as the header donate button, just changing `utm+content=header` to `=participate`?

alanmoo:
I have no opinions on the UTM, so sure.

mmmavis:
<img width="899" alt="image" src="https://user-images.githubusercontent.com/2896608/48034885-10809000-e159-11e8-8b19-18bcccf73f1b.png">


CTA button text and URL are set via Wagtail. Should we just change the URL to from `https://donate.mozilla.org` to`https://donate.mozilla.org?utm_source=foundation.mozilla.org&utm_medium=referral&utm_content=participate` then or?

alanmoo:
That makes sense to me.

kristinashu:
@mmmavis is this all done?

mmmavis:
Oops forgot to close this one.

PR #2125, opened 2018-11-06
Track donate button taps on Participate page

mmmavis:
Related to #1843 

URL for the donate button will be updated (to `https://donate.mozilla.org?utm_source=foundation.mozilla.org&utm_medium=referral&utm_content=participate`) once this has been pushed to production.


source/js/analytics.js
```diff
@@ -1,5 +1,3 @@
-import ReactGA from 'react-ga';
```
mmmavis: `ReactGA` not being used anywhere in this file

mmmavis:
@alanmoo let me know if looking at the code is enough for you to review this PR. If not, I'll manually create a Participate page that is identical to what we have on production.

source/js/main.js
```diff
@@ -203,6 +203,16 @@ let main = {
         Analytics.sendGAEvent(`navigation`, `page footer cta`, label);
       });
     }
+
+    if (document.querySelector(`#view-participate .card-cta .btn[href*="donate.mozilla.org"]`)) {
+      document.querySelector(`#view-participate .card-cta .btn[href*="donate.mozilla.org"]`).addEventListener(`click`, () => {
```
alanmoo: Can this be saved as a variable so we don't have to fetch it twice? Not necessarily from a performance standpoint, but from a readability one.

mmmavis: Yea I thought about doing this but decided to go with what we already have in the rest of the code. (No variable is declared for any DOM nodes.) Maybe we should file a followup ticket to tidy things up all at once?

Pomax:
looks like this still needs that refactor to cache the query selector result, @mmmavis?

Pomax:
looks like this still needs that refactor to cache the query selector result, @mmmavis?

Issue #2001, opened 2018-10-22
Buyers Guide: add seal of approval


kristinashu:
Based on design in https://github.com/mozilla/foundation.mozilla.org/issues/2091

Gavin already added a field to the cms product page with the three answers `yes` `no` `unknown`

<img width="956" alt="screen shot 2018-10-22 at 12 31 42 pm" src="https://user-images.githubusercontent.com/1068783/47314430-7831c980-d5f6-11e8-8a55-2708165e109d.png">

But there has been an update and we will now only have one badge. Badge will say "Meets Our Minimum Security Standards" and only appear if the answer is `yes`.

### New requirements are:
- `yes` shows badge on product page
-  `unknown` and `no` show nothing (could remove `unknown` and replace with `no`)
- svg and pnd for `yes` [badge.zip](https://github.com/mozilla/foundation.mozilla.org/files/2555087/badge.zip)
- placement for [wide](https://redpen.io/bcd47ecb6d670654da) screen (140x140px)
<img width="1012" alt="image" src="https://user-images.githubusercontent.com/1068783/48094847-16618a00-e1c8-11e8-959c-989c38c40494.png">

- placement for [narrow](https://redpen.io/fk5813dda329fda302) screen (76x76px)
<img width="478" alt="image" src="https://user-images.githubusercontent.com/1068783/48094865-224d4c00-e1c8-11e8-8507-48c6d3402337.png">

kristinashu:
@mmmavis I've update the first comment with the new badge design so this is all final now.

kristinashu:
ᶘ ᵒᴥᵒᶅ

PR #2126, opened 2018-11-07
Buyer's Guide - added seal of approval

mmmavis:
Related to #2001

I also reduced padding for mobile in the product details section so it looks closer to what we have on the mockup. Let me know if I should revert this styling fix back.


network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -11,12 +11,21 @@
 
 {% block guts %}
 <div class="container-fluid text-center product-header bg-gray">
-  <img class="mb-0" src="{{mediaUrl}}{{"AWS_LOCATION"|env}}/{{product.image}}" width="400" alt="{{product.name}}"/>
+  <div class="container">
+    <div class="row">
+      <div class="col-12 col-md-10 offset-md-1">
+        <img class="mb-0" src="{{mediaUrl}}{{"AWS_LOCATION"|env}}/{{product.image}}" width="400" alt="{{product.name}}"/>
+        {% if product.meets_minimum_security_standards %}
```
Pomax: This works, but given that this is a tristate value, it's better to be explicit about which we expect to find:
```suggestion
        {% if product.meets_minimum_security_standards is True %}
```

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -11,12 +11,21 @@
...
+        <img src="/_images/buyers-guide/badge.svg" class="seal-of-approval" alt="this product meets minimum security standards" />
```
Pomax: Do we want to be more descriptive in this alt text?
```suggestion
        <img src="/_images/buyers-guide/badge.svg" class="seal-of-approval" alt="this product meets Mozilla's minimum security standards" />
```
/cc @kristinashu

mmmavis: I'm gonna go with `This product meets our minimum security standards` since that's what it says in the seal

Issue #2131, opened 2018-11-07
Buyer's Guide - .jsx, .js, and .scss files don't get linted


mmmavis:
In `package.json`

```
"test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/components/**/*.jsx\" webpack.config.js",
"test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/components/**/*.scss\" --syntax scss",
```

should probably be changed to 

```
"test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
"test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" --syntax scss",
```
so files under the `source/js/buyers-guide` directory get covered by npm test.

Pomax:
good call.

PR #2132, opened 2018-11-07
update the rebuild scripts

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/2131


source/js/buyers-guide/components/filter/filter.scss
```diff
@@ -53,20 +53,17 @@
     right: 0;
     left: 0;
     width: 100%;
-    background: white;
```
mmmavis: a dupe

PR #2133, opened 2018-11-08
[WIP] Bubbles

Pomax:
Closes #

[ ] Migrations squashed


kristinashu:
Sweet! I love it.

Just seems like the font for `A little creepy!` (and also `Filter`) is too bold. Can you make it a font weight less?

screen-shot-2018-11-08-at-12 11 12-pm



Also doesn't show up on mobile, but that can be worked out later as a follow up.

image

Pomax:
They were set to 600, so I've set them to 400, does that look better? (heroku's still updating at the time of this comment)

kristinashu:
Opened a ticket for mobile (low priority) https://github.com/mozilla/foundation.mozilla.org/issues/2141

PR #2149, opened 2018-11-09
CLOUDINARY_URL does not need to be set in the environment

cadecairos:
Removes `CLOUDINARY_URL` as a required environment variable for review apps, and stops trying to load it in settings.py.


patjouk:
👍

PR #2174, opened 2018-11-13
add testing for the Procfile

Pomax:
Follow-up on https://github.com/mozilla/foundation.mozilla.org/pull/2167

This hardens the Procfile by adding a dumb, but effective, test script that checks whether there are only lines of the form `keyword: ...` in it. Empty lines are ignored, but lines that don't start with a `{keyword, colon}` combination are considered build-breakingly bad.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,9 +23,10 @@
     "server": "pipenv run python network-api/manage.py runserver",
     "start": "npm i && npm run build-uncompressed && run-p server watch:**",
     "snyk": "snyk test --file=package.json",
+    "test:procfile": "node test/test-procfile.js",
     "test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,9 +23,10 @@
     "server": "pipenv run python network-api/manage.py runserver",
     "start": "npm i && npm run build-uncompressed && run-p server watch:**",
     "snyk": "snyk test --file=package.json",
+    "test:procfile": "node test/test-procfile.js",
     "test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
     "test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" --syntax scss",
-    "test": "npm run build && run-s test:**",
+    "test": "run-s test:** build",
```
Pomax: I don't understand why we ran the build before we ran the linting, so this line change ensures we never even build if linting fails.

Issue #2087, opened 2018-10-31
Rename GA Event Category


xmatthewx:
Change the Event Category `site` to `buyersguide` for less ambiguity. Why? These events will be mixed with foundation site events. This was my oversight on the initial request. 

The change is noted in [column E of the spreadsheet](https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=980429434) for each affected event.

After this is done, let's update the spreadsheet and delete that column. 

Reminder: After this lands in production, I will need to update the datastudio dashboard.

Issue #2024, opened 2018-10-23
Buyers Guide: GA con't


kristinashu:
Key events have been done https://github.com/mozilla/foundation.mozilla.org/issues/1925 but still a few more in the doc https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=980429434

xmatthewx:
I set up a segment so you can filter most analytics views just for users who hit at least one `/privacynotincluded/` page in their session. 

To view segment (see screenshot)
 1. Tap All Users
 2. Uncheck All Users from list
 2. Search privacy
 3. Select Privacy Not Included
 4. Hit Apply

<img width="986" alt="image" src="https://user-images.githubusercontent.com/53411/47392466-b3adc000-d6ea-11e8-9ae0-596b0da09246.png">

<img width="734" alt="image" src="https://user-images.githubusercontent.com/53411/47392674-43536e80-d6eb-11e8-9eed-68494c73b044.png">

....I believe segments are shared between users, but let me know if that doesn't seem work.

xmatthewx:
Of all these, the site-wide share button will be most important to add, since these shares are part of the key engagement metric we're tracking.

Assigning @alanmoo to triage 😉

Pomax:
@xmatthewx hm, we added event for those already though, are they not showing up? (the primary nav has GA hooked up for both the desktop and mobile views as far as I can see)

kristinashu:
The header shares always share the homepage. But it doesn't look like they are showing up in GA.

xmatthewx:
There were no events for this yesterday. I now see 2 events from today. I think we're good for GA on site wide shares and all the events critical launch. 

Events > Action: Share tap > Category: Site

<img width="321" alt="image" src="https://user-images.githubusercontent.com/53411/47514877-6b5be280-d84f-11e8-80a1-e17641dbbe72.png">

kristinashu:
Ok awesome!

xmatthewx:
[...]

PR #2175, opened 2018-11-13
rename 'site' to 'buyersguide' and add filter event

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2087 and addresses the filter event from the spreadsheet (for the homepage only).

Partially addresses https://github.com/mozilla/foundation.mozilla.org/issues/2024


source/js/buyers-guide/components/filter/filter.jsx
```diff
@@ -82,6 +84,14 @@ export default class Filter extends React.Component {
 
       this.updateWithCSSvalues(update);
       this.setState(update);
+
+      if(DNT.allowTracking) {
```
mmmavis: in `bg-main.js`

```js
if (DNT.allowTracking) {
  ReactGA.initialize(`UA-87658599-6`);
  ReactGA.pageview(window.location.pathname);
}
```

ReactGA won't get initiated if DNT.allowTracking is turned off. I wonder if we still need to check for that value here?

Pomax: Ah, it won't get _initialized to our ga-id_ but it'll still be an active object, so its better to not even trigger it at all if we know that tracking is not appreciated. It's probably even better to make a little wrapper lib that proxies ReactGA calls, but a little more work than should happen in this PR specifically (so I filed https://github.com/mozilla/foundation.mozilla.org/issues/2193 for that work)

mmmavis: Cool. Thanks for filing that ticket!

PR #2178, opened 2018-11-13
Clicking outside the donate modal should not close it

### LINKED ISSUE #2171: Clicking outside the donate modal should not close it
Opinion bug here, @kristinashu what do you think? I think the donate modal should be a bit more permanent- we don't want it going away when someone clicks to vote, for example. Usually, I'm all about giving options to get rid of modals, but this one doesn't block the entire page and we really want to encourage people to donate.

mmmavis:
Closes #2171


Pomax:
lgtm!

Issue #2143, opened 2018-11-09
Fix emoji sprite offset in Firefox and Chrome


kristinashu:
On desktop, the super creepy face disappears at the very bottom, you should still be able to see it.
<img width="1159" alt="screen shot 2018-11-08 at 4 33 44 pm" src="https://user-images.githubusercontent.com/1068783/48236060-75183680-e375-11e8-9663-e5762ef3ab42.png">

On mobile, I see the happy face at the very very bottom of scrolling, should see the super creepy face.

<img width="324" alt="image" src="https://user-images.githubusercontent.com/1068783/48236128-b577b480-e375-11e8-8be7-2d5ab4e030de.png">

kristinashu:
Doesn't look fixed on staging or prod.

Pomax:
It would appear that this is a mismatch between Chrome and Firefox, where:
- chrome: keeps the face, does not seem to compute the sprite offset correctly
- firefox: computes the sprite offset correctly, but is off by one for the most-creepy frame it looks like

PR #2187, opened 2018-11-14
Fix emoji placement and offset

Pomax:
**Actually** closes https://github.com/mozilla/foundation.mozilla.org/issues/2143


kristinashu:
I still wouldn't know if it's working or not but let's see on staging!

image

Issue #2196, opened 2018-11-15
Add mechanism to control which average creepiness is "happy"


Pomax:
Right now frame 0 is reserved for creepiness values 0-2.5 which means there is effectively no way for people to ever see the happy face.

Adding a step that remaps creepiness at the time of reading-in would allow us to control at which creepiness we "start" moving off of the happiest-face towards. Might even give us some A/B testing opportunities?

kristinashu:
Yes! Jen just mentioned she would love for the first few products to be "not creepy". So we're not so ominous.

Pomax:
I've set up a PR for this, on https://foundation-mofostaging-pr-2198.herokuapp.com/privacynotincluded, with the range for cosmetic purposes changed to 25-80, so the emoji will change as:

- any creepiness < 25 is considered not creepy, super happy
- any creepiness in the range 25-80 maps the the emoji transitions if it's the range 0-100
- any creepiness > 80 is considered super creepy recoily face.

PR #2198, opened 2018-11-15
Custom emoji mapping

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2196, as long as we can make sure the minimum and maximum values are acceptable =)

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "optimize:png": "find source/images -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
     "optimize": "run-p optimize:**",
     "server": "pipenv run python network-api/manage.py runserver",
-    "start": "npm i && npm run build-uncompressed && run-p server watch:**",
+    "start": "npm i && run-p build-uncompressed server watch:**",
     "snyk": "snyk test --file=package.json",
     "test:procfile": "node test/test-procfile.js",
     "test:eslint": "eslint --config ./.eslintrc.yaml \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -21,7 +21,7 @@
     "optimize:png": "find source/images -type f -name '*.png' -print0 | xargs -0 -n 1 -P 6 optipng",
     "optimize": "run-p optimize:**",
     "server": "pipenv run python network-api/manage.py runserver",
-    "start": "npm i && npm run build-uncompressed && run-p server watch:**",
+    "start": "npm i && run-p build-uncompressed server watch:**",
```
Pomax: mostly a dev Quality of Life change - this starts the server _way_ earlier, so you can just reload the tab once webpack finishes the build, instead of getting a "server not found" for like... a minute

source/js/buyers-guide/homepage-c-slider.js
```diff
@@ -1,13 +1,28 @@
...
+// the emoji while scrolling.
+//
+// Note: this is a cosmetic value for scroll only.
+const MINIMUM_HAPPINESS_RATING = 25;
+
+// Our threshold value beyond which everything is super
+// creepy by default.
+//
+// Note: this is a cosmetic value for scroll only.
+const MAXIMUM_CREEPINESS_RATING = 80;
```
mmmavis: Are `MINIMUM_HAPPINESS_RATING` and `MAXIMUM_CREEPINESS_RATING` both fixed values unrelated to `EMOJI_FRAME_HEIGHT` and `SPRITE_FRAME_COUNT`? If there's any relationship between them I think it's worth adding some inline comments in case we decide to change the values for frame height and frame count in the future.

Pomax: There is zero relation between the two sets of numbers.

Issue #2199, opened 2018-11-15
Buyers Guide: add a link to the github repo to the footer


Pomax:
Having an email address at the bottom of the about page is not enough to get people to send feedback. We've had numerous people complain that we don't have any information on where they should go with bug reports, which is a problem we already solved in the past by adding a github link to our footer. 

So: let's copy what we do on https://www.mozillapulse.org to the PNI footer.

PR #2200, opened 2018-11-15
add our gh link to the footer

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2199

Adds a GH link next to twitter, and moves "license" so that it's next to legal (since it's important to keep GH and CoC links next to each other, and licenses + legal form a logical set).

image


Pomax:
Different templates so if we do, let's file that as a separate issue.

kristinashu:
Ok!

PR #2205, opened 2018-11-15
Set the default CSP directive to self (properly quoted this time)

### LINKED ISSUE #2168: Fix buyer's guide console errors
Looks like there's a few errors on the buyer's guide around font-src CSP, and some logging related to Talk. Can we clean that up?

cadecairos:
Fixes #2168

I've modified the default CSP directive, and I've added a cononical link tag to product pages. This should stop the logging from the browser about "self" and from talk about the missing tag.


network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -7,6 +7,7 @@
 
 {% block og-title %}
   <meta property="og:title" content="privacy not included - {{ product.name }}" />
+  <link rel="canonical" href="{{ request.scheme }}://{{ request.get_host }}{{ request.get_full_path }}" />
```
Pomax: Does this require canonical links for the homepage, too?

cadecairos: I added this here because it's something Talk asks for specifically, and it's only on product pages.

cadecairos: The block name doesn't quite make sense. I'm going to update that

cadecairos:
@alanmoo / @Pomax  can I get a review?

Issue #2197, opened 2018-11-15
Remove thumbs from homepage


kristinashu:
So everyone is confused by the thumbs on the homepage. They think they can click it to up/down vote, they think it's mozilla rating the products... We should remove them since the cost of the majority of users being confused by the thumbs is more than the cost of the few people who use the filters to filter by likelihood to buy and could be confused by not seeing the thumbs on the product.

- [ ] Remove thumbs from products on homepage (but keep them as a filter)

We are thinking of add product titles as a hover states so could maybe show the thumbs there too, but need to mock it up first.

Pomax:
Removing any indication of what those icons stood for makes it unclear what the filter does, filtering on a secret property. Can we instead replace the thumbs with something that is perceived as an informative element rather than an interactive one? E.g. replace them with the phrase "A likely purchase" or "An unlikely purchase" or something?

kristinashu:
Do you mean show that text over top every image, always? I think that would work well on hover (with the product titles when we do that - working on design now).

Pomax:
Ah, putting that in the hover is a good idea - shall we block this issue on the hover title etc. then?

kristinashu:
It's tough because right now everyone is confused by having the thumbs there so we're keen to remove it asap. Some people will be confused by not seeing the thumbs when they use the filters, but since fewer people use the filter so seems like an ok trade off.

Pomax:
kk, PR incoming.

kristinashu:
One thought... would it be possible to not show the thumbs. But then if a user filters so they only see "Likely to buy" then they would see those thumbs over the product. And vise versa for "Not likely to buy". But then if they click "Both" then they don't see any thumbs?

Pomax:
it's a little more work, but certainly possible.

kristinashu:
This is done and it's great 🎉

PR #2207, opened 2018-11-15
remove the thumb icons from the homepage

Pomax:
Closes #https://github.com/mozilla/foundation.mozilla.org/issues/2197


kristinashu:
I see what you mean. This PR is good but would need to revisit mini seal PR. Do you have to add that extra gray bar? I think the seal can fit within the gray border that surrounds the entire image.

Pomax:
PR updated, we'll have to see what it looks like with some actual images on staging =)

Pomax:
@mavis added you for code review, since this PR also updates how the filter works: as long as the filter is set to "both" for buying likelihood, there will not be any thumb up/down icons shown. When someone picks "likely" or "not lilkely", the appropriate icon is shown on all the resultant products.

Also since it's _technically_ possible to not have a recommendation HTML element, I added an `if` to make sure that if the queryselector fails, the filter doesn't break and disappear =)

Issue #2192, opened 2018-11-14
Buyers Guide: horizontal scrollbar when browser window is "narrow"


Pomax:
On windows, if I resize Chrome or Firefox to 767px wide, a horizontal scrollbar pops into existence for the document (not the body O_o) that, when used, scrolls the entire body, revealing a black document background.

Issue #2120, opened 2018-11-05
Add dot background to Updates section 


kristinashu:
Assets for dot background:
- should be a black square 3x3px centered in a 9x9px white square
- jpg and svg of the repeatable pattern [dot.zip](https://github.com/mozilla/foundation.mozilla.org/files/2412860/dot.zip)
- Pattern is for "updates" section only. Height for the pattern at 228px. It is only visible if there are updates in that section.

- once repeated, it should look like this
<img width="933" alt="image" src="https://user-images.githubusercontent.com/1068783/47882557-24d62d00-dde7-11e8-993a-31aca4f239f9.png">

PR #2208, opened 2018-11-15
Add a full-width dot pattern for updates

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2120, closes https://github.com/mozilla/foundation.mozilla.org/issues/2192

### DIFF
--- a/source/sass/buyers-guide/views/product.scss
+++ b/source/sass/buyers-guide/views/product.scss
@@ -176,7 +176,7 @@
   }
 }
 
-.dot-pattern {
+.row.dot-pattern {
   position: absolute;
   top: 0;
   left: -10em;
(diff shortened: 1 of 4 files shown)


kristinashu:
Added an Update to https://foundation-mofostaging-pr-2208.herokuapp.com/en/privacynotincluded/products/institution-available/ and something odd has happened. There's a bunch of black space on the right side of the page which I can scroll over to.

image

Pomax:
interesting. I can't reopen this PR?

Pomax:
And now I can. Weeeeird

kristinashu:
Desktop looks good!

But would it be possible to add the white box to mobile?

Comp https://redpen.io/fk5813dda329fda302
image

Screen cap from staging
image

mmmavis:
true full width*

kristinashu:
Oooh yes, it should! I didn't notice that.

source/sass/buyers-guide/views/product.scss
```diff
@@ -176,6 +176,26 @@
   }
 }
 
+.row.dot-pattern {
```
mmmavis: Since the dot pattern is solely for decoration purposes, I would add it in as a pseudo element in CSS like `:before` instead of adding extra DOM nodes just to add background pattern in.

Pomax: done'd

PR #2217, opened 2018-11-17
Change "replace by" to "replace with" in README

ZekeMedley:
Changes wording from passive voice to active voice more appropriate for instructions.


Pomax:
thanks!

PR #2233, opened 2018-11-19
fix to the donate popup GA label

Pomax:
Addresses the "donate tap" event not saying which page it was triggered on.

See https://github.com/mozilla/foundation.mozilla.org/issues/2024

Also see L15 in https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=980429434

### DIFF
--- a/source/js/buyers-guide/components/donate-modal/donate-modal.jsx
+++ b/source/js/buyers-guide/components/donate-modal/donate-modal.jsx
@@ -52,10 +52,12 @@ export default class DonateModal extends React.Component {
 
   handleBtnClick() {
     if (DNT.allowTracking) {
+      let url = window.location.pathname.replace(/\w\w(-\W\W)?\/privacynotincluded\//,'');
+
       ReactGA.event({
         category: `buyersguide`,
         action: `donate tap`,
-        label: `donate popup`
+        label: `donate popup on ${url}`
       });
     }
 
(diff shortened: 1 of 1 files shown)


source/js/buyers-guide/components/donate-modal/donate-modal.jsx
```diff
@@ -52,10 +52,12 @@ export default class DonateModal extends React.Component {
 
   handleBtnClick() {
     if (DNT.allowTracking) {
+      let url = window.location.pathname.replace(/\w\w(-\W\W)?\/privacynotincluded\//,``);
+
       ReactGA.event({
         category: `buyersguide`,
         action: `donate tap`,
-        label: `donate popup`
+        label: `donate popup on ${url}`
```
mmmavis: in https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit#gid=980429434 it seems like we want "page title" instead?

row 15: `donate popup on [page title]`

Pomax: however, the page title is a constant string, so it would only ever be "donate popup on Mozilla - *privacy not included" which I assume is not what we want. I figured the "url fragment" that actually tells us which page the donate modal ended up being clicked on made more sense, but maybe @xmatthewx can "make a ruling" here =)

xmatthewx: Yes, that makes sense. Let's use whatever string will give us easiest at-a-glance understanding of when/where people are inspired to donate.

Issue #2183, opened 2018-11-14
Filters should fill screen at narrow widths


kristinashu:
Filters at narrow widths fill the screen
<img width="665" alt="image" src="https://user-images.githubusercontent.com/1068783/48097722-80316200-e1cf-11e8-893d-945e1edeaeea.png">

PR #2237, opened 2018-11-20
Filter tweak

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2183 by having the rectangular filter style kick in at small rather than medium width, and having it go from transitioning to immediate, with a full screen "overlay".

### DIFF
--- a/source/js/buyers-guide/components/filter/filter.scss
+++ b/source/js/buyers-guide/components/filter/filter.scss
@@ -1,4 +1,6 @@
 #product-filter {
+  $bp-cs: 500px;
+
   $filter-bubble-dimension-mobile: 70px;
   $filter-bubble-dimension: 6em;
   $filter-box-dimension: 700px;
@@ -12,7 +14,7 @@
   z-index: 200;
   font-family: "Nunito Sans", Helvetica, Arial, sans-serif;
 
-  @media (min-width: $bp-md) {
+  @media (min-width: $bp-cs) {
     top: calc(-1 * #{$filter-bubble-dimension} + 0.8em);
   }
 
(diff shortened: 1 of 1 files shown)


kristinashu:
Hmm yeah I see what you mean. It looks good a small widths. 

It also looks good at around 582px with the edges cutoff. Screen cap from PR:
screen shot 2018-11-19 at 5 24 06 pm

But at around 430px it feels a little too blank. Screen cap from PR:
image

If possible, I think the edges could be cutoff more. Maybe as much as like this (430px):
screen shot 2018-11-19 at 5 22 28 pm

Hope that makes sense.

Pomax:
I'll see if that's a built in bootstrap value - if not adding a sass var to represent (effectively) "extra small" should be relatively straight forward

mmmavis:
Not sure which PR gets merged first but there will be code conflicts with this PR ["Added product name to homepage & thumbnail sizing"](https://github.com/mozilla/foundation.mozilla.org/pull/2232) (which extends the filter popup to full width). Just fyi

Pomax:
As a larger PR, let's land yours first.

Blocked by https://github.com/mozilla/foundation.mozilla.org/pull/2232

mmmavis:
@Pomax Thanks! 

lol I just realized that PR accidentally addressed 70% what we are trying to fix here (I made some DOM elements full width so that might be why). We need to fix the height of the filter bubble on mobile and probably other tweaks. https://foundation-mofostaging-pr-2232.herokuapp.com/en/privacynotincluded/

Pomax:
Yeah, I'm more than happy to just see what needs doing after yours lands.

Issue #2236, opened 2018-11-19
Add mini seal to Related Products


kristinashu:
Similar to https://github.com/mozilla/foundation.mozilla.org/issues/2234

- [x] Add mini seal to categories - same size as homepage (30px)
- [x] Add gray padding around image

<img width="196" alt="image" src="https://user-images.githubusercontent.com/1068783/48739543-2be3a480-ec09-11e8-88dc-1c437fbba7c2.png">

<img width="925" alt="image" src="https://user-images.githubusercontent.com/1068783/48739470-efb04400-ec08-11e8-8dcc-8c137dbbb08d.png">

PR #2238, opened 2018-11-20
Related mini seals

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2236 by adding the "seal of approval" to the related product images at the bottom of a product page.


kristinashu:
This should be the mini seal like on the homepage, not the full seal. 

Screen cap from pr
image

kristinashu:
Sorry, I should have caught this on the first round but I'm worried the product images will be really small. Could you make the padding less at narrower widths? Because most products have a gray background in their top left corner, it's ok if the seal overlaps the image a little.

image

Pomax:
Does this look better?

image

kristinashu:
Yes, looks better!

Issue #2213, opened 2018-11-16
Add 'Minimum Security Standards' seal to filter


beccaklam:
- [ ] Add Minimum Security Standards section to filter

- [ ] Default is that the box is unchecked and people can see products that do and do not meet the standards
- [ ] People can use the filter to only see products that meet the standards (box is checked)
### **Default**
![home - filters open 2x](https://user-images.githubusercontent.com/20977376/48650462-035e6f00-e9ab-11e8-9cee-a6b8693ce8d9.png)
### **Minimum Security Standards Seal Checked**
![home - filters open copy 2x](https://user-images.githubusercontent.com/20977376/48650536-75cf4f00-e9ab-11e8-9d06-df3e91b262ee.png)

PR #2239, opened 2018-11-20
add seal of approval to filter

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2213 by adding the seal of approval as a filter option. 

- When disabled, it doesn't do anything
- When enabled, it filters out all non-sealed-of-approval'd

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,4 +1,4 @@
-# EditorConfig : http://EditorConfig.org
+# EditorConfig : https://EditorConfig.org
 root = true
 
 [*]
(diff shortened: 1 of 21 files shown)


beccaklam:
Hi @Pomax the filter is looking good and working well! Two things:

- [ ] The spacing is a _little_ off, it should be 24px under 'Filter' and 48px under 'Minimum Security Standards' section and 'Creepiness' section
filters open 2x

- [ ] We're hoping to do a little check box like this for the 'Minimum Security Standards', would that be possible?
screen shot 2018-11-20 at 1 12 39 pm

Pomax:
Having the actual checkmark is possible, but definitely more work - right now the checked state is just a simple bit of CSS that we can reuse from the "radio buttons" below by removing the border radius, putting in an actual checkmark will increase the complexity a little. I'll see by how much =)

Pomax:
dimensions updated, and the checkbox looks like it has a checkmark now (using a checkmark svg overlay)

PR #2244, opened 2018-11-20
Test for running python test on a windows env on Travis

patjouk:


patjouk:
That's way too slow and buggy. Let's wait for the official release from Travis.

PR #2259, opened 2018-11-22
Adjusted project list layout on XL screen

### LINKED ISSUE #2249: Homepage on wide screens
Follow up from https://github.com/mozilla/foundation.mozilla.org/issues/2124

Medium and narrow widths looks awesome and super responsive but the full-width is a little too busy looking at wide screen widths.

@mmmavis please update the widest break point to:

- [ ] 4 products across
- [ ] products should be contained within the grid (see screen cap below)
- [ ] make white cutout box and copy wider too

Comps https://redpen.io/nv740c162fdcf1606c

home copy 2 2x

Grid width
screen shot 2018-11-21 at 11 30 38 am

mmmavis:
Closes #2249

### DIFF
--- a/source/sass/buyers-guide/views/homepage.scss
+++ b/source/sass/buyers-guide/views/homepage.scss
@@ -135,9 +135,14 @@
     @media (min-width: $bp-lg) {
       @include product-box-size(1px, 4);
     }
+  }
 
+  .project-list-section {
     @media (min-width: $bp-xl) {
-      @include product-box-size(1px, 5);
+      // on extra large screen we want project-list-section to be the same width
+      width: calc(#{map-get($container-max-widths, xl)} - #{map-get($grid-gutter-widths, xl)});
+      max-width: 100%;
+      margin: 0 auto;
     }
   }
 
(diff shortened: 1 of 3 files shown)


Pomax:
Looks pretty good! Do we still want some "breathing room" between the products and the footer though?

image

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -135,9 +135,15 @@
...
+  }
 
+  .project-list-section {
     @media (min-width: $bp-xl) {
-      @include product-box-size(1px, 5);
+      // on extra large screen we want project-list-section's width to match
+      // Bootstrap's .container's width
+      width: calc(#{map-get($container-max-widths, xl)} - #{map-get($grid-gutter-widths, xl)});
+      max-width: 100%;
+      margin: 0 auto;
```
Pomax: If there are margin requirements we should probably use an `m-...` rule in the HTML

kristinashu:
Nah I think it's ok without the gap before the footer.

source/sass/buyers-guide/views/homepage.scss
mmmavis: Yea I thought about it.  But since we are adding styling rules only for XL screens, I think leaving them all in homepage.scss makes it easier to glance vs adding `mx-xl-auto` class in DOM that deals with horizontal margin + the rest of the rules in .scss

Issue #2260, opened 2018-11-22
Donate popup follow up


kristinashu:
- [x] bottom is cut off on iPhone (see screen cap)

- [x] don't show pop up on campaign pages

- [x] add button hover state like on pni site

<img width="376" alt="screen shot 2018-11-22 at 1 34 47 pm" src="https://user-images.githubusercontent.com/1068783/48922434-7fe5c780-ee5b-11e8-8019-228f83a2788a.png">

PR #2262, opened 2018-11-22
Update the donate modal

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2260

- bottom should no longer be cut off on iPhone
- should no longer show pop up on campaign pages
- now has button hover state like on pni site

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -1,9 +1,11 @@
 {% extends "./mini_site_name_space.html" %}
 {% load wagtailcore_tags mini_site_tags %}
 
-Note: we don't want the donation modal to pop up on
-pages with petitions, for specifically for campaign
-pages, we blank out the donation block:
+{% comment %}
+  Note: we don't want the donation modal to pop up on
+  pages with petitions, for specifically for campaign
+  pages, we blank out the donation block.
+{% endcomment %}
 
 {% block donate %}{% endblock %}
 
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
@@ -1,6 +1,13 @@
 {% extends "./mini_site_name_space.html" %}
 {% load wagtailcore_tags mini_site_tags %}
 
+Note: we don't want the donation modal to pop up on
+pages with petitions, for specifically for campaign
+pages, we blank out the donation block:
+
+{% block donate %}{% endblock %}
+
+
 {% block subcontent %}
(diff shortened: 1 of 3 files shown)


kristinashu:
There aren't any campaign pages in this PR so I can't test that, so guess we'll see on staging.

The button isn't cut off but it still seems very close to the bottom. Not a blocker but wonder if you can make the `Chip in to help us keep the web healthy, wonderful, and welcoming to all.` the same font styling as the `Let's protect the world's...`. This way the text will be small and take up less room. This would be for the narrow width only. Do you think that would help?

screen shot 2018-11-22 at 2 38 27 pm

Pomax:
I'd recommend tweaking the placement CSS next week, looking at each of the widths that bootstrap supports and seeing which special affordances are necessary.

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
```diff
@@ -1,6 +1,12 @@
 {% extends "./mini_site_name_space.html" %}
 {% load wagtailcore_tags mini_site_tags %}
 
+Note: we don't want the donation modal to pop up on
```
mmmavis: should this be in a comment block?

source/js/donate-modal/donate-modal.scss
```diff
@@ -81,6 +85,12 @@
     color: white;
     padding: 15px 18px;
     border-radius: calc((20px + 18px * 2) / 2); // half of button's height
+
+    &:focus,
```
mmmavis: we can do

```scss
@include hover-focus-active {
  background: #f00;
}
```

source/js/donate-modal/donate-modal.scss
```diff
@@ -97,5 +107,13 @@
     font-weight: 400;
 
     @include scaleText(22px, 28px, 22px, 28px);
+
+    @media (max-width: $bp-sm) {
+      font-family: "Nunito Sans", Helvetica, Arial, sans-serif;
+      font-size: 16px;
+      line-height: 1.5;
```
mmmavis: we don't need
```css
font-size: 16px;
line-height: 1.5;
```
having the following line should be sufficient
```css
 @include scaleText(16px, 24px);
```

network-api/networkapi/wagtailpages/templates/wagtailpages/campaign_page.html
Pomax: As an `extends`, nothing outside of `{%block%}` codes is active code, so it doesn't have to be marked as comment. That said, trivial to add it in so why not.

source/js/donate-modal/donate-modal.scss
Pomax: oh nice!

Issue #2234, opened 2018-11-19
Add mini seal to categories 


kristinashu:
Similar to https://github.com/mozilla/foundation.mozilla.org/issues/2236

- [x] Add mini seal to categories - 40px wide
- [x] Update company name from .h6-heading-uppercase to .h6-heading
- [x] Add gray padding around image 

<img width="394" alt="image" src="https://user-images.githubusercontent.com/1068783/48739093-bb885380-ec07-11e8-9805-d04c8ca6d452.png">

Comp https://redpen.io/mb5979d9d3273ab852

![category copy 2 2x](https://user-images.githubusercontent.com/1068783/48738708-af4fc680-ec06-11e8-8801-315b60dd7437.jpg)

Pomax:
Would these effectively be the same panels we use on the homepage? (if so, we might be able to turn them into a template fragment and have the homepage and category views both use the same thing, just with an argument to show/hide the title)

kristinashu:
Not too sure what you mean my panels. It is the same content as home page (image, mini seal, company name title, product name title, and gray padding). Differences are that thumbs, the product title, and the white space between products on the Category pages is larger. Another small difference is that titles on the homepage will be hidden at narrow mobile widths, where as they must always be shown on the Category pages.

Pomax:
That's effectively what I was looking for as answer - I'll investigate if we can do some unification there, so we can take advantage of the code we already wrote for showing seals-of-approval without repeating code/scss!

Pomax:
Hm, pulling things apart turned out not to be a real option, so adding things specifically to the category page.

PR #2280, opened 2018-11-27
add seal of approval to category product boxes

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2234


kristinashu:
Few styling things:

### Images shifted over
- [ ] Looks like the images are shifted over to the right slightly, they should align with the text.

Screen cap from PR
image

### More responsive
- [ ] Make seal smaller and product image bigger at narrow widths

Screen cap from PR
screen shot 2018-11-27 at 12 45 27 pm

### Add more padding to images at wide screen widths
- Hard to tell with placeholder images but this might need more padding, is it ok to wait and see once it's on staging?

kristinashu:
Here it is will some real images. Noticing the seal for products on the right side is overhanging past the image where as products on the left side the seal is within the gray (but both need more padding).

image

Pomax:
@kristina updated as per checkboxes =)

kristinashu:
Nice! Mobile is looking good.

Sorry, but I think desktop could use even more grey padding around the images. Maybe even double what you have now! But for width widths only (narrow is looking good).

- [ ] Add even more padding

- [ ] _Slightly_ reduce the space between the image, Company Name, and Product Name (see comp below)

- [ ] Category title should be .h3-heading instead of .h1-heading, it's very large right now (this is not a new things so feel free to do as follow up ticket if you prefer)

PR
screen shot 2018-11-28 at 12 56 46 pm

Comp
screen-shot-2018-11-28-at-12 56 36-pm

Pomax:
@kristina should... look right now? =)

kristinashu:
Soooo close! I promise this is the last thing. Can you bring the Company Name and Product Name a little closer together? Not sure if this is the best way but I updated margin bottom from 0.5 to 0.2rem and it looks good (again, not sure that is the best way to do it but just so you can see how close I'm talking).

image

Pomax:
the spacing comes from bootstrap's `mb-...` class, but I think your change is literally the difference between `mb-2` and `mb-1` so let me update that =D

source/sass/buyers-guide/views/category.scss
```diff
@@ -7,7 +7,7 @@
     margin: 0 auto;
     max-width: 545px;
 
-    @media screen and (min-width: 750px) {
+    @media screen and (min-width: $bp-md) {
```
mmmavis: Not directly related to this code change but we actually don't need the class `page-width`. We can just move all the CSS rules `.category-listing`.

source/sass/buyers-guide/views/category.scss
```diff
@@ -48,28 +42,53 @@
...
         display: block;
         width: 100%;
         height: auto;
+
+        @media (max-width: $bp-md) {
+          transform: scale(1.25);
+        }
       }
 
       .category-item-label {
```
mmmavis: Ummm... I couldn't seem to find any element with class name `category-item-label` on the category page. Is this old code to be deleted?

image

source/sass/buyers-guide/views/category.scss
```diff
@@ -78,7 +97,7 @@
         font-weight: 700;
         margin-bottom: 15px;
 
-        @media screen and (min-width: 750px) {
+        @media screen and (min-width: $bp-md) {
```
mmmavis: Same, category page doesn't seem to have any `.category-item-header` elements. Is this entire `.category-item-header { }` block old code to be deleted?

source/sass/buyers-guide/views/category.scss
Pomax: hm, it might very well be. We should probably figure out if there's something we can do around dead rule detection etc.

source/sass/buyers-guide/views/category.scss
Pomax: That... is a good point.

source/sass/buyers-guide/views/category.scss
Pomax: This was a hold over from when we weren't using `h*-heading` instead, so I'll remove it + the header class which is also unused now

source/sass/buyers-guide/views/category.scss
Pomax: it is, removing.

Issue #2231, opened 2018-11-19
Add 'creepy face' beside 'What could happen if something goes wrong' header


beccaklam:
- [ ] Add the creepy face beside the 'What could happen if something goes wrong' header to give it more emphasis. SVG is [here](https://drive.google.com/open?id=1HzVsb2fdlRPVfVnvmrQoU3hYmmcu-Tzh).
![what-could-go-wrong 2x](https://user-images.githubusercontent.com/20977376/48731500-08612f80-ebf2-11e8-810e-a088b4e7f966.jpg)

Pomax:
This feels like this forces the emoji into the page a little - do we need this emphasis? If so, would it make sense to only show it if the product is creepy enough (e.g. not have our seal of approval), so as not to introduce bias into the product's perception?

beccaklam:
Hey @Pomax there was a request from @caltrider to add emphasis to the 'What could happen if something goes wrong' section. There's a concern it's getting passed over and it contains important information. We would prefer to keep the face consistent across the pages so people know what to look for. The choice to use the emoji is to, again, stay consistent with highlighting creepiness with the emoji face. cc: @kristinashu

Pomax:
Alright, I'll add the emoji in, but I'd like to also pitch the idea that right now it just looks like detached text  due to there only being a header and a single sentence, and adding an emoji won't affect that too much. It might be worth turning that into a kind of "alert box" (similar to how textbooks deal with conclusions or break-outs) and style it so it can't be ignored but instead draws eyes to itself.

caltrider:
I like that idea @Pomax .

Pomax:
silly mock-up, but something that pops out like this?

![image](https://user-images.githubusercontent.com/177243/49108111-2eaf4c80-f23c-11e8-959e-14fd964abdbb.png)

kristinashu:
I don't think it needs to stand out **that** much!

Pomax:
I wasn't sure about the bold, but I kind of liked having the emoji on both sides =P

Pomax:
Is it worth waiting for a new design before implementing this, or do we want the intermediary "just the emoji" first and file a followup for more emphasis?

kristinashu:
I think _just the emoji_ and then see if it still comes up as an issue in user testing.

PR #2302, opened 2018-11-28
add creepy emoji to 'what could go wrong' + refactor for that

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2231


beccaklam:
@Pomax Thanks for implementing this! Two small nit-picky things but would it be possible to:

- [ ] Get rid the small bit of 0.2em margin we have on the left side of the creepy emoji (It should left-align perfectly with the rest of the headers) and 

- [ ] increase the right side margin to 0.25em? 

Thanks!

source/sass/buyers-guide/views/product.scss
```diff
@@ -84,6 +84,14 @@
   h4 {
     font-family: "Nunito Sans";
   }
+
+  @import "../creep-face-emoji";
```
mmmavis: image


We need to move this `@import` line to top of the page. We shouldn't include `@import` in a declaration block as what we wanna do here is to inherit and override the existing CSS rules from `creep-face-emoji.scss` and not declare the same rules for `.product-detail .creep-face-emoji`.

source/sass/buyers-guide/views/product.scss
```diff
@@ -84,6 +84,14 @@
...
+
+  .creep-face-emoji {
+    top: 3px;
+    margin: 0;
+    margin-right: 0.25em;
```
mmmavis: we can just do `margin: 0 0.25em 0;` here

source/sass/buyers-guide/views/product.scss
mmmavis: image


We need to move this `@import` line to top of the page. We shouldn't include `@import` in a declaration block as what we wanna do here is to inherit and override the existing CSS rules from `creep-face-emoji.scss` and not declare the same rules for `.product-detail .creep-face-emoji` -- we should keep specificity low.

source/sass/buyers-guide/views/product.scss
mmmavis: the 2 margin rules can be merged into `margin: 0 0.25em 0;`

source/sass/buyers-guide/views/product.scss
Pomax: As I read the Sass documentation, this is the way you should do an import if you need a specific class imported, so as far as I understood, this would have added the `.creep-face-emoji` class at this point in the product-detail stylesheet tree. From your screenshot, that is exactly what it does, so I'm not sure I understand: I _intended_ for this specificity, so that the base class definition, and its overrides, happen in exactly the same place.

source/sass/buyers-guide/views/product.scss
Pomax: they can, but it's always seemed weird to me to list "0" three times when there's a way to set all of them at once and then only add your exception. Do we have a preferred format?

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -24,20 +24,7 @@
...
-        top: 0.25em;
-        margin: 0 0.2em;
-        font-variant: normal;
-        color: transparent;
-        background-image: url(/_images/buyers-guide/faces/most-emoji.png);
-        background-size: 100% 100%;
-        background-position: center center;
-        background-repeat: no-repeat;
-      }
+      @import "../creep-face-emoji";
```
mmmavis: screen shot 2018-12-19 at 10 45 48 am

maybe as a followup? can we make sure the selector is not so long and specific?

source/sass/buyers-guide/views/product.scss
mmmavis: Yes. We shouldn't be overriding rules within the same declaration block. So in this case we should only have one rule that deals with `margin`

Pomax:
oops, good catch, forgot to remove the import from homepage.

Issue #2326, opened 2018-12-07
Python 3.6.7 breaks the build on Windows


Pomax:
clean install log:
```
c:\Users\Mike\Documents\git\collaborations\mozilla>git clone git@github.com:mozilla/foundation.mozilla.org.git
Cloning into 'foundation.mozilla.org'...
remote: Enumerating objects: 184, done.
remote: Counting objects: 100% (184/184), done.
remote: Compressing objects: 100% (87/87), done.
remote: Total 10655 (delta 153), reused 97 (delta 97), pack-reused 10471
Receiving objects: 100% (10655/10655), 23.07 MiB | 8.51 MiB/s, done.
Resolving deltas: 100% (6893/6893), done.

c:\Users\Mike\Documents\git\collaborations\mozilla>cd foundation.mozilla.org

c:\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org>inv setup
Copying default environment variables.
Installing npm dependencies and build.

> node-sass@4.10.0 install c:\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org\node_modules\node-sass
> node scripts/install.js

Cached binary found at C:\Users\Mike\AppData\Roaming\npm-cache\node-sass\4.10.0\win32-x64-57_binding.node

> node-sass@4.10.0 postinstall c:\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org\node_modules\node-sass
> node scripts/build.js

Binary found at J:\Junctions\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org\node_modules\node-sass\vendor\win32-x64-57\binding.node
Testing binary
Binary is fine

> optipng-bin@5.0.0 postinstall c:\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org\node_modules\optipng-bin
> node lib/install.js

  âˆš optipng pre-build test passed successfully
npm WARN optional SKIPPING OPTIONAL DEPENDENCY: fsevents@1.2.4 (node_modules\fsevents):
npm WARN notsup SKIPPING OPTIONAL DEPENDENCY: Unsupported platform for fsevents@1.2.4: wanted {"os":"darwin","arch":"any"} (current: {"os":"win32","arch":"x64"})

added 1281 packages in 49.087s

> network@1.0.0 build c:\Users\Mike\Documents\Git\collaborations\mozilla\foundation.mozilla.org
[...]

PR #2327, opened 2018-12-07
A small fix that unbreaks windows

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2326, or at least seems to, but I'm not entirely happy about the fact that I don't know _why_ it fixes it.

A lot of stackoverflow answers suggest this would be the culprit, though, and `inv setup` runs to completion with it applied, rather than throwing the error listed in the abovementioned issue...

This was on Windows 10 Pro x64, with Python 3.6.7, pipenv v2018.11.26, pip 18.1

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -409,7 +409,7 @@ if USE_S3:
 
 else:
     # Otherwise use the default filesystem storage
-    MEDIA_ROOT = root('media/')
+    MEDIA_ROOT = root('media')
     MEDIA_URL = '/media/'
 
 # CORS


Pomax:
Turns out that if I run Python 3.6.6 with master, things work just fine, and this error only manifests with Python 3.6.7 so I don't know if we want to land this, or whether we need to somehow figure out an MCVE and file that over on Python's issue tracker?

Pomax:
This is python... 3.6.7 received https://bugs.python.org/issue31047, which changes how file paths are handled on Windows, which is pretty backward compat breaking.

Issue #2342, opened 2018-12-13
Add some code that lets us curate our domain whitelist from inside Wagtail


Pomax:
We're using an `env` at the moment to curate the whitelist of allowed embeddable domains and it's kind of a pain? We should totes fix that to become CMS-list-enrichable, where we can do something like "go to /cms, open settings, open whitelist, add domain, good to go".

And then have that kick in _in addition to_ what we preclear in the env, so that we can ensure there are a few domains that are always allowe, with the wagtail page showing those as non-editable entries or something

/cc @alanmoo @cadecairos

alanmoo:
Sure, this seems like it would be nice but I don't think it's particularly worth spending time on. More like a thing that might be nice to add upstream to Wagtail itself

cadecairos:
Yeah I'd love to be able to update the CSP via the admin interface (this is how we do it on the internet health report).

That said, I'm not really willing at the moment to maintain that project. might be nice to have it as an upstream feature.

Pomax:
assigning myself so I remember to file a feature request on the wagtail repo.

kristinashu:
@Pomax did you file it with wagtail? If so then it sounds like we won't be taking this any further.

gideonthomas:
Closing as this should be done upstream in wagtail. @Pomax just a reminder in case you haven't filed it yet.

Pomax:
That's the wrong order in which to do things. File first, then close.

PR #2345, opened 2018-12-13
Airtable block MVP

### LINKED ISSUE #2344: Create airtable streamfield component
To enable embedding of Airtable widgets (like a project brief form), I'd like to add a streamfield component. I've got it partially built, but am mainly looking for a bit of feedback- The way I've built it to start is a bit less-than-obvious, in which the user needs to copy the "embed" url of a view out of the Airtable-suggested script tag, along with the height. For example

```
<iframe class="airtable-embed" src="https://airtable.com/embed/shrSXc1ahbBaG1hYF?backgroundColor=gray" frameborder="0" onmousewheel="" width="100%" height="1561" style="background: transparent; border: 1px solid #ccc;"></iframe>
```

The user would paste `https://airtable.com/embed/shrSXc1ahbBaG1hYF` and `1561` into fields in the component. While it's reasonable for an MVP, it's hard to change later on. But I also don't want to go too far down a rabbit hole on this.

Anyone have specific thoughts on this going? @gideonthomas? @kristinashu?

alanmoo:
Closes #2344 

None below are applicable?

[ ] Migrations squashed


network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -41,6 +41,15 @@ class Meta:
         template = 'wagtailpages/blocks/image_block.html'
 
 
+class AirTableBlock(blocks.StructBlock):
+    url = blocks.URLBlock()
+    height = blocks.IntegerBlock()
```
gideonthomas: is there a default we can maybe set?

alanmoo: Ah, good call; I meant to do that and add a helper description

alanmoo:
@gideonthomas Can you take another look on Monday? I squashed migrations and added help text and want a second set of eyes on it.

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -41,6 +41,20 @@ class Meta:
         template = 'wagtailpages/blocks/image_block.html'
 
 
+class AirTableBlock(blocks.StructBlock):
+    url = blocks.URLBlock(
+        help_text="Copied from the Airtable embed code. The word 'embed' will be in the url"
+    )
+    height = blocks.IntegerBlock(
+        default=533,
+        help_text="The height of the view on a desktop, usually coped from the Airtable embed code",
```
gideonthomas: typo - copied*

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -41,6 +41,20 @@ class Meta:
         template = 'wagtailpages/blocks/image_block.html'
 
 
+class AirTableBlock(blocks.StructBlock):
```
Pomax: Should this be `class AirTableBlock(iFrameBlock):`?

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/airtable_block.html
```diff
@@ -0,0 +1 @@
+<script src="https://static.airtable.com/js/embed/embed_snippet_v1.js"></script><iframe class="airtable-embed airtable-dynamic-height" src="{{ value.url }}" frameborder="0" onmousewheel="" width="100%" height="{{ value.height }}" style="background: transparent; border: 1px solid #ccc;"></iframe>
```
Pomax: should this extend `wagtailpages/blocks/iframe_block.html`?

network-api/networkapi/wagtailpages/customblocks.py
alanmoo: Possibly- what's the difference?

Pomax: Exactly: we already have an iframe block for embedding an iframe with some arbitrary source into a page, so if this injects an iframe then most of this class code is just duplication and we don't need to write it.

gideonthomas: one thing I just remember, we'll need to add airtable to our CSP config (however we're doing that) if we're loading it in an iframe

Pomax: yeah, that's part of what the iframe block makes clear: its helptext informs users that whatever domain they add, it _must_ get added to the domain whitelist.

There's an issue open about seeing if we can make start making that a CMS-managable (or really, enrichable) list over on https://github.com/mozilla/foundation.mozilla.org/issues/2342

alanmoo: After looking at the iFrameBlock model and template, I'd strongly argue that this is different enough that it shouldn't inherit. The use case for this is more of a single item on a page. It also embeds a script on the page specifically for AirTable iframes, with some custom attributes that Airtable embeds give us. There's nothing of the existing iFrameBlock that I want to reuse, including the help text. If a user is selecting an element called "Airtable", they shouldn't be reminded that it needs to be approved; that's on us to do the one time we land this code. (With that in mind, I've added a commit to add Airtable to the CSP defaults)

Pomax: It feels like it might be worth inventing some kind of iframe factory since we now have iframe, youtube, and airtable, and they all "Sort of do the same, with different parameters", but that can be a story for a different year.

alanmoo: I can see why you'd say that. But given the troubles we've had in the past, would the extra layer of abstraction actually yield any benefits or would it just get in our way when it comes to enhancing/changing the models in the future?

Issue #1767, opened 2018-08-23
Deprecate old Participate page


gvn:
Remove the old Participate page template, model, translation bindings, etc from codebase.

mmmavis:
Dupe of #1906 . Closing.

PR #2361, opened 2018-12-18
(WIP) Deprecated old Participate page and factory

### LINKED ISSUE #1906: Update Participate factory
The ParticipatePage factory is still generating the original page. It needs to generate a ParticipatePage2 instead.

While we're at it, we should entirely remove the original Participate model.

mmmavis:
# 🚧 WIP

Closes #1906, #1767 

@Pomax 
I don't know what I should do to make sure pages that are using `ParticipatePage2` don't break after this PR gets landed. Can you advice?


Pomax:
I would recommend doing this in two stages:

1. create a database, and make a copy so you can restore it lots of times! =D
2. set up an "empty" migration with sole responsibility to take old Participate pages, and create new Participate2 pages from them. Example of an empty migration "that just does a thing" without touching model definitions https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/migrations/0009_auto_20181018_2047.py, example of making things based on things: https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/migrations/0006_add_product_categories.py
3. set up another "empty" migration that just runs code that deletes all the old participate objects.
4. update the model.py file, removing old participate, renaming participate2 to participate, etc. which is now safe because there are no old participate objects anymore.

So you'd end up with three migrations for getting this work done.

Pomax:
Is this still a WIP?

mmmavis:
Thanks @Pomax ! I will start a new PR for this.

PR #2372, opened 2018-12-20
Make footer appear over nav when vh is small

### LINKED ISSUE #2359: Donate modal can't be closed on iPhone landscape mode
Probably other devices too- the combination of the fixed nav and fixed modal conflict. 
image

@kristinashu Possible solution: How would you feel if the nav wasn't sticky on mobile?

alanmoo:
Like landscape on a phone. Closes #2359

To verify: Open responsive design mode in Safari and view as an iPhone 8 and SE in landscape mode.

### DIFF
--- a/source/js/donate-modal/donate-modal.scss
+++ b/source/js/donate-modal/donate-modal.scss
@@ -1,5 +1,5 @@
 .donate-modal {
-  $modal-z-index: 10;
+  $modal-z-index: 1040; //War of escalation with .sticky-top
 
   position: fixed;
   z-index: $modal-z-index;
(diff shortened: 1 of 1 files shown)


source/js/donate-modal/donate-modal.scss
```diff
@@ -1,5 +1,5 @@
 .donate-modal {
-  $modal-z-index: 10;
+  $modal-z-index: 1040; //War of escalation with .sticky-top
```
Pomax: Yeah let's not do this. We're in sass, let's use a calc like `calc(#{$sticky-index} + 1)` =)

alanmoo: Ok fair enough. Turns out Bootstrap has a thing built in for this already. See updated PR.

PR #2400, opened 2019-01-03
Logging Improvements

cadecairos:
As we get moved over to Scalyr for our application logging and alerting based on logs we need to make sure that they're as useful to us as possible, and are not dumping data to the log stream that we don't want to record (emails, usernames, etc)

~~This first commit makes some logging configuration changes and adds the Heroku request ID to Django request logs.~~

This PR removes the default logging configuration provided by Django and redefines it with the following attributes:
    1. Only uncaught exceptions are logged on production (DEBUG=False - Heroku router logs have the information we need.)
    2. In development, requests logging in enabled (DEBUG=True)


network-api/networkapi/settings.py
```diff
@@ -462,30 +462,32 @@
 # LOGGING
 LOGGING = {
     'version': 1,
-    'disable_existing_loggers': False,
+    'disable_existing_loggers': True,
```
cadecairos: This ensures we only use these loggers, so that production logging can be shut down (and handled by the Heroku router)

cadecairos:
So, this issue evolved a bit as I worked things out, and no longer has anything to do with Heroku request IDs, but essentially, I realized we didn't need any request logging on production because the Heroku router handles that for us. The new configuration ensures that we won't be accidentally logging request data to console in production environments, and should help keep our logs clean.

network-api/networkapi/settings.py
```diff
@@ -462,30 +462,42 @@
...
+        'require_debug_true': {
+            '()': 'django.utils.log.RequireDebugTrue',
+        }
+    },
     'formatters': {
-        'simple': {
-            'format': '[%(asctime)s] [%(levelname)s] %(message)s'
-        },
+        'verbose': {
+            'format': '%(asctime)s [%(levelname)s] %(message)s',
```
cadecairos: lol, so much more verbose than before.

network-api/networkapi/settings.py
```diff
@@ -462,30 +462,42 @@
...
+            'datefmt': '%Y-%m-%d %H:%M:%S'
```
Pomax: do we want ms just in case we need to debug timing issues?

network-api/networkapi/settings.py
```diff
@@ -462,30 +462,42 @@
...
+            'level': 'DEBUG',
             'class': 'logging.StreamHandler',
-            'formatter': 'simple',
+            'filters': ['require_debug_true'],
+            'formatter': 'verbose'
         },
+        'error': {
+            'level': 'ERROR',
+            'class': 'logging.StreamHandler',
+            'filters': ['require_debug_true']
```
Pomax: Do we really only wan to see errors when debug is on? This feels like a log level that should kick in regardless of debug value.

network-api/networkapi/settings.py
patjouk: I was checking Heroku recommended configuration for django project, and they keep the flag at [False](https://github.com/heroku/django-heroku/blob/master/django_heroku/core.py#L119). Are you sure we won't be missing any logs by silently silencing them?

network-api/networkapi/settings.py
patjouk: Same as Pomax: I would prefer to see errors messages too on prod.

network-api/networkapi/settings.py
cadecairos: These logs will only be present in dev. in Production we'll see logs like this: 
```
2019-01-16T19:14:45.762692+00:00 heroku[router]: at=info method=GET path="/_images/buyers-guide/gradient-bar.svg" host=foundation-mofostaging-pr-2400.herokuapp.com request_id=aebc9f38-b22c-47da-bdad-f4a7e76ef813 fwd="1.1.1.1" dyno=web.1 connect=0ms service=8ms status=200 bytes=1294 protocol=https
```

cadecairos: ...but yeah lets get more detailed timing info

network-api/networkapi/settings.py
```diff
@@ -458,36 +459,64 @@
...
+        'django.db.backends': {
+            'handlers': ['debug-error'],
+            'level': 'ERROR'
+        },
+    }
 }
+DJANGO_LOG_LEVEL = env('DJANGO_LOG_LEVEL')
+logging.config.dictConfig(LOGGING)
+
+# Override all the default Django loggers
```
patjouk: You might want to move that comment a few lines above

cadecairos:
@Pomax which changes are you referring to?

cadecairos:
@Pomax can you assist with Percy approval?

PR #2414, opened 2019-01-07
Python dependencies update

patjouk:
Closes #2410 and closes #2383


patjouk:
Both dependabot PRs worked but had merge conflict in the Pipefile lock
so...

Issue #2428, opened 2019-01-08
Add wagtail-experiments to our setup for A/B testing


Pomax:
Now that https://github.com/torchbox/wagtail-experiments is effectively out of beta, and is compatible with our setup, it's probably a good time to add it in.

It introduces a new settings option called (surprise!) "experiments", which lets you A/B(/C/D/E/...) test page conversion in the sense that you can set up a set of pages, pegged to a "control page", and users hitting the control page's URL will be served either the control page, or any of the alternate pages you specified. Metrics are gathered for the purpose of determining conversion rates, either based on a target CMS-managed URL that acts as "goal", or custom JS that tells wagtail-experiments there has been a completion (e.g. signalling completion if a user clicks "donate", as part of the click handler).

The reporting interface also comes with a way to not just look at the conversions, but also peg which of the various test cases should replace the original page handling, if it turns out the control page does not perform as well as one of the alternative test cases.

It's all pretty sleek!

/cc @alanmoo

Pomax:
How much detail do we want to add to our own README.md when it comes to explaining A/B testing?

alanmoo:
At least a heading and a link to Wagtail Experiments documentation.

PR #2430, opened 2019-01-08
add experiments to our setup

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2428 by adding https://github.com/torchbox/wagtail-experiments to our setup.


alanmoo:
(I was standing next to Chris while he was looking at this, so other than his suggestion, this looks great to me as long as he approves of the code too.)

Pomax:
I was thinking of adding that in follow up, once we actually have an experiment that requires the JS end point, but it's so little work you're right, we should just add that now.

Where do we want to document the fact that we'll now have this functionality?

cadecairos:
Let's document that we use it for A/B testing in the readme and link to the documentation.

Another thought, can we limit access to experiments to specific roles in django? I spoke with @alanmoo and he suggested we might want to limit it to the producer role until we're ready to train people on using it.

Pomax:
As far as I understand it only superusers and "explicitly experiments-granted" users get access to the experiments interface.

Pomax:
image

Issue #2434, opened 2019-01-09
Enable "draft" Buyer's Guide products


alanmoo:
As the team prepares to launch updates to the guide, we need a way to add products that aren't live yet.

I'm not tied to a specific solution, but some ideas that might be useful

* Boolean "is live" or "publish date" could be used
* It's probably useful for unpublished products to be visible to logged in users for previews.

alanmoo:
@caltrider Any other thoughts?

caltrider:
I'm most familiar with something like Medium does with a "publish" feature. So that you could input all the info and have it available to review/double check/preview, and then publish when we're ready to roll out the launch. That being said, since we won't be promoting the guide before the V-Day launch, I don't imagine we'll have too much traffic to worry about if we add things and make updates to get things right before a launch promotion.

caltrider:
I do love deliciously obvious!

Pomax:
@kristinashu right now I've added some dummy CSS to effect "making it obvious that some products are draft" but I don't actually know how much attention should be drawn to them, or what the "least intrusive, super obvious" way would be. Would you have any advice? =)

kristinashu:
Looks good! I think your red text is enough, doesn't even need the extra red border.

Pomax:
excellent, I shall take out the border CSS

PR #2439, opened 2019-01-09
added a draft property to products, defaulting to True

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2434 by giving products a `draft` boolean, which hides the product from any page viewed by an unauthenticated user. Those users will also see a 404 if they directly navigate to a draft product's URL

This updates the tests as well, since products are now draft-by-default which broke a whole bunch of tests that were based on products being live simply by being `create()`'d


source/sass/buyers-guide/views/category.scss
```diff
@@ -39,6 +39,24 @@
         padding-right: 3%;
       }
 
+      &.draft {
```
Pomax: This is duplicated in a few places mostly because there's no design yet and the final CSS will likely be different from what this does.

cadecairos:
Will this PR automatically set existing products to `draft=False`?

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -49,6 +49,24 @@
         padding: $padding-y $padding-x;
         width: calc((100% - #{$row-total-gutter-width}) / #{$items-per-row});
 
+        &.draft {
+          border: 1px solid #ff000054;
```
mmmavis: this HEX code looks too long for me?

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -49,6 +49,24 @@
...
+
+          &::before {
+            content: 'This product is in draft';
+            color: #ff000054;
```
mmmavis: this HEX code looks too long for me?

source/sass/buyers-guide/views/product.scss
```diff
@@ -2,6 +2,20 @@
   padding-bottom: 88px;
   position: relative;
 
+  &.draft::before {
+    content: 'This product is in draft and only visible to CMS-authorised users.';
+    color: #ff000054;
```
mmmavis: this HEX code looks too long for me?

source/sass/buyers-guide/views/category.scss
```diff
@@ -39,6 +39,24 @@
...
+        border: 1px solid #ff000054;
```
mmmavis: this HEX code `#ff000054` looks too long for me?

source/sass/buyers-guide/views/category.scss
```diff
@@ -39,6 +39,24 @@
...
+
+        &::before {
+          content: 'This product is in draft';
+          color: #ff000054;
```
mmmavis: wrong HEX code?

source/sass/buyers-guide/views/homepage.scss
Pomax: last two digits are the transparency.

mmmavis: ah that's something new to me! 
ignore all my other comment then :D

cadecairos:
@Pomax is it trivial to add a draft column to the products view of the wagtail admin page for PNI?

Pomax:
Ooh, that's a good question. The products are plain Django so I hope so, let me see if I can add that.

Pomax:
@cadecairos added

@mmmavis I've updated the CSS based on Kristina's comments and put all the rules in their own sass partial

source/sass/buyers-guide/fragments/_draft_product.scss
```diff
@@ -0,0 +1,14 @@
+// product draft rules
+&.draft::before {
+  content: 'This is a draft.';
```
mmmavis: Although not a huge problem in this case - this is bad for accessibility as content generated in CSS pseudo selectors don't usually get picked up by screen reader. I'd suggest that we use JS to inject a node for this "This is a draft." note instead.

source/sass/buyers-guide/views/category.scss
```diff
@@ -39,6 +39,8 @@
         padding-right: 3%;
       }
 
+      @import "../fragments/_draft_product.scss";
```
mmmavis: We shouldn't `@import` within a declaration block as it copies and includes the same set of rules every time we do this.

image

mmmavis: With the accessibility concern above I think we can probably just have CSS rules for `.draft-note` and have JS inject a `draft-note` node wherever appropriate.

source/sass/buyers-guide/fragments/_draft_product.scss
```diff
@@ -0,0 +1,14 @@
...
+  color: #ff000054;
```
mmmavis: Just a note that generally speaking we should always use 100% opaque colours (unless opacity is requested by design).

source/sass/buyers-guide/fragments/_draft_product.scss
Pomax: Given that this is _not_ for real users, I'm not too concerned with this. I'd rather keep it "out of content" until it becomes necessary to make it real content, since it's purely a visual cue, not "page content".

source/sass/buyers-guide/views/category.scss
Pomax: Whether you use an import or a mixin, the final CSS has the code duplicated as many times as necessary  (both includes and mixins are sass constructs, not CSS constructs) so unless I'm missing something that got landed in CSS itself that allows for reusables now, using one or the other has no effect on the final CSS, only on SASS (in which case this is a normal partial, detected by SASS through the leading underscore in the file name)

source/sass/buyers-guide/fragments/_draft_product.scss
Pomax: /cc @kristinashu @alanmoo

alanmoo: I get both sides, but yeah, we already use pseudoelements to indicate draft pages on opportunity pages so let's go with this as-is.

source/sass/buyers-guide/views/category.scss
alanmoo: I'm with Mavis on this one- an `@import` block in the middle of the sheet feels weird, for some reason. I might even argue that this should just be a class that's applied to a bunch of things in HTML rather than being imported in multiple CSS files. Though I haven't looked extremely closely at where/how this is used.

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -149,7 +151,7 @@
   }
 
   .creepiness-slider {
-    z-index: 1;
+    z-index: 2;
```
Pomax: "this is a draft" would show up on top, which was a little odd.

source/js/buyers-guide/homepage-c-slider.js
```diff
@@ -62,12 +62,17 @@ export default {
...
 
-      // shortcut this scroll update if there are no products
+      // Shortcut this scroll update if there are no products
       if (n===0) { return; }
 
-      let averageCreepiness = visible.reduce( (tally, v) => tally + parseFloat(v.dataset.creepiness)/n, 0);
+      // Reduce the creepiness, treating any product without votes
+      // as a "neutral" product with creepiness = 50
+      let averageCreepiness = visible.reduce( (tally, v) => {
+        let value = parseFloat(v.dataset.creepiness) || 50;
```
Pomax: without this check the resultant creepiness for products without a value (yet) would end up being "undefined", which is not one of the five possible creepiness labels =)

Pomax:
turned the draft styling into a mixin, and noticed the jsx needed an update too because having draft products without votes would cause the homepage slider smiley face to report "undefined" as a creepiness value, rather than one of the five labels it should be using.

mmmavis:
Thanks for updating SCSS @Pomax . I've been cleaning up front end code & working with designers to streamline styling (colours, font, etc) across Foundation site. After that part gets sorted out I will schedule a meeting with you and Youri for us to talk about how we can organize our SCSS code better. Would love to know your thought! 👂

source/sass/buyers-guide/mixins/draft-product.scss
```diff
@@ -0,0 +1,16 @@
+// product draft rules
+@mixin draft-product {
```
mmmavis: If there's only one version of `.draft`, for the SCSS part I think we can simply have rule set for `.draft` (without the mixin). 

```scss
.draft::before {
    content: 'This is a draft.';
    color: #ffa7a7;
    font-weight: bold;
    font-size: 120%;
    font-family: 'Zilla Slab';
    z-index: 1;
    position: absolute;
    top: 0;
    left: 0;
    right: 0;
    text-align: center;
}
```
Then we can simply apply this classname `.draft` to wrapper element of our choice (e.g., `<div class="product-header draft" ...>`). This `<div>` will auto pick up the rules above as it has `draft` as its class (we don't need to  do `@include` or `@import` multiple times)

Pomax:
updated. @cadecairos was this R+ from your end?

source/sass/buyers-guide/bg-main.scss
```diff
@@ -84,6 +84,9 @@ img.thumb-border {
   border: 16px solid #f4f4f4;
 }
 
+// mixins
+@import "./includes/draft-product";
```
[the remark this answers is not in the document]
Pomax: done =P

PR #2443, opened 2019-01-11
Removed ul reset rule

youriwims:
Closes #1550

To add margin-bottom to the ul tags, I removed the ul's reset rule. Now, the ul & ol have identical rules.

PR #2449, opened 2019-01-12
Revert "Removed ul reset rule"

mmmavis:
Reverts mozilla/foundation.mozilla.org#2443

Reason: https://github.com/mozilla/foundation.mozilla.org/pull/2443#issuecomment-453696739

### DIFF
--- a/source/sass/resets.scss
+++ b/source/sass/resets.scss
@@ -13,3 +13,9 @@ img {
   max-width: 100%;
   height: auto;
 }
+
+ul {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}


gideonthomas:
@mmmavis what was the intention of the original ticket? Was it to keep the ol/ul rules the same or was it just to keep ol/ul the same for the bottom margin?

cc @kristinashu

mmmavis:
@gideonthomas my understanding is that we want to make sure both `<ul>` and `<ol>` **added in the CMS** come with bottom margin.

But either way, a fix to one problem shouldn't introduce more problems as side effects. Removing all reset rules for `<ul>` will affect all `<ul>` elements on the site, not just the ones added in CMS. This for sure will introduce unexpected styling changes and redundant code (e.g., unnecessary classnames or CSS rules) to all non-CMS generated `<ul>` (e.g., `<ul>` in side-wide footer, `<ul>` on fellowships pages... etc). We could remove the reset rules for good, but we have to make sure we evaluate all `<ul>` elements on the site and apply necessary changes as part of the PR.

youriwims:
@mmmavis & @gideonthomas , I can definitely make this change and add the margin-bottom to the cms `ul` only. However, it may be worth looking into a bit further as `ol` `ul` and `dl` tags are set to receive the same rules of `margin-top: 0` and `margin-bottom: 1rem` via normalize.css (from what I can see). But `ul` is the only tag out of those three listed in resets.scss which is undoing the `margin-bottom` it should have.

@mmmavis Let me know how you'd like me to go about this.

youriwims:
Thanks @gideonthomas & @mmmavis. Please see #2452 for updates on this.

mmmavis:
Awesome! Thank you thank you! ✨

Issue #1550, opened 2018-05-22
[Wagtail: vertical spacing] Missing margin bottom in <ul> element


taisdesouzalessa:
### Issue
Currently our `<ul>` element has no margin-bottom in the CMS. 

<img width="1022" alt="image" src="https://user-images.githubusercontent.com/22157921/40389665-e2717d4c-5dc7-11e8-8279-0aa4ff6cec8e.png">

This element should be like our `<ol>` element (which has a margin-bottom of 1rem = 16px):

<img width="1064" alt="image" src="https://user-images.githubusercontent.com/22157921/40389757-0636add8-5dc8-11e8-8baa-7cbff408dd75.png">

youriwims:
@taisdesouzalessa or @gvn, on which page can I find this spacing issue? Thanks!

youriwims:
@kristinashu, do the ol and ul still need to have identical rules?

kristinashu:
Hmmm did a little test and looks like this has been fixed!

https://foundation.mofostaging.net/en/initiatives/test-ul-vs-ul/

<img width="945" alt="image" src="https://user-images.githubusercontent.com/1068783/50997247-e5a81480-14d8-11e9-8269-7f619ac2a225.png">

kristinashu:
Ok thank Pomax, yeah let's align them! @youriwims please add bottom margin to `ul` so it matches `ol`.

PR #2452, opened 2019-01-14
Added margin-bottom to CMS ul

youriwims:
Closes #2449 

The original PR #2443 referencing issue #1550 removed the site-wide `ul` reset rules whereas only the CMS `ul` needed to be updated with `margin-bottom`. So, this is a follow-up on PR #2449 which reverted the changes made in PR #2443 and updated the `margin-bottom` on CMS `ul` tags only.

cc: @mmmavis


mmmavis:
Yea, will do after my morning meetings!

mmmavis:
Ummm I took a look at the codebase and realized not all CMS generated content is wrapped in a `.cms` wrapper, so adding a rule for `.cms` won't cover everything. I will create a ticket to discuss whether or not we wanna keep the `.cms` class.

Since we wanna align CMS generated`ul` and `ol`'s styling, I think the prefer way of doing it is to

1.  remove `ul`'s reset rules in `resets.scss`
2. search through all `ul` elements in `*.html, *.js, *.jsx, *.py` in the codebase and modify their classname/SCSS rules where appropriate.

This way we don't have reset/override rules just for `ul` and not `ol`.

I'll create a PR just to show you what I meant. @youriwims let's chat tomorrow and see if you wanna pick up my to-be-filed PR or start a new PR on your own.

Pomax:
Also note that the `.cms` class is for all CMS streamfield content. Not for any template code outside of that.

youriwims:
Ref #2455

youriwims:
Resolved in PR #2455

PR #2460, opened 2019-01-15
[WIP] Refactor & Optimize Home Hero Image

youriwims:
Closes #2425 

- [x] Clean up code (#2462)

- [ ] Optimize images (resolution)

- [ ] Confirm 16:9 image aspect ratio (#2426)

PR #2462, opened 2019-01-15
Merge branch 'clean-up-code' into remove-cloudinary

youriwims:
Based on #2460, Item 1:

1. The code was cleaned up using a singular img tag 
2. Multiple sources were added to the img tag
3. A breakpoint (575px) was established to help the browser choose the best image

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -31,6 +31,7 @@ whitenoise = "*"
 wagtail-modeltranslation = "*"
 "psycopg2-binary" = "*"
 cloudinary = "*"
+sorl-thumbnail = "*"
 
 [dev-packages]
 "flake8" = "*"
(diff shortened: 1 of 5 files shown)


patjouk:
Went into git hell when I tried to get the remove-cloudinary branch back to the state it was before I started adding changes to wagtail files (they are out of scope of the work I was doing on the buyers guide). I didn't realize that your PR #2460 was using my branch. Sorry for that mess >.<
It would be better to close PR #2460 for now, create a new branch and start again from there. You can refer to the wagtail-images branch to see the changes we did but don't use it: I'm only keeping it as a reference but will trash it in the future.

youriwims:
Ahhh, so sorry—but thank you! :bowtie:

PR #2472, opened 2019-01-17
(WIP) Update Participate factory

### LINKED ISSUE #1767: Deprecate old Participate page

### LINKED ISSUE #1906: Update Participate factory

mmmavis:
# 🚧 WIP

Closes #1767 
Closes #1906 

[ ] Migrations squashed


Pomax:
Hm, applying this to the current stage db causes the CMS to start failing its nav menu API calls... so something weird is going on and I might have to dig into this a bit.

Pomax:
Okay, it _is_ this PR, so... maybe we do still have a participate page somewhere.

Pomax:
So this is apparently not as easy as plain Django, and we may need to ask how to properly do this renaming trick on Stackoverflow (which is apparently the official place to ask wagtail questions)

mmmavis:
k, filed a [question on Stackoverflow](https://stackoverflow.com/questions/54318017/wagtial-how-to-properly-rename-a-wagtail-page-model). @Pomax feel free to chime in and help me explain the question better. To be honest I don't 100% know what I'm doing when it comes to Python/Django/Wagtail 😂

mmmavis:
Closing. This work is more complicated than we expected. Gonna deprioritize it and tackle it another time.

PR #2476, opened 2019-01-17
Use SCSS variables for colours where we can

### LINKED ISSUE #2475: Make sure we use SCSS variables for colours where we can
Related to the style guide cleanup ticket https://github.com/mozilla/foundation.mozilla.org/issues/2440

---

from https://foundation.mozilla.org/en/style-guide/

image

e.g. instead of using `#666666`, we use just use `$dark-gray`

mmmavis:
Closes #2475


mmmavis:
Awesome! Thanks for the speedy review!

PR #2514, opened 2019-01-23
Update package.json + .travis.yml to enable percy.io runs

Pomax:
Switches the visual test runner from plain Cypress to Percy.io


Pomax:
Hm, this does trigger percy.io, but it also seems to be generating a snapshot off of a blank database, which it shouldn't be doing... it should be bootstrapping a minimal "real" db with our foundation homepage due to the fact that we're asking percy to run `npm run cypress`, which is supposed to fire up the server with a `wait-on` for the port before it allows cypress to actually kick in.

Pomax:
somehow travis is not picking up the load_fake_data in this PR. closing and refiling.

PR #2531, opened 2019-01-28
Moved nav tag and its relevant template to /utility and /templates directories respectively

mmmavis:
I moved 2 files `nav_tags.py` and `horizontal_nav.html` away from the `/wagtailpages` directory since this `nav_tags.py` library will be used by non-wagtailpages as well.

image

I couldn't decide if `horizontal_nav.html` should be in `/templates/tags` or `/utility/tags` (or another dir?) though.


Pomax:
The whole reason we got a CMS was so that we could stop having pages that aren't CMS managed, so I'm a little worried about the need to move things out of wagtail (Rather than doing a layout refactor, since we should really stop calling this thing "networkapi" =D)

/cc @alanmoo

alanmoo:
What if we spun up a second wagtail managed site with different page templates, but wanted to reuse components like this? Would this directory structure then make sense?

mmmavis:
I was only thinking about the shared nav tags & templates use case. Since those will be used on both wagtail pages & non wagtail pages like Fellowship and Buyer's Guide pages.

Pomax:
The fellowship pages are getting moved into CMS so we shouldn't optimize for that I think. Buyers guide is also a "not CMS managed" set of pages because that's what we started with, but there is no good reason for that to keep living outside the CMS, too... 

The ideal dir structure here isn't `networkapi/network-api/wagtailpages/` but `foundation/apps/` and then moving this up makes a little more sense.

Ultimately it doesn't make too much of a difference as far as the code cares, but if we're moving things around to get code that's based on CMS content to be usable by non-CMS content, there should be a discussion on why that content isn't in the CMS, and whether that's due to "that's what it is now" or due to "it cannot be CMS content".

@alanmoo this feels like something that's worth figuring out relatively soon, it'd be nice to have a codebase with a dir structure that no longer harkens back to "þe olden dayes" =)

mmmavis:
let's chat about this during today's meeting regarding templates

Pomax:
@mavis do we want to land or close this one?

mmmavis:
I would like to land this as it will help unblock some of my nav clean up work. Will fix merging issues soon.

When we figure out how we want to organize templates for the Foundation site we can move these nav related files around accordingly.

Issue #2536, opened 2019-01-29
"fix" dates and placeholder "image" bg-colors for the data faker


Pomax:
It looks like the fake data generator is not entirely respecting the fact that there is a fixed seed, leading to different dates and image bg-color on auto generated data despite using the same seed value.

Pomax:
@cadecairos if you happen to know how to effect that off hand, let me know. If not, I'll dig into this a little.

PR #2541, opened 2019-01-29
added a RANDOM_SEED as env-level seed for the data faker

Pomax:
Partial fix for https://github.com/mozilla/foundation.mozilla.org/issues/2536

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,7 @@ script:
 - npm test
 - pipenv run flake8 tasks.py network-api/
 - pipenv run coverage run --source './network-api/networkapi' network-api/manage.py test networkapi
-- pipenv run python network-api/manage.py load_fake_data --seed 1
+- pipenv run python network-api/manage.py load_fake_data
 - npm run percy
 after_success:
 - coveralls
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,6 +11,7 @@ install:
 - nvm use --lts=carbon
 - npm install -g npm@latest # Needed to use npm ci
 - npm ci # use package-lock.json
+- npm i @percy/cypress@0.2.3 # version also noted in package.json -> peerDependencies
 - pipenv run python network-api/manage.py collectstatic --no-input
 before_script:
 - psql -c 'create database network;' -U postgres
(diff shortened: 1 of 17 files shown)


Pomax:
Nope, I'm down with that.

Pomax:
Added seed value 530910203, which is the CRC16 for "foundation.mozilla.org" =)

.travis.yml
```diff
@@ -19,29 +19,30 @@ script:
...
 - pipenv run flake8 tasks.py network-api/
 - pipenv run coverage run --source './network-api/networkapi' network-api/manage.py test networkapi
-- pipenv run python network-api/manage.py load_fake_data --seed 1
+- pipenv run python network-api/manage.py load_fake_data
 - npm run percy
 after_success:
 - coveralls
 env:
   global:
+  - ALLOWED_HOSTS=localhost
```
Pomax: sorted them alphabetically just because I like being able to find things =D

Pomax:
Hm, it looks like I/we need to figure out how to make the RANDOM_SEED value get ignored during the standard `inv test` part of the run. Because a value is set at the env. level for Travis, `inv test` will use that seed value, and so dates in News won't follow the testing constraints that should apply when no seed is specified =S

Pomax:
Resolved this by making faker instructions that need to be "compliant" during Django's `manage.py test` run but need to be deterministically "random" for visual regressiont esting tap into the `settings.TESTING` var, which tells us whether or not we're in a true test run, or whether we're "just running code".

network-api/networkapi/settings.py
```diff
@@ -538,10 +540,9 @@
 # Buyers Guide Rate Limit Setting
 BUYERS_GUIDE_VOTE_RATE_LIMIT = env('BUYERS_GUIDE_VOTE_RATE_LIMIT')
 
-# Detect if we're testing
+# Detect if we're in official testing mode
```
patjouk: Nitpicking after PR is merged: can you define official here? Does it mean CI or testing on dev machines?

Pomax: Neither - the only testing Django knows of is the kind that happens when you run `manage.py test`, which this lets you detect.

PR #2543, opened 2019-01-30
[WIP] a profile directory stub

Pomax:
⚠️ Do not merge ⚠️ 

@youriwims this is an example of the stubbed out profile directory component, with:

1. a new `ProfileDirectory` component in `customblocks.py`, with a corresponding
2. entry in `models.py` to make it show up in the streamfield block picker when editing a page in wagtail.
3. a template for the `ProfileDirectory` (explicitly indicated in its Meta class) found in `wagtailpages/templates/blocks/profile_directory.html` that does really not all that much:
```
{% load static %}

<div class="profile-directory">
  <div class="profile-filters">
    {% for label in filters %}
    <button class="filter-button">{{ label }}</button>
    {% endfor %}

    {% if filters|length > 0 %}
    <script src="{% static 'js/directory-listing-filters.js' %}"></script>
    {% endif %}
  </div>

  {% include "./profile_blocks.html" %}
</div>
```
4. that `directory-listing-filters.js` file, in a directory that Django knows how to "harvest" when it collects all the static assets it needs to expose for a site (for our `wagtailpages` app that's `wagtailpages/static` and then because it's a `js` file, it's put in a dir called `js`. That could be any name though, as you can see that the dirname is included in the argument to `static`).

So it builds some buttons for whatever we're filtering on, and then it just goes "and for the actual profiles list... lol do whatever profile_blocks.html knows how to do with the profiles I got back from the pulse API" =P

One thing this also does is load a javascript file that can hook its way into the filter buttons, so that we can add in "real code" to start toggling visbility on/off for profiles that are/aren't relevant based on the selected filters.

## Testing this component

Check out the `directory-listing` branch and then fire up http://localhost:8000/cms and pick a page to edit such as **campaigns**→ **single page**, and then in the body of that page, click `(+)` to reveal the block picker, and pick the "profile directory" component.

For its filters, just copy-paste `2013,2014,2015,2016,2017,2018` into the field and hit publish, then visit the live page after publish.

You should see something like this:

image

And clicking the buttons should console log their value.

## So what's left?

- making the filter bar and buttons look good.
- making sure that there are properties on the profile block DOM that allow for easy toggling (like a data-attribute), which we can do by updating the `profile_blocks` template - I would suggest we add `data-profile`, `data-type`, and `data-year`, and then as initial pass we can just "assume" the filters are going to be years and tie them into `[data-year=...]` query selectors


network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -488,3 +488,21 @@ class Meta:
...
+
+
+class ProfileDirectory(LatestProfileList):
+    filter_values = blocks.CharBlock(
+        required=True,
+        default='',
+        help_text='Example: 2013,2014,2015,2016,2017,2018'
+    )
+
+    def get_context(self, value, parent_context=None):
```
Pomax: This is a bit of a funky thing: the component lets people type strings, but we want to work with an array of strings in the template, so we override the definition of `get_context`, which is the dictionary that django uses as "variables available inside templates", and then we add in a context variable called `filters` which is the original string, split up into an array instead.

Pomax:
superceded by the actual work PRs

PR #2547, opened 2019-01-30
made things that weren't tied into the PRNG seed, tied into the PRN seed

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/2536 by applying the final fixes necessary for the random data to be tied to the PRNG (for the test cases we have right now)

### DIFF
--- a/network-api/networkapi/utility/faker/streamfield_provider.py
+++ b/network-api/networkapi/utility/faker/streamfield_provider.py
@@ -1,12 +1,17 @@
 import json
-from random import randint
 
+from random import randint
+from django.conf import settings
 from faker import Faker
 from faker.providers import BaseProvider
-
 from wagtail.images.models import Image
 
+seed = randint(0, 5000000)
+if settings.RANDOM_SEED is not None:
+    seed = settings.RANDOM_SEED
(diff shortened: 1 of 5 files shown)


network-api/networkapi/utility/faker/streamfield_provider.py
```diff
@@ -67,7 +78,7 @@ def generate_image_text2_field():
 
 
 def generate_spacer_field():
-    size = randint(1, 5)
+    size = fake.unix_time() % 5
```
Pomax:
```suggestion
    size =  1 + (fake.unix_time() % 5)
```

network-api/networkapi/utility/faker/streamfield_provider.py
```diff
@@ -1,12 +1,17 @@
...
-from random import randint
 
+from random import randint
+from django.conf import settings
 from faker import Faker
 from faker.providers import BaseProvider
-
 from wagtail.images.models import Image
 
+seed = randint(0, 5000000)
```
Pomax: hm, are these lines still required now...?

Issue #2550, opened 2019-01-31
Turn the cypress fixed length wait into a "JSX is done" event listener.


Pomax:
A 10 second delay is kind of long, we should be able to listen for `main.js` to say "Okay I'm done the page is now stable" and use that as signal to trigger a Cypress/Percy snapshot

PR #2553, opened 2019-01-31
add an event trigger for when main.js is done

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/2550 by adding a custom event at the end of the main.js run, which cypress can pick up on in order to do its visual regression testing, cutting down the travis-ci runtime by quite a bit.

### DIFF
--- a/cypress/integration/multi-page.js
+++ b/cypress/integration/multi-page.js
@@ -9,11 +9,16 @@ describe('multipage visual regression tests', () => {
     // Load the multipage campaign
     cy.visit(`/en/campaigns/multi-page/`);
 
-    // Give the browser a few seconds for JSX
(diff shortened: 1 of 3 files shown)


Pomax:
Hm, looks like this isn't ending up triggering.

Pomax:
Looks like it's not kicking in late enough now, because in addition to the main.js bundle, we also have to wait for the pulse results to come in.

Pomax:
Left: "what it should look like", right "what we're snapshooting": clearly too early

image

mmmavis:
do we need to check if images etc have been loaded as well?

Pomax:
It's mostly a "just in case" pause, not because something is still waiting to load.

PR #2567, opened 2019-02-01
Remove “featured” property of news

### LINKED ISSUE #721: API: Remove `featured` property from News model
Since we're switching to an admin managed homepage which deals with the relationship between the homepage and people, news, and highlights using foreign key relationships, there is no need for the `featured` property that we currently have in the News and Person models.

We would need to note down the currently featured news and person instances so that we can migrate/translate that data over to the new admin managed homepage (which does not use this `featured` flag).

The front-end would also need to be updated to not rely on the `feature` flag of these two models before we can remove this.

**Edit by Alan: For this ticket, let's just remove it from the News model. The People model will eventually entirely change.**

alanmoo:
Closes #721


Pomax:
Note that the percy differences are caused by the fact that removing a field changes which elements receive which pseudo-random numbers, thus changing basically everything about the pages that `load_fake_data` generates.

Issue #2561, opened 2019-01-31
Increase max number of highlights on Initiatives Featured Programs and Projects section


kristinashu:
I've added a 4th highlight to [initiatives](https://foundation.mozilla.org/en/initiatives/) `Featured Programs and Projects` section but it isn't showing up on the page. Is there a max of 3 features for this section? If so, please increase it to 9.

Screen cap of cms:
<img width="1172" alt="image" src="https://user-images.githubusercontent.com/1068783/52086707-7dd76d80-255c-11e9-9390-9afdf21253e6.png">

Screen cap of prod:
<img width="1285" alt="image" src="https://user-images.githubusercontent.com/1068783/52086733-90ea3d80-255c-11e9-91c9-e6ca0804e9d7.png">

Pomax:
Yeah, the template limits it to 3 (see [here](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/static/initiatives_page.html#L72-L82)) but I can't remember why we did that rather than just trusting people who have the rights to modify the initiatives page to know what they're doing.

@alanmoo objects to removing that count limit and just rendering `.all()`?

alanmoo:
My only guess is "if there's 5 it looks weird", but yeah, I'm fine letting that minor breakage happen if it's our own doing. (Though ideally those might be flex'd items or something based on a design recommendation, it's not necessary to address right now.)

alanmoo:
It seems like the 3 was, in part, a way to enforce it looking good at all screen sizes. Currently, we hide the third item on "medium width" devices, so that it doesn't end up looking like this: 
<img width="609" alt="image" src="https://user-images.githubusercontent.com/1682681/52244377-5ad5f200-28ab-11e9-8ab8-9573dd07a594.png">

If we unlock this to allow an arbitrary number, there's always going to be a case where there's a widowed item or two at some screen size. Is this ok @kristinashu?

kristinashu:
Yeah, I think that's ok if there are gaps sometimes.

Pomax:
roger - I shall land this PR, then, and if it turns out to be a problem we can address it in a new issue?

PR #2583, opened 2019-02-04
remove the hardcoded limit for highlights

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2561

Rather than hardcode the limit, it's nicer to have whoever updates the initiatives page decide how much content should be shown.


network-api/networkapi/wagtailpages/templates/wagtailpages/static/initiatives_page.html
```diff
@@ -72,11 +72,9 @@ <h3 class="h2-heading">{{ page.h3 }}</h3>
   <div class="highlights mb-5">
     <div class="row">
       {% for highlight in page.featured_highlights.all %}
-      {% if forloop.counter <= 3 %}
       <div class="col-md-6 col-lg-4 {% if forloop.counter == 3 %}d-md-none d-lg-block{% endif %}">
```
alanmoo: It's not obvious to me what this is doing, but...I feel like it needs to be changed too. Yes/no?

alanmoo: (The forloop.counter check)

Pomax: it just goes "is this highlight 0, 1, or 2? Cool, make a div" and even if there are more, it won't build divs for them.

alanmoo: False: See medium viewports (this is a bootstrap class, the `-md-` tipped me off). Following up in a comment, as we need to pull in Kristina.

Pomax: gah, always forget about those breakpoints =_=

looks like the removal got okayed, though, so I shall land this.

Issue #2510, opened 2019-01-23
Increase shares after voting implementation 


beccaklam:
This is the implementation ticket for issue #2446 

- [x] Change header to h3, [M]h3 for mobile, however, since things are centered in this section of the page can we change the text-align also to center?
- [x] Add vote count to above header
- [x] Delete 'View comments or share your results'
- [x] Change social icon button styles (follows the would buy/wouldn't buy button styling, except only one option per button)

## Desktop:
![screen shot 2019-01-23 at 2 44 52 pm](https://user-images.githubusercontent.com/20977376/51634170-92df4b80-1f21-11e9-849b-0268ed848773.png)

## Mobile:
![screen shot 2019-01-23 at 2 45 20 pm](https://user-images.githubusercontent.com/20977376/51634191-9d014a00-1f21-11e9-80db-8a66ca141327.png)

You can find the file in abstract here:
<img width="343" alt="screen shot 2019-01-23 at 3 15 53 pm" src="https://user-images.githubusercontent.com/20977376/51634254-cd48e880-1f21-11e9-9874-41a3b00fa7f7.png">

**Assets:** Wasn't sure if any new assets were needed let me (@beccaklam) know if anything is needed.

kristinashu:
@Pomax are you working on this? We're aiming to have it done for launch early this week.

Pomax:
Ah! I read this issue wrong, but I can definitely take this one.

kristinashu:
Lol thank you!

Pomax:
Where might I find the "link" icon? (we have the fb, twitter, and email already)

kristinashu:
Could you use the one at the top of each product page (but make it black)? 
<img width="943" alt="image" src="https://user-images.githubusercontent.com/1068783/52245517-c497d100-2896-11e9-8fb9-c070dc5ecc10.png">

Pomax:
That should work - I'll make it a different file though, so it can sit alongside the other social media icons in the social media icon dir.

Pomax:
PR up - I added several breakpoints to keep things from looking weird at intermediate sizes, each size documented in the PR with a screenshot.

PR #2584, opened 2019-02-05
change the share CTA in accordance with #2510

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2510 by switching over the CTA to match the new comps, with several breakpoints.

**updated images**

Desktop (990px and higher):

image

max large (768px - 989px):

image

max medium (576px - 767px):

image

x-small:

image


kristinashu:
Want to make sure you've kept the current share analytic utms and have added a new one for the link share?

Pomax:
@kristina the events are still there, and the link has a share event identical to the other three except for the lable, which is "share vote using a link" in this PR.

Pomax:
@beccaklam good catch, I completely forgot to wire up the same copy functionality. The spacing is autogenerated by flexbox, I'll turn that off and try to make it a fixed 16px at desktop size. As for the ipad, this might be due to the CSS using `max blah: X else Y` rather than going "min blah: Y else X". so at the exact value itself it's doing the wrong thing. I'll see if I can switch those around!

kristinashu:
Awesome, nice catch and fix on the link share!

Layout issues Rebecca mentioned look good now but now at narrow widths, the text and icon look slightly lower on the button when it should be vertically centered.

image

Pomax:
@kristinashu fixed, should look right again.

kristinashu:
Bah I can't see 
image

Pomax:
@kristinashu @beccaklam @mavis github and heroku are both "green" again so the review app should be good to go

beccaklam:
@Pomax Thanks for making those changes!  The 'Copied' feedback is good. Desktop looks good. 768px width looks good.

Only thing seems to be that at 320px, the Facebook share button still feels a little tight. Could we shrink the icons maybe a tiny bit so the word 'Facebook' isn't right up against the edge of the button?

screen shot 2019-02-06 at 2 51 23 pm

Otherwise I think we're good to go :)

Pomax:
should be entirely possible!

Pomax:
@beccaklam done, this is probably the best I can make it behave for 320 screens (bootstrap itself only goes down to 576, anything below that is assumed equally tiny, so I had to introduce a custom breakpoint to make things look right at 320 without making the 400-576 range look a little off.

xmatthewx:
So is this ready for review by @mmmavis? Thanks!

mmmavis:
On it. I think the SCSS code can probably be simplified a bit so things can adjust nicely themselves without us setting fixed width and changing font/icon sizes for different screens. I'll leave my suggestion soon.

source/js/buyers-guide/components/social-share/social-share.scss
```diff
@@ -0,0 +1,47 @@
+.pni-share-buttons {
```
mmmavis: It'd better if we let elements adjust width themselves instead of setting fixed width and margin. I would simply the following SCSS to

```scss
.pni-share-buttons {
  max-width: 568px; // width from comps
  margin: 0 auto;

  .pni-s-link {
    $margin-x: 0.5rem;

    border: 1px solid black;
    border-bottom-width: $btn-shadow-width;
    color: black;
    transition: all 0.1s linear;
    border-radius: 2em;
    white-space: nowrap;
    margin-bottom: 1rem;
    padding: 0.5em 0;
    font-weight: 600; // this should be semi-bold so 600
    flex: 1;

    @media screen and (max-width: $bp-md) { // Generally we want to avoid using "max-width" since we write our SCSS in mobile first style. This is an exception here since we want these rulesets to apply only on mobile and using "max-width" is the cleanest solution here.
      &:nth-child(odd) {
        margin-right: $margin-x;
      }

      &:nth-child(even) {
        margin-left: $margin-x;
      }
    }

    @media screen and (min-width: $bp-md) {
      margin-left: 0.5rem;
      margin-right: 0.5rem;

      &:first-child {
        margin-left: 0;
      }

      &:last-child {
        margin-right: 0;
      }
    }

    @include hover-focus-active {
      ...
```
(don't forget to remove my inline comments)

mmmavis: # 🔷 mobile (buttons are now wider and they fill the row)


# 🔷 desktop
screen shot 2019-02-06 at 15 41 50

mmmavis: We'll still have to adjust the icon sizes and position since they differ quite a bit from comps. I can take this on and tackle it as a followup since it requires access to Sketch 👀

Pomax: Could I ask you to just commit that on this branch?

mmmavis: Oops, 2 mins. I think my suggested code above can be simplified further.

Pomax: let's do this in follow-up and get this in today still

Pomax: follow-up filed: https://github.com/mozilla/foundation.mozilla.org/issues/2618

Issue #2587, opened 2019-02-05
🌡 URL encoding funkiness for buyer's guide


alanmoo:
I copied a URL from my browser and pasted it into Slack, only to discover it wasn't working:

https://foundation.mofostaging.net/en/privacynotincluded/categories/Valentine%E2%80%99s%20Day/

The correct link is 
https://foundation.mofostaging.net/en/privacynotincluded/categories/Valentine's%20Day/

The trick here is that somewhere along the way, the apostrophe got converted into a "smart apostrophe" (That's probably the wrong word. It's the curly kind.)

Given that this is going to start being shared narrowly tomorrow and broadly in the following few days, what's the best thing we can do to make sure we don't have users stumbling on this and getting 404's? Can we easily sanitize the string that generates the category URL to strip that character altogether (thus making the url `/categories/Valentines%20Day`)? Or should we simply catch it with a redirect in the CMS admin for now?

@patjouk @cadecairos @Pomax @mmmavis

Pomax:
This isn't CMS-managed, so we're currently beholden to whatever Django does. We're creating the actual link over in https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/templates/category_nav_links.html#L5-L6, so one thing we can probably do here is use a sluggifying tag, with an unsluggifier on the views.py side.

Pomax:
Yeah, I just added that.

PR #2592, opened 2019-02-05
give categories a slug

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2587 by giving categories a `slug` field, and switching to using that rather than names for URL purposes.

Do we want to add this for products, too? /cc @alanmoo


network-api/networkapi/buyersguide/models.py
```diff
@@ -93,6 +98,10 @@ def published_product_count(self):
     def __str__(self):
         return self.name
 
+    def save(self, *args, **kwargs):
+        self.slug = slugify(self.name)
```
Pomax: one risk of doing it this way is that the slug can change if we rename a category, so the URL for that category will change. Alternatively, we set the slug once, at object creation (e.g. `if not self.id`) but then of course the slug will no longer reflect the category itself if we rename it.

I suspect we won't be renaming categories all too often though, so this feels... fine? @cadecairos @alanmoo

cadecairos:
@pomax should we consider falling back to name just in case of links out in the wild?

Pomax:
Good call, let me add that.

network-api/networkapi/buyersguide/models.py
cadecairos: that was sort of the thing we agreed upon when this was added to product names, that they wouldn't need to change. We could try to make it only on create, but if we're changing category names infrequently, I think our redirect middleware should be "good enough" to put in a redirect for a category change.

network-api/networkapi/buyersguide/views.py
```diff
@@ -89,7 +89,12 @@ def buyersguide_home(request):
 def category_view(request, categoryname):
```
cadecairos: Can we change the parameter name from `categoryname` to `category_slug` while we're here?

Issue #2608, opened 2019-02-06
Remove "npm install" from "npm start"


Pomax:
We're currently running `npm install` every time you run `npm start`, and at this point we don't add new libraries on a daily basis, so making sure the libraries are in sync doesn't offer any real benefits, but does constantly update package-lock.json, and makes it take longer before things actually start up.

PR #2609, opened 2019-02-06
remove "npm install" from the start command

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2608

Everyone tagged for signoff mostly because it's a change to how to install and run the foundation site.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,8 @@ If you're on windows, you need an extra step: run `inv manage createsuperuser` t
 
 You're done :tada:
 
+To catch up on new dependencies, migrations, etc. after initial setup, you can use the `inv catch-up` command.
+
 ## Development
 
 ### Pipenv and Invoke commands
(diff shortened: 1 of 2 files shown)


alanmoo:
There isn't anything that's assuming `npm i` gets run with `npm start` in the deployment flow, is there?

cadecairos:
> There isn't anything that's assuming `npm i` gets run with `npm start` in the deployment flow, is there?

There isn't, the Heroku buildpack for node projectsdoesn't call `npm start`, it's only used in a dev context.

mmmavis:
> There isn't, the Heroku buildpack for node projectsdoesn't call `npm start`, it's only used in a dev context.

Ah cool that was my only concern.

mmmavis:
Ah one note -
Now that `npm install` is not part of the dev flow, we need to find a way to ensure `package-lock.json` is always up-to-date.

Pomax:
If you need npm installs, `inv catch-up` still has you covered. updating package-lock.json when there's _actually_ a change (e.g. dependabot) is perfectly fine.

Issue #2615, opened 2019-02-06
Pseudo-randomise the PNI product factory, rather than using true random


Pomax:
If we don't, every single PR is going to get flagged as bad because the visual regression screenshots are different =D

PR #2617, opened 2019-02-06
replace random with pseudorandom

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2615

### how to test

1. add RANDOM_SEED=530910203 to your .env
2.  `rm network-api\db.sqlite3 && inv migrate && inv manage load_fake_data && inv runserver`
3. open `http://localhost:8000/en/privacynotincluded` in a tab,
4. rerun (1)
5. open the pni homepage in another tab
6. verify that the pages from different runs have identical product listing.

### what it should look like

using that specific random seed value, things should like this:

screenshot_2019-02-07 privacy not included a buyer s guide for connected products


Pomax:
@patjouk you might be interested in the python bits in `.travis.yml` which should work better than the bit we already had.

patjouk:
I'm also testing this solution but I'm still getting errors https://travis-ci.org/mozilla/foundation.mozilla.org/builds/490070041 Looks like something is changing the flag at some point. Like I suggested on slack, we could add a verbose 0 to collectstatics instead.

Pomax:
Yeah, if you're still getting fails and this just "happened to pass", we might just need to actually get someone's attention at Travis, and set verbose 0 on our end for now.

Pomax:
removed that change from this PR again.

network-api/networkapi/buyersguide/factory.py
```diff
@@ -17,13 +19,31 @@
...
+seed = randint(0, 5000000)
+if settings.RANDOM_SEED is not None:
+    seed = settings.RANDOM_SEED
+
+fake = ValueFaker()
+fake.random.seed(seed)
+
+
 def get_random_category():
     all = BuyersGuideProductCategory.objects.all()
```
patjouk: Out of scope of this PR but it would be nice to rename this variable to something that does not shadow all()

Pomax: good catch, filed https://github.com/mozilla/foundation.mozilla.org/issues/2625

Pomax:
follow-up for replacing unix_time() calls with random itself: https://github.com/mozilla/foundation.mozilla.org/issues/2627

Pomax:
@patjouk went over the code, it'll get cleaned up in the follow-up issue, right after I land this.

PR #2620, opened 2019-02-07
Simplified SCSS code for post vote section

### LINKED ISSUE #2618: improve the CSS for the PNI post-vost share dialog
Based on https://github.com/mozilla/foundation.mozilla.org/pull/2584#discussion_r254499647:

---

<@mmmavis> It'd better if we let elements adjust width themselves instead of setting multiple sets of fixed width and margin. I would simplify the following SCSS to

```scss
.pni-share-buttons {
  max-width: 568px; // width from comps
  margin: 0 auto;

  .pni-s-link {
    $margin-x: 0.5rem;

    border: 1px solid black;
    border-bottom-width: $btn-shadow-width;
    color: black;
    transition: all 0.1s linear;
    border-radius: 2em;
    white-space: nowrap;
    margin-bottom: 1rem;
    padding: 0.5em 0;
    font-weight: 600; // this should be semi-bold so 600
    flex: 1;

    @media screen and (max-width: $bp-md) { // Generally we want to avoid using "max-width" since we write our SCSS in mobile first style. This is an exception here since we want these rulesets to apply only on mobile and using "max-width" is the cleanest solution here.
      &:nth-child(odd) {
        margin-right: $margin-x;
      }

      &:nth-child(even) {
        margin-left: $margin-x;
      }
    }

    @media screen and (min-width: $bp-md) {
      margin-left: $margin-x;
      margin-right: $margin-x;

      &:first-child {
        margin-left: 0;
      }

      &:last-child {
        margin-right: 0;
      }
    }

    @include hover-focus-active {
      ...
```
(don't forget to remove my inline comments)

mmmavis:
Closes #2618

### DIFF
--- a/source/js/buyers-guide/components/social-share/social-share.scss
+++ b/source/js/buyers-guide/components/social-share/social-share.scss
@@ -1,33 +1,36 @@
 .pni-share-buttons {
+  max-width: 568px;
+  margin: 0 auto;
+
   .pni-s-link {
+    $margin-x: 0.5rem;
+
     border: 1px solid black;
     border-bottom-width: $btn-shadow-width;
     color: black;
-    text-transform: initial;
     transition: all 0.1s linear;
     border-radius: 2em;
     white-space: nowrap;
-    font-size: 1em;
-    width: 43%;
-    margin: 0;
+    margin-bottom: 1rem;
     padding: 0.5em 0;
+    font-weight: 600;
 
-    @media screen and (min-width: $bp-xs) {
-      margin: 0.5em 8px;
-    }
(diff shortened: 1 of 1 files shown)


Pomax:
looks good at all breakpoints!

Pomax:
@beccaklam can you confirm that this looks the way we wanted? It takes what we already landed but cleans it up a bit and should even work a little better at the small resolutions.

Pomax:
@kristinashu can you confirm that this looks the way we wanted? It takes what we already landed but cleans it up a bit and should even work a little better at the small resolutions.

mmmavis:
🎉

Issue #2594, opened 2019-02-05
CI is failing _a lot_


Pomax:
There's a lot of CI failure happening recently, so I figured I'd file this with some logs so we have something to look at @alanmoo @cadecairos @patjouk 

Travis failures:
```
Post-processed 'wagtailadmin/css/normalize.css' as 'wagtailadmin/css/normalize.aa75c284e281.css'
Post-processed 'wagtailadmin/css/core.css' as 'wagtailadmin/css/core.27ba9a2168bf.css'
Post-processed 'wagtailadmin/css/userbar.css' as 'wagtailadmin/css/userbar.15d766f7de4f.css'
Post-processed 'wagtailstyleguide/css/styleguide.css' as 'wagtailstyleguide/css/styleguide.917c29e92d5d.css'
Post-processed 'wagtailmodeladmin/css/index.css' as 'wagtailmodeladmin/css/index.73ce0bf177ce.css'
Post-processed 'wagtailmodeladmin/css/choose_parent_page.css' as 'wagtailmodeladmin/css/choose_parent_page.408cbfc7829f.css'
Post-processed 'rest_framework/css/prettify.css' as 'rest_framework/css/prettify.a987f72342ee.css'
Post-processed 'rest_framework/css/bootstrap-theme.min.css' as 'rest_framework/css/bootstrap-theme.min.66b84a04375e.css'
Post-processed 'rest_framework/css/default.css' as 'rest_framework/css/default.8d5591a6aabc.css'
Post-processed 'rest_framework/css/bootstrap.min.css' as 'rest_framework/css/bootstrap.min.77017a69879a.css'
Post-processed 'rest_framework/css/bootstrap-tweaks.css' as 'rest_framework/css/bootstrap-tweaks.46ed116b0edd.css'
Post-processed 'rest_framework/css/bootstrap-theme.css' as 'rest_framework/css/bootstrap-theme.a279820bbeb5.css'
Post-processed 'rest_framework/css/font-awesome-4.0.3.css' as 'rest_framework/css/font-awesome-4.0.3.c1e1ea213abf.css'
Post-processed 'rest_framework/css/bootstrap.css' as 'rest_framework/css/bootstrap.23773f1bb199.css'
Post-processed 'adminsortable/css/admin.sortable.css' as 'adminsortable/css/admin.sortable.82662a72ea4a.css'
[...]

PR #2621, opened 2019-02-07
set collectstatic to minimal verbosity

Pomax:
Helps (hopefully) with the Travis failures we're seeing, ref: https://github.com/mozilla/foundation.mozilla.org/issues/2594

Let's see what this does across CI runs.


Pomax:
ack, yes... I totlaly did that, to the wrong branch. Let me do that here too.

Pomax:
@patjouk thusly removed.

Issue #2622, opened 2019-02-07
Collapse the various multiple Cypress tests into a single file


Pomax:
We shouldn't really need multiple files, just multiple tests in the same file. Given that each file incurs a bootstrapping delay and postprocess delay, it's worth putting all the tests in one file, instead. Until we start doing "real" interaction based testing, which would definitely go in its own file.

PR #2623, opened 2019-02-07
consolidated visual tests

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2622

Note that this will go "I have no screenshots to compare to yet" because all tests are now in a new file, making them all "new tests".


Pomax:
> We shouldn't really need multiple files, just multiple tests in the same file. Given that each file incurs a bootstrapping delay and postprocess delay, it's worth putting all the tests in one file, instead. Until we start doing "real" interaction based testing, which would definitely go in its own file.

PR #2425, opened 2019-01-08
Hero Image Scaling Progress

youriwims:
Closes #2279 
Closes #1485

@alanmoo I'm not too familiar with this templating system, but when using the same image for both mobile & desktop, there's no issue (when tested on the real site in devtools). However, when doing this in the testing environment, it creates additional overscroll issues.

PR #2632, opened 2019-02-07
Updated image scaling

youriwims:
See #2425

cc: @Pomax


network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
```diff
@@ -11,18 +11,15 @@
 
 {% block content %}
 <div class="container">
-  <div class="row mb-5 mt-0 mt-sm-5 d-flex align-items-center homepage-feature">
-    <div class="col-md-6 mb-3 d-sm-none">
-      {% image page.hero_image fill-1152x770-c100 format-jpeg class="full-bleed-xs" %}
+  <div class="row mb-5 mt-0 mt-sm-5 d-flex align-items-center flex-sm-row-reverse homepage-feature">
+    <div class="col-sm-6 mb-3 mb-sm-0">
```
alanmoo: This line change seems to cause the Percy regression. Is that an intentional change?

youriwims: Yes, these changes were intentional.

Pomax:
Yeah this is looking a lot better on windows for me, too, hurray! We can add a 3x sources if we think that's worth it, not sure if we want to tackle that in this PR or rather land this, and file an image audit issue where we look at which images can easily be 3x'd and which images would be a lot more work.

network-api/networkapi/wagtailpages/templates/wagtailpages/tags/picture_ratios.html
```diff
@@ -0,0 +1,21 @@
+{% load wagtailimages_tags %}
+
+
+<picture class="cms-image {{class_names}}">
+
```
Pomax: Since the alt text is the same for all these images, I would recommend a small refactor:
```
{% with alt_text = unoptimized_image.alt %}
<picture class="cms-image {{class_names}}">

  {% image unoptimized_image fill-.... %}
  {% image unoptimized_image fill-.... %}
  <source ...  alt="{{ alt_text }}">

  {% image unoptimized_image fill-.... %}
  {% image unoptimized_image fill-.... %}
  <source ...  alt="{{ alt_text }}">

  {% image unoptimized_image fill-.... %}
  {% image unoptimized_image fill-.... %}
  <source ...  alt="{{ alt_text }}">

</picture>
{% endwith %}
```

youriwims: done.

youriwims:
I agree with Pomax, I think it'd be best to do 3x in another pr post image audit.

Pomax:
Code looks good to me! Adding @kristinashu for final visual pass

Issue #2630, opened 2019-02-07
Turn off donate banner for now


alanmoo:
Maybe related to an approach in #2451, but let's just turn it off for the time being. Ideally we'll get to a point where we can administer it in Wagtail.

cc @xmatthewx

PR #2634, opened 2019-02-07
disable modal slideup for both the main site and PNI

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2630

I assumed the ticket was to disable the modal on the site as a whole, including the buyers guide, let me know if it wasn't!


Pomax:
Legitimate visual regression error! O_O

Pomax:
It looks like what percy's showing as "the new screenshots" are the correct ones, with the reference screens having seemingly been taken in the middle of the modal popping up (ironically). Can you confirm, @alanmoo?

Issue #2645, opened 2019-02-08
Expose more petition strings for localization


alanmoo:
We should be able to localize the fields for Checkbox 1, 2, and "Thank you". Possibly the shareprogress links too (so that we can have different campaigns for different languages, though we should confirm with @TheoChevalier about the execution of this part)

Pomax:
is this on the CMS side, or in the JSX side?

alanmoo:
When toggling languages in the CMS, those fields don't have alternates. Presumably once that happens, the JSX field would be passed the right string when getting rendered, but we should double check that too. (Aside: the checkbox fields don't do anything right now, so this is...less pressing for those items. Probably useful to wait for that functionality, which is a ticket that hasn't been filed yet.)

TheoChevalier:
Pretty sure I synced with you Alan on this at some point, but wanted to make it explicit in this ticket that the approach for ShareProgress should indeed be making the `Share twitter`, `Share Facebook` and `Share email` fields localizable, then we would set up individual buttons for each language in Share Progress. I just had a call with @kaili09 confirming this approach would work for campaigns

So I think the two remaining tasks in this ticket are ShareProgress + the Thank you field (the checkboxes are localized)

kaili09:
It would be great if we could get this working for recently launched campaigns we're planning to localize in the coming weeks (although if that timeline is not feasible, I understand). It's also relevant for our H2 plans as it relates to our team OKRs.

alanmoo:
A note on these petitions right now: The checkboxes don't actually get saved anywhere. If there's going to be a need to make them do something, we're going to have to put some work into that again. Regarding the button fields: that should be eaiser. Let me see what I can do about it getting onto our schedule coming up.

TheoChevalier:
Right, I think I confused “checkbox 1 & 2” with the Privacy Policy + email opt-in checkboxes (which are localized).
The custom checkboxes are only used occasionally, so it’s totally fine not to prioritize that

alanmoo:
@TheoChevalier did anything ever happen to this or is the above still accurate?

TheoChevalier:
Yes, thanks!

This has been fixed by https://github.com/mozilla/foundation.mozilla.org/pull/3422

PR #2647, opened 2019-02-10
Update translations for petition snippet

TheoChevalier:
Adding more translations + ~~localizing 2 more strings on https://foundation.mozilla.org/en/campaigns/eu-misinformation/ “Add you name” + 3rd checkbox.~~

~~I couldn’t find those two strings anywhere in the repo, I guess they need to go elsewhere?~~


TheoChevalier:
Nevermind, just realized those two strings were coming from the snippet in wagtail. Making the checkbox localizable will be done in #2645

Since this is now `es` and `pt`, there’s no rush for this to go live.

PR #2658, opened 2019-02-13
Docs for tricky django migrations

patjouk:
Add info on how to handle tricky migrations in Django.

As you can see, we will need multiple deploys if we want to avoid downtime. Process for those migrations needs to be discussed: it will require multiple PRs that needs to be deployed to prod in a certain order.


docs/deployment.md
```diff
@@ -55,3 +55,32 @@ Folder | URL prefix
...
 
+## Django Migrations: is my deploy safe?
+
+### Context
+
+[Preboot](https://devcenter.heroku.com/articles/preboot) is a feature from Heroku that prevents downtime when deploying: Heroku still serves traffic to old dynos while creating new dynos at the same time. For up to 3 minutes, two versions of the code are running and talking to the same DB: backward-incompatible changes to the DB result in internal server errors from the old code.
+
+### How to apply backward-incompatible migrations
+
+Always announce in `#mofo-engineering` when you deploy backward-incompatible changes.
```
Pomax: Do we want to change this to something slightly stronger, like "Announce your intention to deploy in `#mofo-engineering`, and hold off deploying until people have indicated they've seen your announcement. If there is no one around, do not deploy."?

docs/deployment.md
```diff
@@ -55,3 +55,32 @@ Folder | URL prefix
...
+
+#### Rename or change a field type
+
+**Three deploys:**
```
Pomax: It feels like we probably want a section in the development.md docs as well, that explains the process for working on a task that requires multiple deploys: we want that to happen over multiple PRs, so something like "rename blah.moo to blah.flurp" should end up being a tracking issue for multiple PRs

.github/pull_request_template.md
```diff
@@ -1,5 +1,7 @@
 Closes #
 
+
+[ ] Check if this PR introduces [backward-incompatible DB changes](../docs/deployment#django-migrations-is-my-deploy-safe)
```
Pomax: Can we add a `  - [ ] Turn the issue this is a PR for into a tracking issue, and make sure there are 2 or 3 separate PRs to handle the changes` here, too?

docs/deployment.md
patjouk: I don't think it's a good idea in the current state of those dev/deploy docs: they're super disorganized and splitting them made things even worst. I've moved my changes in a "workflow.md" doc but the whole thing needs a serious reorg/rewrite.

.github/pull_request_template.md
patjouk: I've added a "related issue/PR" field, as it's more versatile. I don't expect the majority of our PRs to introduce backward-incompatible migrations.

docs/deployment.md
patjouk: done!

docs/deployment.md
Pomax: I down for a serious rewrite, but it'd be good to at least have a mention in the development text about that fact that if you're touching models/fields, please make sure to have read this part of the deployment docs, too.

PR #2669, opened 2019-02-15
Building out the fellow directory component

youriwims:
This pr is focused solely on building out the directory component whereas the styles (including social icons & fellow aspect ratio updates) are included in #2685 and the functionality of the filter bar will also be in a separate pr.


Pomax:
Looking at the commit history, this looks like it's the demonstration PR, did you add anything that you would like my review on?

youriwims:
No, nothing additional was added here.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/profile_directory.html
```diff
@@ -0,0 +1,11 @@
+{% load static %}
+
+<div class="profile-directory">
+  <div class="profile-filters">
+    {% for label in filters %}
+    <button class="filter-button">{{ label }}</button>
+    {% endfor %}
+  </div>
+
+  {% include "./profile_blocks.html" %}
```
Pomax: even if the JS comes in a different PR, this PR should probably have the changes to `profile_blocks` that adds a `data-year`, `data-profile-type`, and `data-program-type` that the filtering will be tapping into.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -52,6 +52,7 @@
     ('pulse_listing', customblocks.PulseProjectList()),
     ('profile_listing', customblocks.LatestProfileList()),
     ('profile_by_id', customblocks.ProfileById()),
+    ('profile_directory', customblocks.ProfileDirectory()),
```
Pomax: Since we'll be landing this in parts, let's lock this behind a feature flag:
```
base_fields = base_fields = [field for field in [
    ...
    ('profile_by_id', customblocks.ProfileById()),
    ('profile_directory', customblocks.ProfileDirectory()) if settings.ENABLE_PROFILE_DIRECTORY_BLOCK else None,
    ('airtable', customblocks.AirTableBlock()),
] if field is not None]
```
which requires a small update to `settings.py`. Add this in the `env` block at the top:
```
    ...
    DOMAIN_REDIRECT_MIDDLWARE_ENABLED=(bool, False),
    ENABLE_PROFILE_DIRECTORY_BLOCK =(bool, False),
    FILEBROWSER_DEBUG=(bool, False),
    ...
```
and then below the INSTALLED_APPS section, add this:
```
# Wagtail feature flags
ENABLE_PROFILE_DIRECTORY_BLOCK = env('ENABLE_PROFILE_DIRECTORY_BLOCK ')
```
That way, we can land this code, and even promote it to production, without it actually doing anything until we're either ready to flip that switch, or we decide to do a final code update to remove the feature flag condition.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -52,8 +52,9 @@
...
```
Pomax: ammending this:
```suggestion
   ('profile_directory', customblocks.ProfileDirectory()) if settings.ENABLE_PROFILE_DIRECTORY_BLOCK else None,
```
and then adding `ENABLE_PROFILE_DIRECTORY_BLOCK=True` to your .env file should show the profile directory in the CMS streamfield component picker, and setting it to `False` or omitting it from your .env file should show it _not_ being in the CMS streamfield component picker now.

youriwims: Got it. 👍

Pomax:
However, remember to run `npm test` and `inv test`: there's a linting error for the python code that needs to be fixed before this will pass CI

PR #2670, opened 2019-02-16
Fixes #1647: Removed `noValidate=true` from the petition form

### LINKED ISSUE #1647: Remove `noValidate=true` from the petition form
Right now the form we're using is set to bypass all field validations so that even if an HTML5 field has good built-in validation, we can't take advantage of that.

We should refactor the `Petition.jsx` code so that it no longer bypasses validation, and remove any code that we put in that basically duplicates validation that could be handled by the browser instead.

suhailsinghbains:
Fixes #1647 
Remove `noValidate=true` from the petition form


Pomax:
Unfortunately the reason we were using `noValidate={true}` is because we were doing custom validation in the react handling code for this form, so we can't just remove the attribute and trust things to work, hence the question on the original issue about whether you had any experience with html and React.

suhailsinghbains:
Thanks @Pomax I do have bit experience with HTML and React I would like to try again.

suhailsinghbains:
I should be removing validatesAsEmail(input) along with the call of this function. 
We already have type="email" in the html form, so, I don't think I should be removing that. if I am not wrong.
Please correct me if I am missing something.

Pomax:
That's the thing: no one's looked at what can safely be removed (without affecting the validation behaviour) so the first part of the work would be to, before actually _changing_ any code, audit which functions are currently involved in validating fields, and how many of those might be duplicating the work that browser-based form validation is already able to do. Then we can see whether we can even do this work in a single PR, or whether we need multiple PRs to keep things manageable (both in terms of working on, and in terms of reviewing).

Pomax:
hi @suhailsinghbains, so this PR needs more comments as part of the PR process itself: why did you remove what you removed? What did that leave? What was the result of your validation audit? Etc. 

And if that is turning into too much work, it's fine closing this one.

source/js/components/petition/petition.jsx
```diff
@@ -275,9 +275,6 @@ export default class Petition extends React.Component {
     // validate data here. Do not continue unless we're cool.
     let hasName = this.givenNames.element.value && this.surname.element.value;
     let email = this.email.element.value;
-
-    email = email && this.validatesAsEmail(email);
```
suhailsinghbains: I removed this line as we were performing logically "AND" between email and validatesAsEmail(), now as we are removing this function, so we have to remove this too.

source/js/components/petition/petition.jsx
```diff
@@ -318,18 +315,6 @@ export default class Petition extends React.Component {
     });
   }
 
-  /**
-   * Performs very simple validation for emails.
-   * @param {string} input the string that should be validated as an email address
-   * @returns {boolean} true if the input is a legal-enough email address, else false
-   */
-  validatesAsEmail(input) {
```
suhailsinghbains: This function has to be removed as the HTML itself does the checking of the E-mail.

source/js/components/petition/petition.jsx
```diff
@@ -528,7 +513,7 @@ export default class Petition extends React.Component {
     });
 
     let emailGroupClass = classNames({
-      'has-danger': this.state.userTriedSubmitting && (!this.email.element.value || !this.validatesAsEmail(this.email.element.value))
```
suhailsinghbains: As we are removing the function, but, we still want the email to not be null, hence I left the checking of the element's value.

source/js/components/petition/petition.jsx
```diff
@@ -584,7 +569,7 @@ export default class Petition extends React.Component {
                 disabled={disableFields}
                 onFocus={this.onInputFocus}
               />
-              {this.state.userTriedSubmitting && (!this.email.element.value || !this.validatesAsEmail(this.email.element.value)) && <small className="form-check form-control-feedback">{ get(`Please enter your email`) }</small>}
```
suhailsinghbains: As we are removing the function, but, we still want the email to not be null, hence I left the checking of the element's value.

suhailsinghbains:
Hi @Pomax 
I didn't understand the last commit :(

kristinashu:
cc @Pomax

Pomax:
I'm going to close this for now, as we need to do a deeper audit around how jsx forms vs. html forms are used. Many apologies, but thank you for being willing to help us.

PR #2685, opened 2019-02-19
Styling fellow directory component

youriwims:
Related PRs/issues #2669, #2684

## Checklist

- [x] Add social icons
- [x] Update Fellow aspect ratio to 1:1
- [x] Style directory component

PR #2698, opened 2019-02-20
Filter bar for the directory component

Pomax:
Related PRs/issues #2669 #2685

**Changes in Models:**
There is a migration in this PR, but that migration is already covered by #2669 and will disappear from this PR on a rebase.

**Documentation:**
- [x] Is my code documented?

**Commits**
This builds on another, not yet merged, PR, so most of the "files changed" are not actually part of this PR. The relevant changes are specifically in https://github.com/mozilla/foundation.mozilla.org/pull/2698/commits/0286cd09fb68dfb5cc7272d2580f964bbdd4e479

**How to test**
short version: https://foundation-mofostaging-pr-2698.herokuapp.com/cms/pages/21/edit/ for the CMS side, https://foundation-mofostaging-pr-2698.herokuapp.com/en/campaigns/single-page/ for the example

long version: check out the branch, and...
- add `ENABLE_PROFILE_DIRECTORY_BLOCK=True` to your `.env` file (hurray, feature flags)
- run `inv catch-up` (just to make sure. This also runs the migrations)
- if you started on an empty database, `inv manage load_fake_data` is a good idea, so you have something to edit
- run `npm start` to start everything up and then once running head over the `http://localhost:8000/cms" and log in as usual (if you had an empty db, you might need to run the createsuperuser command)
- head over to the single-page campaign page, and edit it: add a `profile directory` component to the streamfield body, and set the profile type to `fellow`, and the year filter to `2013,2014,2015,2016,2017,2018,2019`, then "save draft" (or publish) and then view the draft (or live page).

Scroll to the profile listing, and hit the various filter buttons to see content getting swapped in based on the year you picked.


alanmoo:
Other than the obvious presentation issues, one thing that jumps out at me is I have no idea what set of profiles I'm looking at by default.

source/js/components/fellow-list/fellow-list.scss
```diff
@@ -1,5 +1,7 @@
 .fellowships-directory-filter {
-  background: $subtle-gray;
+  display: flex;
+  flex-wrap: wrap;
```
alanmoo: Can this line and the previous be applied as bootstrap classes on the element? (`.d-flex` and `.flex-wrap`? (https://getbootstrap.com/docs/4.1/utilities/flex/)

Pomax: absolutely

Pomax: This is code from a different PR for the styling, but we'll make sure to take that on board there.

Pomax:
The default set is simply the latest X profiles from pulse, so we probably want to come up with a sensible "default view" policy. E.g. if we order the filter buttons 2019,...,2013 and then have all 2019 results loaded, with the 2019 button styled to suggest "it is selected", would that make more sense? 

Maybe @xmatthewx and @kristinashu has some thoughts around what the initial batch of profiles should be for this particular component, too.

kristinashu:
Not sure if this is what you're asking but the filter bar should basically be how it is on prod https://foundation.mozilla.org/fellowships/directory/. By default 2019 is listed first in the nav and is selected (selected state is bolder font). Is that what you're asking?

Pomax:
that's the perfect answer, thanks!

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -458,9 +458,6 @@ def get_context(self, value, parent_context=None):
         # filter out emptish values
         query_args = {k: v for k, v in query_args.items() if v}
 
-        # FIXME: the protocol should be part of the pulse api variable.
```
Pomax: we did that very thing quite a while back, but forgot to remove this comment

Pomax:
CI keeps failing on something that I literally cannot reproduce locally, no matter what I try, so I just redid the whole thing on top of a fresh clone master and pushed a new PR up.

Issue #2700, opened 2019-02-21
update ProfileDirectory component to allow for more than "year" filtering


Pomax:
Currently the ProfileDirectory has been primed to allow for year filtering (specifically to make porting the fellowhship pages possible), but this should be extended to filtering on any of the fields that can be specified,  eventually.

PR #2701, opened 2019-02-21
Filter bar

Pomax:
Redid #2698 on top of the known working master in the hopes of no longer hitting that weird migration CI failure, over migrations that `manage makemigrations` says do not exist and so cannot be generated O_o

Related PRs/issues #2669 #2685

**Documentation:**

- [x] Is my code documented?

**Commits**

This builds on another, not yet merged, PR, so most of the "files changed" are not actually part of this PR. The relevant changes are specifically in https://github.com/mozilla/foundation.mozilla.org/pull/2698/commits/0286cd09fb68dfb5cc7272d2580f964bbdd4e479

**How to test**

short version:
- https://foundation-mofostaging-pr-2701.herokuapp.com/cms/pages/16/edit/ for the CMS side,
- https://foundation-mofostaging-pr-2701.herokuapp.com/en/campaigns/until-event-evidence/ for the live example

long version: check out the branch, and...
- add `ENABLE_PROFILE_DIRECTORY_BLOCK=True` to your `.env` file (hurray, feature flags)
- run `inv catch-up` (just to make sure. This also runs the migrations)
- if you started on an empty database, `inv manage load_fake_data` is a good idea, so you have something to edit
- run `npm start` to start everything up and then once running head over the `http://localhost:8000/cms" and log in as usual (if you had an empty db, you might need to run the createsuperuser command)
- head over to the single-page campaign page, and edit it: add a `profile directory` component to the streamfield body, and set the profile type to `fellow`, and the year filter to `2013,2014,2015,2016,2017,2018,2019`, then "save draft" (or publish) and then view the draft (or live page).

Scroll to the profile listing, and hit the various filter buttons to see content getting swapped in based on the year you picked.

### DIFF
--- a/network-api/networkapi/wagtailpages/customblocks.py
+++ b/network-api/networkapi/wagtailpages/customblocks.py
@@ -443,23 +443,26 @@ class LatestProfileList(blocks.StructBlock):
         default=''
     )
 
-    def get_context(self, value, parent_context=None):
+    def get_context(self, value, parent_context=None, no_limit=False, initial_year=False, ordering=False):
         context = super().get_context(value, parent_context=parent_context)
+
         query_args = {
-            'limit': context['block'].value['max_number_of_results'],
-            'profile_type': context['block'].value['profile_type'],
-            'program_type': context['block'].value['program_type'],
-            'program_year': context['block'].value['year'],
-            'ordering': '-id',
+            'limit': value['max_number_of_results'],
+            'profile_type': value['profile_type'],
+            'program_type': value['program_type'],
+            'program_year': initial_year if initial_year else value['year'],
+            'ordering': ordering if ordering else '-id',
             'is_active': 'true',
             'format': 'json',
         }
 
-        # filter out emptish values
-        query_args = {k: v for k, v in query_args.items() if v}
+        # Removing after the fact is actually easier than
+        # conditionally adding and then filtering the list.
+        if no_limit:
+            query_args.pop('limit')
 
-        # FIXME: the protocol should be part of the pulse api variable.
-        #   see: https://github.com/mozilla/foundation.mozilla.org/issues/1824
+        # Filter out emptish values
+        query_args = {k: v for k, v in query_args.items() if v}
 
         url = "{pulse_api}/api/pulse/v2/profiles/?{query}".format(
             pulse_api=settings.FRONTEND['PULSE_API_DOMAIN'],
@@ -482,6 +485,10 @@ class LatestProfileList(blocks.StructBlock):
             pass
 
         context['profiles'] = data
+        context['profile_type'] = value['profile_type']
+        context['program_type'] = value['program_type']
+        context['program_year'] = value['year']
+
         return context
 
     class Meta:
--- a/network-api/networkapi/wagtailpages/customblocks.py
+++ b/network-api/networkapi/wagtailpages/customblocks.py
@@ -498,16 +498,42 @@ class LatestProfileList(blocks.StructBlock):
 
 
 class ProfileDirectory(LatestProfileList):
+    """
+    NOTE:
+    this component has been set up specifically for year
+    filtering for its initial pass. It does not do any
+    kind of filter detection and query argument juggling
+    yet, which is why the default filter_values text is
+    literally the string that matches the filter that
+    was used on the fellowship directory page prior to
+    porting to the CMS.
+    There is an issue open to make this component more
+    generic. See:
+    https://github.com/mozilla/foundation.mozilla.org/issues/2700
+    """
+
     filter_values = blocks.CharBlock(
         required=True,
-        default='',
-        help_text='Example: 2013,2014,2015,2016,2017,2018'
+        default='2019,2018,2017,2016,2015,2014,2013',
+        help_text='Example: 2019,2018,2017,2016,2015,2014,2013'
     )
 
     def get_context(self, value, parent_context=None):
-        context = super().get_context(value, parent_context=parent_context)
-        filter_values = context['block'].value['filter_values']
-        context['filters'] = filter_values.split(",")
+        pulse_api = settings.FRONTEND['PULSE_API_DOMAIN']
+        filter_values = value['filter_values']
+        years = filter_values.split(",")
+        initial_year = years[0]
+
+        context = super().get_context(
+            value,
+            parent_context=parent_context,
+            no_limit=True,
+            initial_year=initial_year,
+            ordering='custom_name'
+        )
+
+        context['filters'] = years
+        context['api_endpoint'] = f"{pulse_api}/api/pulse/v2/profiles/?ordering=custom_name&is_active=true&format=json"
         return context
 
     class Meta:
--- a/network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
+++ b/network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
@@ -1,0 +1,140 @@
+(function() {
+  // start of function
+
+  const profileCache = {};
+  const filters = document.querySelectorAll(`.profile-directory .fellowships-directory-filter .filter-option button`);
+  const profileContainer = document.querySelector(`.profiles .row`);
+  const { profileType, programType } = document.querySelector(`.profiles`).dataset;
+  const API_ENDPOINT = document.querySelector(`[data-api-endpoint]`).dataset.apiEndpoint;
+
+  /**
+   * After initial page load, the filter buttons are responsible for
+   * fetching results "per filter entry".
+   */
+  function getData(programYear) {
+    // set up a url for performing an API call:
+    let url = API_ENDPOINT;
+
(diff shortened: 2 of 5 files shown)


network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -443,23 +443,26 @@ class LatestProfileList(blocks.StructBlock):
         default=''
     )
 
-    def get_context(self, value, parent_context=None):
+    def get_context(self, value, parent_context=None, no_limit=False, initial_year=False):
```
alanmoo: I'm confused by this. 
a.) What is `no_limit`?
b.) Why are we passing initial_year in instead of using the first filter item? (That's what we're going to want to do once we can filter by more than just year anyway)

network-api/networkapi/wagtailpages/customblocks.py
```diff
@@ -443,23 +443,26 @@ class LatestProfileList(blocks.StructBlock):
...
+
         query_args = {
-            'limit': context['block'].value['max_number_of_results'],
-            'profile_type': context['block'].value['profile_type'],
-            'program_type': context['block'].value['program_type'],
-            'program_year': context['block'].value['year'],
+            'limit': value['max_number_of_results'],
+            'profile_type': value['profile_type'],
+            'program_type': value['program_type'],
+            'program_year': initial_year if initial_year else value['year'],
```
alanmoo: Having this parameter if we can _only_ filter by year is misleading

alanmoo: JK, I get it

network-api/networkapi/wagtailpages/customblocks.py
Pomax: no_limit bypasses the limit query argument so we get "all 2019 entries". We're passing the first filter item, but I've called it year for now because _right now_ it's explicitly set up to deal with years, with https://github.com/mozilla/foundation.mozilla.org/issues/2700 files to expand it to "whatever" instead of just year.

network-api/networkapi/wagtailpages/customblocks.py
Pomax: No it isn't: this is the super class, which absolutely needs to respect its own `year` value. `initial_year` only gets used if the subclass passes that along through its `super.get_context` call.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/profile_directory.html
```diff
@@ -1,11 +1,19 @@
...
+    <div class="filter-option">
+      <input type="radio" name="program_year" id="fellow-list-program_year-{{label}}" value="{{label}}" {% if forloop.first %}checked{% endif %}>
+      <label for="fellow-list-program_year-{{label}}">{{label}}</label>
+    </div>
     {% endfor %}
   </div>
 
   {% include "./profile_blocks.html" %}
+
+  {% if filters|length > 0 %}
```
alanmoo: Just making sure I understand this: this check is technically unnecessary, since "filters" is a required field on the backend of the component, correct? (If that's the case, I still understand why it's here, I just want to make sure I'm reading this right)

Pomax: correct

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/profile_directory.html
```diff
@@ -1,11 +1,19 @@
...
+  <input type="hidden" data-api-endpoint="{{ api_endpoint }}">
```
alanmoo: This is...an interesting approach

network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
```diff
@@ -0,0 +1,145 @@
...
+    ].filter(v => v).join('&');
+
+    // and then perform our API call using the Fetch API
+    return new Promise(
+      (resolve, reject) => {
+        return fetch(`${url}&${query}`)
+               // conver the resonse from JSON to real data
+               .then(response => response.json())
+               // all went well? resolve.
+               .then(obj => resolve(obj))
```
alanmoo: It's been a while: what's `resolve(obj)` do here?

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/profile_directory.html
Pomax: `input` and `meta` are fairly standard, but as a block, rather than page level, `meta` seemed wrong.

It's either that, or make the JS itself a template that django can render and.. ehhhh... no? =(

network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
Pomax: It's the "trigger `then` handling" function, with `reject` being the "trigger `catch` handling" function, which in this case it's the thing that lets https://github.com/mozilla/foundation.mozilla.org/pull/2701/files/bcbe62a6790f1f96304d02177168d8b875acb22b#diff-70b039d3f247fb0182821822739840d5R123 happen

alanmoo: Thank you, I forgot this was the definition of the `getData` function, not the function call.

network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
```diff
@@ -0,0 +1,148 @@
...
+  };
+
+  /**
+   * Load an array of profiles into the profile car container.
+   * We first map the profiles array, turning profile objects
+   * into templated HTML using the same HTML as we have in
+   * profile_blocks.html
+   */
+  function loadResults(profiles) {
+    let cards = profiles.map(profile => {
```
Pomax: Note that this block (L46 to L88) is effectively "blocked by whatever is the final form implemented in the fellow-style PR", since it needs to be the same HTLM code as used in the loop in `profile_blocks.html`

Pomax:
@kristina this should be ready for design review

kristinashu:
@Pomax the filter bar looks and works well when I click on your link. But when I sign into the cms and click preview (without changing anythig in the cms) all the other years disappear. Am I missing something?

image

alanmoo:
No, the draft different than the published page. This is working as expected.

kristinashu:
Ok yeah, adding additional years in seems to work!

kristinashu:
I'm not sure which PR I should ask this in but is there a reason the order of the Fellows is different than on the live site? Example for 2019 the PR shows Karen Kornbluh first where as the live site shows Alexander Morley (in alphabetical order).

Pomax:
Good catch, the filter's set to `ordering=-id` so it orders "most recent fellow first", but if we're filtering on year that should just be alphabetical. This would be the right PR for that.

Pomax:
@kristinashu switched over the ordering, should look correct now!

kristinashu:
Perfect!

kristinashu:
Sorry noticed one more thing. Does it matter that this PR and the #2685 break the year filter we currently have?
https://foundation-mofostaging-pr-2701.herokuapp.com/fellowships/directory/
https://foundation-mofostaging-pr-2685.herokuapp.com/fellowships/directory/

There will be some time in between pushing this component and switching out the CMS pages for the hardcoded ones.

Pomax:
I'm in the process of rebasing things so the filter header looks the way it's supposed to again.

Pomax:
@kristinashu oh man, that is an _excellent_ catch: no those should not suddenly reveal their radio buttons at all, let me make sure that stays the case.

Pomax:
@kristinashu https://foundation-mofostaging-pr-2701.herokuapp.com/fellowships/directory should look the way it looked again. Again, thanks for spotting that!

kristinashu:
Awesome! Is that the type of thing Percy would catch?

Pomax:
It is, although since this is the work that was preventing us from moving over the fellowship pages we'll have to talk through whether it's worth the screenshot allotment to add the old fellowship pages as part of the standard visual regression set set, or whether those pages won't be around for long enough for that to be worth it. I've filed https://github.com/mozilla/foundation.mozilla.org/issues/2703 for us to see which way we want to go.

Pomax:
@alanmoo @kristinashu I've rebased this PR on the styling PR that landed, so that the profile cards that get generated have the right html, so just for sanity this should get one more review pass.

source/js/components/fellow-list/fellow-list.scss
```diff
@@ -2,32 +2,53 @@
...
       &:hover {
         text-decoration: underline;
       }
+
+      &.active {
+        font-weight: 700;
+      }
+    }
+
+    /* Old page rules */
```
alanmoo: When can these be removed?

Issue #2725, opened 2019-02-26
Create new template: Primary Page banner, campaign page below


beccaklam:
For the Misinfo campaign we need an intermediate solution for its campaign landing page. We need a Primary Page that will have the option of adding a petition. The URL for this page should be the same as what we currently have for the landing page: https://foundation.mozilla.org/campaigns/eu-misinformation/

**List of requirements:**
- [x] Primary Page look and functionality
- [x] Can add a petition
~~- [ ] Can change banner image (related to #2661)~~

Redpen (first three designs):
https://redpen.io/tb154f87eb91f8fb07

Mobile:
![image](https://user-images.githubusercontent.com/20977376/53437254-11208900-39cb-11e9-9fe9-822848ea6468.png)

Desktop:
![image](https://user-images.githubusercontent.com/20977376/53437301-272e4980-39cb-11e9-91d2-e753491a1cca.png)

cc: @xmatthewx

kristinashu:
Assigned to @alanmoo to triage.

So basically we need a new template where the banner and subnav look like a primary page and below the subnav will look like campaign pages.

alanmoo:
I've updated the title of this to be more clear that it's a new template type entirely.

@kristinashu @beccaklam: Where else do you anticipate this template being used in the future (once it evolves)?

beccaklam:
I think the campaigns team would use it again, so in campaigns for sure. @kristinashu mentioned something about MozFest?

Pomax:
I assume that this is "not a primary page" for the purposes of being part of the main site navigation (like the intiatives/participate/etc primary pages)?

alanmoo:
Right, this is more along the lines of a sibling to the existing campaign/opportunity pages.

xmatthewx:
I see it as landing page for long-term initiatives that warrant a home. I could see it used for Machine Decision Making as we build up content and perspective in that space. I could see it replacing the internet health report parent site (while keeping reports themselves elsewhere). It could be used to establish the new home for mozfest.

kristinashu:
Naming could be something along the lines of "special landing page"? 

Yes to what Matthew said, I think something like CS Challenge using this too.

kristinashu:
This might be helpful https://redpen.io/hx70e9eb24e0534b72

xmatthewx:
Litmus test: could our team _and_ stakeholders imagine buying a vanity url to redirect to this page? If yes, then this template is the right choice. 

[...]

PR #2733, opened 2019-02-27
created a BanneredCampaign page type (campaign page with primary nav banner

Pomax:
Related PRs/issues #2725

This ended up being a model extension of the `PrimaryPage`, with a template extension on `CampaignPage`. It's a beautiful example of how MVC separation means the cleanest impementation doesn't have to make sense from "just M" or "or V" perspectives. That said, it would be nicer from a code design perspective if both the CampaignPage and BanneredCampaign inherited the cta functionality through a Mixin, but that would certainly overload the issue with "refactoring the campaign page".

### as main page under homepage

live: https://foundation-mofostaging-pr-2733.herokuapp.com/en/venmo-our-privacy-first/
edit: https://foundation-mofostaging-pr-2733.herokuapp.com/cms/pages/38/

(see the review apps channel for the admin login for review app 2733)

### as a campaign

live: https://foundation-mofostaging-pr-2733.herokuapp.com/en/campaigns/venmo-our-privacy-first/
edit: https://foundation-mofostaging-pr-2733.herokuapp.com/cms/pages/41/

tests and fake_data in followup #2742


Pomax:
I have no idea how this passed CI, looking at the heroku logs this is a mess of migration failures =P

alanmoo:
Long term, this will not continue to simply extend existing templates. Is that going to be easy enough to change with this approach?

Pomax:
V is entirely detached of M in Django, so: yes. However, the actual model extension might be harder to easily switch over in the future, so if the actual model fields are going to drastically change we should think that through now. Although I would argue that if they do, that would essentially be a completely different model anyway.

alanmoo:
The model may need a new field or something- looking at the direction we're likely headed, there's a possibility of offering the ability to pick a hero background color with [this template](https://redpen.io/xn8fcd1a5fa6365218)

@beccaklam Are there any other fields for customizing this page type you can easily anticipate needing?

Pomax:
Adding an eventual banner picker (as it were) shouldn't be a problem.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -486,6 +487,43 @@ def get_context(self, request):
...
+    cta = models.ForeignKey(
+        'Petition',
+        related_name='bcpage',
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        help_text='Choose existing or create new sign-up form'
+    )
+
+    content_panels = PrimaryPage.content_panels + [
```
Pomax: does it make sense to put this in between the parent panels, so that the cta is above the body field?

alanmoo: Just moving it up one notch? 
image

I don't think that's necessary, but a designer might have a thought.

Pomax: yeah, since the body can run pretty long. Then again, if we don't want CTAs on every single of these pages, it might make sense to keep it at the bottom instead?

alanmoo:
@beccaklam just an option that wasn't checked in the admin. They're both Zen mode now.

Pomax:
Seeing appveyor die in a way that suggests not all is well with the migration files in this PR, so there's probably a conflict and I'll try to get that fixed up.

Pomax:
There we go. @beccaklam was this good to go (with zen mode turned on?)

network-api/networkapi/wagtailpages/models.py
```diff
@@ -486,6 +487,43 @@ def get_context(self, request):
...
+    ]
+
+    parent_page_types = [
+        'HomePage',
+        'MiniSiteNameSpace',
+        'BanneredCampaignPage',
+    ]
+
+    subpage_types = [
+        'BanneredCampaignPage',
```
alanmoo: This may not be true for long, but it's fine to start with.

Pomax: I want to audit this in general, because we're mixing parent and subpage lists and it's really hard to control what goes where when parents get to say who their children are AND children get to say who their parents are allowed to be.

Pomax: https://github.com/mozilla/foundation.mozilla.org/issues/2734

Issue #2692, opened 2019-02-20
RIP V-Day on Feb 28 - hide category from nav


kristinashu:
At the end of the month we should de-valentines the site. Related to Hero Banner Swap #2705.

- [x] hide V-Day category from nav (keep url and it's contents https://foundation.mozilla.org/en/privacynotincluded/categories/valentines-day/)

Summary from convo below: add a flag to the category models for whether or not to show up in the nav menu at all.

cc @caltrider @alanmoo

alanmoo:
I was thinking that the category page should continue to work, we just remove it from the navigation.

Pomax:
Hm, can we easily do that? As far as I know the navigation gets autogenerated based on whether there are products tied to a particular category, so as long as we have products tied to it, the nav generation will pick up on that category existing and add it to the nav bar.

alanmoo:
Right. There is currently some logic that makes Valentine's day be the first category which we'd have to remove. The easiest thing that comes to mind is a "permanent" category name and a site-level redirect that hands visitors to the "Valentine's" category to the new one.

That being said, I'm only at like 💩 💩 out of 5 on this. So I'll yield to whatever Jen/Kristina decide, with a minimum requirement on my part of making the category URL redirect to the homepage.

caltrider:
I took a look back at the Wired article and it links directly to the Valentine's Day category page as does the Boing Boing article. I think the landing page is super interesting, so I'd be OK to redirect there. However, that might be confusing for folks who link from these articles (not sure how many people that will be, but I did want to take it into consideration). So, i guess I'm of the mind that if there is a pretty easy way to keep the Vday tab link while removing it from the nav, that would be best. If there's not an easy way, redirecting to the landing page is a fine option that I think works well.

Pomax:
In terms of quick/easy solutions, we could put in an instruction to "skip over building this nav entry if its text is exactly: <...> ", which shouldn't need more than just an update to the page template. 

Alternatively, and probably easier to support longer term, is adding a flag to the category models for whether or not to show up in the nav menu at all, which shouldn't be all that much more work, and is a little more future proof.

alanmoo:
>Alternatively, and probably easier to support longer term, is adding a flag to the category models for whether or not to show up in the nav menu at all, which shouldn't be all that much more work, and is a little more future proof.

@Pomax I had that thought earlier too. Since you came to it independently, let's go with that.

natalieworth:
[...]

PR #2741, opened 2019-02-28
add 'hidden' flag to PNI product categories

Pomax:
Tis ensures that categories can continue to exist, with products tagged as belonging to that category, which not showing up as a selectable category in the category nav bar.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/2692

This can be tested by going to the review app's PNI page, as well as the /cms route in a second tab, and toggling the `hidden` checkbox for buyers guide categories, saving, and reloading the PNI page to see the resulting change in the category nav bar.


network-api/networkapi/buyersguide/models.py
```diff
@@ -76,11 +76,17 @@ class BuyersGuideProductCategory(models.Model):
...
         blank=True
     )
+
     featured = models.BooleanField(
         default=False,
         help_text='Featured category will appear first on Buyer\'s Guide site nav'
     )
 
+    hidden = models.BooleanField(
+        default=False,
```
Pomax: due to the default being `False`, we don't need an explicit migration function that ensures all pre-existing categories get the correct value assigned.

Issue #2661, opened 2019-02-14
Make header image on primary pages cms editable


kristinashu:
We would like to be able to update the background image on individual primary pages. 

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1068783/52802652-3ad5c980-3035-11e9-94c6-cff63450a5bb.png">

mmmavis:
I think we can keep the existing banner (like the one in the screecap above) as fallback image and add an optional banner image field to CMS so people can override default with the image they want.

alanmoo:
Suggestion: for the template implementation, use a `<style>` tag to override the CSS applied to `#hero::before`'s background.

kristinashu:
Not sure if this is helpful but it is currently possible to update the hero on [/initiatives ](https://foundation.mozilla.org/en/initiatives/) in the cms.

<img width="1439" alt="image" src="https://user-images.githubusercontent.com/1068783/53269339-7576dc00-369d-11e9-97ac-e8755e2020ed.png">

alanmoo:
Only kind of. They use different markup to achieve the same effect. I'd like to avoid rewriting markup just for this ticket because that could be a rabbit hole.

PR #2765, opened 2019-03-06
Make the primary page banner cms-controllable

Pomax:
Initial stab: this makes the primary page banner a real image element rather than a CSS pseudo-element. From what I can tell this still looks the same, but that needs confirmation.

Related PRs/issues https://github.com/mozilla/foundation.mozilla.org/issues/2661


alanmoo:
Can Percy tell us if it's the same?

Pomax:
I've tested this by comparing the PR's `about` page with the https://foundation.mozilla.org/en/about page at various breakpoint sizes, and there is marginal different in image scaling at the min-width breakpoint, but nothing that would seem to be so different as to qualify as "needs to be redone".

If you agree, then I can add the actual background-from-CMS functionality on top of this.

Pomax:
@alanmoo it'll at least be able to tell us what's different, let me add an "about page" screenshot to the percy battery

Pomax:
@alanmoo added a test, but because it's a first-time test it won't help us in this specific case, and we still need the visual confirmation that things look "the same enough".

kristinashu:
Sorry but I'm not too sure what and how to test...

1. About https://foundation.mozilla.org/en/about/
Looks good but I don't see where to update the banner image in the cms.

2. Initiatives & Participate https://foundation-mofostaging-pr-2765.herokuapp.com/en/initiatives/ & https://foundation-mofostaging-pr-2765.herokuapp.com/en/participate/
Should not be affected by this update, their banners should be unique (do they look different because of the review app?).

3. Where can I test the new multi-page campaign page https://foundation.mozilla.org/cms/pages/331/view_draft/?

Pomax:
@kristina there is no new functionality yet in this pr, all it does is change the html+css so that the banner is still hardcoded, just not as a CSS `:before` pseudo element, but as a real `` element that we can swap the image source of in immediate follow-up.

1. it should look the same as before. if it doesn't, that's bad.
2. initiatives being empty is due to the load_fake_data process for the review app.  the participate page is definitely also the load_fake_data process but looks like a completely different page, so that's probably use never having updated what the content for that should be. I've filed that as https://github.com/mozilla/foundation.mozilla.org/issues/2767
3. not in this PR

kristinashu:
We might need to clarify what we mean by "primary page banner" because it is confusing. It's anything with this style banner is considered "primary page banner" (even if it isn't in the main nav).

https://redpen.io/hx70e9eb24e0534b72

image

Pomax:
The definition I'm working with is the one that maps to our `class PrimaryPage` model, which defines the data for anything that is page of type "Primary Page" in the CMS, which includes the following top level pages:

- the about page
- the fellowships page
- internet heath
- mozilla awards

image

kristinashu:
Ah got it! Then this all looks and sounds good to me. About page banner looks good and I'll trust you two that there aren't any unexpected side effects.

Pomax:
As far as I can tell, all those pages marked as "primary page" have that same header, so I'm pretty sure we're talking about the same thing, but this PR is not "the work to replace the banner image", this is scaffolding work that needs to happen _before_ the banner image can be made replacable, so all those pages should still look the same as before, even though I had to change the underlying code to make it possible for us to start writing code that lets the banner image be CMS-controllable.

PR #2766, opened 2019-03-06
Prevent duplicate image renditions using AWS_S3_FILE_OVERWRITE

### LINKED ISSUE #2761: CMS images caching
From @sabrinang:

Regarding images in the CMS / thumbnail issue:

I see the same image for all the images but in the editing view I have uploaded different PNGs but the live link is displaying one thumbnail or multiples

See pages: https://foundation.mozilla.org/en/docs/brand/design/templates/ or https://foundation.mozilla.org/en/docs/brand/design/logos/ 

Image is wrong on prod:
image

Image is correct in editor:
image

cc @beccaklam because she was having similar issue on staging

cadecairos:
Closes #2761 

## Checklist

**Documentation:**
- [x] Is my code documented?

## Testing:
You'll need two different images, both with the same 17 character (at least) file names. Add one as an image on a page, then check that it's working. Then, change the image to the second. The image should update after publishing


cadecairos:
So I tested this out on the review app using https://foundation-mofostaging-pr-2766.herokuapp.com/en/opportunity/single-page/

Some notes:

- It appears to do what I expected, but we need to be conscious of the fact that this means ANY file we upload to S3 from wagtail is capable of overwriting existing content in the bucket.
- This doesn't solve the problem that images are supposed to be cached on the edge of CloudFront, so on Prod it will take time for updates to propagate, unless we invalidate the cached image.

Issue #1437, opened 2018-04-20
[Wagtail Component] Remove Heading Component


kristinashu:
~If we add H1 to the Paragraph styles #1389 then should we remove the Heading component since a Paragraph component can do everything it can do and more!~

Having fewer components will be easier for people to use and as is Heading component only allows for H1 which isn't useful.

Updated: Only things this ticket needs to do is remove the 'Heading' component.

gvn:
I'm fine either way. If we keep the heading we should probably allow people to choose the heading level, otherwise it's not that useful.

kristinashu:
For simplicity, I think we should remove it. We could always rename `Paragraph` to `Text` to make it more generic.

Pomax:
@kristina this kind of fell off the roadmap, do we still want to remove this?

youriwims:
@kristinashu have we settled on renaming the `Paragraph` component? If so, I'll open a separate issue for that.

youriwims:
What's the best way to do a block inventory search for `Heading`?

/cc: @mmmavis @Pomax

kristinashu:
I don't think renaming it will be necessary, let's just keep the name `Paragraph`. Also don't think we need to add H1 to the Paragraph component since the Title/Header field using H1 and we should restrict it's use.

So the only things this ticket needs to do is remove the 'Heading' component. I don't see any instances of this in the Block Inventory search but let me know if you see any and I'll update them to Paragraphs.

Thank you!

youriwims:
From what I can tell, it looks like `Heading` is still active on these pages:

- [ ] https://foundation.mozilla.org/en/news/
- [ ] https://foundation.mozilla.org/en/style-guide/
- [ ] https://foundation.mozilla.org/cms/pages/169/view_draft/ (Draft page...may not be important)

I'm going to do a bit more digging to make sure there's no more instances of this component.

kristinashu:
I've just deleted that old draft page. The first two on your list aren't cms pages so I don't think they are relevant, but @mmmavis any thoughts?

mmmavis:
> https://foundation.mozilla.org/en/news/
Headings in the article list shouldn't be affected as the list is generated via JS.

> https://foundation.mozilla.org/en/style-guide/
This is a hardcoded static page. Things here shouldn't be affected.

I'm not sure what the proper way is to remove a component on CMS without breaking pages that are already using them. @Pomax can you chime in here?

youriwims:
[...]

PR #2788, opened 2019-03-11
Wrap the wagtail RichText __html__ function to effect heading ids

### LINKED ISSUE #2743: Generate id attribute for headings in streamfields (anchor links)
Problem: In order to be able to link directly to sections of our pages, we need headings to have an ID attribute.

Suggested solution: Use JS to create IDs based on the text of `h1` and `h2` elements (at least) of streamfield content.

This will be used for the EU-misinfo campaign

cc @beccaklam

Pomax:
Closes #2743 by getting creative: we can't override the `richtext` filter, because while that's a necessary filter tag in any template we wring, including block template, it does _not_ get used for pure rich text streamfield content. So, the only way to ensure that _every_ rich text field has linkable id attributes for its headings is to take the RichtText class's html serialization function, and wrap it in something that "calls that, then forwards the result on after performing a replacement".

e.g. https://foundation-mofostaging-pr-2788.herokuapp.com/en/campaigns/multi-page/pass-vote-wonder/#pick-all-beautiful-science-cold

So this works, and it works_really_ well, without actually changing how Wagtail works, in a way that _should_ be pretty future proof by virtue of technically not changing any of wagtail's internal class code...

...much

One thing that is a problem, however, is the fact that our CSS for the page top nav is now very clearly wonky: instead of the top nav being a real element, it's an absolutely positioned layer "on top of" actual page content, so navigation to any `#heading-ref-here` means the page correctly scrolls that to the top of the window, where it gets promptly hidden by the nav layer.

So we might need some extra CSS that puts `padding-top` on the body while the nav menu is visible? @mmmavis any ideas here by any chance?


Pomax:
Bonus points to percy for catching the font suddenly going bold, too.

alanmoo:
Can we change the markup we're injecting so that it accounts for the nav?

Similar to [this approach](https://www.caktusgroup.com/blog/2017/10/23/css-tip-fixed-headers-and-section-anchors/)

Pomax:
Added in a temporary bit of HTML with a relocated anchor, but the real solution is using `scroll-padding` once firefox adds support for it. At least two years late.

source/sass/cms.scss
```diff
@@ -1,6 +1,26 @@
 // Resets / sensible defaults for CMS-managed content
 
 .cms {
+  h1,
```
alanmoo: These only target elements inside `.cms` tags, but the markup replacement seems like it could catch elements outside that. Is that right? What happens then?

source/sass/cms.scss
```diff
@@ -1,6 +1,26 @@
...
+    position: relative;
+
+    a {
+      text-decoration: inherit;
+      color: inherit;
+      font: inherit;
+
+      &[id] {
+        position: absolute;
+        top: -85px;
```
mmmavis: This fixed offset might not work responsively. (try it on 874px wide screen, for example)

source/sass/cms.scss
Pomax: not right: this only affects streamfield RichtText blocks, and streamfield content always ends up in a `<div class="cms">` given the templates we've written.

mmmavis:
I had similar issue with anchors on Buyer's Guide before and I had to come up with a JS solution to make sure things work responsively: 

https://github.com/mozilla/foundation.mozilla.org/blob/3d90047ca998c3e04ad75da119adc6c562833030/source/js/buyers-guide/components/creep-vote/creep-vote.jsx#L208-L218

source/sass/cms.scss
Pomax: happy to work in suggestions, is there any way for us to get the "correct" size at specific breakpoints? (just a bunch of media queries would be fine for instance)

source/sass/cms.scss
alanmoo: I guess it's that latter clause that concerns me slightly. But not enough to block this?

Pomax:
I don't want to have to resort to JS solutions if there is a pure markup solution. Ideally we use `scroll-padding`, but FF is dragging its feet so that only works in Chrome/Opera/Safari atm.

source/sass/cms.scss
Pomax: Not sure I understand the concern: we wrote templates in a way that streamfield blocks are _always_ in a `.cms` class, and this CSS has been written to _only_ apply to content in a `.cms` class... so how would this CSS end up applying to elements outside of our cms content?

alanmoo: Right, we _should_ be doing that but that's not enforced. But again, outside the scope of this issue.

source/sass/cms.scss
alanmoo: Two things to consider: 

1.) The `DEV HELP` item makes this worse at more resolutions than it is in reality
2.) The fact that the nav needs to wrap in the first place seems like a bug, possibly a result of a switch to the items in the nav that we made a few months(?) ago.

Between those two, I'd argue that this isn't an issue to solve here, and we should instead file a follow up about the banner itself looking wonky at just above 768px.

source/sass/cms.scss
Pomax: Outside the scope, but I still don't understand what you're asking. Can this affect content on our pages outside of CMS streamfields? No.

Might it in the future? Extremely unlikely, we'd have to abandon all common sense about css classes.

source/sass/cms.scss
Pomax: That sounds quite reasonable: filed https://github.com/mozilla/foundation.mozilla.org/issues/2823

alanmoo: @mmmavis this was your objection- do you agree with me?

Pomax:
Hm, I'll see if I can add a test that verifies that the output of `RichText.__html__()` is plain string data, as anything else would greatly upset `re.sub`

Pomax:
@cadecairos added a typing test to ensure that the output of the original `__html__` function is a string, before running it through re.sub

source/sass/cms.scss
```diff
@@ -1,6 +1,26 @@
...
+  h5,
+  h6 {
+    position: relative;
+
+    a {
+      text-decoration: inherit;
+      color: inherit;
+      font: inherit;
+
+      &[id] {
```
mmmavis: update the selector here so it has the classname

source/sass/cms.scss
mmmavis: yup yup let's do that in a follow-up

network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
```diff
@@ -0,0 +1,67 @@
...
+
+        <h1><a id="some-text"></a><a href="#some-text">some text</a></h1>
+
+    where the id attribute value is generated by running
+    the heading text through Django's slugify() function.
+    """
+    n = match.group(1)
+    text_content = match.group(2)
+    id = slugify(text_content)
+    return f'<h{n}><a id="{id}""></a><a href="#{id}">{text_content}</a></h{n}>'
```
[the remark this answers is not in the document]
Pomax: Hmm, I don't think adding a class adds anything that `.cms { h1...h6 { a[id]:not([href]) { ... }}}` doesn't already catch?

Pomax:
Putting a link inside a heading feels like a "you should not do this" policy case that I'm okay with completely breaking, but let's get @alanmoo and @kristinashu's input here, too.

network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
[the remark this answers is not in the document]
alanmoo: I'm confused about the two anchor elements here. Can't we put the id on the heading element directly?

alanmoo:
We don't actually need the heading to be a link, so I'd be ok with us not inserting it; I feel like most of our pages aren't so long that this is _that_ useful.

Pomax:
you need links for staff and editors to actually be able to copy heading links, so in this case I would argue that putting "real" links inside headings is questionable at best, and is an edge case we should be able to safely ignore.

network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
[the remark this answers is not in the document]
Pomax: in terms of discoverability a class is better. Adding one.

kristinashu:
Not sure I totally follow but links in headings are very common on our site. Example https://foundation.mozilla.org/en/initiatives/internet-health-research-archive/ 
I want to confirm this will still be possible? It would be unfeasible to update all instances of this.

Also, one example of where the jump links will be used is here https://foundation.mozilla.org/en/fellowships/faq/ Staff would like to be able to link directly to a specific FAQ question. I assume we would need to update those questions to be styled as headings but that's a quick update.

Pomax:
@kristina it looks like all of those are h4 (elements, not css), would it be safe to say that h1 and h2 never have links in them, as they are "our content" headings?

Pomax:
@kristinashu it looks like all of those are h4 (elements, not css), would it be safe to say that h1 through h3 never have links in them, as they are "our content" headings?

If so, I can update the PR to only rewrite h1-h3 headings.

kristinashu:
Probably isn't even necessary to add a jump link to h1 since once https://github.com/mozilla/foundation.mozilla.org/issues/1437 is done they will only appear via the title/heading at the top of the page. But they definitely won't have a url link since it isn't possible in that field.

h2 and h3 are trickier. I suspect they are less likely to have url links but would need to look into it more. But even if we say not to use them as links I'm not sure how we can enforce that since the Paragraph component allows it.

Pomax:
For now I've updated the PR to forego linkwrapping, and only adding the fragment-id anchor. That makes it a little hard to "find" the link to use, but won't interfere with any preexisting headings with links in them.

I've also restricted this to only h1/h2/h3 for now, and made sure to strip any html tags before creating the id slug, so that things like italics/bold/etc don't cause problems either.

network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
```diff
@@ -0,0 +1,74 @@
...
+    This replacement will only kick in for h1/h2/h3.
+    """
+    n = match.group(1)
+    text_content = match.group(2)
+
+    if int(n) > 3:
+        return text_content
+
+    id = slugify(strip_tags(text_content))
+    return f'<h{n}><a class="fragment-id" id="{id}"></a>{text_content}</h{n}>'
```
alanmoo: I'm still confused to the need for the anchor element here. Why can't we put the class and id on the heading element itself?

Pomax: because the anchor gets offset by 85px above the heading. Otherwise navigating to it puts the heading under the sticky nav bar.

alanmoo: Duh. Right, it's been a while since we opened this PR. Since it's non-obvious, can you put that in a comment between 46 and 47?

Pomax:
okay, that is a disconcerting percy error. Why did the mozilla logo disappear O_o

network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
Pomax: I've update the code comments.

PR #2791, opened 2019-03-12
Make BanneredCampaignPage usable everywhere

alanmoo:
So we can use it as the homepage of the Mozfest site


Pomax:
R+ but this feels wrong somehow. If this is the role we want the BanneredCampaignPage to play, we should really give it a much better name.

alanmoo:
We need to do some renaming all around, but this is already in use. You're not wrong, though.

PR #2801, opened 2019-03-13
rogue django {{ ... }} instead of js ${ ... }

### LINKED ISSUE #2798: People Name Links on People Cards (Fellowships)
On the fellowships page, https://foundation.mozilla.org/en/fellowships/meet-our-fellows/, when you click into another cohort year and click on the name of the fellow the URL `https://www.mozillapulse.org/profile/{{ profile.profile_id }}` leads me to a _profile not found_
image

When I first go into the fellowships page the 2019 cohort name links work for me, it is when I click into another year and go back it becomes broken.

/cc @youriwims @pomax @kristinashu

Pomax:
Closes #2798

Classic porting mistake: all-but-one django template statements had been converted to js template literal... all but one.

STR: load fellowship page, note initial data works fine, click "2019", note that on the live site that has `{{ profile.profile_id }}` in the URL, whereas this fix .. err.. fixes that.


alanmoo:
Flagging as an example of the type of Percy page we need- something that auto-generates a profile directory component so this is easier to test.

PR #2803, opened 2019-03-13
Profile Directory Google Analytics

### LINKED ISSUE #2533: Fellowships: Add analytics to directory and profile component
We will soon build a new directory component that leverages an update pulse profile component. Let's wire it up for GA to measure our success with engaging visitors and extending fellow promotion.

We will want to attach this to:
 - profile
   - tap on photo or name --> pulse profile
   - tap on social icon --> linkedin or twitter
 - directory
   - tap on directory tab/filter --> year by year view

Component | Category | Action | Label
:-- | :-- | :-- | :--
Profile | profiles | profile tap | [page title] - [name of person] - [destination]
Directory | profiles | directory filter | [page title] - [filter value]

Example | GA Event
:-- | :-- 
2015 | label: "profiles", action: "directory filter", label: "fellowship directory - 2015"
Renee twitter | label: "profiles", action: "profile tap", label: "fellowship directory - renee - twitter"
Meghan pulse | label: "profiles", action: "profile tap", label: "fellowship directory - meghan - pulse profile"

youriwims:
Closes #2533

[Test Page](https://foundation-mofostaging-pr-2803.herokuapp.com/en/opportunity/profile-ga-test/)

cc: @xmatthewx


youriwims:
The alert works now on the initial pageview, but if I click on another year and test the event again (on both headshot + name), it doesn't work. Do you know why that is @Pomax ?

youriwims:
@Pomax , I can't tell exactly what was changed in the 'small rework' to resolve the file conflicts above. Please advise—thanks!

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
       );
     });
 
+    function profileCardSocialAnalytics(socialTwitter, socialLinkedIn) {
+      if (socialTwitter) {
+        socialTwitter.addEventListener(`click`, evt => {
+          evt.preventDefault();
```
Pomax: we don't want this line here anymore, because it turns off the link

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
...
+            action: `profile tap`,
+            label: `${document.title} ${name} twitter`,
+            transport: `beacon`
+          });
+        });
+      }
+
+      if (socialLinkedIn) {
+        socialLinkedIn.addEventListener(`click`, evt => {
+          evt.preventDefault();
```
Pomax: we don't want this line here anymore either, because it turns off the link

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
...
+    // And start listening for profile filter events,
+    // in case profile cards get updated.
+    document.addEventListener(`profiles:list-updated`, evt => {
+      evt.preventDefault();
+
+      //Profile Directory Filter Analytics
+      const filters = document.querySelectorAll(
+        `.profile-directory .fellowships-directory-filter .filter-option button`
+      );
+      let labels = Array.from(filters);
```
Pomax: we should be able to say `filters.forEach` directly here, since querySelectorAll returns an array-like that comes with forEach support (thankfully)

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
...
+      bindProfileCardAnalytics(profileCards);
+    }
+
+    // And start listening for profile filter events,
+    // in case profile cards get updated.
+    document.addEventListener(`profiles:list-updated`, evt => {
+      evt.preventDefault();
+
+      //Profile Directory Filter Analytics
+      const filters = document.querySelectorAll(
```
Pomax: We don't need this to happen "whenever `profiles:list-updated` fires", we just need to run it once, as the filter buttons never get swapped out.

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
...
     });
 
+    function profileCardSocialAnalytics(socialTwitter, socialLinkedIn) {
+      if (socialTwitter) {
+        socialTwitter.addEventListener(`click`, evt => {
+          evt.preventDefault();
+          ReactGA.event({
+            category: `profiles`,
+            action: `profile tap`,
+            label: `${document.title} ${name} twitter`,
```
Pomax: where is `name` coming from?

source/js/main.js
```diff
@@ -379,6 +381,95 @@ let main = {
...
+        });
+      }
+
+      if (socialLinkedIn) {
+        socialLinkedIn.addEventListener(`click`, evt => {
+          evt.preventDefault();
+          ReactGA.event({
+            category: `profiles`,
+            action: `profile tap`,
+            label: `${document.title} ${name} linkedin`,
```
Pomax: where is `name` coming from?

source/js/main.js
```diff
@@ -237,8 +237,10 @@ let main = {
     }
 
     // Embed additional instances of the Join Us box that don't need an API exposed (eg: Homepage)
-    if (document.querySelectorAll(`.join-us`)) {
-      var elements = Array.from(document.querySelectorAll(`.join-us`));
+    if (document.querySelectorAll(`.join-us:not(#join-us)`)) {
+      var elements = Array.from(
```
Pomax: We should be able to do this without Array "casting".

youriwims: what do you mean?

source/js/main.js
youriwims: removed it

source/js/main.js
youriwims: removed

source/js/main.js
youriwims: ✔️

source/js/main.js
youriwims: ✅

source/js/main.js
youriwims: valid question.let me fix that.

source/js/main.js
youriwims: fixed!

source/js/main.js
```diff
@@ -379,6 +379,96 @@ let main = {
       );
     });
 
+    function profileDirectoryFilterAnalytics() {
+      const filters = document.querySelectorAll(
+        `.profile-directory .fellowships-directory-filter .filter-option button`
+      );
+      filters.forEach(filter => {
+        filter.addEventListener("click", () => {
+          let year = filter.textContent;
```
Pomax: Just to be sure, let's `trim()` this, too

source/js/main.js
```diff
@@ -379,6 +379,96 @@ let main = {
       );
     });
 
+    function profileDirectoryFilterAnalytics() {
```
Pomax: we don't ever call this code more than once, so in this case let's not wrap this in a function.

Issue #2805, opened 2019-03-13
Add prettier as CSS formatter


Pomax:
With prettier in control of our JS/JSX, it'll be nice to also make it responsible for CSS/SCSS formatting. Just so we don't ever have to think about that again, either.

PR #2807, opened 2019-03-13
prettier for css/scss

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2805

All updates to (s)css files came from loading each file with errors flagged by prettier, and making prettier reformat them.

Note that this preserves `stylelint` as linter, since that can catch things that a pure formatter can't.


.stylelintrc
```diff
@@ -1,9 +1,6 @@
 {
-  "extends": "stylelint-config-standard",
+  "plugins": ["stylelint-prettier"],
   "rules": {
-    "at-rule-no-unknown": null,
-    "font-family-no-missing-generic-family-keyword": null,
-    "no-descending-specificity": null,
-    "color-hex-length": "long"
```
mmmavis: I'd like to keep this rule `"color-hex-length": "long"` as it will really prevent us from adding custom colors to the codebase in the future.

https://github.com/prettier/stylelint-prettier#options says code editors will ignore custom configs but as least `npm test` and Travis will still respect them.

Pomax: makes sense, although I'd also strongly advocate that we start replacing any "raw color value" in our CSS (outside of the color variable definitions) with named variables. In a different issue of course.

Pomax:
just noticed this forgot to also include the networkapi dir (which has a few custom .css files) so I'll be updating the branch to include those, too

Issue #2808, opened 2019-03-13
add more pages for percy to check


Pomax:
we're currently running on 5 pages, and it would be good to add additional pages that can act as class representatives for their CMS page type:

- homepage
- about page
- multi-page campaign
- PNI homepage
- PNI category page

Pomax:
Important to add: a profile directory component with probably three entries so we get both "the look" and "the reflow" across multiple columns.

Issue #2767, opened 2019-03-06
update the fake "Participate" page, because it looks very different these days


Pomax:
current load_fake_data page:

![image](https://user-images.githubusercontent.com/177243/53906631-4dbe3700-4000-11e9-8055-bd5f0d24881c.png)

Actual page on foundation site:

![image](https://user-images.githubusercontent.com/177243/53906658-57479f00-4000-11e9-871a-3cc77279e15f.png)

PR #2809, opened 2019-03-13
added several representative pages to catch page-type level visual regression

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2808

**Tests**
- [x] the code I'm adding _is_ tests


Pomax:
In this case I'd say it's asking for the correct URL, but that URL is generating the wrong page, so as a test it'll "work" but it'll flag a change when we follow up by fixing the participate page?

Pomax:
marking as blocked on https://github.com/mozilla/foundation.mozilla.org/issues/2767 for the moment

Pomax:
unblocked, but waiting for master to fully build and deploy to staging, so that percy has a chance to regenerate its base images.

Pomax:
@mmmavis @alanmoo this is ready for review agian.

The percy "fails" are the newly added screens, which as far as I can tell all look right.

alanmoo:
How often is the fake data going to flag Percy errors? All the time or when generators change?

Pomax:
every time we change things that affect random sequence generation. The main things that'll affect that is adding/removing fields from page models, and adding/removing streamfield blocks. In both cases, the number of possible things the faker can generate changes, and so the sequence for each faker call will change.

We can mitigate this to some degree by reseeding the faker at every step but I'd like to do that in followup.

PR #2832, opened 2019-03-15
fixed profile directory spacing

### LINKED ISSUE #2707: Fellowships directory spacing between cards
Spacing between individual profile cards is pretty inconsistent. They should be evenly spaced our between each other.

Screen cap from prod on Desktop:
image

Screen cap from prod on Mobile
image

youriwims:
Closes #2707 

Test fixed component [here](https://foundation-mofostaging-pr-2832.herokuapp.com/en/opportunity/directory-spacing-test/).


youriwims:
Agreed, I wanted to have a conversation about this and was hoping to open another ticket to get the conversation started about limiting characters/words for the bios we're pulling in. The goal is to have equal spacing between cards.

@mmmavis _Pinging you about possible solutions!_

kristinashu:
> I'm not sure if the goal is trying to make sure spacing between each card row is consistent or to make all profile cards have equal height. Happy to hop on a call to discuss about this.

Goals is to make spacing between cards consistent. So ideally, the lowest point of one card (whether it's the image or the text) will have the same space to the top of the next card.

Also happy to chat!

mmmavis:
@youriwims 

(Equal card height is not necessary, but if you wanna chat more, @kristinashu and I are available.)

For this PR let's focus on getting equal spacing between rows to work. I'm not super familiar with CSS grid so I can't give you code suggestion right off the bat. Let me know if you have any questions though. Maybe we can work through this together.

kristinashu:
Yes!!! Design looks perfect! 

https://foundation-mofostaging-pr-2832.herokuapp.com/en/opportunity/directory-spacing-test/

youriwims:
@mmmavis what's the general rule-of-thumb when using rem vs px?

mmmavis:
@youriwims for border width we wanna use `px` since it's usually under 2px and using `px` makes the value easy to understand. For other things... I don't think we are being very consistent in this codebase. For me I tend to use `rem` for spacing related style and `px` for dimension related stuff.

youriwims:
Okay, thanks @mmmavis! 😊

PR #2859, opened 2019-03-20
JS updates

alanmoo:
I ran `npm update`


alanmoo:
@patjouk This looks good, but I can't R+

PR #2860, opened 2019-03-20
remove news.featured

Pomax:
remove featured flag use across the codebase, as prerequisite to removing the field itself from news.models in followup.

Part 1 of https://github.com/mozilla/foundation.mozilla.org/issues/721
Unblocks https://github.com/mozilla/foundation.mozilla.org/pull/2861


alanmoo:
That's an odd Travis failure

alanmoo:
Also, is this dramatic Percy change the normal "random data seed generation" thing? If so, this is getting to be...not useful.

Pomax:
Errrr... I'll hard-coding a PNI product, because we shouldn't guess which one the faker invents for us >_>

alanmoo:
Ah, so that's still to be done on this PR?

cypress/integration/endpoint-tests.js
```diff
@@ -104,7 +104,7 @@ describe(`Visual regression testing for foundation.mozilla.org`, () => {
   });
 
   it(`PNI product page`, function() {
-    cy.visit(`/en/privacynotincluded/products/home-great/`);
+    cy.visit(`/en/privacynotincluded/products/near-reality/`);
```
alanmoo: I see you changed this after your comment, but I assume there's another commit to come that manually adds this product to the fake database?

Pomax:
blocked until the reseed PR (#2897) lands.

cypress/integration/endpoint-tests.js
Pomax: yeah, handled in #2897 after which I'll rebase this.

Issue #1099, opened 2018-02-24
Figure out how to remove the `unsafe-inline` CSP rule for scripts


gideonthomas:
We either set it for django admin or for mezzanine, not sure. But this CSP rule is also being set for user facing pages (e.g. petitions, opportunities, etc.) which isn't great. Let's try to figure out if we can remove it entirely or at least isolate them to the features that actually need it.

**Deliverable 1**: Research output of what happens when we remove it and list of pages that it is required for.
**Deliverable 2**: Implementation - Either we can't do this or we remove it entirely or we isolate it to pages that need it.

Pomax:
This still seems relevant - wagtail needed it for a while because of its image browser, but as far as I know that got fixed, so we should revisit this and see if we can remove these CSP rules now.

harshs14:
hi @Pomax, I am new here, can I take this up if no one is working on it and can u explain me on how to solve this

gideonthomas:
@Pomax is this still something that's worth pursuing or give up on this?

Pomax:
Wagtail apparently still relies on this, so.... closing with a sigh.

PR #2862, opened 2019-03-20
[WIP] disallow unsafe-inline scripts

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/1099 by removing the `unsafe-inline` CSP keyword for scripts. Wagtail itself already removed reliance on that, and so the only thing I could find that violated it was our polyshim check in the buyers guide base template, which was trivially moved to its own static file and src-linked rather than inlined.

### EDIT

turns out there is some script injection for the language picker code that we added to the CMS, so this PR is still WIP.

### DIFF
--- a/env.default
+++ b/env.default
@@ -46,7 +46,7 @@ CRM_PETITION_SQS_QUEUE_URL=""
 # CSP config
 
 CSP_DEFAULT_SRC="'none'"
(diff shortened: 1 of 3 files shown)


Pomax:
even on wagtail 2.4, we still can't remove this due to the admin interface using bare code in <script> tags.

PR #2868, opened 2019-03-21
[Wagtail Component Removal] Heading Part 2 of 2

### LINKED ISSUE #1437: [Wagtail Component] Remove Heading Component

youriwims:
Closes #1437 
[Part 1 of 2](https://github.com/mozilla/foundation.mozilla.org/issues/1437#issuecomment-475364033)

## Checklist

**Changes in Models:**
- [x] If my changes are not backward-compatible, did I schedule a deploy with the rest of the team?

cc: @mmmavis @Pomax @kristinashu


Pomax:
let's get it in there.

PR #2869, opened 2019-03-21
Add static maintenance page + auto upload to S3

### LINKED ISSUE #1742: Custom Heroku error page
If there's ever a platform outage or database outage, we can serve a static page of our own design, that's hosted on a highly reliable service (in our case, S3). Lets create a stand alone error page for the foundation site.

Ideally, the page can be a part of this project, and one of the build steps on staging and production is to sync it to a pre-configured place on S3.

I can set things up, but I'm pretty sure we'll need some design + front end skills so that the page is consistent with the rest of the site, and that messaging is good.

cadecairos:
Closes #1742 

This PR adds a `maintenance` folder to the root of the project which contains an html file and supporting resources that are going to be uploaded to an AWS S3 bucket every time a build of the master branch successfully completes.

The travis configuration and S3 bucket are ready to go. you can check the static page by checking out this branch and opening the html file in your web browser. it has no external dependencies, and should load without an issue. It's designed to look like the actual site, and has functioning links in the header and footer.


cadecairos:
image

.travis.yml
```diff
@@ -1,10 +1,8 @@
 language: python
 python:
-  - '3.6'
-
+- '3.6'
```
patjouk: Is it a voluntary change to remove all indents? I kind of want to keep them for readability reason.

.travis.yml
```diff
@@ -27,48 +25,57 @@ env:
...
+    - npm run build
+    - pipenv run python network-api/manage.py collectstatic --no-input --verbosity
+      0
+    - pipenv run python network-api/manage.py migrate --no-input
+    - pipenv run python network-api/manage.py block_inventory
+    - pipenv run python network-api/manage.py sync_page_translation_fields
+    - pipenv run python network-api/manage.py update_translation_fields
+    - pipenv run python network-api/manage.py load_fake_data
+    script: npm run percy
+deploy:
```
patjouk: Travis docs mention something about adding the AWS region for S3 hosted site: https://docs.travis-ci.com/user/deployment/s3/#deploy-to-a-s3-hosted-website
Do we need it?

cadecairos: yeah lets be explicit about region

.travis.yml
cadecairos: oh. hmm, I used an auto-formatter for yaml, and this was the output. I'll try to configure it to have indents

cadecairos: should be fixed now.

PR #2865, opened 2019-03-21
Remove Fellowship Directory & Related Files

youriwims:
Closes #2812 
Related PRs/issues #2849

PR #2870, opened 2019-03-21
Remove people.jsx, people.scs, and related code

### LINKED ISSUE #2686: remove people.jsx, refactor people.scss
We're no longer using the people page, but some of its SCSS is used by other pages/components, so we should pull those two things apart, remove the people page code itself, and make sure we relocate the still-used styling.

youriwims:
Closes #2686 

_Feel free to let me know if I've missed anything or removed something in error._

cc: @mmmavis @Pomax


youriwims:
This pr is blocked by the [Fellows files removal](https://github.com/mozilla/foundation.mozilla.org/pull/2865). It'd be best to land that one first as to avoid any weird conflict errors.

youriwims:
@mmmavis & @Pomax I'm setting you both as reviewers for this since you are both more familiar with these files and what they affect/are affected by more than I am. Please let me know if there is something I overlooked or may have missed. Thanks! :)

source/js/components/people/people.scss
```diff
@@ -1,91 +0,0 @@
...
-        content: "";
-        margin-right: 5px;
-        background-size: contain;
-        background-repeat: no-repeat;
-      }
-    }
-
-    .meta-block-item-role {
-      &::before {
-        background-image: url("/_images/glyphs/mozilla-icon.svg");
```
mmmavis: This image file should be removed as well as it's not being used anymore. I would search the file name (e.g., `mozilla-icon`) in the codebase to decide whether an image file can be removed or not.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -788,7 +788,6 @@ class ParticipatePage2(PrimaryPage):
 
 class PeoplePage(PrimaryPage):
```
mmmavis: If we no longer needs this `PeoplePage` we should fully remove this class.

Pomax: But only as a separate PR, because is a potentially problematic migration.

source/js/main.js
```diff
@@ -7,7 +7,6 @@ import Analytics from "./analytics.js";
 
 import JoinUs from "./components/join/join.jsx";
 import Petition from "./components/petition/petition.jsx";
-import People from "./components/people/people.jsx";
```
Pomax: People.jsx relies on `person.jsx`, so if that's now orphaned, we should remove that, too.

network-api/networkapi/wagtailpages/models.py
youriwims: I'll open a separate issue for this.

youriwims:
@Pomax follow-up filed in #2898, although I'll put that specific model in the initial comment.

Issue #2846, opened 2019-03-18
"harden" load_fake_data wrt changes in models and streamfield blocks


Pomax:
Every time we add/remove models, or model fields, or streamfield blocks, load_faka_data in its current form will change all the random numbers and effective generate completely new data "downstream" from a change. That means that if we, for example, deprecate and remove a custom streamfield block used in modular pages, then all the Buyers Guide pages will suddenly have different content, too, which makes visual CI throw a bit of a fit (it'll flag a ton of diffs. All of them correct! But many of them not undesired)

Let's update load_fake_data so that we "reseed" the base python random number generator at each step, and at least ensure that changes to models in one place don't affect random data for another.

cadecairos:
This was done already

PR #2871, opened 2019-03-21
reseed at each step, to ensure changes to one model stay local to that model

Pomax:
closes https://github.com/mozilla/foundation.mozilla.org/issues/2846

I've added explicit reseeding between every faking step, to make sure that whenever models change, which changes what faker generates, which changes the pseudo-random number sequence, which changes _everything downstream_, no longer happens.

Instead the PRNG is reseeded and changes to a model only affect the PRNG sequence for faked data involving that model.


Pomax:
superceded by #2897

PR #2872, opened 2019-03-22
Updated spacing on news page

### LINKED ISSUE #1304: News Page Updates
### Issue:
Since we made some changes in the "Latest News" section in our Foundation site, we also need to update the "News" page to reflect that.

Here is a list of suggested changes:
- [x] the “mb-3” should become “mb2”, so it matches the style of the home page.
image

- [x] We should use the same text format that the home page has (1 line instead of 2)
image

@alanmoo assigning to you for triage. I am assuming we can skip design because the requirements seem clear. Let me know if I am wrong, though :)

mmmavis:
Closes #1304

Note that there are 2 cases - with and without article author

image


mmmavis:
@kristinashu would be lovely if we could have the Sketch file updated as well 🙄 🙄 🙄

kristinashu:
@mmmavis yeah it's just a screen cap in sketch 😬. I'll mock up it!

Issue #2812, opened 2019-03-14
Remove old fellowships related files and code


mmmavis:
Now that fellowships bits have been CMSified, please remove old static/hardcoded fellowships related files and code. Thanks!

youriwims:
Just want to confirm that everything in `networkapi/fellows/templates` can be removed?

mmmavis:
@alanmoo @Pomax tagging you both as I wasn't involved with CMSifying fellowships work so I'm not 100% sure

My assumption is that since we have CMSified fellowships pages, we now can safely remove **everything** in `network-api/networkapi/fellows` and old SCSS ?

mmmavis:
Things in `network-api/networkapi/fellows` include

```
network-api/networkapi/fellows
├── __init__.py
├── apps.py
├── migrations
│   └── __init__.py
├── templates
│   ├── application_cta_banner.html
│   ├── application_deadline.html
│   ├── contact_banner.html
│   ├── fellows_apply.html
│   ├── fellows_directory.html
│   ├── fellows_home.html
│   ├── fellows_host_org.html
│   ├── fellows_media.html
│   ├── fellows_open_web.html
│   ├── fellows_science.html
│   ├── fellows_support.html
│   ├── fellows_tech_policy.html
│   └── fellows_type.html
├── urls.py
└── views.py
```

alanmoo:
Sounds right to me, but duplicate of #2737. Though this has more detail, so I'll close that.

HSR05:
@alanmoo I was previously on #2737 . This has made it clear. So, do i keep the `network-api/networkapi/fellows` and remove the old static/hardcoded files in it or should i remove the whole fellows folder present in `network-api/networkapi`.
Kindly confirm this and i'll create a PR for it. Thankyou.

HSR05:
@alanmoo I removed `network-api/networkapi/fellows/templates` file. Is there anything that can be safely removed?

alanmoo:
Ideally we'll be removing that entire `/fellows` directory, though I recognize it might need to happen in pieces. @cadecairos or @patjouk may be able to provide some guidance.

PR #2886, opened 2019-03-25
Restored & Renamed Filter bar files

youriwims:
Accidentally removed the filter bar styles. Also, renamed the file for easy finding in the future 😬 
Related PRs/issues #2812


youriwims:
Moved fix to #2887

PR #2897, opened 2019-03-25
Reseed load fake data 2

Pomax:
Just trying to get percy to understand that https://github.com/mozilla/foundation.mozilla.org/pull/2871 got updated...


Pomax:
Right, I have run out of ideas on why the hell that product page, which works locally, is a 404 once it hits CI

PR #2902, opened 2019-03-26
Make the deploy step a stage

cadecairos:
If my understanding of this is correct, this change should wait for the "Tests" stage to complete before running the "Deploy" stage.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/1742


cadecairos:
yay! 
image

PR #2906, opened 2019-03-27
Remove Email CTA Part 1 of 2

youriwims:
Related PRs/issues #2899 

This pr removes the email cta from the `Opportunity` template as requested [here](https://github.com/mozilla/foundation.mozilla.org/issues/2899#issuecomment-477209814). Part 2 will remove it from models in a separate pr. 

[Test page](https://foundation-mofostaging-pr-2906.herokuapp.com/en/opportunity/no-sign-up-cta/) and/or feel free to check out [Percy](https://percy.io/Mozilla-Foundation/foundation.mozilla.org/builds/1648225?utm_campaign=Mozilla-Foundation&utm_content=foundation.mozilla.org&utm_source=github_status_public).

cc: @kristinashu @alanmoo

PR #2913, opened 2019-03-28
Remove Email CTA Part 2 of 2

### LINKED ISSUE #2899: Remove email cta snippet
To streamline and grow our main email list, we'll no longer encourage individual/independent email lists. Soon we will have email signup in the footer and a link in the nav so we don't need any in-body within a cms page. Although mozfest will be an exception https://github.com/mozilla/foundation.mozilla.org/issues/2874.

Please determine if and where the following email cta snippet are being used, delete those instances, and then completely remove them as an option.
- Internet Health basic form
- MITI
- Fellowship basic form
- Default Form

image

image

image

cc @alanmoo please update this ticket as you see fit.

youriwims:
Closes #2899
[Part 1 of 2](https://github.com/mozilla/foundation.mozilla.org/pull/2906)

This pr removes the use of the email cta in the`Opportunity` template's model. Removing the snippet from the template was completed previously in Part 1 ↖️ .

**To Test**
Create new opportunity page in app and there will be no option to add email cta.

## Checklist

**Changes in Models:**
- [x] If my changes are not backward-compatible, did I schedule a deploy with the rest of the team?


network-api/networkapi/wagtailpages/models.py
```diff
@@ -219,18 +219,9 @@ class OpportunityPage(MiniSiteNameSpace):
     """
```
alanmoo: This comment is irrelevant now too

PR #2933, opened 2019-04-01
Fixing IH CTA Width

### LINKED ISSUE #2909: Update IH CTA Snippet to span width of body content
**Bug Description** 🐛 
Currently, the IH CTA Snippet doesn't line up with the primary body content and instead spans the full width of the page's parent container. So, we want to fix that.

**To Reproduce** 🌱 
1. Follow this [link](https://foundation.mozilla.org/en/opportunity/web-literacy/)
2. Scroll down to the bottom of the page.
3. Error: There you will see that the snippet spans the full width of the navigation and body content.

**Expected behavior** 🌟 
Fix: The snippet should only span the main body content. Per the example ⬆️, the snippet should always be the same width as the `Web Literacy - Overview` section.

_To-Do Checklist_ ✔️ 

- [x] Add field to `ModularPage` model
- [x] Update `settings_panel`
- [x] Include snippet template in `primary_page.html`
- [x] Test feature (If successful, add to all templates)

**Additional context** 📖 

Related Issue #2900 
[Reference Comment](https://github.com/mozilla/foundation.mozilla.org/issues/2900#issuecomment-476908641) ⬇️ 
> The IH Snippet should continue to appear on all the same pages it does now (all the /opportunity, /initiative, and /campaign pages) which means it would need to fit the width on pages that have nav and ones that don't, and pages with the "make narrow" checkbox checked or not checked. 
cc: @kristinashu

youriwims:
Closes #2909
Related PRs/issues #2918 

Here we've templated the IH CTA to easily include it in `/opportunity`(which includes Initiative child pages) & `/campaigns` pages. In doing this, we're removing it from the footer and instead adding it within the primary body content which will also fix the width. 

[Opportunity Template Test](https://foundation-mofostaging-pr-2933.herokuapp.com/en/ih-cta---opportunity-page-test/)
[Campaign Template Test](https://foundation-mofostaging-pr-2933.herokuapp.com/en/campaigns/ih-cta---camapaign-page-test/)
[Test Page w/ Sticky CTA](https://foundation-mofostaging-pr-2933.herokuapp.com/en/campaigns/man-science-office/)


kristinashu:
Layout-wise this is perfect, it is resizing perfectly to all possible page layouts/widths. But...

> Here we're implementing the IH CTA via optional checkbox to add it onto page templates. 

I am wondering where this decision came from because the functionality of where and how this CTA appears should not have changed. From the ticket: "IH Snippet should continue to appear on all the same pages it does now (all the /opportunity, /initiative, and /campaign pages)..."

I would also like to flag that I see the checkbox on `primary` pages but not on `mini site` pages. `primary` pages do not need this CTA, it is the `mini site` pages that should have the cta.

youriwims:
@kristinashu, I misunderstood dev implementation. I'll go ahead and fix that and remove the checkbox functionality.

kristinashu:
Oh excellent, thank you!

network-api/networkapi/wagtailpages/templates/wagtailpages/opportunity_page.html
```diff
@@ -35,5 +35,6 @@ <h1>{{ block.value }}</h1>
         {% endif %}
       {% endfor %}
     </div>
+    {% include "./static/ih_cta.html" %}
```
Pomax: we;re using the `static` dir for purely static page templates, so this should live in `fragments` instead.

Pomax:
Rest of the code looks good! R+ after moving the file.

Issue #2908, opened 2019-03-27
Update opportunity snippet and embed on BanneredCampaignPage


alanmoo:
In order to allow people to sign up for Mozfest emails, we need to update the signup snippet and enable use on the BanneredCampaignPage, below the petition form (though neither will be used at the same time)

![image](https://user-images.githubusercontent.com/1682681/55106652-df701000-50a5-11e9-82ed-68a94ad19f8a.png)

---

**🗒  Design -> Dev Handoff Checklist**

- [x] Design has been finalized
- [x] Link to original design ticket - #2874
- [ ] Link to Redpen
- [ ] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Assets are included
- [ ] Hover state for elements are designed
- [ ] Other states for elements are designed (e.g., error state for form)

alanmoo:
@stephaniemcv As I understand, we should be signing people up for the `mozilla-foundation` newsletter with an "interest" in the festival. It's not entirely clear how this translates to what we submit to Basket. Do we need to check with Pmac?

alanmoo:
@mmmavis I believe the new button styling is part of the work you've been doing, correct?

stephaniemcv:
Yes we should check with Pmac, as I'm not sure on if you're just submitting two slugs mozilla-foundation and mozilla-festival or something else.

alanmoo:
Paging @pmac

pmac:
Basket doesn't really know anything other than what newsletters should be set to True, so if you need for 2 newsletters to be checked then it's probably best to just send `newsletters=mozilla-foundation,mozilla-festival`. They just need to be comma separated.

That said, one alternative is a concept called "newsletter groups" that we can setup on basket. It acts like a newsletter in that it's a single slug, but it can subscribe multiple newsletters at once which are administered centrally via basket. If you'd like one of those so that you can just send "mozilla-festival" and have it also include "mozilla-foundation" I can set that up, and then if it needed to change in future we could do it via the basket admin instead of a code change on your end.

alanmoo:
I think we can handle the comma separated list, but @stephaniemcv does that handle things the way you want?

Pomax:
Reading through our signup component, it looks like all it currently does is send a request to `https://www.mozilla.org/en-US/newsletter/`, so I assume we actually want that switched over to the same SQS post we're using for posting petition-signage-data?

pmac:
[...]

PR #2937, opened 2019-04-01
Signup bcpage

Pomax:
Related PRs/issues https://github.com/mozilla/foundation.mozilla.org/issues/2908

This is a (draft) PR to add signup functionality based on basket-managed newsletters to the bannered campaign page.

This code is not ready to land, but it _is_ ready for a sanity check around how the signup form (from join.jsx) calls the API server's new `/signups/<id>` route, which is hanlded by `signup_submission_view` in `networkapi.campaign.views`.

There is a minimal amount of duplication in the views.py file, falling through to the function that petitions were already using to send data over to SQS, but with a new event type, `newsletter_signup_data`, rather than the existing `crm_petition_data`.

## Checklist

**Tests**

**Documentation:**
- [ ] Is my code documented?


network-api/networkapi/wagtailpages/models.py
```diff
@@ -502,12 +502,25 @@ class BanneredCampaignPage(PrimaryPage):
...
+    signup = models.ForeignKey(
+        'Signup',
+        related_name='bcpage',
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        help_text='Choose an existing, or create a new, sign-up form'
+    )
+
+    panel_count = len(PrimaryPage.content_panels)
```
Pomax: "put things in between other panels" - this adds the signup option between the existing cta option, and the body field.

network-api/networkapi/campaign/views.py
```diff
@@ -30,6 +30,21 @@
 logger = logging.getLogger(__name__)
 
 
+@api_view(['POST'])
+@parser_classes((JSONParser,))
+@permission_classes((permissions.AllowAny,))
+def signup_submission_view(request, pk):
```
Pomax: this is essentially a duplication of the petition code, but I wanted to keep the route view "its own thing" instead of making both signup and petition go to the same def and have its code unravel the actual url things came in on, which would make things far harder to read.

Pomax:
to be clear, @alanmoo @cadecairos this needs review, despite styling not being done.

Pomax:
test page with both CTA and signup component: https://foundation-mofostaging-pr-2937.herokuapp.com/en/signup-test-page/

network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html
```diff
@@ -23,3 +23,45 @@
     </div>
   </div>
 {% endblock %}
+
+The following block has been borrowed from the templates/tags/cta/petition.html file,
```
alanmoo: Why doesn't this (or the above) need to be in some sort of comment tag?

network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html
```diff
@@ -23,3 +23,45 @@
...
+named "cta", which is an assumption that the bannered campaign page squarely violates.
+
+As such, the CTA tag needs to be refactored to take this into account, however this
+work should not be done at the same time as adding signup CTAs to the b.c. pages.
+
+
+{% block cta %}
+  {% if page.cta != None or page.signup != None %}
+  <div class="cta
+    {% if page.specific.narrowed_page_content == False %}
```
alanmoo: Possibly out of scope, but I think this can be removed/made just one class- since if you're seeing the right column, the page content width is a fixed size anyway.

Pomax: I'd say out of scope - it's a copy of what we already had so I'd rather leave it the same as much as possible until we tackle the cleanup.

alanmoo: Does the templating support wrapping `block` tags inside the `if` statement? Or can you call a block's super content manually? Just trying to reduce the WETness here.

network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html
Pomax: as an `extends` template, anything that isn't inside a block gets entirely ignored. (based on "where would it even render it?" logic)

Pomax:
@mmmavis is helping with the markup because we'll need the signup component in the footer etc too.

network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html
Pomax: any toplevel block declaration means "I am overriding the content of this block: ..." and any block declaration inside of templating code means "I am defining a block here, that extending templates can override".

alanmoo: Right, I wasn't sure if there was a secret manual call to `super`. Not a big deal, when you land this file a ticket to remember to DRY it out.

Pomax:
superceded

PR #2939, opened 2019-04-02
Updated Style Guide

### LINKED ISSUE #2903: Update style guide
Follow up from - Design audit to make sure design follows site's style guide #2440 

Update [/style-guide](https://foundation.mozilla.org/en/style-guide/) to reflect our [latest styles](https://foundation.mozilla.org/en/docs/brand/ ) and remove the component portion.

cc @mmmavis

mmmavis:
Closes #2903

- removed old colour (`$green`) from style guide
- removed component section from style guide

Review app page: https://foundation-mofostaging-pr-2939.herokuapp.com/en/styleguide/


mmmavis:
@sabrinang thanks for raising that. As we have discussed it before, buttons by default don't come with margin. I didn't wanna add margin to the buttons demoed on Style Guide just for them to display nicely on that page as I'm worried this could be misleading. (People might think buttons always come with margin?)

If we do want all buttons to come with margin by default we should tackle it as a new ticket. This styling change will require us to visit all pages that currently have buttons and adjust each button's classname and margin accordingly.

mmmavis:
🎉

PR #2940, opened 2019-04-02
Caching Cypress binary

patjouk:
Unzipping Cypress takes a few minutes and it slowing everything down. Caching the binary like recommended in Cypress doc: https://docs.cypress.io/guides/guides/continuous-integration.html#Travis


patjouk:
Wait we're already doing it >.< Nevermind.

PR #2947, opened 2019-04-02
Tidy up type.scss by using mixins

### LINKED ISSUE #2946: Create mixin to tidy up type.scss
https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/type.scss can be further simplified by using mixins.

mmmavis:
Closes #2946

### DIFF
--- a/source/sass/type.scss
+++ b/source/sass/type.scss
@@ -21,176 +21,77 @@ a {
 }
 
 h1,
-h2,
-h3,
-h4,
-h5,
-h6,
-.h1-heading,
-.h2-heading,
-.h3-heading,
-.h4-heading,
-.h5-heading,
-.h6-heading {
-  color: $black; // Bootstrap override
-  margin-bottom: 1rem;
-
-  a {
-    color: inherit;
-    font-weight: inherit;
-
-    @include hover-focus-active {
-      color: inherit;
-      text-decoration: underline;
-    }
-  }
-
-  &.dark-theme {
-    a {
-      @include dark-theme-inline-anchor();
-    }
-  }
-}
-
-h1,
 .h1-heading {
-  @include heading("Zilla Slab", 400, 0);
+  @include heading("Zilla Slab", 400, 0, $black, $white);
   @include scaleText(36px, 40px, 48px, 56px);
-
-  &.dark-theme {
-    color: $white;
-  }
 }
 
 h2,
 .h2-heading {
-  @include heading("Zilla Slab", 300, 0);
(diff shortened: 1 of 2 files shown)


source/sass/type.scss
```diff
@@ -20,177 +20,78 @@ a {
...
-h4,
-h5,
-h6,
-.h1-heading,
-.h2-heading,
-.h3-heading,
-.h4-heading,
-.h5-heading,
-.h6-heading {
-  color: $black; // Bootstrap override
```
mmmavis: these rules have been moved to `@mixin heading`

source/sass/type.scss
```diff
@@ -20,177 +20,78 @@ a {
...
-    a {
-      @include dark-theme-inline-anchor();
-    }
-  }
-}
-
 h1,
 .h1-heading {
-  @include heading("Zilla Slab", 400, 0);
+  @include heading("Zilla Slab", 400, 0, $black, $white);
```
Pomax: This feels like the kind of duplicated rule that should have a dedicated shorter-hand mixing, like `@include defaultHeading(400)`

mmmavis: As not all headings share the same letter spacing and colours, I'd like to keep one heading mixin for all headings for clarity and to avoid overrides. This mixin will and should only be used by headings so once this PR is in we won't have to 'deal with it' again (in case you are finding the long parameter list 😐😒).

Pomax: No, but at least five of them have identical invocations except for the weight, so all of those suggest it makes a lot of sense to add a `default-heading` that only takes a weight value.

mmmavis: Are you suggesting we should have a `default-heading` mixin just for `h1` to `h5` and let `h6` be the only exception? If so we will end up having two copies of the almost identical rules for these headings, one set for `h1-h5` and one set for `h6` and I think this will make things much more to maintain. I might have missed some key points here though. Can you provide code snippets to demostrate how the code here can be further improved?

mmmavis: One improvement I can think of is to make some params have default values so we can turn them into optional params.

Pomax: oh if that's possible, that'd be super sweet. I shouldn't be holding up this PR though, so we can keep it the way it is now, and do more tidying up in followup if we remember?

mmmavis: Cool. I'm gonna keep it the way it is now as it's easier to compare values at a glance. Feel free to open a follow-up ticket if you'd like to chat about this further 😃

Issue #2952, opened 2019-04-03
Our page spawns an hscrollbar on Windows at <=768px widths


Pomax:
Chrome: 

![image](https://user-images.githubusercontent.com/177243/55490365-83971100-55e8-11e9-875b-2332a295d6c4.png)

Firefox:

![image](https://user-images.githubusercontent.com/177243/55490409-8e51a600-55e8-11e9-9d9b-046e0a2ac1d1.png)

cadecairos:
Closing this issue due to its age, if this is work that we still want to do, please refile this with up to date information so that it can be put in the correct milestone with up to date estimates.

PR #2953, opened 2019-04-03
fix scrollbar <= 768px on Windows

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2952 by switching the mobile nav element from using anchoring position+dimensions (`top:0; left: 0; width: 100vw; height: 100vh;`) to using pure anchoring position (`top:0; right: 0; bottom:0; left:0;`, effecting the same full width/height nav without the possibility of running into errors due to using a fixed number of units

### DIFF
--- a/source/sass/components/primary-nav.scss
+++ b/source/sass/components/primary-nav.scss
@@ -214,10 +214,10 @@
         padding-top: 72px;
       }
 
-      height: 100vh;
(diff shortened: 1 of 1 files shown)


Pomax:
closing, this is not the right fix, no matter how much I try to fiddle with units or positioning types.

Issue #2941, opened 2019-04-02
Enable localization of "Intro" field on BanneredCampaignPages


alanmoo:
**Describe the bug**
The field labeled "Intro" on BanneredCampaignPage can't be localized. Let's make that possible.
<img width="1244" alt="image" src="https://user-images.githubusercontent.com/1682681/55406561-508e4800-552a-11e9-9c1a-cea4489bd961.png">

Pomax:
This is an inherited field from PrimaryPage, so we'd have to enable it for all primary pages. If that's fine then this should be a quick thing to do.

alanmoo:
Yeah, that's good.

Pomax:
accidentally landed the wrong code.

PR #2954, opened 2019-04-03
localise the primarypage intro field

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2941

image


TheoChevalier:
Thanks for fixing this! Any estimate as to when this could hit production?

Pomax:
continued in #2986

PR #2955, opened 2019-04-03
Maintaining Directory Height when empty

### LINKED ISSUE #2738: Directory component improvement: hold past height while fetching new list
When switching from one year in the filter component to the next, the page "jumps" because the div is suddenly emptied out while fetching the next group of profiles. Can we reduce the visual impact of this? Perhaps by grabbing the height of the div when switching, manually setting it with an inline style, then removing it once the new data has loaded. (Feel free to go about it another way, I just wanted to explain the steps as I see them for functional clarity.)

jumpy-filter

youriwims:
Closes #2738 

**To Test:**
1. Follow this [link](https://foundation-mofostaging-pr-2955.herokuapp.com/en/directory-height/).
2. Click a different year filter button.
3. When spinner is active, the profile container maintains its height and the footer doesn't jump because it's empty.
4. Once results are loaded, the fixed height is removed.


youriwims:
Good point @mmmavis, I'll try that.

network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
```diff
@@ -105,11 +105,16 @@
...
+    profileContainer.style.height = null;
     profileContainer.innerHTML = cards.join("\n");
     document.dispatchEvent(new CustomEvent("profiles:list-updated"));
   }
 
   function showLoadSpinner() {
+    let containerHeight = profileContainer.offsetHeight;
+    let containerFullHeight = `${containerHeight}px`;
+
+    profileContainer.style.height = containerFullHeight;
```
mmmavis: Looking good!

Let's merge the three lines into 

```jsx
profileContainer.style.height = `${profileContainer.offsetHeight}px`
```

as the two variables `containerHeight` and `containerFullHeight` aren't used elsewhere & `profileContainer.offsetHeight` is pretty self explanatory.

youriwims: sounds good :)

network-api/networkapi/wagtailpages/static/js/directory-listing-filters.js
```diff
@@ -105,11 +105,13 @@
     });
 
     // And then we update the content that the user sees:
+    profileContainer.style.height = null;
```
Pomax: setting to `null` is a little tricky, better to use this:
```suggestion
    profileContainer.style.removeProperty(`height`);
```

youriwims: great suggestion 👍

PR #2963, opened 2019-04-04
Forcing highlight cards equal height

### LINKED ISSUE #2568: Highlights aren't forced to be the same height
Discovered during a visual regression test, we've got nothing forcing the height of highlights to be the same on the homepage:

image 1

Am I correct in assuming that this should be enforced by the template's rendering of consistently sized images? Is this linked to the work @youriwims started, @kristinashu?

youriwims:
Closes #2568

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/card.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/card.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags wagtailimages_tags %}
 
-<div class="card-regular">
+<div class="card-regular h-100 d-flex flex-column">
 
   {% if commitment %}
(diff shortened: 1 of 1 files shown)


youriwims:
@kristinashu should there be space between the home highlight cards on mobile? I figure since I'm fixing their height, I can fix this as well since it's a small fix.
[**PR App**](https://foundation-mofostaging-pr-2963.herokuapp.com/en/)
**On Prod:** There doesn't seem to be any spacing below the first highlight card's gray line

Screen Shot 2019-04-08 at 11 56 09 AM

kristinashu:
Thank you for working on this!

Looking at the sketch file, the spacing on desktop should be: 

image


and on mobile:
image

Pomax:
looks like the percy screenshot had the wrong font for the baseline. Its webfont handling leaves something to be desired...

PR #2956, opened 2019-04-03
[Do Not Merge!} Get sign up component to look right

mmmavis:
### ⚠️ ⚠️ ⚠️ DO NOT MERGE YET ⚠️ ⚠️ ⚠️ 

Test page: https://foundation-mofostaging-pr-2956.herokuapp.com/en/campaigns/test-bannered-campaign-page/

---

The UI/UX portion of PR #2937. Branched off of `signup-bcpage`. I ended up including new footer and donate banner work in this PR so I don't have to deal with the to-be-removed existing sign up banner on production homepage.

image

This PR also closes #2833 #2834 (updated footer & added evergreen donate banner to site)

---

### Work that I _think_ can be done as follow-up

- update footer on `maintenance.html` (maybe as followup?)
- update footer on PNI? (maybe as followup?)
- fine tune footer styling (let's do it in a follow-up ticket)
- code cleanup for 'dark theme' version of sign up (need to wait for https://github.com/mozilla/foundation.mozilla.org/issues/2959 is done first)

PR #2982, opened 2019-04-08
signup was missing O_o

Pomax:
Discovered as part of https://github.com/mozilla/foundation.mozilla.org/pull/2956

The `Signup` snippet does not have localization enabled, which means it cannot be saved without manually duplicating information across the base fields, as well as the `[EN]` copy fields.

This fixes that: test it by trying to create a new Signup snippet but only filling in the normal fields - you should get a confusing error about fields that cannot be empty, despite not being empty when you try to save. Pull this code change, run `inv catch-up` to make sure it gets applied correctly, and then try to make the same signup snippet with the same "only the base fields" data. Saving should now work fine.

### DIFF
--- a/network-api/networkapi/wagtailpages/translation.py
+++ b/network-api/networkapi/wagtailpages/translation.py
@@ -15,6 +15,7 @@ from .models import (
 
     CTA,
     Petition,
+    Signup,
 
     # DEPRECATED
     ParticipatePage,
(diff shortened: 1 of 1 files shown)


mmmavis:
@youriwims I was able to reproduce the issue on both staging and production.

e.g,. 

1. go to https://foundation.mofostaging.net/cms/snippets/wagtailpages/signup/add/
2. fill out the two required fields and hit save

youriwims:
Thanks @mmmavis !

PR #2990, opened 2019-04-09
Removing location pin if there's no meta location

### LINKED ISSUE #2958: Make location display conditional on profile cards
Profiles include an icon even when location is empty (profile by id  streamfield component). 
 

**To Reproduce**

1. Go to https://foundation.mozilla.org/en/docs/how-do-i-wagtail/profiles-from-pulse/
1. Scroll down and compare Peter and Luca
4. See placemarker by Luca's empty location field. 

**Expected behavior**
When profile has no location, there should be no icon and no extraneous space. 

**Screenshots**

image

youriwims:
Closes #2958


youriwims:
Moved to #2997

PR #2993, opened 2019-04-09
Wagtail django uplift

cadecairos:
Updates Django to 2.1.7 (wagtail will support 2.2 on their next release), and wagtail to 2.4.

Also includes changes to some field definitions that now require on_delete to be specified, and a change in settings.py that removes SessionAuthenticationMiddleware (it was deprecated in 1.10 and removed in 2.0, and the functionality is enabled by default in Django now.)


network-api/networkapi/wagtailpages/models.py
```diff
@@ -782,7 +782,7 @@ class HomepageFeaturedNews(WagtailOrderable, models.Model):
         'wagtailpages.Homepage',
         related_name='featured_news',
     )
-    news = models.ForeignKey('news.News', related_name='+')
+    news = models.ForeignKey('news.News', on_delete=models.CASCADE, related_name='+')
```
Pomax: does this require a migration?

cadecairos: nope, I believe this was the default behaviour, they've just made it mandatory to specify now.

Pomax:
suuuuuuuuuuuperceded~~

PR #3023, opened 2019-04-16
Updated footer on PNI and maintenance page

### LINKED ISSUE #3006: Update footer on PNI and maintenance.html
Follow-up for PR https://github.com/mozilla/foundation.mozilla.org/pull/2998

- [x] Update footer on PNI
- [x] Update footer on `maintenance.html`

mmmavis:
Closes #3006


maintenance/maintenance.html
```diff
@@ -100,37 +100,61 @@ <h2 class="h3-header">We Are Undergoing Maintenance</h1>
       </div>
     </div>
 
-    <footer class="mofo-footer">
+    <div class="donate-banner dark-theme py-5">
```
mmmavis: copied markups from `footer.html`

maintenance/static/css/main.css
```diff
@@ -7375,8 +7375,11 @@ a.text-gray-dark:focus, a.text-gray-dark:hover {
   display: inline-block;
   width: 20px;
   height: 20px;
-  margin-right: 10px;
```
mmmavis: I replaced code in this file with what's in `https://foundation.mozilla.org/_css/main.compiled.css` & updated a few image urls to make sure these images show up on maintenance page.

source/js/buyers-guide/bg-main.js
```diff
@@ -16,64 +17,79 @@ import ProductGA from "./product-analytics.js";
 // until all the React stuff is _actually_ done.
 const apps = [];
 
+let env, networkSiteURL, csrfToken;
+
 let main = {
   init() {
-    ReactGA.initialize(`UA-87658599-6`);
-    ReactGA.pageview(window.location.pathname);
+    this.fetchEnv(envData => {
```
mmmavis: Mimicked what we have in `main.js`. 

@Pomax (probably as a followup). Should we think of a way to deal with site-wide shared JS component (like footer) instead of duplicating code in both `main.js` and `bg-main.js`?

mmmavis:
😃 
@cadecairos - can you review the `maintenance` page related bits? 
@Pomax - I have a question for you: https://github.com/mozilla/foundation.mozilla.org/pull/3023/files#r276014451. It would be great if you could also take a look at the rest of the code in this PR!

maintenance/maintenance.html
mmmavis: I assumed we wouldn't wanna load javascript on this page so I didn't include `<div class="join-us mb-5" data-layout="side-button"></div>` here as it serves as React component inject target.

source/js/buyers-guide/bg-main.js
Pomax: Yeah, if you want to file an issue around making things that can be shared, shared, we can do some investigations there.

source/js/buyers-guide/bg-main.js
```diff
@@ -16,64 +17,81 @@ import ProductGA from "./product-analytics.js";
...
+  },
+
+  fetchEnv(callback) {
+    let envReq = new XMLHttpRequest();
+
+    envReq.addEventListener(`load`, () => {
+      callback.call(this, JSON.parse(envReq.response));
     });
+
+    envReq.open(`GET`, `/environment.json`);
```
Pomax: This returns `{"PULSE_API_DOMAIN": "https://network-pulse-api-production.herokuapp.com", "PULSE_DOMAIN": "www.mozillapulse.org", "NETWORK_SITE_URL": "", "TARGET_DOMAINS": null}`, and that last one looks very wrong. Is there a way for us to make that be `[]` instead of `null`?

mmmavis: PR updated to address this!

source/js/buyers-guide/bg-main.js
mmmavis: Follow-up ticket filed: https://github.com/mozilla/foundation.mozilla.org/issues/3055

source/js/buyers-guide/bg-main.js
```diff
@@ -194,6 +212,34 @@ let main = {
...
+            0}/`;
+
+          props.csrfToken = props.csrfToken || csrfToken;
+
+          apps.push(
+            new Promise(resolve => {
+              ReactDOM.render(
+                <JoinUs
+                  {...props}
+                  isHidden={false}
```
Pomax: should we move this into props, too, since we're already working with a separate var?
```
var props = element.dataset;
...
props.isHidden = false;
props.whenLoaded = () => resolve();

apps.push( ... <JoinUs {...props}/>, element);
```

mmmavis: Not sure why but if I do `props.whenLoaded = resolve;` it actually transforms `resolve` function into a string. 👀 

```js
console.log(typeof props.whenLoaded); // returns string
```

mmmavis:
@Pomax PR updated again!

maintenance/maintenance.html
Pomax: yeah, when we're down for maintenance, we don't want a component that needs to talk to the server.

source/js/buyers-guide/bg-main.js
```diff
@@ -194,6 +223,31 @@ let main = {
         injectDonateModal(donationModal, modalOptions);
       }
     */
+
+    if (document.querySelectorAll(`.join-us`)) {
+      var elements = Array.from(document.querySelectorAll(`.join-us`));
```
Pomax: It turns out that `querySelectorAll` returns a static NodeList, so we don't need the `Array.from` anymore. We might want to file a separate issue to clean up `Array.from` everywhere, though, and add one here even though it's new code.

mmmavis: I really wanna get this PR merged so let's tackle ^ as a follow-up: https://github.com/mozilla/foundation.mozilla.org/issues/3109

PR #3026, opened 2019-04-17
Footer hover state improvement & dark theme paragraph link updates

### LINKED ISSUE #3007: Fine tune footer styling
Follow-up for PR https://github.com/mozilla/foundation.mozilla.org/pull/2998

To make sure things match what we have on the mockups. 

- hover state
- ~form error state~ for form related stuff we'll revisit when form standardization work is done.

mmmavis:
Closes #3007

- did **not** fix anything form related - we'll revisit when form standardization is done
- added hover state for social icons
- updated dark theme paragraph links' stylings
- updated the two mozilla.org links (logo & "Visit us at mozilla.org") to go to foundation.mozilla.org instead and say "Visit us at foundation.mozilla.org". Abigail from legal says if we're talking about the not for profit then we shouldn't be linking to moz.org.


mmmavis:
> checkbox small body copy looks like it goes up to the edge of the button

It was an adjustment we had to make due to tech complexity. When I implemented the form I've chatted with both Kristina and Natalie on this already.

> other small body copy aligns to bottom edge to where the mozilla logo is

Might be a bit tricky to do this but I'll see if I can come up with a not complicated solution next week.

sabrinang:
@mmmavis Thanks for aligning the small body copy to the bottom edge to where the mozilla logo, looks good!

For the checkbox area, that's good you checked with Nat already. There's one breakpoint that stands out that has a really short line length, perhaps the button can be full-width below to allow the field and text to be longer here as a suggestion but will defer to @natalieworth 
image

mmmavis:
@sabrinang good suggestion! Do you remember which breakpoint your screencap was taken from? If we need a transition state/layout for tablet view let's open a followup ticket for that as it'll require designers to create mockup & devs for implementation.

sabrinang:
I believe its from `@media screen and (min-width: 768px)` the second smallest breakpoint @mmmavis, sounds good for separate followup

natalieworth:
Hey @mmmavis Just had a couple last updates. So close and it's looking so good!

- 'Donate Now' text can be in our blue. (Sorry, it may not have been like that in our style guide button file since Sketch requires a button background color)
- Form input text can be black (but also feel free to ignore this if you want to wait til we have all the form designs updated)
Artboard 2@2x

The error message still displays as a box:
Artboard 3@2x

Could we make the icon hover the light blue #B7B9FA
Artboard 4@2x

mmmavis:
@natalieworth thanks for the comments.

### Form related stuff

- UI tweaks: Let's leave them until brand standardization stuff is done.
- Error bubble for email field: I'll create a bug ticket for it and include the link here (ticket to be filed).

### Donate Button / Dark Theme Secondary Nav Design

The donate button is using the default dark theme secondary nav design. I'll create a follow-up ticket if we wanna change the current behaviour of that particular button type. (Not sure if it's technically possible to make text appear transparent to achieve the 'cutout' effect we want. We'll discuss it further in the to-be-created ticket)

### Social icon colour on hover

They are in the default `$light-blue` colour code: `#b7b9fa`: https://foundation.mozilla.org/en/style-guide/ Can you file a follow-up ticket if we wanna change `$light-blue` to `#b7b9f9` instead?

mmmavis:
Thanks! I will update my comments above with link to the follow-up tickets I'm going to create.

mmmavis:
Follow-up tickets filed! See https://github.com/mozilla/foundation.mozilla.org/pull/3026#issuecomment-485593910

Issue #2994, opened 2019-04-10
Add newsletter sign up button and form to nav


kristinashu:
### Add newsletter sign up button to nav, update donate button, and add form.

- [x] Add `Newsletter` button
- [x] Update style of `Donate` button ([icon-heart.svg.zip](https://github.com/mozilla/foundation.mozilla.org/files/3064965/icon-heart.svg.zip)) (hover: dark blue #0d10bf)
- [x] Mobile: `Donate` stays put while `Newsletter` button moves down below nav items
- [x] Clicking on `Newsletter` button animates sign up form to open and page content slides down
- [x] Clicking anywhere outside of the black form box animates the form to close
- [x] Form error/success/hover states should follow footer styles
- [x] Keep Donate UTM
- [ ] Add google event to Newsletter button (clicks and actual signups)
- [ ] Add source tag to Newsletter button for filtering purposes so Will can look up how many signups came from that button (@WillatMozFdn @Pomax please check if I worded this correctly ...)

**🗒  Design -> Dev Handoff Checklist**

- [x] Design has been finalized
- [x] Link to original design ticket #2757 #2948 
- [x] Link to related ticket https://github.com/mozilla/foundation.mozilla.org/pull/2956
- [x] Link to Redpen https://redpen.io/p/uvd561910fde4f4bc9
- [x] Design changes been merged into the master Sketch file _(Home page in Foundation site file)_
- [x] Desktop and mobile mockups are included
- [x] Assets are included _(heart icon above)_
- [x] Hover state for elements are designed
- [x] Other states for elements are designed (e.g., error state for form)

Final comps:
![Home Nav Copy@2x](https://user-images.githubusercontent.com/1068783/55899124-ebb19e00-5b78-11e9-8499-7d8772f7117a.jpg)

Hovers:
![Home Nav Hovers@2x](https://user-images.githubusercontent.com/1068783/55899140-f409d900-5b78-11e9-9ca9-da85b88047bc.jpg)

Newsletter opens on click:
![Nav Open Copy@2x](https://user-images.githubusercontent.com/1068783/55899157-fc621400-5b78-11e9-9333-5f5fbebc27e7.jpg)

Buttons prioritized on mobile & Newsletter opens on tap (closing by clicking on anywhere in nav bar or clicking Sign up button):
[...]

PR #3028, opened 2019-04-17
[Part 1 of 3] Newsletter sign up button/form

youriwims:
Related PRs/issues #2994

Updated donate button styles & extended mobile nav behavior to tablet dimensions.


youriwims:
Screen Shot 2019-04-17 at 4 34 46 PM

@mmmavis , this is caused by a box-shadow on `.btn:focus`, do we want to keep this?

mmmavis:
@sabrinang wondering if we should design another type of button for use cases like above? (button with transparent background, no border)

sabrinang:
For the use case that we need an icon before a link or label, we shouldn't use a button at all then? It should be a link with an in-line icon that comes before it. Similar to how read more →  has an arrow after it, I can see donate having the heart before it

mmmavis:
@youriwims - @sabrinang and I are chatting now. This might be special "anchor" or "button" style that we wanna add to our style guide. Will get back to you!

youriwims:
Okay, not a problem. Can I land this in the meantime so I can move on to Part 2(adding the newsletter button & form) of this?

mmmavis:
@youriwims thanks for flagging it. I chatted with @sabrinang and @kristinashu and we are all okay with making the "donate" link using the default link behaviour for now. i.e., remove `btn` class from it

donate

youriwims:
okay, sounds good. Thank you!

kristinashu:
Hover is looking good but sizing seems generally a little small. The "Donate" text was designed to be the same font size and weight as the nav item. The heart icon also seems a little small, it should be a little bigger so it stands out.

image

youriwims:
Sorry, I was under the impression from Mavis' conversation with you and
Sabrina that I should be removing the `.btn` class from this which gives it
those styles...just want to confirm that it should have the same font &
font weight as nav items?

As for the heart, I swapped the dimensions accidentally 😬.Thanks for
pointing that out—I'll go back and fix that!

>

youriwims:
@kristinashu, all should be fixed now. Let me know if any other updates need to be made :)

Pomax:
Chrome with mobile view set to something with a small screen shows me an empty nav... =S

image

youriwims:
Thanks for pointing this out @Pomax 😬

Pomax:
Looks like there's still something funny happening with the width. 

### foundation site, mobile view:

image

### https://foundation-mofostaging-pr-3028.herokuapp.com/en/, mobile view:

image

PR #3048, opened 2019-04-22
[Part 2 of 3] Newsletter sign up button/form added newsletter button

youriwims:
Related PRs/issues #2994 

https://foundation-mofostaging-pr-3048.herokuapp.com/
## Checklist

- [x] Add Newsletter button
- [x] Mobile: Donate stays put while Newsletter button moves down below nav items
- [x] Clicking on Newsletter button animates sign up form to open and page content slides down
- [x] Clicking anywhere outside of the black form box animates the form to close
- [x] Form error/success/hover states should follow footer styles


mmmavis:
@kristinashu this is ready for design review. Note that the desktop version form is still in one-column layout. I'm working on modifying the Newsletter/Sign Up React component so we can have the option to show it as 2-column. I'm separating the work from this PR as that React component is used in other places (e.g., footer) and it's best that we do it as 2 stages in case I accidentally break something.

Another note is that I _can't_ perfectly vertically centered the form section. We have to reserve some extra space for error messages to go. (Technical context: in order to achieve the slide down effect, we have to give the fully expanded section a fixed height before the animation starts so the slide down animation knows when to 'stop growing/expanding'. If we _really_ want it to look centered of course we can go with some JavaScript help, but hopefully we can do this as a follow-up ticket.)

Artboard 1

kristinashu:
Super slick @mmmavis! Animation and interactions are excellent! 

But is it possible to change the "No thanks" copy on the success screen to something like "Close" or "Back to menu" (but keep "No thanks" on the signup form page)?

image

youriwims:
On the app, the newsletter is overlapping the page content once it opens. Should it be pushing the page content down once it opens like the mocks in the og ticket @mmmavis @kristinashu ?

Screen Shot 2019-05-10 at 3 04 59 PM

kristinashu:
Thanks Youri! We decided the form should not push the content down so that it could always open the same way, whether the user is at the top of the page or if they've scrolled down the page.

But good catch about the newsletter signup on tablet, since it would fit in the nav at that width we should show it.

mmmavis:
> On tablet+, [the newsletter button should appear in the nav](https://github.com/mozilla/foundation.mozilla.org/issues/2994#issuecomment-484658811) btw :) Right now, it's only in the nav at desktop.

@youriwims thanks for pointing that out!

Our nav implementation is suuuuuper complicated already to achieve the current nav design we want (mobile/desktop design, collapsing/expanding menu, zen/not zen, static/sticky elements, etc). Our nav jumps to mobile design at <992px and I'm hoping we can have the same breakpoint for the newsletter section we are adding in this PR. This will significantly help the code from becoming much more crazier 😂 and help streamline things. @kristinashu @youriwims is it okay if we make the compromise here?

mmmavis:
Note: if we really wanna have the newsletter button appear in the nav on tablet+ we can still do it. It doesn't require a lot of code changes. I put my thinking in my previous comment just to see if we can help streamline the existing crazy nav code a little better.

youriwims:
True, I completely understand. Looks like Kristina gave your previous comment a thumbs up—I'm also good with the compromise as well! 👍 Maybe we could look into updating the form at medium device sizes in future enhancement ✨ :)

mmmavis:
@kristinashu @youriwims thanks! Yea we can file an enhancement ticket for medium devices.

mmmavis:
I'm now thinking about turning the JS `nav-newsletter` module into a React component... 🤔 🙄 😬 Will decide next Monday 🤦‍♀ @Pomax thoughts?

source/js/main.js
```diff
@@ -241,8 +243,8 @@ let main = {
     }
 
     // Embed additional instances of the Join Us box that don't need an API exposed (eg: Homepage)
-    if (document.querySelectorAll(`.join-us`)) {
-      var elements = Array.from(document.querySelectorAll(`.join-us`));
+    if (document.querySelectorAll(`.join-us:not(.on-nav)`)) {
```
Pomax: If we're touching this, let's make sure we don't run the same query selector twice:
```
var joinUsElements = document.querySelectorAll(`.join-us:not(.on-nav)`);
if (joinUsElements.length) {
  ..
}
```

source/js/nav-newsletter.js
```diff
@@ -0,0 +1,135 @@
...
+let buttonMobile = narrowMenuContainer.querySelector(".btn-newsletter");
+let joinUs = container.querySelector(`.join-us.on-nav`);
+let buttonDismiss = container.querySelector(".form-dismiss");
+
+let navNewsletter = {
+  getShownState: () => {
+    return isShown;
+  },
+
+  // Inject a new <JoinUs> component each time so we can show <JoinUs> with its initial state
```
Pomax: What does this comment mean with "each time"? It feels like it's not being detailed enough. If the idea is to "replace what is on the page" with a new JoinUs compnent just to reset its state, then this should probably be changed to something like this:
```
class JoinUs extends Component {
  constructor(props) {
    super(props);
    this.state = this.getInitialState();
  } 
  reset() {
    this.setState(this.getInitialState());
  }
  getInitialState() {
    return { ... };
  }
```
so that we can call the JoinUs component's `.reset()` function without having to replace it with an entirely new component.

source/js/nav-newsletter.js
```diff
@@ -0,0 +1,135 @@
+import React from "react";
```
Pomax: This file feels off. Like it's trying to act like a prototyped object, without having been defined as such (using a `class`)

source/js/nav-newsletter.js
mmmavis: @Pomax we can't call a React component's method from outside of the React world. That's why I had to unmount the component in `nav-newsletter.js` instead.

mmmavis: By each time I mean each time the newsletter section is revealed. I can update the comment.

source/js/nav-newsletter.js
mmmavis: Should I turn it into a class?

mmmavis:
@kristinashu this is up for design review again!

Pomax:
It looks like the button stays highlighted if you click it once to open the dialog, and then click it again to dismiss the dialog. Should it reset to outline styling?

source/js/main.js
```diff
@@ -261,29 +263,26 @@ let main = {
...
-              ReactDOM.render(
-                <JoinUs {...props} whenLoaded={() => resolve()} />,
-                element
-              );
-            })
-          );
-        });
-      }
+    var joinUsElements = document.querySelectorAll(`.join-us:not(.on-nav)`);
+    if (joinUsElements && joinUsElements.length) {
```
Pomax: `document.querySelectorAll` always returns a NodeList, so just `if (joinUsElements.length)` is enough (unlike `querySelector()`, which can return `null`). However, since we only use it to check whether to start a `forEach`, we don't even need that: `forEach` will always work, it'll just "do nothing" for an arraylike of length 0, so we don't need any conditional here, just:
```
var joinUsElements = document.querySelectorAll(`.join-us:not(.on-nav)`);
joinUsElements.forEach( element => {
  ...
});
```

source/js/nav-newsletter.js
```diff
@@ -0,0 +1,162 @@
...
+    // close newsletter section if clicking anywhere outside of the section
+    if (!container.contains(event.target) && event.target !== container) {
+      navNewsletter.closeDesktopNewsletter();
+    }
+  }
+
+  checkDomNodes() {
+    let allNodesExisted = false;
+
+    try {
```
Pomax: Note that none of the code in this try block will ever throw, so try/catch wrapping will not do anything meaningful. We can instead rewrite this function to something quite a bit smaller by relying on [Array.every](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/every):
```javascript
/**
 * Find and bind all necessary DOM nodes, returning "false" if not all DOM nodes were found.
 */
checkDomNodes() {
  return Object.keys(elements).every(key => {
    // if this element already resolved to a DOM node, move on to the next
    let value = elements[key];
    if (typeof value !== "string") return true;

    // find this DOM node, and report on the result (binding it if found for later use)
    let element = document.querySelector(value);
    if (element) elements[key] = element;
    return !!element;
  });
}
```

source/js/primary-nav.js
```diff
@@ -1,4 +1,6 @@
 import ReactGA from "react-ga";
+import ReactDOM from "react-dom";
```
Pomax: unused import?

source/js/nav-newsletter.js
```diff
@@ -0,0 +1,162 @@
+import React from "react";
+import ReactGA from "react-ga";
+import ReactDOM from "react-dom";
+import JoinUs from "./components/join/join.jsx";
+
+let primaryNav,
```
Pomax: Let's change this to an object, due to the fact that we can make checkDomNodes a lot smaller:
```javascript
const elements = {
  primaryNav: `#primary-nav-container`,
  narrowMenuContainer: `.wide-screen-menu-container`,
  wideMenuContainer: ...,
  container: ...,
  buttonDesktop: ...,
  buttonMobile: ...,
  joinUs: ...,
  buttonDismiss: ...
};
```
With corresponding changes so that `container..` becomes `elements.container` etc.

source/js/nav-newsletter.js
mmmavis: > Note that none of the code in this try block will ever throw, so try/catch wrapping will not do anything meaningful.

Ah I thought `buttonDesktop = wideMenuContainer.querySelector(".btn-newsletter");` would throw error when `wideMenuContainer` is `null`?

Anyways I will update my PR based on the code snippet above.

source/js/main.js
mmmavis: 😂 i keep forgetting about this

mmmavis: We don't use `joinUsElements` elsewhere so I'm gonna merge them into

```js
document.querySelectorAll(`.join-us:not(.on-nav)`).forEach(element => { ... }
```

mmmavis:
> It looks like the button stays highlighted if you click it once to open the dialog, and then click it again to dismiss the dialog. Should it reset to outline styling?

@Pomax I can't reproduce the issue. Maybe your had your cursor over the button that's why it stayed highlighted?

newsletter-btn-highlighted

Pomax:
both in FF and Chrome the button stays black.

STR:
- load https://foundation-mofostaging-pr-3048.herokuapp.com/en/
- move mouse to over "newsletter button" => button turns black
- click button => newsletter dialog comes up
- click button again => newsletter dialog disappears, button stays focused (focus-outline shows)
- move mouse away from button => button remains focused (focus-outline shows) and filled black until you focus on a different part of the page, or remove focus by clicking an inactive part.

source/js/nav-newsletter.js
Pomax: ahhh missed that - we can turn that into a "full" query selector though and then with the qs mapping object that won't be a problem.

mmmavis:
OMG FINALLY! THANKS TEAM!

PR #3080, opened 2019-04-30
Force image to be 50*50px in image text mini component

### LINKED ISSUE #3076: Image text mini layout bug
**Describe the bug**
- Icon and layout in image text mini component changes depending on copy on right side. 

**To Reproduce**
- I've made this test page to show the three different possible text variations https://foundation.mofostaging.net/en/initiatives/image-text-mini-test/.
- Component looks good in version 1 when text wraps but icon looks stretched (and ratio of icon to text is different) if text doesn't warp (is only on one line or has returns to break lines).

**Expected behavior**
- Icon and layout should look the same regardless of how much text there is and how text is formatted.

**Screenshots**
image

This is happening on prod here https://foundation.mozilla.org/en/fellowships/apply/ (correct) vs https://foundation.mozilla.org/en/fellowships/partner-with-us/ (icon is too large).

cc @youriwims @mmmavis

mmmavis:
Closes #3076 

Test page: https://foundation-mofostaging-pr-3080.herokuapp.com/en/test-page/

I didn't modify this part but note that padding (purple highlighted area) is set to 16px on medium+ devices and 8px for smaller devices. 
image

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/image_text_mini.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/image_text_mini.html
@@ -2,7 +2,7 @@
 
 <div class="my-5">
   <div class="p-2 p-md-3 col-12 d-flex align-items-center img-txt-mini-container">
-    <div class="mr-3 mr-md-4">
+    <div class="img-txt-mini-img-container mr-2">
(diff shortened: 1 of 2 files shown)


kristinashu:
Icon size is looking consistent at all screen widths! 

But is it possible to make the space on the left and right side of the image look the same on both sides?

image

mmmavis:
@kristinashu PR updated!

PR #3088, opened 2019-05-02
Prevent PG port conflicts

Pomax:
This switchies the postgres port to 5439 rather than using the default 5432, so that it does not crash the docker setup process if someone happens to also have postgres installed and running for other work.

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -19,7 +19,7 @@ services:
   postgres:
     image: postgres:9.6
     ports:
-      - "5432"
+      - "5439"
     volumes:
       - postgres_data:/var/lib/postgresql/data/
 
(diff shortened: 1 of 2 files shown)


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
   postgres:
     image: postgres:9.6
     ports:
-      - "5432"
+      - "5439"
```
cadecairos: This will require an additional change to work. We can map the internal container port `5432` to the host port `5439` with the syntax `"5439:5432"`.

patjouk:
Using this syntax, it's [not supposed](https://docs.docker.com/compose/compose-file/#ports) to be conflicting with the host: 
>Either specify both ports (HOST:CONTAINER), or just the container port (an ephemeral host port is chosen).

I tried and could not repoduce it: can you copy past the error message?

docker-compose.yml
Pomax: Not with the docker.env change that makes sure the client gets build with the correct env values too though?

cadecairos: I'm not sure what you mean by that

Pomax:
hm, I'll close this for now, reproducing is proving a little too unpredictable

Issue #555, opened 2017-07-10
Conduct web properties audit


hannahkane:
This is only tangentially related to the Network site, but I think this is the best repo to file this.

Review all existing web properties and compile a list of recommendations about what to do with various content, functionality, etc.

gvn:
I'm kind of vague on the ask here. Can you clarify this more?

hannahkane:
Sure. We're reviewing all of the various Mofo web properties with an eye towards what content should be pulled into the Network site (perhaps allowing us to sunset some other properties).

No specific ask for engineering right now, but let me know if you'd like to be involved in this early on.

hannahkane:
This work has begun and will continue for awhile, so rolling into next heartbeat.

xmatthewx:
Work is ongoing. Had a good kickoff. Roll ticket over or track in planning docs?

hannahkane:
Will roll over.

hannahkane:
Conversations with several stakeholders have happened, and a few more are scheduled for next week. 

One tangible output will be a plan for how issue-specific content (e.g. web literacy map/curriculum) will show up on the Network site in the future.

kristinashu:
I feel like this work has happened, closing but @xmatthewx please reopen if needed.

xmatthewx:
Sydette is working on content migration plan. I don't think we need to track that here.

PR #3090, opened 2019-05-02
Made stylelint throw errors if custom colours are found

### LINKED ISSUE #2663: update linter to forbid raw CSS colours outside of colors.scss
Based on https://github.com/mozilla/foundation.mozilla.org/issues/2636, it will be well-worth the effort to update our linter such that it only allows "raw colours" (rgb/hsl/hex/etc) in `colors.scss`, flagging raw colours anywhere else as a linting error.

mmmavis:
Closes #2663

Stylelint will throw errors if `rgb`, `rgba`, `hsl`, `hsla`, `hwb`, `gray`, HTML color names are found outside of `_variables.scss`. This prevents us from accidentally introduce custom colour to the system.

Design team is in the process of eliminating use cases for using not 100% opaque colours. However, in cases where opacity modification is needed:

- use SASS's [transparentize function](https://sass-lang.com/documentation/functions/color#transparentize) to decrease the alpha channel of a color variable.
- use SASS's [opacify function](https://sass-lang.com/documentation/functions/color#opacify) to increase the alpha channel of a color variable.


package.json
```diff
@@ -30,6 +30,7 @@
     "test:procfile": "node test/test-procfile.js",
     "test:eslint": "eslint --config ./.eslintrc.json \"cypress/integration/*.js\" \"network-api/networkapi/{,!(frontend)/**/}*.js\" \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
     "test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" --syntax scss",
+    "test:scss:styleguide": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" \"!source/sass/**/_variables.scss\" --syntax scss --config .stylelintrc-styleguide.json",
```
mmmavis: We want this task to only lint SCSS files outside of `_variables.scss`

source/js/components/petition/donation-modal.scss
```diff
@@ -5,7 +5,7 @@
   right: 0;
   bottom: 0;
   left: 0;
-  background: rgba(255, 255, 255, 0.9);
+  background: transparentize($white, 0.1);
```
alanmoo: This feels like it violates the spirit of this PR.

mmmavis:
@alanmoo I agree that turning pure `white` and `black` into variables seems too much. Maybe we should make these two colours as exceptions. However, I think this kind of linting is really necessary. In the past designers and devs have struggled so much to keep design (especially colours) consistent across. Check out @sabrina's [audit](https://docs.google.com/presentation/d/1mQJOwKj1OQkXlhHOVa2uLX0gRPQ-PffW0B3ffXztk5w/edit#slide=id.g4da221aa2c_1_0) to get a sense of how many random colours there were on the site. Custom colours in codebase weren't always added by design. A lot of times new colours got added because design specs weren't detailed enough (or devs didn't have access to the Sketch file to inspect colours) that devs had to just guess the colour themselves (like `grey`, `#ccc`, `#eee`, `#555`, `red`). This created issues as designers didn't have tools to verify colours on review apps & PR reviewers wouldn't spot this issue as PR had already passed design review.

I chatted with @sabrinang and it seems like our goal is to make sure we have zero custom colours and to try our best to stick with 100% opaque colours (@sabrinang please chime in in case I said something wrong). Designers are in the process of making these restrictions on Sketch so I'm hoping devs should work from the same restrictions as well. 

> If a new color comes in, an engineer that doesn't know the intent of this PR might simply add the new variable to the one file that they can, and continue on their way, rather than pushing back on it.

We should let devs on our team know going forward we wanna make sure we only use standardized colours. If they see colour in design that falls outside of the styleguide they should flag it to designers. I remember GitHub has the feature to set default reviewer for individual files. Will it help if we make sure by default a designer is set as PR reviewer when `_variables.scss` gets modified? This way we can ensure random colours don't "sneak in" just to pass linting errors.

alanmoo:
That sounds like a reasonable solution, perhaps partnered with a comment near the list of color variables talking about what's going on.

package.json
Pomax: `:styleguide` is pretty good, it leaves the door open for other things like "only allowing font families we want" etc.

source/js/components/multipage-nav-mobile/multipage-nav-mobile.scss
```diff
@@ -45,7 +45,7 @@
   }
 
   .col-12 {
-    border: 2px solid #dedede;
+    border: 2px solid $gray-20;
```
Pomax: Poor king Dedede =(

source/js/components/petition/petition.scss
```diff
@@ -17,7 +17,7 @@
     background: $white;
     padding: 15px;
     width: 100%;
-    box-shadow: 0 -1px 3px rgba(0, 0, 0, 0.15);
+    box-shadow: 0 -1px 3px transparentize($black, 0.85);
```
Pomax: Is this a real function, or did we come up with it? If the latter, can we rename it to something like `transparent()` instead?

source/sass/buyers-guide/includes/draft-product.scss
```diff
@@ -1,7 +1,7 @@
 // product draft rules
 .draft-product::before {
   content: "This is a draft.";
-  color: #ffa7a7;
+  color: $light-red;
```
Pomax: Colour nit: unlike green or blue, there is no "light" red. It goes from red through a few named colours to pink, so this would be something either "salmon" or "rose" =)

source/sass/buyers-guide/views/about.scss
```diff
@@ -64,7 +64,7 @@
         bottom: 0;
 
         &:focus {
-          background: red;
+          background: $dark-red;
```
Pomax: kind of the same nit, but I guess we're using a uniform naming scheme that ignores real world colour naming =(

source/js/components/petition/petition.scss
mmmavis: it's SASS/SCSS' built-in function: https://sass-lang.com/documentation/functions/color#transparentize

note: its opposite function is called [opacify](https://sass-lang.com/documentation/functions/color#opacify)

mmmavis:
@Pomax 

> at least make the few colours that we "transparentize" dedicated named colors in _variables.scss that are "solid equivalents" of the transparent colours we use, in preparation for design saying what they should really be.

Since we are adding this note to `_variables.scss` to make sure non-standardized styles don't sneak in there just to pass linting errors, we should probably create a follow-up ticket to handle what you suggested above as those styles are still to be reviewed and standardized by designers. Ideally we want this file to reflect what designers have in our shared Sketch style library so designers and devs can work from the same style guide.

image

mmmavis:
Maybe I should move those standardized variables to a new file and give it a more "styleguid" specific name?

.stylelintrc-styleguide.js
```diff
@@ -0,0 +1,298 @@
...
+      "declaration-property-value-blacklist": [
+        {
+          "/.*/": [
+            /rgba{0,1}\(/i,
+            /hsla{0,1}\(/i,
+            /hwb\(/i,
+            /gray\(/i,
+            // uses of web color names are not allowed (with 'white' and 'black' being the only exceptions)
+            /[^\$-\w]lavender/i,
+            /^lavender/i,
```
mmmavis: to capture the instances when value of the CSS rule starts with a color name,

e.g.,

```css
property: colorname; // we don't have to worry about the whitespace there as stylelint seems to trim beginning and tailing whitespace before it runs the checks
```

mmmavis:
Updating this PR atm. Will ping people for review again once I'm done!

.stylelintrc-colors.js
```diff
@@ -0,0 +1,298 @@
...
+      "declaration-property-value-blacklist": [
+        {
+          "/.*/": [
+            /rgba{0,1}\(/i,
+            /hsla{0,1}\(/i,
+            /hwb\(/i,
+            /gray\(/i,
+            // uses of web color names are not allowed (with 'white' and 'black' being the only exceptions)
+            /[^\$-\w]lavender/i,
+            /^lavender/i,
```
mmmavis: this is to capture the instances when a rule's value start with `<colorname>`.

e.g.,

```css
color: colorname;
// We don't have to worry about the space after `:`.
// stylelint trims out whitespaces around a rule's `value` before running the test
```

source/sass/buyers-guide/includes/draft-product.scss
mmmavis: I like current our color naming system 'cuz everything looks intuitive to me. However I'm not an expert in this area so if you'd like to propose a new naming convetion we should create a ticket and get designers to chime in. 🎨

mmmavis:
@Pomax @alanmoo after changing things back and forth for a few times I've decided to move all color variables into its own new SCSS file `_color.scss` and make the stylelint color test run against all other SCSS files. Let me know what you think! 👀

.stylelintrc-colors.js
```diff
@@ -0,0 +1,24 @@
...
...ue|pink|lightpink|hotpink|deeppink|palevioletred|mediumvioletred|lightsalmon|salmon|darksalmon|lightcoral|indianred|crimson|firebrick|darkred|red|orangered|tomato|coral|darkorange|orange|yellow|lightyellow|lemonchiffon|lightgoldenrodyellow|papayawhip|moccasin|peachpuff|palegoldenrod|khaki|darkkhaki|gold|cornsilk|blanchedalmond|bisque|navajowhite|wheat|burlywood|tan|rosybrown|sandybrown|goldenrod|darkgoldenrod|per|chocolate|saddlebrown|sienna|brown|maroon|darkolivegreen|olive|olivedrab|yellowgreen|limegreen|lime|lawngreen|chartreuse|greenyellow|springgreen|mediumspringgreen|lightgreen|palegreen|darkseagreen|mediumseagreen|seagreen|forestgreen|green|darkgreen|mediumaquamarine|aqua|cyan|lightcyan|paleturquoise|aquamarine|turquoise|mediumturquoise|darkturquoise|lightseagreen|cadetblue|darkcyan|teal|lightsteelblue|powderblue|lightblue|skyblue|lightskyblue|deepskyblue|dodgerblue|cornflowerblue|steelblue|royalblue|blue|mediumblue|darkblue|navy|midnightblue|snow|honeydew|mintcream|azure|aliceblue|ghostwhite|whitesmoke|seashell|beige|oldlace|floralwhite|ivory|antiquewhite|linen|lavenderblush|mistyrose|gainsboro|lightgray|silver|darkgray|gray|dimgray|lightslategray|slategray|darkslategray)/i
```
mmmavis: this is to capture the instances when a rule's value start with `<colorname>`.

e.g.,

```css
color: colorname;
// We don't have to worry about the space after `:`.
// stylelint trims out whitespaces around a rule's `value` before running the test
```

source/sass/buyers-guide/includes/draft-product.scss
Pomax: the scheme makes sense, it just leads to some color names that out of context don't make a lot of sense =)

.stylelintrc-styleguide.js
Pomax: We don't need two rules per color here, `/[^\$-\w]?lavender/i,` captures both. However, rather than specifying this ourselves, could we just tap into https://github.com/stylelint/stylelint/tree/master/lib/rules/color-named instead and say `never`?

(since we have a $black and $white for use outside of _variables.scss)

.stylelintrc-colors.js
Pomax: Rather than specifying this ourselves, could we just tap into https://github.com/stylelint/stylelint/tree/master/lib/rules/color-named instead and say `never`?

mmmavis: not if we want to have exceptions - in this case `black` and `white`

mmmavis: Yea that was my first attempt. But if we want to have `black` and `white` as exceptions then we can't use Stylelint's `color-named ` 😭

Pomax: Do we, though? Using $black and $white seems pretty okay in everything except the variables file?

alanmoo:
This seems fine but I'll let @pomax give it the final stamp because he was working more closely with it.

mmmavis:
Yea merging!

PR #3091, opened 2019-05-02
Update bannered_campaign_page.html

alanmoo:
Only show fixed "TAKE ACTION" button on mobile for petitions because this is undesirable: 
image

Demo at https://foundation-mofostaging-pr-3091.herokuapp.com/en/mozfest/


network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html
```diff
@@ -38,6 +38,8 @@
 As such, the CTA tag needs to be refactored to take this into account, however this
 work should not be done at the same time as adding signup CTAs to the b.c. pages.
 
+An extra check has been added so as to not show the fixed "Take action" button on mobile pages for email signups.
```
Pomax: this is probably unnecessary, since future code readers won't even know there was ever a time where "Take action" wasn't conditional. This is the kind of comment that is good to put in the commit/PR comment.

PR #3105, opened 2019-05-06
npm updates

alanmoo:
I ran `npm update`. Let's see which dependabots this closes


package-lock.json
```diff
@@ -1527,7 +1449,7 @@
         },
         "util": {
           "version": "0.10.3",
-          "resolved": "https://registry.npmjs.org/util/-/util-0.10.3.tgz",
+          "resolved": "http://registry.npmjs.org/util/-/util-0.10.3.tgz",
```
cadecairos: Why is npm converting urls from https to http

alanmoo: Hmm. Good catch. https://npm.community/t/npm-install-downgrading-resolved-packages-from-https-to-http-registry-in-package-lock-json/1818/7

alanmoo: I ran `sed -i -e 's/http:\/\//https:\/\//g' package-lock.json` and added another commit. Let's confirm it builds.

PR #3107, opened 2019-05-06
Make cms url more forgiving, add page edit link

alanmoo:
These things were annoying me

`/en/cms` now redirects to `/cms`
When logged in, there's an "edit" button at the bottom of Wagtail pages to view the page in the admin.

I made it a link to the "view page" so that a user still goes somewhere useful if they can't edit the page. I also hardcoded the styles because it's so separate from anything else on the site, it didn't feel like it made sense to embed in the page CSS. The edit button is in the bottom left so it doesn't get in the way of the nav.

### DIFF
--- a/network-api/networkapi/templates/pages/base.html
+++ b/network-api/networkapi/templates/pages/base.html
@@ -18,7 +18,7 @@
 
     {% if user.is_authenticated %}
     <div class="authenticated-page-edit" style="position: fixed; bottom:0; background: teal; padding: 0.25em; z-index:1000">
-      <a style="color: white" href="/cms/pages/{{page.id}}/">Edit</a>
+      <a style="color: white" href="/cms/pages/{{page.id}}/edit/">Edit</a>
     </div>
     {% endif %}
 
--- a/network-api/networkapi/templates/pages/base.html
+++ b/network-api/networkapi/templates/pages/base.html
@@ -16,6 +16,12 @@
   </head>
   <body id="view-{% block bodyID %}{% endblock %}">
 
+    {% if user.is_authenticated %}
+    <div class="authenticated-page-edit" style="position: fixed; bottom:0; background: teal; padding: 0.25em; z-index:1000">
+      <a style="color: white" href="/cms/pages/{{page.id}}/">Edit</a>
+    </div>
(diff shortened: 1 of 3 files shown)


xmatthewx:
I’m excited about this. I can’t login to the branch. I hit login with google, leads to a 404. Even without /en in the url, it redirects to https://foundation-mofostaging-pr-3107.herokuapp.com/en/soc/login/google-oauth2/?next=/cms

network-api/networkapi/templates/pages/base.html
```diff
@@ -16,6 +16,12 @@
   </head>
   <body id="view-{% block bodyID %}{% endblock %}">
 
+    {% if user.is_authenticated %}
+    <div class="authenticated-page-edit" style="position: fixed; bottom:0; background: teal; padding: 0.25em; z-index:1000">
+      <a style="color: white" href="/cms/pages/{{page.id}}/">Edit</a>
```
Pomax:
```suggestion
      <a style="color: white" href="/cms/pages/{{page.id}}/edit">Edit</a>
```

network-api/networkapi/urls.py
```diff
@@ -40,6 +40,7 @@
         name='how-do-i-wagtail'
     ),
     url(r'^cms/', include(wagtailadmin_urls)),
+    url(r'^en/cms/', RedirectView.as_view(url='/cms')),
```
Pomax:
```suggestion
    url(r'^en/cms/', RedirectView.as_view(url='/cms/')),
```

network-api/networkapi/templates/pages/base.html
alanmoo: See original comment- this was intentional

Pomax: This does make it fairly cumbersome to edit pages, through - if we want to make sure people without edit rights still see something useful, we should be able to test for that to make sure the template renders the right link up front?

alanmoo:
It looks like the Percy delay still didn't catch everything. I'm going to file a new PR to increase the delay a bit more. If that still doesn't load all the assets, I'm going to reach out to Percy support.

PR #3108, opened 2019-05-06
Will CSS background position fix the logo?

alanmoo:
Trying to make Percy not fail


alanmoo:
It seems like this delay made Percy pick up all the graphics. Hopefully, with this in place, future review apps don't lose them again.

PR #2376, opened 2018-12-20
Refactor load_fake_data

cadecairos:
Refactor load_fake_data so that the factory files can define a callable 'generate' method which the management command can call

Slightly WIP, I want to turn wagtailpages into a module to futher reduce that file's complexity.

Closes #

PR #3126, opened 2019-05-10
Load fake data refactor

### LINKED ISSUE #1708: Improve dev workflow for fake data generation
Right now the fake data script needs to be edited each time a new model needs to be mocked. It feels a bit clunky. It would be great if there was a way of doing this automatically if a certain file exists in an app's directory or something like that.

cadecairos:
Closes #1708
Previous PR #2376 

Still TODO: split up wagtailpages/factory.py


cadecairos:
filed follow-up: https://github.com/mozilla/foundation.mozilla.org/issues/3143

PR #3159, opened 2019-05-15
[WIP] Mozfest test branch 2

Pomax:
test PR because things were very strange for a bit.

Test URL: https://foundation-mofostaging-pr-3159.herokuapp.com/en/

This PR introduces a new `mozest` app to the `INSTALLED_APP` list in our `settings.py`, which houses all the code for the mozfest templates and models.

The models inherit from the base wagtail Page, as well as our Foundation mixin (to ensure all CMS-managed pages have the same base additionals), and model a page type similar to out standard `PrimaryPage` (called, hopefully unsurprisingly, `MozfestPrimaryPage`). The homepage is an extension of this model with a `pass` model body, meaning it does nothing more/less than the superclass, but allows django/wagtail to resolve it to a different template.

Template wise, a new `mozfest-base.html` was made that extends the foundation site's `base.html`. In order to ensure that anything and everything that needs to be changed can be changed, the `base.html` template was given a few more blocks, so that `mozfest-base.html` is mostly just an elaborate sequence of block overrides rather than being an HTML page with template logic. Both the MozfestPrimaryPage and MozfestHompage templates extend the mozfest-base` template, with minimal changes. 

CSS is being tweaked in the `source/sass/mozfest.scss` file, with very little modifications at the moment (and the fewer we end up needing, the better of course! =D)

Input requests for @alanmoo  and @mmmavis, @youriwims added as observer-reviewer

Will close https://github.com/mozilla/foundation.mozilla.org/issues/3160


network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -0,0 +1,70 @@
...
+          </div>
+        </div>
+      </div>
+      <nav class="container">
+        <div class="row">
+          <div class="col">
+            <div class="d-flex flex-row justify-content-between">
+              <div id="primary-nav-links">
+                <div class="d-flex align-items-center flex-wrap">
+                {% if page.zen_nav == True %}
```
Pomax: there's still some cruft like this that needs to get trimmed out because mozfest has no zen/narrow etc, to be confirmed in the comments

Pomax:
@sabrinang @kristinashu Just to confirm, these pages will not have dedicated zen/narrow mode will they?

network-api/networkapi/mozfest/templates/partials/hero/large.html
```diff
@@ -0,0 +1,14 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<picture>
+  {% image page.banner fill-4032x2304 as image_xl %}
```
Pomax: this partial only differs in the "vertical" resolution, but Django templates don't do preprocessing macros or something similar to `calc()`, so there's no way to only make those vertical dimension numbers something that gets controlled by a single value that is passed into the partial.

At least, not without writing more code that adds additional django logic in a tag file which then needs to load another template file, so it just gets too impractical and instead I opted to duplicate the hero image code with large and regular banners, to allow "homepage" vs. "every other page" styling.

network-api/networkapi/mozfest/templates/partials/mozfest_footer.html
```diff
@@ -0,0 +1,43 @@
+{% with wrapper_class="py-5" %}
```
Pomax: While it looks like it isn't, this footer _does_ get used, the React code for the footer just kicks in and replaces its content.

In order to give us some control, I've assigned the `<body>` a class which for mozfest pages always contains `mozfest`, so we can probably update our React code for the footer in `main.js` to do something different for mozfest pages, so they look the way they're supposed to.

network-api/networkapi/mozfest/templates/partials/primary_heroguts.html
```diff
@@ -0,0 +1,30 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<div id="hero">
+  <div class="banner">
+    {% if page.banner %}
```
Pomax: Just to keep partials legible, it might be worth adding this separation to the foundation `primary_heroguts.html` as well.

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -0,0 +1,70 @@
+{% load primary_active_nav %}
+
+<div id="primary-nav-container">
```
Pomax: It feels like there is work we can probably do bringing down the level of div nesting here, but this is not the PR for that.

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -0,0 +1,70 @@
...
+  <div class="wrapper-burger">
+    <div class="menu-container">
+      <div class="narrow-screen-menu hidden d-lg-none">
+        <div class="narrow-screen-menu-background">
+          <div class="narrow-screen-menu-container">
+            <nav class="container">
+              <div class="row">
+                <div class="col">
+                  <div class="nav-links pt-3">
+                    <div><a class="{% primary_active_nav request '/' %}" href="/">Home</a></div>
```
Pomax: This code sort of exists "twice" in the partial, which threw me off the first time I edited this block and didn't see it update the page, only to realise I was editing the code that shows only at mobile widths. 

It might be a good idea to turn these links into an include as well, so that both navs include "the same links" but primed for a different UX role.

package.json
```diff
@@ -65,6 +65,7 @@
     "mofo-bootstrap": "4.1.0",
     "mofo-style": "^2.4.0",
     "moment": "^2.24.0",
+    "node-fetch": "^2.5.0",
```
Pomax: this guy is an accidental npm install in the wrong dir and will be removed.

source/sass/mozfest.scss
```diff
@@ -0,0 +1,18 @@
+body.mozfest {
+  &#view-mozfest-home {
+    #hero {
+      --moz_banner_height: 430px;
```
Pomax: this is a standard CSS variable. I don't know if we want to start using those, but I've been working with them a lot in a different project and they make `calc()` easier to work with. That said, it's probably better to use a SASS variable here, so consider this PoC rather than "done" code.

Actually, I guess that goes for everything in this PR atm =D

kristinashu:
@Pomax that's right, no zen mode nav for this site. cc @sabrinang

network-api/networkapi/mozfest/templates/partials/mozfest_footer.html
youriwims: > While it looks like it isn't, this footer does get used, the React code for the footer just kicks in and replaces its content.

How is the React code kick in and replace this footer content?

network-api/networkapi/mozfest/templates/partials/primary_heroguts.html
youriwims: Agreed.

alanmoo:
Looking pretty good so far

source/sass/mozfest.scss
youriwims: How does this CSS variable make working with `calc()` easier?...just out of curiosity. Also, 
PoC = Proof of Concept?

youriwims:
Same, looks good to me so far 🌟 ... I dropped some random questions here and there 🤓

source/sass/mozfest.scss
Pomax: 1. plain old CSS variables, but _very_ few people seem to know those exist, and that they're mad functional (you can even change them in JS with a simple `element.style.setProperty('--varname', 'new value')` and boom, all the CSS depends on it changes).
2. Yep, Proof of Concept.

network-api/networkapi/mozfest/templates/partials/mozfest_footer.html
Pomax: We're loading the same `main.js` that the foundation site does, so it sees the `join-us` class and goes "aha! I can drop a `<JoinUs>` component in there~"

source/sass/mozfest.scss
youriwims: I _did not_ know that...how cool! I'm going to try to start using them 😅

alanmoo: [This handsome gentleman](https://youtu.be/jWDZP8twWDg?t=611) has known about them for a while.

network-api/networkapi/mozfest/templates/partials/mozfest_footer.html
youriwims: 👏 👏 👏 ✨ ✨

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
mmmavis: _runs away from nav clean up work 😂_

source/sass/mozfest.scss
Pomax: I remember being in the audience. It was a good talk =)

network-api/networkapi/mozfest/templates/partials/primary_heroguts.html
Pomax: https://github.com/mozilla/foundation.mozilla.org/issues/3166

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
Pomax: https://github.com/mozilla/foundation.mozilla.org/issues/3160#issuecomment-493170721

PR #3167, opened 2019-05-16
Mozfest Site Factories

cadecairos:
~WIP~

~I'm going to tweak this a bit tomorrow, and create a way to just generate the mozfest site after a production content clone.~

Ready for review. This PR adds factory classes for the mozfest app classes. It also updates load_fake_data, flush_models.py and adds a new management command that's specific to generating a mozfest site, which we'll use in our site clone task.


cadecairos:
I'll need to land an update to our mofo-cron repo to call the `generate_mozfest_site` management command after a clone, but this lays all the groundwork for it.

network-api/networkapi/management/commands/flush_models.py
```diff
@@ -48,9 +52,17 @@ def handle(self, *args, **options):
         CTA.objects.all().delete()
 
         self.stdout.write('Dropping all Pages')
-        Page.objects.exclude(title='Root').delete()
+        WagtailPage.objects.exclude(title='Root').delete()
 
         self.stdout.write('Dropping all Products')
         Product.objects.all().delete()
 
+        print('Dropping all Mozfest pages')
```
alanmoo: Wouldn't line 55 take care of this?

cadecairos: Ah, good point!

network-api/networkapi/mozfest/factory.py
```diff
@@ -0,0 +1,92 @@
...
+
+    print('Creating MozFest Site record in Wagtail')
+    tds = settings.TARGET_DOMAINS
+    if tds and len(tds) > 1:
+        # Assume that tds[0] is the main mofo domain, and tds[1] is the Mozfest domain
+        hostname = tds[1]
+        port = 80
+    else:
+        # use a localhost domain (must be set in /etc/hosts)
+        hostname = 'mozfest.localhost'
```
alanmoo: What will this do on review apps?

network-api/networkapi/management/commands/generate_mozfest_site.py
```diff
@@ -0,0 +1,37 @@
...
+
+        if options['seed']:
+            seed = options['seed']
+        elif settings.RANDOM_SEED is not None:
+            seed = settings.RANDOM_SEED
+        else:
+            seed = random.randint(0, 5000000)
+
+        print('Dropping all Mozfest pages')
+        MozfestPrimaryPage.objects.all().delete()
```
alanmoo: This has probably already been done too, but seems reasonable to leave this here in case the command is run separately? But maybe it needs a try/except?

network-api/networkapi/mozfest/factory.py
cadecairos: hmm I hadn't considered review apps. We only get one domain per review app though.

We could build an integration that sets up a CNAME record in route53 under our mofodev domain, and add it to the app.

network-api/networkapi/management/commands/generate_mozfest_site.py
cadecairos: I'll see if `.delete()` throws if `.all()` is an epty QuerySet

cadecairos: The plan is to run this separately on mofo-cron

cadecairos: I checked, empty QuerySets do not raise exceptions when you call `delete` on them.

network-api/networkapi/management/commands/flush_models.py
cadecairos: It's removed now.

network-api/networkapi/management/commands/generate_mozfest_site.py
alanmoo: What about MozfestHomepage?

alanmoo: There might be other page types too, is there a way to say "Homepage and all descendants"?

cadecairos: Unfortunately all I think I can do is filter on the immediate parent. If we end up adding new page types we can just call the same methods against the new Page class.

alanmoo: I guess I'm confused as to why this script drops anything, instead of just being for generating the site itself?

cadecairos: Here's a thought, if we've got pages on production now, then we don't need to generate any new ones after a copy. We need to only update the hostname of the Site.

If that's true, then we can probably just delete this command.

alanmoo: But it's useful to generate them for staging/dev. Are you just referring to deleting this line?

cadecairos: As it is now, the load_fake_data management command will generate the mozfest pages locally. Since we added a Mozfest site and pages to prod, those will be copied over on the weekly clone, making this command unnecessary, I think.

alanmoo: Great.

cadecairos: `MozfestHomepage` inherits from `MozfestPrimaryPage`, so that give me the nice benefit of being able to query for the whole lot of Pages through `MozfestPrimaryPage`

alanmoo:
One last update before I leave for the weekend:

`inv docker-catch-up` did what it was supposed to.

Then I ran `inv docker-manage "load_fake_data --delete"` and it seemed to succeed.

Out of curiosity, I then ran `inv docker-nuke-db` which then gave me the same error as above.

Another run of `inv docker-manage "load_fake_data --delete"` then failed too.

Pomax:
Does that happen when you don't have a database to wipe, or also immediately after a load_fake_data? If the first, I suspect we're missing the try/except-pass code that deals with deleting individual model instances, rather than deleting through a queryset.

Pomax:
I'm getting this error on the wagtailpages refactor, too, so there's probably something that's gone wrong with how we try to get the site root.

network-api/networkapi/mozfest/factory.py
```diff
@@ -0,0 +1,106 @@
...
+            root_page=home_page,
+            site_name='Mozilla Festival',
+            is_default_site=False
+        )
+
+    print('Generating Mozfest sub-pages')
+    [MozfestPrimaryPageFactory.create(
+        parent=home_page,
+        title=title
+    ) for title in ['About', 'Spaces', 'Tickets', 'Team', 'Sponsors']]
```
Pomax: there is no need for an `about` page anymore (it's been removed)

alanmoo:
This looks good, but can you put a note in the documentation about fake data and mention that the Mozfest site generates as a child of "Homepage" for the purposes of review apps, since they can only have one domain?

network-api/networkapi/mozfest/factory.py
cadecairos: made the change!

Issue #3160, opened 2019-05-15
Implement new mozfest app


Pomax:
Implement a new mozfest app with a homepage and "other" page model, with associated templates.

Closes: https://github.com/mozilla/foundation.mozilla.org/issues/3170
Related: https://github.com/MozillaFoundation/mozfest-website-2019/issues/1 (includes comps)

Pomax:
Tasks left to do:
- [ ] clean up the footer (does it need a refinement to `<JoinUs>` based on being on the mozfest site?)
- [ ] clean up the nav (make the links includable fragments to DRY out that part of the template)
- [ ] add SCSS for narrow content on all pages, while making the homepage the only page with wide content.
- [ ] make sure GA kicks in for mozfest rather than for the foundation site (this probably involved `main.js` work, probably with a new `MOZFEST_ANALYTICS_ID` or somesuch env var)

mmmavis:
Re-opening to track leftover todos.

Pomax:
Solved the footer by making them always pull the signup details from the `Mozfest` signup snippet.

Pomax:
All caught up!

PR #3169, opened 2019-05-16
make sure the footer uses the correct newsletter signup

Pomax:
Fixes the footer so that it always makes sure to use the Signup snippet with the `Mozfest` name (case insensitive) for the page footers on any of the pages.

Closes: #3170 
Related to: #3160


network-api/networkapi/mozfest/templates/partials/mozfest_footer.html
```diff
@@ -1,43 +1,23 @@
-{% with wrapper_class="py-5" %}
+{% extends "partials/footer.html" %}
```
Pomax: the change to this file is mostly to make sure that all that duplicated code stays in the template I originally copied it all from. No need to have all of that here when really all we want is the left column for the newsletter signup.

network-api/networkapi/mozfest/models.py
```diff
@@ -50,7 +51,11 @@ class MozfestPrimaryPage(FoundationMetadataPageMixin, Page):
 
     def get_context(self, request):
         context = super().get_context(request)
-        return get_page_tree_information(self, context)
+        context = get_page_tree_information(self, context)
+        # Also make sure that these pages always tap into the mozfest newsletter for the footer!
+        mozfest_footer = Signup.objects.filter(name__iexact='mozfest').first()
```
alanmoo: What if there are no results?

alanmoo: (This is looking at the Signup snippet, correct? It seems like the review app doesn't have any snippets, but the submission still "works". I just can't see where it's going)

Pomax: The React conversion [in main.js](https://github.com/mozilla/foundation.mozilla.org/blob/mozfest-footer-update/source/js/main.js#L249-L265) looks at the HTML generated in https://github.com/mozilla/foundation.mozilla.org/pull/3169/files#diff-a7829fada9ac08105a722faacb866cefR13 and will see "there is no id here, I will default to id=0"

Pomax: This is then further corrected for in the signup submission process: if a valid submission comes in (e.g. the csrf token is correct) but it's for an unknown signup snippet, [the code will use the default signup](https://github.com/mozilla/foundation.mozilla.org/blob/mozfest-footer-update/network-api/networkapi/campaign/views.py#L54-L67) by building a throw-away instance so it can get the default newsletter we've set in our Signup model.

alanmoo: So if `mozfest` doesn't exist in the database, it falls back to the MozillaFestival newsletter signup?

Pomax: correct

alanmoo: I think my concern is more around a future-us testing this on a review app or locally and wondering why it's doing the wrong thing. As noted in Slack, let's put this into the migration path.

Pomax: migration added

Pomax:
updated.

PR #3220, opened 2019-05-23
Made MozFest primary page's secondary nav wrap

### LINKED ISSUE #3201: Sub nav at medium widths
Sub nav gets a little odd a medium widths (example https://mozillafestival.mofoprod.net/en/spaces/).

Screen cap from prod:
image

@mmmavis is it feasible to make the nav go onto a new line, something like this? 
image

cc @sabrinang

mmmavis:
Closes #3201 

Based on design tweaks @kristinashu and I chatted about in the office...

secondary-nav-small-gap


mmmavis:
I don't have creds to log in to review app so I can't set it up for live testing. Hopefully the attached gif is good enough... if not, we'll figure out another way 😂

Issue #3118, opened 2019-05-08
MozFest: Social Media Feed


sabrinang:
### Description
For Mozilla Festival 2019, it'd be great to feature more social activity and stories by including a social feed. Specifically, twitter #mozfest hashtag or @mozillafestival handle.

**Feed**
This can be featured above the footer area on the landing page. TBD how it is styled because it may just use twitter's styles.

Wireframe on redpen here: https://redpen.io/smaff9809d00feddb3
<img width="948" alt="image" src="https://user-images.githubusercontent.com/2492510/57413384-ca0ffa80-71a8-11e9-8a67-58e8a6c9745b.png">

alanmoo:
I envision this as a streamfield component, which would be useful across the site. Keep in mind that streamfield components are separate from the template, which defines how they get laid out on a page. So if you're looking for a 2 column page layout, that's a different topic.

Pomax:
this might not be shared, but just something that only lives on the homepage, which means we can probably just "do this in the template" rather than trying to make wagtail slip this into a streamfield

sabrinang:
![image](https://user-images.githubusercontent.com/2492510/57659865-c5788700-7598-11e9-8bc6-98d109eb6810.png)
social feed mockup in landing page on redpen here: https://redpen.io/vuefbd37973f0c2a38
- 2 column since it is only on the homepage
- can the text left of it be editable?
- how does the embed look like? I'm not sure if my screenshot is accurate in the mockup and if it needs any additional styling

kristinashu:
@alanmoo can you please provide some guidance on implementing this?

kristinashu:
On slack @alanmoo said:

> My understanding is we can _just_ embed this in the template and be good

<a class="twitter-timeline" href="https://twitter.com/mozillafestival?ref_src=twsrc%5Etfw">Tweets by mozillafestival</a> <script async src="https://platform.twitter.com/widgets.js" charset="utf-8"></script> 

@Pomax is skeptical and reminded us that CSP clearance will be necessary because that's a third part embed.

Maybe @mmmavis could take a look???

Pomax:
@sabrinang @kristinashu  Styling wise, what is the max-height this embed block should be?

kristinashu:
We might need to adjust it once we see it at all screen widths but could you please start with 336px (this is based on using multiples of 8 so 8x42=336px which is pretty close to what Sabrina mocked up)?

kristinashu:
[...]

Issue #3216, opened 2019-05-22
MozFest: pre-footer homepage styling 


kristinashu:
- [ ] Add black divider line 
- [ ] Add padding below Location text
- [ ] Update location text to body-large
- [ ] Mobile: Location copy should go wide width and stack on top of the twitter feed

### Prod:
<img width="1435" alt="image" src="https://user-images.githubusercontent.com/1068783/58215623-9e643880-7caf-11e9-905e-7d6486df5cc1.png">

### Comp: https://redpen.io/vuefbd37973f0c2a38
<img width="1237" alt="image" src="https://user-images.githubusercontent.com/1068783/58215543-357cc080-7caf-11e9-87cb-04c195801075.png">

PR #3225, opened 2019-05-23
twitter feed on the homepage

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3118 by adding a twitter embed to the homepage prefooter block. 

Note that this needs _quite a bit of CSP updating_ so I've added the clearances to the `env.default` but the staging/prod instances will need updating to allow for these clearances, too.

I've customised the stream to make sure DNT is passed along (even if our users don't have DNT turned on, for the twitter stream we should damn well make sure no tracking is happening if we can help it =D) but there might still be CSP violations due to twitter trying to contact its own syndication tracking while scrolling through the stream: _that's a good thing_

demo: https://foundation-mofostaging-pr-3225.herokuapp.com/en/


kristinashu:
Note: styling for mobile can be done here https://github.com/mozilla/foundation.mozilla.org/issues/3216

Pomax:
@sabrinang it feels like that's probably worth making part of the mobile tweaks rather than this PR?

sabrinang:
@Pomax ah okay, if it isn't part of this PR that's fine then

PR #3240, opened 2019-05-27
Update heroku + travis to python 3.7

patjouk:
Closes MozillaFoundation/mofo-devops#659

The Pipfile and Docker file are already on 3.7. Let's use 3.7 everywhere.


patjouk:
I had to add libgconf because it's not installed by default on Xenial, cf https://github.com/cypress-io/cypress/issues/4069.

PR #3241, opened 2019-05-27
Separating JS dev and prod dependencies

patjouk:
We're installing a bunch of JS deps that are not necessary for prod. I tried to separate dev/prod dependencies, but I might have missed one or two, can someone double check? ~I'm pretty sure using `NODE_ENV=production` for review apps, staging and prod will be okay, but again, correct me if I'm wrong 😅~ It's already running with this setting, never mind :D

It won't change anything for local dev or CI: all dependencies will still be installed when running `npm i` or building the node image. Only review apps, staging and prod will run without dev dependencies.

todos:
- [x] Add `NPM_CONFIG_PRODUCTION=true` to staging,
- [x] Add `NPM_CONFIG_PRODUCTION=true` to prod.


patjouk:
Heroku was still installing all dependencies. [I checked the doc](https://devcenter.heroku.com/articles/nodejs-support#package-installation): we need to use `NPM_CONFIG_PRODUCTION=true` and not `NODE_ENV=production`.

package.json
```diff
@@ -45,43 +45,28 @@
...
-    "dotenv": "^8.0.0",
-    "eslint": "^5.16.0",
-    "eslint-config-prettier": "^4.3.0",
-    "eslint-plugin-prettier": "^3.1.0",
-    "eslint-plugin-react": "^7.13.0",
     "event-stream": "3.3.4",
     "html-entities": "^1.2.1",
     "js-cookie": "2.2.0",
     "mofo-bootstrap": "4.1.0",
     "mofo-style": "^2.4.0",
```
mmmavis: Seems like we no longer uses `mofo-style`. I'll file a separate ticket to remove this.

mmmavis: Filed: #3248

youriwims:
I'll let Mavis & Cade review this since I'm not super knowledgeable on all the packages we utilize. 😅

PR #3262, opened 2019-05-28
Implement a temporary exception for the mozilla festival domain

cadecairos:
This PR adds a temporary exception in the domain redirect middleware that will redirect `https://www.mozillafestival.org/` to `https://foundation.mozilla.org/en/mozfest/` when `MOZFEST_DOMAIN_REDIRECT_ENABLED` is set to True.

@alanmoo if there's a related issue, please link it below.

I'm going to open a follow-up issue to revert this code and the environment variable once the site launches.


alanmoo:
This looks good, though is there a way to test it before we make it live?

cadecairos:
I tested it locally my manually modifying some lines.

1. update `.docker.env` so that `DOMAIN_REDIRECT_MIDDLEWARE_ENABLED=True`, `MOZFEST_DOMAIN_REDIRECT_ENABLED=True` and `TARGET_DOMAINS=localhost:8000,mozillafestival.localhost:8000`
2. update `middleware.py:36` so it reads `if request_url = 'mozillafestival.localhost:8000``
3. If needed, add `mozillafestival.localhost` to `/etc/hosts`
4. visiting `mozillafestival.localhost:8000/` should redirect to `localhost:8000/en/mozfest` (this will 404 since it's not added by default in dev), and visiting `mozillafestival.localhost:8000/foobar` should not redirect, but 404 (just to illustrate that it doesn't redirect anything except `/`

cadecairos:
> 
> 
> Looks good, just put the appropriate env vars on prod before merging this in.

The var is set up to default to False, but I'll set it up to be `True`

PR #3272, opened 2019-05-30
change mozfest page URLs when using a review app

cadecairos:
The URLs for mozfest site pages are different on review apps, so I modified the nav template fragment so that if `HEROKU_APP_NAME` is set, it will use URLs with `/mozilla-festival/` as the prefix.

I'm open to suggestions if anyone can think of a DRY-er way to achieve this..

### DIFF
--- a/network-api/networkapi/mozfest/templates/fragments/nav-links.html
+++ b/network-api/networkapi/mozfest/templates/fragments/nav-links.html
@@ -1,5 +1,7 @@
 {% load primary_active_nav %}
 
+{# TODO: Dynamically generated Nav links: https://github.com/mozilla/foundation.mozilla.org/issues/3289 #}
+
 {% if HEROKU_APP_NAME %}
 
 {{ pre }}<a class="{% primary_active_nav request '/mozilla-festival/spaces' %}" href="/mozilla-festival/spaces">Spaces</a>{{ post }}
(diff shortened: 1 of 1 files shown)


alanmoo:
Can you set a variable in the if block? https://stackoverflow.com/a/1070414

cadecairos:
> 
> 
> Can you set a variable in the if block? https://stackoverflow.com/a/1070414

oh interesting, I could try:

```
{% if  HEROKU_APP_NAME %}
  {% with path_prefix="/mozilla-festival/" %}
{% endif %}
{% if not  HEROKU_APP_NAME %}
  {% with path_prefix="/" %}
{% endif %}
```

cadecairos:
no, I misunderstood how the `with` tag works, It's still scoped to a block

cadecairos:
aha:

```
{% if foo %}
  {% with a as b %}
    {% include "snipet.html" %}
  {% endwith %} 
{% else %}
  {% with bar as b %}
    {% include "snipet.html" %}
  {% endwith %} 
{% endif %}
```

https://stackoverflow.com/a/7081901/4722739

alanmoo:
I was just about to suggest that :)

cadecairos:
@alanmoo should we close this out while we consider how to make the nav dynamic, or land this and open an issue to convert it to a dynamic solution? either way I'd rather not leave this PR here.

alanmoo:
File an issue to make it dynamic, and then add a comment to this code referencing the issue, and we can land this.

cadecairos:
> File an issue to make it dynamic, and then add a comment to this code referencing the issue, and we can land this.

@alanmoo done.

alanmoo:
Looks like this is about to be superceded by #3294

PR #3274, opened 2019-05-30
Enable wide content pages when necessary

### LINKED ISSUE #3237: MozFest: Pulse Profile listing is too small
**Describe the bug**
The photos and spacing is too small and cramped for all pulse profile components. Can we make it bigger?

**To Reproduce**
Here's a page where we use it: https://mozillafestival.mofoprod.net/en/team/wranglers/

**Expected behavior**
Ideally something like what's on the foundation site like https://foundation.mozilla.org/en/fellowships/meet-our-fellows/

cc @kristinashu to triage

alanmoo:
I think this closes #3237 

**Tests**
	I'm not sure we have appropriate tests for this already.

**Changes in Models:**
	This block based approach feels much less fragile for narrow/wide templates

**Documentation:**
- [x] Is my code documented?
	Do we have documentation for Mozfest pages already?

[Example page](https://foundation-mofostaging-pr-3274.herokuapp.com/en/mozilla-festival/team/)


kristinashu:
If we're going with this approach then we should at least use the same full width as the foundation site https://foundation.mozilla.org/en/fellowships/meet-our-fellows/

PR:
image

Foundation Prod:
image

alanmoo:
That's fine, but keep in mind if there's any text on the page it's going to be super wide too. I went for this width because it seemed like a decent compromise but I'm happy to switch to the full width if you give the 👍

kristinashu:
And I still don't think it's ideal because it still doesn't look great at medium widths 
image

kristinashu:
Just to clarify, will this be a checkbox on each page like it is on the foundation site? So we can choose exactly which pages will be narrow vs full width?

network-api/networkapi/mozfest/templates/mozfest/mozfest_primary_page_wide.html
```diff
@@ -0,0 +1,4 @@
+{% extends "./mozfest_primary_page.html" %}
+
+{% block bootstrap_width %}col-lg-10 offset-lg-1{% endblock %}
```
alanmoo:
```suggestion
{% block bootstrap_width %}col-lg-12 offset-xl-1 col-xl-10{% endblock %}
```

alanmoo:
> Just to clarify, will this be a checkbox on each page like it is on the foundation site? So we can choose exactly which pages will be narrow vs full width?

Correct, in the settings panel like it has been previously (but "narrow" is the default here)

I've updated the widths at XL and LG now (it will take a few minutes to build though). It's still not perfect at medium, but I think it's way better, and will get even better when coupled with Mavis's PR.

network-api/networkapi/mozfest/templates/mozfest/mozfest_primary_page_wide.html
```diff
@@ -0,0 +1,4 @@
+{% extends "./mozfest_primary_page.html" %}
+
+{% block bootstrap_width %}col-lg-12 offset-xl-1 col-xl-10{% endblock %}
```
alanmoo:
```suggestion
{% block bootstrap_width %}{% endblock %}
```

PR #3287, opened 2019-06-03
Removing Snyk

cadecairos:
GitHub has built in dependency monitoring tools now, so we don't need to rely on Snyk an longer.


patjouk:
Let's not forget to remove snyk webhook in this repo settings too :)

cadecairos:
> Let's not forget to remove snyk webhook in this repo settings too :)

I think that's already done, but I'll verify.

PR #3290, opened 2019-06-03
Add Google site verification for Mozfest domain

alanmoo:
And make sure it doesn’t have two, for clean separation

Google emailed me about usability issues and this was a super quick fix to verify ownership, so no issue filed.


alanmoo:
@Pomax that’s great but it doesn’t let me merge 😋

cadecairos:
I got you.

Issue #3289, opened 2019-06-03
Dynamic Mozfest header nav


cadecairos:
The nav on the mozfest site is static, but thinking ahead, we want to make it dynamic to support adding and removing sections without the need for a code change.

alanmoo:
This will be useful once the inevitable "tickets are no longer available, let's take that page off the nav" or "add this other page to the nav!" asks come in.

Pomax:
This is relatively easy to do (we didn't do it earlier mostly due to time constraints)

PR #3294, opened 2019-06-04
Make primary nav items dynamic

Pomax:
This PR makes the mozfest menu system dynamic, based on whatever are the children of the highest mozfesthomepage ancestor, or if the current page _is_ the homepage, its own children.

This also updates the `primary_active_nav` tag, because it wasn't stripping out slashes on the left of the strings it checks, which it needs to because otherwise we can't compare against the `Page.slug` value that wagtail presupplies.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/3289

### things

test url: https://foundation-mofostaging-pr-3294.herokuapp.com/en/mozilla-festival

Also we now get persistent top-level `active` for free:
- https://foundation-mofostaging-pr-3294.herokuapp.com/en/mozilla-festival/spaces/cpage-1/
- https://foundation-mofostaging-pr-3294.herokuapp.com/en/about/country-red-smile/

### discoveries

the mobile-width nav uses a black underline for the active page. On a black background. I assume we want to change that @mmmavis?


network-api/networkapi/mozfest/models.py
```diff
@@ -63,12 +63,25 @@ def get_template(self, request):
...
+
         # Also make sure that these pages always tap into the mozfest newsletter for the footer!
         mozfest_footer = Signup.objects.filter(name__iexact='mozfest').first()
         context['mozfest_footer'] = mozfest_footer
+
+        # Find the homepage, and then record all pages that should end up as nav items. Note
+        # that subclasses can bypass this, because the MozfestHomepage doesn't need any of
+        # this work to be done.
+        if not bypass_menu_buildstep:
+            homepage = list(filter(
```
Pomax: I really wish python has a nicer syntax for filtering out data, but I _think_ this is the best we can do without using a for loop...

alanmoo: TIL about `lambda` functions in Python.

network-api/networkapi/mozfest/templates/fragments/nav-links.html
```diff
@@ -1,8 +1,7 @@
 {% load primary_active_nav %}
 
-{{ pre }}<a class="{% primary_active_nav request '/spaces' %}" href="/spaces">Spaces</a>{{ post }}
-{{ pre }}<a class="{% primary_active_nav request '/team' %}" href="/team">Team</a>{{ post }}
-{{ pre }}<a class="{% primary_active_nav request '/tickets' %}" href="/tickets">Tickets</a>{{ post }}
-{{ pre }}<a class="{% primary_active_nav request '/sponsors' %}" href="/sponsors">Sponsors</a>{{ post }}
+{% for item in menu_items %}
+  {{ pre }}<a class="{% primary_active_nav request item.slug %}" href="{{ item.url }}">{{ item.title }}</a>{{ post }}
```
alanmoo: General question: Any idea what the deal is with using `pre` and `post` here vs pulling those `div` elements into the markup here?

Actual request: the `active` class doesn't seem to get applied when using the [review app](https://foundation-mofostaging-pr-3294.herokuapp.com/en/mozilla-festival/spaces/). I haven't spun it up locally to check if it does when the festival is on its own domain, but ideally it should work both ways.

Pomax: Yeah:

https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html#L15

vs.

https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html#L37

One of them need `<div>` wrapping, The other does not.

alanmoo:
In order to land this, please update the review app to reflect the proper menu ordering that Percy caught, and make the same changes on production so we don't forget to do that later on.

Pomax:
Updated the review app and foundation.mozilla.org to make sure the items are ordered correctly and only the primary nav entries are marked as "show in menu".

mmmavis:
> the mobile-width nav uses a black underline for the active page. On a black background. I assume we want to change that @mmmavis?

I'll file a ticket for this.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -1072,6 +1081,8 @@ def get_context(self, request):
         # due to our custom image upload approach pre-wagtail
         context = super(Homepage, self).get_context(request)
         context['MEDIA_URL'] = settings.MEDIA_URL
+        context['menu_root'] = self
+        context['menu_items'] = self.get_children().filter(live=True, show_in_menus=True)
```
alanmoo: Can we use `.live().in_menu()` here too?

network-api/networkapi/utility/templatetags/primary_active_nav.py
```diff
@@ -4,13 +4,12 @@
...
-def primary_active_nav(request, target_path):
+def primary_active_nav(request, root_path, target_path):
 
     if not request:
         return ""
 
-    # remove langauge code segment from path to make path comparison easier
-    lang = f"/{request.LANGUAGE_CODE}"
-    path_info = request.path_info[len(lang):] if request.path_info.startswith(lang) else request.path_info
+    if target_path == root_path:
```
alanmoo: What happened to the localization issue here? Does passing "root_path" do something about that? (If so, can you comment that somewhere around here?)

Pomax: both of these are now "the url for that page, for this user", so both of them will always contain the relevant locale code.

network-api/networkapi/utility/templatetags/primary_active_nav.py
```diff
@@ -4,13 +4,12 @@
...
+        return "active" if request.path == target_path else ""
 
-    return "active" if path_info.rstrip("/") == target_path.rstrip("/") else ""
+    return "active" if request.path.startswith(target_path) else ""
```
alanmoo: Following up on above, how does this function know to grab not just the locale?

alanmoo: (I mean, this works, so R+, but I'm a bit confused as to why)

Pomax: "if the request path that django is serving, including all the everythings, starts with the same path (including all the everythings) that is associated with the target_path, then this page is active in the menu, because the user's either on this exact page, or on one of its children"

alanmoo: Ah. We're comparing strings here. That makes much more sense.

network-api/networkapi/templates/fragments/nav-links.html
```diff
@@ -0,0 +1,7 @@
+{% load primary_active_nav %}
+
+{% for item in menu_items %}
+  {{ pre }}<a class="{% primary_active_nav request menu_root.full_url item.full_url %}" href="{{ item.url }}" title="{{ menu_root.full_url }}, {{ item.full_url }}, {{ request.build_absolute_uri }}">{{ item.title|title }}</a>{{ post }}
```
Pomax: `full_url` is used for code comparison purposes because `.url` is not stable (it might or might not yield a relative or absolute path). However, for the actual `href` attribute it doesn't matter whether the url is relative or absolute.

Pomax:
Added a workaround for https://github.com/wagtail/wagtail/issues/5379 by removing the http(s) protocol from all urls involved, which lets us compare with stable strings.

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -11,8 +11,8 @@
               <div class="row">
                 <div class="col">
                   <div class="nav-links pt-3">
-                    <div><a class="{% primary_active_nav request '/' %}" href="/">{{ root_name }}</a></div>
-                    {% include "../fragments/nav-links.html" with pre="<div>" post="</div>" %}
+                    <div><a class="{% primary_active_nav request menu_root.full_url menu_root.full_url %}" href="{{ menu_root.url }}">{{ root_name }}</a></div>
```
Pomax: `.full_url` is used for code comparison purposes because `.url` is not stable (it might or might not yield a relative or absolute path).

However, for the actual `href` attribute it doesn't matter whether the `.url` value is relative or absolute. All that matters is that it's correct in terms of browser navigation.

network-api/networkapi/utility/templatetags/primary_active_nav.py
```diff
@@ -4,13 +4,21 @@
...
     if not request:
         return ""
 
-    # remove langauge code segment from path to make path comparison easier
-    lang = f"/{request.LANGUAGE_CODE}"
-    path_info = request.path_info[len(lang):] if request.path_info.startswith(lang) else request.path_info
+    request_url = request.build_absolute_uri()
 
-    return "active" if path_info.rstrip("/") == target_path.rstrip("/") else ""
+    # In order to work around https://github.com/wagtail/wagtail/issues/5379
```
alanmoo: 😆 gross

alanmoo:
Looks good, just please the CI overlords.

Pomax:
Nnnnnnnnope, that fix didn't work, so "string replace" it is. So dumb =(

Pomax:
That took a while to finalise, but: in it goes.

PR #3295, opened 2019-06-05
DevContainer configuration for VSCode

cadecairos:
This setup requires Docker, VSCode-insiders, and the remote development extension for VSCode.

https://code.visualstudio.com/insiders/
https://code.visualstudio.com/docs/remote/remote-overview


cadecairos:
For those concerned about telemetry in insider builds, you can still disable it with the following in your config file:

```jsonc
{
    // Disable Telemetry
    "telemetry.enableTelemetry": false,
    "telemetry.enableCrashReporter": false,
    "workbench.statusBar.feedback.visible": false,
}
```

alanmoo:
Will this always require the insiders edition or is it just a feature-in-progress?

cadecairos:
> Will this always require the insiders edition or is it just a feature-in-progress?

It will eventually make it to the stable branch. Haven't been able to find a timeline for that yet though.

patjouk:
To test it:
- update to the latest version of vscode. You **don't need** the insider version: latest stable is enough,
- install the [remote-containers](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.remote-containers) plugin. It should be enough but if not, get the [remote development](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) plugins pack,
- open the foundation site project in vscode,
- a pop-up in the bottom right corner will invite you to reload the project using container. Click on reload/yes (look at [vscode instructions](https://code.visualstudio.com/docs/remote/containers#_getting-started) if something is weird),
- vscode will open a new window and will start building the python containers. Wait until it's done,
- try to edit python files: intellisense should work.

Don't forget to use an external terminal when trying to run the foundation site. Runing invoke commands in vscode's terminal will fail (we're running inside the python container).

If you try to run the foundation site, you might run into an issue with node-sass. If that happens, run `docker-compose run --rm  watch-static-files npm rebuild node-sass`.

docs/local_development_with_docker.md
```diff
@@ -86,6 +87,30 @@ Use `invoke docker-npm update`.
...
+## Connecting Docker to your code editor
+
+
+### Pycharm
+
+This feature is only available for the professional version of Pycharm. Follow the official instructions [available here](https://www.jetbrains.com/help/pycharm/using-docker-as-a-remote-interpreter.html#config-docker)
+
+### Visual Studio Code
+
+Visual Studio Code use a feature called Dev Container to run Docker projects. The configuration files are in the `.devconatainer` directory. This feature is only available starting VSCode 1.35 stable. For now, we're only creating a python container to get Intellisense, we're not running the full project inside VSCode.
```
alanmoo: Can you add some info about why "For now" (like what we're blocked on)?

patjouk: Chris tried to set up docker compose for vscode and made some progress but the workflow was sadly terrible :( The "for now" is me acknowledging that the setup could be better and being okay to revisit it if Docker support for vscode gets better.

alanmoo:
```suggestion
Visual Studio Code use a feature called Dev Container to run Docker projects. The configuration files are in the `.devconatainer` directory. This feature is only available starting VSCode 1.35 stable. For now, we're only creating a python container to get Intellisense, we're not running the full project inside VSCode. We may revisit this in the future if Docker support in VSCode improves.
```

alanmoo:
I'm a bit confused about what's happening here, and how to debug the issues I'm hitting. Is this using a different container than the one that gets spun up with `docker-compose up`? It seems I need to reinstall Python dependencies.

alanmoo:
After doing some installs in a Python env (in which I'm not entirely sure what's happening), things seem to be working! One type of import consistently *doesn't* though: 
image

Any idea what's going on there? Is it an environment issue or does it have to do with how we're writing those import statements?

patjouk:
Intellisense needs to be connected to a python environment to work: the problem is that vscode can't do that out of the box for python environments that are inside a container. To fix that issue, we give vscode a container using the image built using `dockerfile.python`. It's the same image that we're using to run the dev website but those containers won't be talking together. To summarize:
- We have one container used only for vscode and Intellisense,
- We have the 3 services (DB, front and back) for running the website created with `docker-compose up`.
Same image, different  containers, different purposes.

I might have an idea of why you're running into the unresolved import thing, but not sure yet how to fix it. I'll look into it.

patjouk:
All imports work on my laptop and I don't get this warning :(
Can you check that you're using the remote python? At the bottom of vscode, click on python next to `Dev Container`. It should look like this:
Screenshot 2019-06-13 at 10 40 36

If that doesn't work, try closing vscode and reopening it: it fixed many weird issues for me >.<

alanmoo:
Hrm, no, still being weird. Maybe we can pair on this for a few minutes next week.

patjouk:
Turns out it was an issue existing before using Docker for VSCode. Fix is explained [here](https://github.com/Microsoft/vscode-python/issues/3840#issuecomment-455761457).

Issue #3324, opened 2019-06-11
Reactivate donate popup


kristinashu:
Reactivate the donate popup the same as it was before but will new copy.

### Copy (WAITING FOR APPROVAL)
> Big corporations try to restrict how we access the web. 

> Misinformation makes it harder for us to find the truth. Web-connected devices go to market without basic security standards. 

> The non-profit Mozilla Foundation fights for a healthier internet. Will you donate today?

Where "Will you donate today?" is in bold.

### Comp
https://redpen.io/ps2f6569e046e136bf

<img width="1003" alt="image" src="https://user-images.githubusercontent.com/1068783/59299243-08358980-8c41-11e9-9b18-5e81c8247530.png">

Pomax:
I also remember @xmatthewx saying we wanted to have it be dismissable for at most a week, so I will change the session storage approach to a localstorage approach instead, and then if people dismiss the banner, they will see it again one week later (if they happen to load up the site again =)

kristinashu:
Yes! That would be much less annoying.

Pomax:
Do we still want a 10 second timeout, or do we want to "tighten that up"?

kristinashu:
10 sec is good! Want them to spend some time on the site before we ask them for money.

PR #3325, opened 2019-06-11
initial rewrite for one-week dismissals

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3324

This switches the donate modal to one that can be dismissed for a week at a time. It will pop up after 10 seconds, and will keep popping up until dismissed. Once dismissed, the time of dismissal is stored in localStorage, and once a week has passed, the modal will start popping up again until dismissed again.

### DIFF
--- a/source/js/donate-modal/donate-modal.jsx
+++ b/source/js/donate-modal/donate-modal.jsx
@@ -11,8 +11,8 @@ class DonateModal extends React.Component {
   constructor(props) {
     super(props);
 
-    const INITIAL_DELAY = parseInt(this.props.delay) || 10 * MILLISECONDS; // 10 seconds, in ms
-    const DISMISSAL_DELAY = parseInt(this.props.hideFor) || 7; // one week, in days
+    const INITIAL_DELAY_IN_MILLISECONDS = parseInt(this.props.delay) || 10 * MILLISECONDS; // 10 seconds, in ms
+    const DISMISSAL_DELAY_IN_DAYS = parseInt(this.props.hideFor) || 7; // one week, in days
 
     // Don't treat the modal as dismissed if:
     // 1. the user never dismissed it before, or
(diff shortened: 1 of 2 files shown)


kristinashu:
Looks good at all screen widths except at super narrow (<335px) when the button gets cut off. But I think that's ok because I doubt anyone would go that narrow.

image

source/js/donate-modal/donate-modal.jsx
```diff
@@ -2,58 +2,47 @@ import React from "react";
...
+    if (lastLoad) {
+      var diff = (Date.now() - parseInt(lastLoad, 10)) / DAY;
+      if (diff < DISMISSAL_DELAY) {
+        dismissed = true;
+      }
+    }
 
     this.state = {
-      delay: DELAY - prevTimer,
+      delay: INITIAL_DELAY,
```
mmmavis: if value of `delay` won't be changed, should we turn it into a class property instead a state?

source/js/donate-modal/donate-modal.jsx
```diff
@@ -2,58 +2,47 @@ import React from "react";
...
+const localStorage = typeof window === "undefined" ? {} : window.localStorage;
 
 class DonateModal extends React.Component {
   constructor(props) {
     super(props);
 
-    let dismissed = !!sessionStorage.getItem(KEY_STATE) || false;
-    let prevTimer = this.getPrevTimer();
+    const INITIAL_DELAY = parseInt(this.props.delay) || 10 * MILLISECONDS; // 10 seconds, in ms
+    const DISMISSAL_DELAY = parseInt(this.props.hideFor) || 7; // one week, in days
```
mmmavis: I was thinking maybe we could rename these two variables into `INITIAL_DELAY_IN_MILLISECONDS` and `DISMISSAL_DELAY_IN_DAYS` for clarity?

source/js/donate-modal/donate-modal.jsx
Pomax: it can be changed through a prop, though.

source/js/donate-modal/donate-modal.jsx
Pomax: good call

Issue #3323, opened 2019-06-10
Implement Social Buttons


beccaklam:
We have defined some new button styles (more like filling in gaps we had before in our style guide) and they need to be implemented!

## Social Buttons

We reuse these buttons a lot so let's standardize them:

### Regular Buttons:
- [ ] Facebook

- [ ] Twitter

- [ ] Email

- [ ] Link

![social-buttons](https://user-images.githubusercontent.com/20977376/59519376-b6d20800-8e95-11e9-9dad-15cf7d0ec2af.jpg)

### Circular Buttons:
- [ ] Facebook

- [ ] Twitter

- [ ] Email

- [ ] Link
![circular-social-buttons](https://user-images.githubusercontent.com/20977376/59519568-21834380-8e96-11e9-9610-8372327e362d.jpg)

### Participate Page:
![participate@2x](https://user-images.githubusercontent.com/20977376/59223565-42981b80-8b9a-11e9-90b1-52c1ef7eb94e.jpg)
### Tablet
![tablet@2x](https://user-images.githubusercontent.com/20977376/59223571-4461df00-8b9a-11e9-8252-d0506ffa366a.jpg)
### Mobile
![mobile@2x](https://user-images.githubusercontent.com/20977376/59223574-462ba280-8b9a-11e9-8843-ab90621a9bbf.jpg)

Related ticket: MozillaFoundation/Design#371
Redpen: https://redpen.io/p/knefb82f90a4716342
Assets: 
[social-button-icons.zip](https://github.com/mozilla/foundation.mozilla.org/files/3291243/social-button-icons.zip)

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] Link to original design ticket
- [x] Link to related ticket
- [x] Link to Redpen
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Assets are included

suhailsinghbains:
Hi, Can I please work on this?
Thanks

mmmavis:
Hi @suhailsinghbains This ticket is part of a bigger work that is currently being tackled by our team. Thanks for offering help but I think this might not be the best ticket to assign to contributors.

Issue #3322, opened 2019-06-10
Implement Grouped Button 'Style'


beccaklam:
We have defined some new button styles (more like filling in gaps we had before in our style guide) and they need to be implemented!

## Grouped Buttons
Grouped in-line icon buttons should be 16px apart from the left/right, and top/bottom. On desktop grouped buttons are laid out in a line, on mobile the group can switch to two rows. Grouped buttons are left aligned (not centered).

How wide the buttons are depends on the longest word in the grouped button set (with the other buttons following its width), aiming to have 16px of spacing between the word and the right and left side of the button (see the Facebook button as an example).

![grouped-button-style](https://user-images.githubusercontent.com/20977376/59221431-42495180-8b95-11e9-9056-fce6600d33f7.jpg)

## Affected Page
Post-sign Campaign Social Share
![image](https://user-images.githubusercontent.com/20977376/59221465-53925e00-8b95-11e9-9cb7-c451047a2d89.png)

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [ ] Design has been finalized
- [ ] Link to original design ticket
- [ ] Link to related ticket
- [ ] Link to Redpen
- [ ] Design changes been merged into the master Sketch file
- [ ] Desktop and mobile mockups are included
- [ ] Hover state for elements are designed

mmmavis:
@beccaklam I tested a petition form locally and the three share buttons look very different from what you included in the ticket description + there's no "link" button. Did you take your [screencap](https://user-images.githubusercontent.com/20977376/59221465-53925e00-8b95-11e9-9cb7-c451047a2d89.png) from a live campaign page or from Sketch?

<img width="1095" alt="image" src="https://user-images.githubusercontent.com/2896608/60144696-3f935280-9778-11e9-977e-6ecf42d9d0e0.png">

beccaklam:
Hi @mmmavis, sorry if this wasn't clear. What you have above is the current version, but we want to update to the screen cap that I provided! I guess there perhaps on order to all the button updates with this being one of the later ones?

beccaklam:
![image](https://user-images.githubusercontent.com/20977376/60190056-00a8df80-9800-11e9-9bd3-f45960ef397a.png)
☝️ So here are all the button implementation tickets that stemmed from this ticket: https://github.com/MozillaFoundation/Design/issues/371
[...]

PR #3352, opened 2019-06-26
Added 'share button group' style + example on /styleguide

### LINKED ISSUE #3321: Implement In-line Icon Button Style
We have defined some new button styles (more like filling in gaps we had before in our style guide) and they need to be implemented!

## In-line Icon Button

- [x] Icon should be 16px in height, width can vary (with 16px of spacing between the grouped icon and label and the right/left edges of the button, see grouped buttons ticket for reference: #3322)

- [x] Icon should be spaced 8px from the label

- [x] Label and Icon should then be grouped and centered in button
in-line-icon-button

Related ticket: MozillaFoundation/Design#371
Redpen: https://redpen.io/p/knefb82f90a4716342

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] Link to original design ticket
- [x] Link to related ticket
- [x] Link to Redpen
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Hover state for elements are designed

mmmavis:
(Will assign a code reviewer once this passes design review)

Related issues: #3321 #3322 #3323 

---

In this PR I've introduced a new component called "share-button-group". Currently there are four versions: see examples at the bottom of the [style guide](https://foundation-mofostaging-pr-3352.herokuapp.com/en/styleguide/)

@beccaklam can you review and make sure design is as expected? This PR doesn't close any ticket. My plan is to get this component in and

1. get share buttons on blog pages
2. update existing share buttons on the site (this will take longer so I wanna do it one by one) and eventually close  #3321 #3322 #3323 

---

@Pomax :

I went with the React component approach because

- There are multiple versions of this share button group. In order for the right style to kick in, markups need to follow particular pattern + have the right `class` on elements.
- React component vs Django template: If it's a Django template we can't "import" and use it in a React component.


source/js/copy-to-clipboard.js
```diff
@@ -0,0 +1,83 @@
+const copyToClipboard = (linkElement, textToCopy) => {
```
mmmavis: [to future code reviewer]

This is exactly the same copy of [`source/js/buyers-guide/copy-to-clipboard.js`](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/buyers-guide/copy-to-clipboard.js). Ultimately I want PNI to use this new file that I created here so we don't have duplicate of the same file. I did't wanna do it in this PR to avoid complication. If this sounds good I'll create a follow-up ticket and tackle it right after this PR gets merged.

network-api/networkapi/wagtailpages/templates/wagtailpages/static/styleguide.html
```diff
@@ -307,6 +307,38 @@ <h3 class="h2-heading mt-5">Dark Theme</h3>
...
+      <h3 class="h2-heading mt-5">Full Version & Stacked</h3>
+      <hr />
+
+      <div class="mb-5">
+        <div class="share-button-group-wrapper" data-layout="stacked"></div>
+      </div>
+
+
+      <h3 class="h2-heading mt-5">Mini Version</h3>
+      <hr />
```
Pomax:
```suggestion
      <hr>
```

network-api/networkapi/wagtailpages/templates/wagtailpages/static/styleguide.html
```diff
@@ -307,6 +307,38 @@ <h3 class="h2-heading mt-5">Dark Theme</h3>
...
+
+      <div class="mb-5">
+        <div class="share-button-group-wrapper" data-version="mini"></div>
+      </div>
+
+      <h3 class="h2-heading mt-5">Mini Version & Stacked</h3>
+      <hr />
```
Pomax:
```suggestion
      <hr>
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+  componentDidMount() {
+    if (this.props.whenLoaded) {
+      this.props.whenLoaded();
+    }
+  }
+
+  renderFacebookButton() {
+    let label =
+      this.props.version === `mini` ? (
+        <span class="sr-only">Share by Facebook</span>
```
Pomax: share "on" facebook?

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+      ) : (
+        `Faceook`
+      );
+
+    let link = this.props.link ? this.props.link : ``;
```
Pomax:
```suggestion
    let link = this.props.link || ``;
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+
+    return (
+      <a
+        class="btn btn-secondary btn-share facebook-share"
+        href={`https://www.facebook.com/sharer/sharer.php?u=https://${link}`}
```
Pomax:
```suggestion
        href={`https://www.facebook.com/sharer/sharer.php?u=${ encodeURIComponent(`https://${link}`) }`}
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+      >
+        {label}
+      </a>
+    );
+  }
+
+  renderTwitterButton() {
+    let shareText = this.props.shareText ? this.props.shareText : ``;
```
Pomax:
```suggestion
    let shareText = this.props.shareText || ``;
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+    let label =
+      this.props.version === `mini` ? (
+        <span class="sr-only">Share by Twitter</span>
```
Pomax: probably also share "on" twitter, though I've seen "via" as well... maybe needs a quick check with someone who writes copy

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+          shareText
+        )}`}
+      >
+        {label}
+      </a>
+    );
+  }
+
+  renderEmailButton() {
+    let shareText = this.props.shareText ? this.props.shareText : ``;
```
Pomax:
```suggestion
    let shareText = this.props.shareText || ``;
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+        <span class="sr-only">{label} page link</span>
+      ) : (
+        label
+      );
+
+    let classes = classNames(`btn btn-secondary btn-share link-share`, {
+      copied: this.state.linkCopied
+    });
+
+    let onClickHandler = event => {
```
Pomax: Does this need to be inlined, or will this work as a class level function, so we can use `onClick={ () => this.handleLinkClicked() }`?

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+      event.preventDefault();
+
+      copyToClipboard(event.target, window.location.href);
+      this.setState({
+        linkCopied: true
+      });
+    };
+
+    return (
+      <a class={classes} href="#" onClick={onClickHandler}>
```
Pomax: Under HTML5 rules, `href="#"` means "navigate to the top of the page" so if this is not _actually_ a link, we should probably use a `<button>` here instead.

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+        {this.renderTwitterButton()}
+        {this.renderEmailButton()}
+        {this.renderLinkButton()}
+      </div>
+    );
+  }
+
+  render() {
+    return (
+      <>
```
Pomax: let's not use the short code here... no other SGML/HTML/XML specification allows empty tags, so in order to keep the code readable it's better to make it explicit what we're doing here.
```suggestion
      <React.Fragment>
```

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,152 @@
...
+        {this.props.version === `mini`
+          ? this.renderCircleButtons()
+          : this.renderRectangleButtons()}
+      </>
```
Pomax:
```suggestion
      </React.Fragment>
```

source/js/components/share-button-group/share-button-group.jsx
mmmavis: I was trying to follow how PNI's copy link does it so things can be more aligned. Can we file a follow-up ticket for this so we can fix both `href="#"` occurances?

source/js/copy-to-clipboard.js
mmmavis: Ticket filed!

source/js/components/share-button-group/share-button-group.jsx
Pomax: we totally can - https://github.com/mozilla/foundation.mozilla.org/issues/3366

Pomax:
@beccaklam @mavis for the mini versions, should those have a "tooltip" that says what they are? Or, at the very least the copy link button? (we might need to think about how that one works in follow up, too, because as "copy link" it feels like that should also just ... be a real link? which is hard to combine with click-to-copy functionality)

mmmavis:
@Pomax 

> [...] as "copy link" it feels like that should also just ... be a real link? which is hard to combine with click-to-copy functionality)

Not sure if I follow 😬 Did you mean clicking the "copy link" button should expose the full page URL somehow + auto copy the url to clipboard?

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -0,0 +1,158 @@
...
+  componentDidMount() {
+    if (this.props.whenLoaded) {
+      this.props.whenLoaded();
+    }
+  }
+
+  renderFacebookButton() {
+    let label =
+      this.props.version === `mini` ? (
+        <span class="sr-only">Share on Facebook</span>
```
mmmavis: Changed this from `Share by Facebook` to `Share on Facebook`. I can ping Kevin tomorrow to verify the wording.

Pomax:
Yeah... if you think about what "copy link" implies, it's that there is a real link (e.g. `<a href="moz/whatever/page.html">` that you can copy so you can then paste it somewhere. In this case, the chain icon would suggest that the button _is_ that link element, but instead it's a non-link that you need to click in order to copy the link.

That makes a ton of sense on mobile, but for desktop users that's kind of weird? I don't know what the best thing to do there is, though.

beccaklam:
> Yeah... if you think about what "copy link" implies, it's that there is a real link (e.g. `<a href="moz/whatever/page.html">` that you can copy so you can then paste it somewhere. In this case, the chain icon would suggest that the button _is_ that link element, but instead it's a non-link that you need to click in order to copy the link into your clipboard.
> 
> That makes a ton of sense on mobile, but for desktop users that's kind of weird? I don't know what the best thing to do there is, though.

Sorry @Pomax I'm still not following. Are you saying a user would expect the small link button to open a new link instead of producing a link that is copied to the user's clipboard that they can share?

mmmavis:
@beccaklam from @Pomax 's [comment](https://github.com/mozilla/foundation.mozilla.org/pull/3352#issuecomment-508271940)

> @beccaklam @mmmavis for the mini versions, should those have a "tooltip" that says what they are? Or, at the very least the copy link button? (we might need to think about how that one works in follow up, too, because as "copy link" it feels like that should also just ... be a real link? which is hard to combine with click-to-copy functionality)

What do you think about adding a tooltip to the mini version copy link button? Something like below
tooltip-link-button

beccaklam:
> @beccaklam from @Pomax 's [comment](https://github.com/mozilla/foundation.mozilla.org/pull/3352#issuecomment-508271940)
> 
> > @beccaklam @mmmavis for the mini versions, should those have a "tooltip" that says what they are? Or, at the very least the copy link button? (we might need to think about how that one works in follow up, too, because as "copy link" it feels like that should also just ... be a real link? which is hard to combine with click-to-copy functionality)
> 
> What do you think about adding a tooltip to the mini version copy link button? Something like below
> tooltip-link-button

That's fine by me! Seems useful :)

PR #3354, opened 2019-06-27
Selenium-based Interaction testing

Pomax:
This is the start of full site interaction testing using Selenium. It's not comprehensive, but covers the bases: Selenium (tapping into gecko and chrome driver), using chai's `assert` style of assertion testing, and npm scripts to kick everything off.

1. `npm install` because there's a few new dependencies. Thankfully, none of them 100s of MBs
2. Visit https://github.com/mozilla/geckodriver/releases and download the geckoDriver for your platform, then copy it into the repo's root dir
3, Visit http://chromedriver.chromium.org and do the same for the chromeDriver
4. Update your `.env` and add the following keys, with the locations of the binary executables for firefox and chrome filled in:

```
FIREFOX_BINARY="..."
CHROME_BINARY="..."
```

5. Provided you've done those things, `npm run selenium` should now kick off a site interaction test

The tests are currently simple:

- load the homepage
- verify that the `About Us` primary nav entry exists and follow it
- verify that the `About Us` page has a secondary nave, with 7 nav items
- verify that the first item in the secondary nav is marked as `active` through a class.

And then we can start building this out.

### Still left

Update our docs, but it's important to verify things work independently of that.


mmmavis:
@Pomax I got `ERROR: Task not found: "test:site:selenium"` from running `npm run selenium`

package.json
```diff
@@ -29,6 +29,9 @@
     "test:scss": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" --syntax scss",
     "test:scss:styleguide:color": "stylelint \"source/sass/**/*.scss\" \"source/js/**/*.scss\" \"!source/sass/**/_colors.scss\" --syntax scss --config .stylelintrc-colors.js",
     "test:css": "stylelint \"network-api/networkapi/{,!(frontend)/**/}*.css\" --syntax scss",
+    "selenium": "run-p --race server test:site:selenium",
```
mmmavis: @Pomax I got `ERROR: Task not found: "test:site:selenium"` from running `npm run selenium`.

package.json
```diff
@@ -29,6 +29,9 @@
...
+    "selenium:run": "mocha ./test/selenium --timeout 60000",
+    "selenium:site:run": "wait-on tcp:8000 && npm run test:selenium",
```
mmmavis: Is this task `test:selenium` missing too?

.gitignore
```diff
@@ -116,3 +116,6 @@ network-api/networkapi/frontend/*
 
 # Cypress visual regression testing
 cypress/videos
+
+# Selenium drivers
+*driver.*
```
patjouk:
```suggestion
*driver*
```
That will ignore drivers on non windows system.

env.default
```diff
@@ -75,3 +75,7 @@ USE_CLOUDINARY=False
 CLOUDINARY_CLOUD_NAME=""
 CLOUDINARY_API_KEY=""
 CLOUDINARY_API_SECRET=""
+
+# Selenium testing
+FIREFOX_BINARY=
```
patjouk: If you don't add quotes, dotenv inside pipenv will make a mess and think that CHROME_BINARY is the value of FIREFOX_BINARY. FUN STUFF 😞

test/selenium/main.js
```diff
@@ -0,0 +1,82 @@
+// Before we do ANYTHING else, verify that there are drivers in the main dir.
+// If there aren't, then this is going to crash and burn, and whoever's trying
+// to run this should "install" the firefox and chrome drivers for their OS.
+
+const fs = require('fs');
+
+const geckoDriver = fs.existsSync('./geckoDriver') || fs.existsSync('./geckoDriver.exe');
```
patjouk: Could we use the env value from the .env and .docker-env instead? Just in case people want to put it elsewhere.

patjouk:
I forgot to add: Why not write those tests in python? It would give the opportunity to write those tests to more people on the team.

package.json
Pomax: no, because there's a `test:*` as part of the main test running, and selenium should _not_ run on every test run.

env.default
Pomax: ... what, that's crazy. Empty quotes it is =_=

test/selenium/main.js
Pomax: Based on how the selenium webdriver code works. It _only_ works with the drivers either in the dir you run selenium, which is easy for people to effect, or with the webdrivers installed somewhere else and then with the system PATH updated, which should never be a requirement for individual projects.

Pomax:
Mostly because in the world of front end testing, Selenium is the main player, and all its tutorials and examples are based on JS. Even MDN [advocates its use for testing](https://developer.mozilla.org/en-US/docs/Learn/Tools_and_testing/Cross_browser_testing/Your_own_automation_environment) so I figured we'd be in good company just sticking with that.

package.json
Pomax: that should be `selenium:site:run`

env.default
Pomax:
```suggestion
FIREFOX_BINARY=""
```

env.default
```diff
@@ -75,3 +75,7 @@ USE_CLOUDINARY=False
...
+CHROME_BINARY=
```
Pomax:
```suggestion
CHROME_BINARY=""
```

package.json
Pomax:
```suggestion
    "selenium": "run-p --race server selenium:site:run",
```

package.json
Pomax:
```suggestion
    "selenium:site:run": "wait-on tcp:8000 && npm run selenium:run",
```

package.json
Pomax: updated

env.default
Pomax: updated

test/selenium/main.js
patjouk: OK! So why do we need those two env vars for?

Pomax: telling selenium where to find the actual Firefox and Chrome executables, as the webdriver will need to fire those up in headless mode.

alanmoo:
Reading through the setup instructions, I'm concerned about how far away this moves us from our current "Configuration as code" setup that we've worked hard to get to. How can we bring this all into Docker?

patjouk:
Like Alan mentioned, it's not possible to run it with Docker for the moment. Happy to pair for an hour if you're not sure how to start.

alanmoo:
We may have opportunity to pick this thread back up again next sprint. @patjouk is the next action for this "Get it working in Docker" or is there more Ops side planning to do before continuing down this path?

Pomax:
I'm closing this PR in favour of using cypress, which does not require the webdriver shenanigans, but does offer full interaction testing.

Issue #3275, opened 2019-05-30
Create Mozfest 404 page


alanmoo:
If someone hits a 404 on Mozfest, they should see the error message in a Mozfest header/footer instead of the foundation site's.

alanmoo:
@Pomax Any thoughts on how this might work? I couldn't figure out how we can change the template based on the request domain.

Pomax:
I can't see to find any documentation on how to achieve this, so I've posed https://stackoverflow.com/questions/56449188/having-a-404-for-each-site-in-a-wagtail-multisite-setup and hopefully there is a way to easily achieve this, I just can't find what it is.

Pomax:
unblocked by https://github.com/wagtail/wagtail/pull/5375/files

PR #3378, opened 2019-07-09
create a dedicated mozfest 404 page, triggering on mozillafestival.org

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3275

I'm not actually sure how this can best be tested: the 404 routing now relies on host filtering, which of course ideally we want to do based on the `sites` entries in wagtail. This code does not do that right now (even better would be if wagtail let you specify a 404 in its sites admin, but we're not that far along yet).

The easiest way to test this right now is to check out the branch, chack the `mozillafestival.org` host to `localhost` instead in the new `wagtailpages/views.py` 404 handler, and then loading up a URL you know is bogus **with DEBUG=False** because I haven't figured out a way to make 404 pages show up propertly with DEBUG=True yet - there's a number of supposedly working ways to make that happen by changing the 404 url handling, but I haven't managed to make any of them work.

I suspect this is a WIP PR because it'd be nice to still add in that `sites` filtering, and ideally, 404-with-DEBUG=True, but given that mozfest currently has no 404 of its own at all, it might be worth landing first with immediate followup for those issues.

**A concern**

The package-lock.json is showing _a ton_ of dependencies suddenly resolving on `http` rather than `https`. I have no idea why, npm install manages that for you, but the fact that it's suddenly no longer https is ... worrisome?

**A confusion**

Travis is failing on something that I'm not sure I understand. I can't reproduce it locally (`inv test` passes just fine) and the error messages are... puzzling


Pomax:
Screen Shot 2019-07-08 at 5 46 04 PM

@gideonthomas @kristinashu, I based the 404 off of what we have on the foundation site - do we need anything more tailored to mozfest or does this look fine?

kristinashu:
Looks good to me (as long as the "Go to home page" button links to the mozfest home page)!

Pomax:
yep, it simply points to `/` on whatever domain is loading it, so on mozillafestival.org that's what it'll point to.

network-api/networkapi/wagtailpages/views.py
```diff
@@ -0,0 +1,24 @@
...
+
+    It would be great if we could pull the "which domain uses
+    which 404 template" information from the wagtail "sites"
+    configuration, but there is no way to know which template
+    belongs to which site, as "a site" is not tied to "a django
+    app" in the wagtail way of things.
+    """
+    if request.site.hostname == 'mozillafestival.org':
+        html = render(request, 'mozfest/404.html')
+        return HttpResponseNotFound(html)
```
patjouk:
```suggestion
        return HttpResponseNotFound(html.content)
```

cf https://code.djangoproject.com/ticket/29238

network-api/networkapi/wagtailpages/views.py
```diff
@@ -0,0 +1,24 @@
...
+    else:
+        html = render(request, '404.html')
+        return HttpResponseNotFound(html)
```
patjouk:
```suggestion
        return HttpResponseNotFound(html.content)
```

patjouk:
Can you make sure to use postgresql locally? It's what we're using in prod and CI and it would have helped you reproduce this bug locally.

Pomax:
I was, which is why I was so confused. I run 10.6 with a copy of the staging db because I can't really do meaningful dev work otherwise, so I don't understand why this wasn't throwing any errors before. I blasted away my repo, and my db, then recloned and checked out this branch, and did a new staging db copy, and now the error does show up locally so I have _no_ idea why I wasn't see it before.

Thanks for helping debug this.

patjouk:
Those DBs are not connected: [Django creates](https://docs.djangoproject.com/en/2.2/topics/testing/overview/#the-test-database) a new empty db for running tests.

Issue #3379, opened 2019-07-09
Remove `featured` property from News model


Pomax:
Do-over of https://github.com/mozilla/foundation.mozilla.org/issues/721, which ran into enough problems that it's far easier to retry this "fresh" than it is to continue adding to that issue.

Pomax:
as far as I can tell we already removed all use of the news.featured flag, so the only work left here is a PR that removes the field itself.

alanmoo:
![dragons](https://user-images.githubusercontent.com/1682681/60987370-74bea980-a30f-11e9-9756-9787eb815b19.jpg)

Pomax:
estimated 1 because the work's technically already done...

alanmoo:
![image](https://user-images.githubusercontent.com/1682681/62656711-d2f0a380-b932-11e9-96ef-3bb112cb0932.png)

Issue #3296, opened 2019-06-05
Confirm deadness of and remove vestigial file


alanmoo:
**Describe the bug**
The [signup component](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta/signup.html) previously used in the Opportunity page templates appears to be vestigial. This needs to be confirmed and any dead code around it should be shaken out.

Sharma-palak:
Hi @alanmoo ,i am new to open source ,can I take this issue??

alanmoo:
You're welcome to open a PR if the description is clear enough, but I don't have capacity to support figuring out what needs to happen right now.

Sharma-palak:
ok @alanmoo , I want to give it a try.

youriwims:
@alanmoo as far as I can tell, it looks like this is currently in use on two bannered campaign pages:

1. https://foundation.mozilla.org/en/fellowships/newsletter/
2. https://foundation.mozilla.org/en/newsletter/

But they don't appear to be _real_ pages...or are they?

alanmoo:
Those are very real, but are you sure that code is used? I thought they're using [this code](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html#L64-L71) to render the div that React hooks into (notice the extra data attributes.)

In addition, (this file)[https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/basket-signup.js] is another one I'm suspect of its actual use these days (and the one I was thinking might actually be useful in Pulse, @mmmavis )

youriwims:
Okay, makes more sense. Yeah, it doesn't look like basket is being used anymore either.

alanmoo:
![](https://media.giphy.com/media/65NO1TrKrJUT6/giphy.gif)

PR #3380, opened 2019-07-09
remove news.featured property

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3379
Related PRs/issues #721, https://github.com/mozilla/foundation.mozilla.org/issues/3296

The last time we landed this we ran into a caching/desync issue where the old code was using the updated db and things crashed because the old code was trying to load a column that the new db didn't have anymore...

I don't know if that's still going to be the case, but last time we discovered that rather than rolling back the code change, we could have waited 2 minutes for things to sync back up, so.... maybe we'll have to do this again when we deploy?


Pomax:
Yeah, I'm pretty sure we're not using it anywhere anymore, which is why I'm hopeful it won't be a problem this time round anymore. But that template update was a great idea!

PR #3406, opened 2019-07-12
Fix Comment field localization in petition forms

TheoChevalier:
The new linter in #2796 removed quotes around the comment field translations, and it’s showing in English on production (see https://foundation.mozilla.org/fr/campaigns/tell-amazon-connected-devices-must-have-privacy-policies/chat-amazon-customer-service/ for instance)

I’m assuming adding quotes back will fix the issue. @pomax, any exclusion rule to be added somewhere for the linter? Or should we use something else than double quotes?


Pomax:
hm, that's a tricky one. What are we currently using to read in this jsx? Does that have the ability to use `Object.keys`?

alanmoo:
Found it. See #3466.

PR #3420, opened 2019-07-15
Do pulse project ordering via the API

### LINKED ISSUE #3419: Pulse project ordering doesn't work
**Describe the bug**
This was noticed on the MozFest site. https://www.mozillapulse.org/entry/1200 was a post whose created date was set to the end of the year in the admin interface (so that it would show up first on the site). This worked fine until we hit the limit per page in the API for pulse projects...that post went to the second page and hence no longer showed up on the MozFest site.

Reason:
For the pulse project block, we sort the projects in JS after we fetch the, from the API. This becomes problematic since the projects sent by the API are paginated. So when we sort projects in JS, we're only sorting a subset of projects.

**Expected behavior**
Projects that should be ordered by created date should be ordered for all projects that match the filter query, not just those on the first page.

gideonthomas:
Fix #3419 

This PR fixes an ordering issue with loading pulse projects on the foundation site. It basically changes the way we order projects by offloading that logic to the API vs. doing it on the front-end. See the ticket for why.

Assigning to @alanmoo and @mmmavis...whoever gets to it first.

### DIFF
--- a/source/js/components/pulse-project-list/pulse-project-list.jsx
+++ b/source/js/components/pulse-project-list/pulse-project-list.jsx
@@ -43,7 +43,7 @@ export default class PulseProjectList extends React.Component {
       page_size: this.props.max ? this.props.max : 12,
       search: this.props.query,
       featured: this.props.featured && `True`,
-      ordering: this.props.reverseChronological ? `-created`, `created`
+      ordering: this.props.reverseChronological ? `-created` : `created`
     };
 
     // Serialize parameters into a query string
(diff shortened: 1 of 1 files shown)


alanmoo:
Travis failed. Can you also include a link to a page on the review app that demonstrates it functioning?

alanmoo:
(Code looks reasonable though)

gideonthomas:
For review, compare the contents of the pulse listing here: https://www.mozillafestival.org/en/#latest-festival-news vs. here: https://foundation-mofostaging-pr-3420.herokuapp.com/en/mozilla-festival/#latest-festival-news. Notice that the second one has Mark's post in it which is correct since its created date is set to Dec 31, 2019.

Issue #3377, opened 2019-07-08
Add localization support for ShareProgress buttons in petition snippets


alanmoo:
**Describe the bug**
We need to be able to add a different URL for each shareProgress button in different languages.

**To Reproduce**
1. Edit a petition
2. Note the Share link fields aren't localized
<img width="929" alt="image" src="https://user-images.githubusercontent.com/1682681/60828352-2c24b600-a180-11e9-82d4-34a3fae9bfb2.png">

**Expected behavior**
When toggling languages, a URL field for the share link should exist for each supported language

alanmoo:
@TheoChevalier This has landed on master, will be live by Monday morning EDT, maybe earlier.

TheoChevalier:
> @TheoChevalier This has landed on master, will be live by Monday morning EDT, maybe earlier.

Yay! Thanks so much folks!

Issue #70, opened 2017-02-10
Identify featured projects for launch


xmatthewx:
Currently, featured items on Pulse are picked by yours truly without much rigor. We should look at items and pick 3 for homepage and 9-18 for the projects page. Using a similar rubric that we hacked together for offerings. 

We should avoid Mozilla things and focus on stuff by the network. I think MozFest talks are an exception to that rule, though I would choose image and copy to emphasize the person.

@KevZawacki – how should we do this? Async. On a call? Anyone else you think should join us on the task?

KevZawacki:
@xmatthewx We can collaborate asynchronously, in a spreadsheet.

I think the most important characteristic is accessibility: Is it clear what the project is and why it exists?

I've highlighted Pulse projects that meet that criteria **[here](https://docs.google.com/a/mozilla.com/spreadsheets/d/1MVOTTxNH4FvyoiBIR--F8HqZYhBvlL2LSKi4XDa1LKo/edit?usp=sharing
)**. Feel free to add / subtract. We can tinker with their eventual arrangement in the second tab. 

@Lottarao and @HPaulJohnson may want to weigh in, too.

xmatthewx:
@KevZawacki – I've [featured](http://mozilla.github.io/network-pulse/) the Zeynep Tufekc talk you posted to Pulse now that we have an image. Looking good. 

I'd like to suggest that we remove the second half of this item title, and do the same for other entries. The Algorithmic Spiral of Silence ~~| Zeynep Tufekci at MozFest~~

I think the descriptions are short enough that we don't need a subhead. If you do think we need subheads, I'd like to design it as a distinct field.

KevZawacki:
@xmatthewx Lookin' fresh!

Agreed that we can do without the subhead.

xmatthewx:
Thanks @KevZawacki. I'll will flag your suggests as featured after we launch new Pulse this week. 😄

PR #3422, opened 2019-07-16
made share links/text localizable

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3377

Ignore the branch name, it's just "share", not "shareprogress" - this adds several of our share fields to the set of localizable fields, so that we can serve different urls/text for different locales where needed.


Pomax:
1. huh, I could have sworn I added that, so: yes, absolutely.
2. as long as l10n-sync and l10n-update run (oh right... remember to run those locally! just `migrate` is not enough when updating localization!) the english field becomes the default text for the other locales.

Pomax:
Yeah, that's because of the migration changeover - the deploy script only runs on review app creation so it got stuck on the fact that I deleted the old #70 and built a new one. Deleting and recreating the app shows https://foundation-mofostaging-pr-3422.herokuapp.com/cms/pages/14/edit/ working.

alanmoo:
Not surprising, considering the newsletter ID was "start". I do want to get better at testing these, but I don't see how this would break submission; I'll confirm on staging.

Issue #3371, opened 2019-07-05
Page for /blog?tag=[name]


kristinashu:
Blocked by related implementation ticket for /blog page #3370

Related design ticket https://github.com/mozilla/foundation.mozilla.org/issues/3165
Related blog epic ticket https://github.com/mozilla/foundation.mozilla.org/issues/2797

- styles should follow the /blog page
- should show the number of results and the selected tag (should match the same style as on pulse)
- clicking `clear filter tag` should show users /blog

### Comp
https://redpen.io/ztf6d6d69bc4aa274f

![](https://s3.amazonaws.com/redpen-prod/red-pen-40388170-6400-40e5-cd45-dfbad50c1f1b.jpg)

cc @alanmoo

alanmoo:
This will likely be a part of the same template as the main blog page, as some extra functionality supported by query parameters or something like that.

Pomax:
Given that results will be paginated, how do we want to indicate this? E.g. "Showing 12 of 50 results found", and updating that value as the user clicks "load more results"?

Pomax:
Also, given that the tag filtering is now a properly different URL, `clear filters` is probably either no longer necessary, or needs a design tweak. The base page is still `/en/blog` but the filtered pages now (necessarily) live on `/en/blog/tags/(tags)`, so "clearing" is actually a navigation to `/en/blog`

PR #3427, opened 2019-07-16
Generating a generic listing page and blog listing specific page

### LINKED ISSUE #3370: /blog page
Related design ticket https://github.com/mozilla/foundation.mozilla.org/issues/3165
Related implementation ticket for /blog?tag=[name] #3371
Related blog epic ticket https://github.com/mozilla/foundation.mozilla.org/issues/2797

### The following should follow the same rules as our existing related posts
- image pulled from "promote" tab, if no image than it's the generic m image
- hover states
- font styles
- responsive states
- grid

### Other notes
- blue text is the first tag listed for that post
- "see more" button loads another 9 posts below (example https://thewirecutter.com/blog/)
- Blog will not appear in the main nav until we have more posts (people will only discover /blog if they type it in) 

### Note from @alanmoo (please edit as you see fit): 
My instinct is to make the "blog" page a generic page type that shows the preview of all its children pages, so that we can use it for other page types like an Advocacy campaigns index. But because pages would have different types (example: blog cards show a tag and advocacy _might_ not), I believe we could embed a check in the page where, if the card is a blog, we show some extra information.

### Comp
https://redpen.io/mo33ed7416a225ea57

Pomax:
Closes #3370 
Related PRs/issues #3371 

This does not do any fancy "load initial subset, JS-enabled MORE button to load more results", given that the first use of this model is for blog posts, of which we only have a handful. 

https://github.com/mozilla/foundation.mozilla.org/issues/3443 has been filed for the JS functionality.

### Test

https://foundation-mofostaging-pr-3427.herokuapp.com/en/blog/


network-api/networkapi/wagtailpages/models.py
```diff
@@ -160,6 +161,7 @@ class MiniSiteNameSpace(ModularPage):
         'BanneredCampaignPage',
         'CampaignPage',
         'OpportunityPage',
+        'BlogListingPage',
```
alanmoo: This needs to be a subpage of the Homepage type

Pomax: Just homepage, or also others?

Pomax:
@kristinashu this should be ready for design review

network-api/networkapi/wagtailpages/templates/wagtailpages/listing_page.html
```diff
@@ -0,0 +1,31 @@
+{% extends "./modular_page.html" %}
+
+{% load wagtailcore_tags mini_site_tags %}
+
+
+{% block content %}
```
Pomax: It appears that the main nav is missing... which is kind of silly.

kristinashu:
Thanks Pomax! A few questions and thoughts:
## Tags
For the /blog page, can we show the first tag listed in the posts promote tab for each post?

Comp:
image

## Filter view
Have you added the filter by tag view yet? I added the tag 'test' to one of the post then tried the url https://foundation-mofostaging-pr-3427.herokuapp.com/en/blog-list/?tag=[test] but it didn't change anything.

Comp: 
image

## Spacing
Just a bit of finessing is needed for the gaps. I'm reluctant to give exact px spacing but here is a comp with some guidelines. I know this will very slightly because of the preexisting spacing associated with the text so just use this comp as a guideline. Note, we stick to multiples of 8px.

image

mmmavis:
(Let me know if I should review this sooner. I usually wait for PR to pass design review first.)

Pomax:
- tags: can do, but is the first tag always the most appropriate?
- filtering: that's been filed as separate issue, so that while we still have a low number of blog posts we can have a page up that we can refine, rather than wait with landing this until all the parts are done
- spacing wise, I'm relying primarily on the default styling for header/intro, so hopefully the bootstrap padding/margin classes line up with our grid? I'll try to get as close as I can.

Pomax:
@mmmavis you can wait until all the design review has been addressed; most of the code won't change, but it's not a huge PR so there's no need to "catch things early on" as it were =)

kristinashu:
> tags: can do, but is the first tag always the most appropriate?

I'm open to suggestions but on prod I have gone through each post manually and made sure the first tag is the most relevant and broad.

Pomax:
@kristinashu it looks like the spacing around the horizontal rule is fixed in the `intro-and-content-divider` hr class, in a way that adding bootstrap classes for margin/padding can't change (they're not specific enough so the media queries "win" in terms of which rules are applies). I can change this to a different class, but then if we change that rule's styling at some point, this page will not automatically update looks, of course. 

Also for the cards: it's currently using the same spacing/styling that we use for "related cards" on blog pages - do we want to change the spacing specific for this page, for "collections of cards" in general, or stick with the spacing as is to keep it uniform?

kristinashu:
> @kristinashu it looks like the spacing around the horizontal rule is fixed in the `intro-and-content-divider` hr class, in a way that adding bootstrap classes for margin/padding can't change (they're not specific enough so the media queries "win" in terms of which rules are applies). I can change this to a different class, but then if we change that rule's styling at some point, this page will not automatically update looks, of course. 

Ok makes sense, I'm ok with keeping the way it is and not doing any overrides.

> Also for the cards: it's currently using the same spacing/styling that we use for "related cards" on blog pages - do we want to change the spacing specific for this page, for "collections of cards" in general, or stick with the spacing as is to keep it uniform?

Updating all "collections of cards" is good!

> (the heading top spacing, at least, was entirely doable with bootstrap instructions =)

Sweet!

Pomax:
@alanmoo @kristinashu @mavis I rewrote this PR in a way that _should_ have preserved the look, but under the hood uses very different code to effect the index page, so basically the entire PR needs a rereview =)

(I've also changed the test URL to https://foundation-mofostaging-pr-3427.herokuapp.com/en/blog/ to better reflect its intended place)

kristinashu:
Spacing looks good!

## Grey bar
I'm noticing a grey bar just below the main nav that appears at medium and narrow widths. @mmmavis I feel like we've seen this before? Something about reminisce of the mobile drop down sub nav?

Scree cap from PR:
image

## Line length for intro text
At wide widths the intro text should only cross 8 of the 12 bootstrap columns, instead of all 12 like in the PR (this is the keep the line length short and easier to read). But at medium and narrow widths it should continue to fill all the columns (since line length is shorter). 
image

Pomax:
Hm, the page template is based on the primary page, minus the "free content" parts, do primary pages also show this incorrect line sizes at those breakpoints in the review app? If so, we might have a site-wide regression for these templates.

Pomax:
Aha! Probably just this template's fault: https://github.com/mozilla/foundation.mozilla.org/pull/3427/files#diff-aa22ca63badaf6d5490c9347c4e8a92eR13 uses only `col-12` instead of the additional breakpoint size instructions.

Pomax:
@kristinashu fixed the line widths, it should now be 8 at medium and above, and 12 at smaller than that.

Pomax:
The grey bar looks like the minisite nav without any content (because there is none) so I'll see if I can remove that nav from this template entirely.

Pomax:
it was - removed the minisite-specific bits because this is not a minisite and what were they doing in the template in the first place here =P

Pomax:
@kristinashu @mavis the Percy failure looks like a genuine concern: the "participation guidelines" gets spread out over two columns...

kristinashu:
Why is this PR changing the footer?

Pomax:
Your guess is as good as mine: it's only showing changes in the footer for two (out of eleven?) pages, both entirely unrelated to this, and if I check those pages on the review app, things look perfectly fine, so.... it could just be Percy being wrong? =/

kristinashu:
Hmm weird, should we push to staging and then check those pages there?

Pomax:
That feels reasonable, although @mmmavis and @alanmoo will still need to R+ before we land.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic-card.html
```diff
@@ -0,0 +1,16 @@
+{% load wagtailcore_tags wagtailimages_tags %}
+
+<div class="child-card col-6 col-md-4 mb-5">
+  <div>
+    <a href="{{ page.url }}"">
```
mmmavis: extra `"`

mmmavis: Does it need quotation marks at all though?

alanmoo:
> Two questions:
> 
>     * In order for a page to show up on an index page, that page has to be child of the index page. This means child page urls will become `/index-page/child-page`. Will we running into issues with URL if we are planning to create index page of existing pages on production (e.g., their URL will change and we'll have to deal with redirects???)?

We shouldn't; We've had mini-site namespace pages in place for a while, so this will be replacing some of those, making previously dead URLs into real ones.

> 
>     * Will there be cases where we want a page to belong to more than one index page?

I think that's beyond the scope of intent for this page.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic-card.html
Pomax: it does, this isn't React JSX, Django templating does not automagically add the necessary quotes because it has no idea "where" the macro is (it can't tell the difference between `<a {{ href }}>` and `<a href="{{ href }}">`, which React can).

Pomax:
```suggestion
    <a href="{{ page.url }}">
```

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic-card.html
```diff
@@ -0,0 +1,16 @@
...
+      {% image page.specific.get_meta_image fill-400x225 %}
+    </a>
+  </div>
+  {% block tags %}{% endblock %}
+  <a class="h4-heading d-inline-block my-2" href={{ page.url }}>{{ page.title }}</a>
+  <p class="body-small mb-0">
+    {% block byline %}
+      Published on {{ page.first_published_at|date:"F j, Y" }}
```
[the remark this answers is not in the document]
Pomax: Yeah, you're right.

Issue #3005, opened 2019-04-11
Align implmentation of petition form and sign up form


mmmavis:
Follow-up for PR https://github.com/mozilla/foundation.mozilla.org/pull/2998

Align implementation of petition form and sign up form so UI and UX can be the same between the two types of forms.

### Back-end

- Not sure if anything can be done here? (cc @Pomax )

### Front-end

**Might be blocked by the upcoming form standardization work https://github.com/MozillaFoundation/Design/issues/344**

- Make the "in-page" sign up form's design follow what we have for petition form (text style, spacing, field size, animation of placeholder/label text, error state behaviour etc)
- Create shared React components that can be used by the two forms

*Note: by "in-page" sign up form I mean the sign up form that appears in the main body of the page (vs sign up form on footer or header)

---

- Live petition form example (on production): https://foundation.mozilla.org/en/campaigns/aadhaar/
- Live "in-page" sign-up from (currently on a PR review app): https://foundation-mofostaging-pr-2998.herokuapp.com/en/campaigns/test-bannered-campaign-page/

![form](https://user-images.githubusercontent.com/2896608/55989807-09efca80-5c5b-11e9-9f6c-c9354858937f.png)

mmmavis:
We should consider using dummy form validation: https://github.com/mozilla/foundation.mozilla.org/pull/3430#discussion_r305007499 /cc @gideonthomas

cadecairos:
Closing due to inactivity. lets reopen if we want to re-visit this refactor.

PR #3430, opened 2019-07-17
Fix sign up component's browser error bubble issue

### LINKED ISSUE #3051: Sign-up component bug - email validation error for invalid email should be inline
**Describe the bug**
According to design - email validation error should be inline and not appear in a browser default error bubble.

**To Reproduce**
1. Go to https://foundation.mozilla.org
2. Scroll down to the footer section
3. Type something random in email field

**Expected behavior**
Error message for invalid email should be inline.

**Screenshots**
sign up error bubble

mmmavis:
Closes #3051


source/js/components/join/join.jsx
```diff
@@ -48,6 +48,18 @@ export default class JoinUs extends React.Component {
     );
   }
 
+  /**
```
mmmavis: Copied from `petition.jsx`: https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/petition/petition.jsx#L262-L272

mmmavis:
@alanmoo the bug this PR is trying to fix is to prevent the browser bubble from showing up. If you check footer sign up on prod the submit button is already doing the things you described in your comment above. If we wanna fix it I think it should be tackled as a separate ticket.

mmmavis:
Sorta related: I filed a ticket a while back for aligning sign up / petition form on Foundation site: https://github.com/mozilla/foundation.mozilla.org/issues/3005

mmmavis:
@alanmoo nvm i misunderstood your comment lol.

source/js/components/join/join.jsx
gideonthomas: interesting...wonder why we didn't use dummy form field validation i.e. something like:

```js
function isValidEmail(email) {
  const el = document.createElement('input');
  el.type = `email`;
  el.required = true;
  el.value = email;
  return el.validity.valid;
}
```

either way, this is still fine for majority emails.

mmmavis: I'll mention this in this ticket: #3005

PR #3455, opened 2019-07-26
Fix mobile nav link styling

### LINKED ISSUE #3298: Mobile version primary nav - active link treatment
Not sure if this is a bug. @kristinashu do you know what the expected on hover treatment is for `active page link` on mobile version of primary nav (e.g, "Home" in the attached gif below)? Currently we add a black underline to it on hover which doesn't make any visual difference since the whole nav section on black background.

mobile-nav-active-hover

mmmavis:
Closes #3298

- [x] Add a Selected state that looks like the Hover state
- [x] Update the bar to be white
- [x] Update the bar to be 3 px
- [x] Update the bar to be closer to the text

### DIFF
--- a/source/sass/components/primary-nav.scss
+++ b/source/sass/components/primary-nav.scss
@@ -259,7 +259,7 @@
           }
 
           &.active {
-            border-color: $white;
+            border-bottom-color: $white;
           }
         }
       }
--- a/source/sass/components/primary-nav.scss
+++ b/source/sass/components/primary-nav.scss
@@ -250,7 +250,17 @@
           transition: transform 0.15s;
           display: inline-block;
           margin: 10px 0;
+          padding-bottom: 0;
+          border-width: 3px;
           color: $white;
+
+          @include hover-focus-active {
+            border-bottom-color: $white;
+          }
+
+          &.active {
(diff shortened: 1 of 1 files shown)


source/sass/components/primary-nav.scss
```diff
@@ -250,7 +250,17 @@
           transition: transform 0.15s;
           display: inline-block;
           margin: 10px 0;
+          padding-bottom: 0;
+          border-width: 3px;
           color: $white;
+
+          @include hover-focus-active {
+            border-bottom-color: $white;
```
alanmoo: Maybe it's been too long since I've written CSS, but how does an include statement also have a content block?

mmmavis: oh this is a Bootstrap's mixin

```scss
@mixin hover-focus-active {
  @if $enable-hover-media-query {
    &:focus,
    &:active {
      @content
    }
    @include hover { @content }
  }
  @else {
    &:focus,
    &:active,
    &:hover {
      @content
    }
  }
}
```

alanmoo: Ah, I forgot about the @content directive.

PR #3467, opened 2019-07-29
order index page entries by most-recently-created

### LINKED ISSUE #3465: sort index page listings (blog etc) on publish date, newest first
It's pulling up the records in database ordering at the moment, and that should be published_data, most recent first.

@kristinashu @alanmoo do we want true publication date, or "latest edit date"? I assume the first so that the ordering doesn't seemingly-randomly change?

Pomax:
Closes #3465  by ordering the set of entries by publication data. As we can't do ordering in the template itself, this is now done as part of the template context preprocessing function. Thankfully, that means no migrations are required.

### DIFF
--- a/network-api/networkapi/wagtailpages/models.py
+++ b/network-api/networkapi/wagtailpages/models.py
@@ -560,6 +560,7 @@ class IndexPage(FoundationMetadataPageMixin, Page):
         context = super().get_context(request)
         context = set_main_site_nav_information(self, context, 'Homepage')
         context = get_page_tree_information(self, context)
+        context['entries'] = self.get_children().live().order_by('-first_published_at')
         return context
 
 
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/models.py
```diff
@@ -560,6 +560,7 @@ def get_context(self, request):
         context = super().get_context(request)
         context = set_main_site_nav_information(self, context, 'Homepage')
         context = get_page_tree_information(self, context)
+        context['entries'] = self.get_children().live().order_by('-first_published_at')
```
youriwims: Why the hyphen before `first_published_at`?

alanmoo: Newest first

youriwims: Yeah, but I'm asking is that just the syntax for how to write it here?...I looked at the docs and none of the fields had a hyphen before them.

alanmoo: https://docs.djangoproject.com/en/2.2/ref/models/querysets/#order-by

Issue #3468, opened 2019-07-30
Change fake data generator to add “index page” at /blog


alanmoo:
So that we can see the names of our fake blog posts.

Let’s also generate a few more blog posts while we are at it.

Pomax:
done and done, PR waiting for your review.

PR #3469, opened 2019-07-30
use IndexPage for the blog namespace, and generate 7 blog posts (to verify listing)

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3468

This switches the blog namespace from a minisitenamespace page type to the new indexpage page type, and generates 7 blog posts so that listing can be checked (both in terms of "does it look right?" as well as "is it ordered correctly")

https://foundation-mofostaging-pr-3469.herokuapp.com/en/blog/

### DIFF
--- a/network-api/networkapi/wagtailpages/factory/blog.py
+++ b/network-api/networkapi/wagtailpages/factory/blog.py
@@ -58,7 +58,7 @@ def generate(seed):
 
     print('Generating blog posts under namespace')
     for i in range(7):
-        title = f'post {i}'
+        title = Faker('sentence', nb_words=6, variable_nb_words=False)
         try:
             BlogPage.objects.get(title=title)
         except BlogPage.DoesNotExist:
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -49,15 +49,17 @@ def generate(seed):
...
 
-    try:
-        BlogPage.objects.get(title='post')
-        print('a post page (BlogPage) exists')
-    except BlogPage.DoesNotExist:
-        print('Generating a post page (BlogPage) under namespace')
-        BlogPageFactory.create(parent=blog_namespace, title='post')
+    print('Generating blog posts under namespace')
+    for i in range(7):
+        title = f'post {i}'
```
alanmoo: Let's add a few random words to this so it's more representative of title length on live blog posts.

Issue #3424, opened 2019-07-16
Make the main CTA button for MozFest CMS editable


gideonthomas:
<img width="1438" alt="Screen Shot 2019-07-16 at 2 55 15 PM" src="https://user-images.githubusercontent.com/5816130/61321604-e0a28580-a7d9-11e9-9a30-c039165b87e1.png">

We will need to update the text and link of the button on the top-right when ticket sales open up, the schedule goes live, and after the festival.

Alan mentioned that it would be useful to make it CMS editable so that we don't have to change it every time (including next year).

cc @alanmoo and @kristinashu to triage. The first update needs to happen by the first week of August.

alanmoo:
@kristinashu Any design requirements for this other than "be able to update the text and the URL"?

kristinashu:
Nope but I would like to clarify that this is only for the MozFest site. Buttons on the main foundation site are much more fixed and the average Wagtail staff should not be able to update them.

alanmoo:
I'd guess we probably want to use [Site settings](https://docs.wagtail.io/en/v2.5.1/reference/contrib/settings.html) for this.

Pomax:
I'd like to clarify: does the _text_ change, or does the _button_ change? Because if the text changes from "submit proposal" to "buy tickets", then really that's a different button, and it might make more sense to define a collection of buttons of which only one can be picked to show up on the homepage at a time, similar to "news items" or "campaign CTAs" on the foundation site.

alanmoo:
Good point. The button changes.

Pomax:
sounds good

Pomax:
Do we want it to point to "any URL", or should it point to a specific mozfest app page? (e.g. should people type the URL and hope it's right, or should they be able to pick a page in the CMS and the button will point to that page, whatever its URL happens to be?)

alanmoo:
I imagine it could be external links, so url string

mmmavis:
A design question for you @sabrinang : do we want the nav CTA button (text + link) to stay the same across the site? Or allow each MozFest page to configure the button so it can link to different URL?

sabrinang:
@mmmavis Keep the same nav CTA button to be consistent across the site.

PR #3470, opened 2019-07-30
Cms controlled mozfest cta button

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3424 by adding a cta_button label and destination field that can be configure so that all mozfest pages use those values for the button in the upper right in the nav bar.


network-api/networkapi/mozfest/migrations/0005_auto_20190730_2045.py
```diff
@@ -0,0 +1,33 @@
...
+            model_name='mozfestprimarypage',
+            name='cta_button_destination',
+            field=models.CharField(blank=True, help_text='The URL for the page that the CTA button in the primary nav bar should redirect to', max_length=2048),
+        ),
+        migrations.AddField(
+            model_name='mozfestprimarypage',
+            name='cta_button_label',
+            field=models.CharField(blank=True, help_text='Label text for the CTA button in the primary nav bar', max_length=250),
+        ),
+        migrations.AlterField(
```
Pomax: I'm not sure why wagtail/django altered these fields but... this is an auto-generated migration, so I guess it feels it needs those alterations? =S

network-api/networkapi/mozfest/models.py
```diff
@@ -69,6 +69,7 @@ def get_template(self, request):
 
     def get_context(self, request, bypass_menu_buildstep=False):
         context = super().get_context(request)
+        context = set_main_site_nav_information(self, context, 'MozfestHomepage')
```
mmmavis: A question just because I'm not familiar with Django / our existing setup -
Why do we need this now? Is it because the MozFest primary pages need to know about the `root` for the CTA button on nav to render properly?

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -6,14 +6,14 @@ def set_main_site_nav_information(page, context, homepage_class_name):
...
-    homepage = list(filter(
-        lambda x: x.specific.__class__.__name__ is homepage_class_name,
-        page.get_ancestors()
-    ))[0]
 
-    context['menu_root'] = homepage
-    context['menu_items'] = homepage.get_children().live().in_menu()
+    root = list(filter(
+        lambda x: x.specific.__class__.__name__ == homepage_class_name,
+        page.get_ancestors(True)  # incude ourselves!
```
mmmavis: Will this affect other page classes which use this method? e.g., `IndexPage`, `ModularPage`, `PrimaryPage`?

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -39,7 +39,7 @@
                   </div>
                 </div>
               </div>
-              <a class="btn btn-secondary" href="/proposals">Submit proposal</a>
+              <a class="btn btn-secondary" href="{{ menu_root.specific.cta_button_destination }}">{{ menu_root.specific.cta_button_label }}</a>
```
mmmavis: Can be a followup - can we make external link to open on a new tab?

network-api/networkapi/mozfest/models.py
Pomax: correct - all pages need to know where the homepage lives, so they can access its `cta_button_label` and `cta_button_destination` values

network-api/networkapi/wagtailpages/utils.py
Pomax: we have a dedicated `Homepage` class for the foundation site as well, so this should not interfere.

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
Pomax: That feels like something we should be looking at when we actually change the button to link to an external site, because it's entirely possible we'll never do and always funnel to our own pages first.

Pomax: /cc @alanmoo @gideonthomas

gideonthomas: yeah I wouldn't worry about it

network-api/networkapi/mozfest/models.py
```diff
@@ -82,6 +83,18 @@ def get_context(self, request, bypass_menu_buildstep=False):
 
 
 class MozfestHomepage(MozfestPrimaryPage):
+    cta_button_label = models.CharField(
```
alanmoo: Why are we adding this to the Homepage rather than using [Base settings](https://docs.wagtail.io/en/v2.5.1/reference/contrib/settings.html#defining-settings)?

Pomax: No idea: that link goes nowhere?

Pomax: investigation planned in https://github.com/mozilla/foundation.mozilla.org/issues/3477

Issue #3443, opened 2019-07-22
Add "load more" JS functionality to the listing page


Pomax:
Related to #3370, the listing page currently does not load X initial results with additional results loaded through a JS-enabled `MORE` button, which for now is fine, but we're going to have more blog posts, as well as other page types, and so it would be useful to think about how we want to deal with large pages (e.g does 200 blog posts as initial view make sense? 30 campaigns might. What about a listing page for all listing pages? etc)

Pomax:
Blocked on https://stackoverflow.com/questions/57299153/getting-all-api-fields-for-a-wagtail-api-child-of-call as it appears for all intents and purposes that we can get all the fields we need to generate more cards client-side only if we explicitly ask for named models using `type=<model name>`, rather than `child_of=<indexpage pk>`, which means that the actual list we want doesn't have the data we need, and the list that has the data we need is not scoped to an index page parent. =S

Pomax:
Currently using a workaround by discovering, up front, all the different types that children might have. This works, but is hardly optimal.

Pomax:
@alanmoo @mmmavis @youriwims I'm assuming that between https://github.com/mozilla/foundation.mozilla.org/pull/3476 and https://github.com/mozilla/foundation.mozilla.org/pull/3485 we want the latter, using non-API code to make the fetching happen.

PR #3476, opened 2019-07-31
[WIP] enabling progressive loading of index page content (API-based)

Pomax:
Related PRs/issues https://github.com/mozilla/foundation.mozilla.org/issues/3443

Covers some of the work necessary to make iterative loading of index page content possible
- Wagtail API v2
- main.js extension for the "load more" button on index pages
- IndexPage class and associated temples updated 

open questions:

- how on earth am I going to make JS generate the same cards when JS has no knowledge of django template fragments... I'm not super high up on "mirroring HTML in JSX" again =(

One option is to have Django render the html and transmit that, rather than the JSON data, e.g. 
```
from django.template import loader
entries_html = loader.render_to_string(
    'wagtailpages/fragments/entry_cards.html',
    {'entries':  ... }
)
return HTTPResponse(entries, content_type="text/html")
```
 (e.g. [like this](https://alphacoder.xyz/lazy-loading-with-django-and-jquery/)) and then inject that directly into the page... in which case security needs to be evaluated here (does Django safify the output? it probably does, but let's confirm)

### test

load https://foundation-mofostaging-pr-3476.herokuapp.com/en/blog/, open dev tools, press "load more" button.


Pomax:
Tagging a bunch of folks mostly opinions, rather than review. Have a look at the code and voice any concerns while I'm working on this, so dumb things can be caught early =P

network-api/networkapi/wagtailpages/models.py
```diff
@@ -574,7 +582,6 @@ class BlogPageTag(TaggedItemBase):
 
 
 class BlogPage(FoundationMetadataPageMixin, Page):
-    template = 'wagtailpages/blog_page.html'
```
Pomax: as a normal page model (rather than page with a static template backing it) this line was not necessary.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -600,23 +607,24 @@ class BlogPage(FoundationMetadataPageMixin, Page):
...
-    tags = ClusterTaggableManager(through=BlogPageTag, blank=True)
-
     promote_panels = FoundationMetadataPageMixin.promote_panels + [
         FieldPanel('tags'),
     ]
 
-    # Database fields
-
-    zen_nav = True
+    api_fields = [
```
Pomax: these will show up in the API request **if** `?types=...` is specified with `wagtailpages.BlogPage` as one of the types, **and** `?fields=*` is used (or `?fields=title,author,tags`, but that would break selecting children of different types)

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/og_metadata.html
```diff
@@ -0,0 +1,8 @@
+{% load wagtailmetadata_tags i18n %}
+
+ {% get_current_language as lang_code %}
```
Pomax: if we're already doing this somewhere else I'll happily use that instead, but I couldn't find us marking the locale we're using to generate content anywhere =S

source/js/main.js
```diff
@@ -520,6 +520,50 @@ let main = {
...
+          data.index
+        }${types}&fields=*&order=-first_published_at&format=json`;
+
+        // And then fetch the results and render them into the page.
+        fetch(url)
+          .then(result => result.json())
+          .then(data => data.items)
+          .then(entries => {
+            // TODO: load this data into appropriate card form...
+            // TODO: how do we GET the appropriate card form?
```
Pomax: Yeeeeeeah, so this one's a big'n. How on earth are we going to do this cleanly. One thought was to front-load the index page with `<script type="html/template"> ... </script>` that templates the skeleton for each card type in, but I'd really rather take something better.

source/js/main.js
```diff
@@ -520,6 +520,50 @@ let main = {
...
+            console.log(entries);
+          })
+          .catch(err => {
+            // TODO: what do we want to do in this case?
```
Pomax: Also this is an open question. What _do_ we do when trying to fetch more results errors out?

source/js/main.js
```diff
@@ -520,6 +520,50 @@ let main = {
...
+            console.error(err);
+          });
+      });
```
Pomax: Also, this still needs the pagination filtering added, and a front-loading bit of code in the IndexPage model, so that we don't generate the 'load more' button if there's nothing more to load on initial page generation, as well as hide the button once there's nothing left to load because we reach the end of the paginated content through the API.

network-api/networkapi/api.py
```diff
@@ -0,0 +1,6 @@
+from wagtail.api.v2.router import WagtailAPIRouter
```
Pomax: see http://docs.wagtail.io/en/v2.0/advanced_topics/api/v2/configuration.html

network-api/networkapi/wagtailpages/models.py
Pomax: For more details, see http://docs.wagtail.io/en/v2.1.1/advanced_topics/api/v2/usage.html

source/js/main.js
youriwims: ...might be a silly question, but does this include when there _aren't_ any more results to show? Or are we just talking about when we run into a legitimate error while fetching?

youriwims: ...might be a silly question, but does this include when there aren't any more results to show? Or are we just talking about when we run into a legitimate error while fetching?

Pomax:
Closing in favour of https://github.com/mozilla/foundation.mozilla.org/pull/3485

PR #3482, opened 2019-08-01
For newsletter signup: exclude query params in 'source url' sent to Basket/Salesforce

### LINKED ISSUE #3481: Newsletter sign-up component: exclude query params in "source_url" sent to Salesforce
Based on this [comment](https://github.com/mozilla/foundation.mozilla.org/issues/2994#issuecomment-516997473): 

For newsletter signups, we should send along the source URL as the base URL without all the query params, as having more data in Salesforce makes things harder to parse.

(GA should still do its thing with query params)

mmmavis:
Closes #3481

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -97,7 +97,8 @@ export default class JoinUs extends React.Component {
     return new Promise((resolve, reject) => {
       let payload = {
         email: this.email.value,
-        source: window.location.toString()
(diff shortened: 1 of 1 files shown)


mmmavis:
I think all newsletter signup forms on the site use this `join.jsx` file. 

Do we also wanna update this? I think this is for petition though? https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/petition/petition.jsx#L187

alanmoo:
Good question. I've asked in Slack, but I think the original issue has been addressed.

Issue #3479, opened 2019-08-01
Add tags to load_fake_data blog posts


Pomax:
Currently blog pages are generated without any tags, which makes testing for tag filtering impossible

PR #3464, opened 2019-07-29
Blog index with tags rebase

Pomax:
Closes #3371 
Related PRs/issues https://github.com/mozilla/foundation.mozilla.org/pull/3427 (#3370)

This adds tag filtering to any page of type `IndexPage` - e.g. if there is a listing page with slug `/blog`, visiting `/blog/tags/tag1` or `/blog/tags/tag1/tag2/tag3/...` effects listing only pages posts that match that specific tag, or any of the specified tags, respectively.

(tags as used in urls should be the tag slug: `Internet Health` becomes `internet-health`, etc.)

Test URL

https://foundation-mofostaging-pr-3464.herokuapp.com/en/blog/

Also click on any of the blog cards' main tag to go to the dedicated tag page, e.g:

https://foundation-mofostaging-pr-3464.herokuapp.com/en/blog/tags/iot

Also also, verify that "related posts" blog cards still work:

https://foundation-mofostaging-pr-3464.herokuapp.com/en/blog/initial-test-blog-post-with-fixed-title/

### Deviations from the design

As tags are no longer "filters" but real URLs, so the "clear filters" link has been (for now) omitted, in favour of people using their browser's "back" button.

The "result" phrasing has been changed to "X pages for (tag)", as the URL and page content does not suggest any searching/filtering happens, so talking about "results" without the user performing a search or active filter was problematic.

On any tag-specific page, blog cards show all tags, not just their "main" tag, or just the tag that page is for.

### Design questions

- which styling is suppose to be used for the "pages found" line? Using `body` makes it feel very small.
- should cards show all tags when specifically searching for a tag? If not, things like https://foundation-mofostaging-pr-3464.herokuapp.com/en/blog/tags/iot/ get confusion, as the page is supposedly for `iot` posts, but because that's not the main tag for some pages, it'll look like there are pages listed that don't belong on that page.

PR #3484, opened 2019-08-01
Add tags to fake blog posts

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3479, by adding tag binding in the blog factory


Pomax:
Note that the code in this PR has also been worked into https://github.com/mozilla/foundation.mozilla.org/pull/3464 in order to very that PR works on heroku, but I'd rather land this PR explicitly, and then land the tag-filtering PR with this PR's code taken back out, because it'll already be in master.

PR #3485, opened 2019-08-01
Enabling progressive loading of index page content

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3443

This PR adds in a "load more results" button to the `IndexPage` template, with code hooked up in `main.js` to make that an active button that fetches paged entry sets from the index page for progressively loading the backlog of entries.

### test

- main blog page: https://foundation-mofostaging-pr-3485.herokuapp.com/en/blog (7 total)
- filtered page: https://foundation-mofostaging-pr-3485.herokuapp.com/en/blog/tags/iot (5 total)
- filtered page: https://foundation-mofostaging-pr-3485.herokuapp.com/en/blog/tags/policy/ (2 results, no "load more" button because 2<=4)

### details

the index page for blog has been set to a "page size" of 4 in the CMS (this is a per-index-page setting that defaults to 12 results per page if left untouched). Currently possible page sizes are 4, 8, 12 (which is probably best), and 24

Screen Shot 2019-08-07 at 12 21 25 PM


network-api/networkapi/wagtailpages/models.py
```diff
@@ -556,13 +560,45 @@ class IndexPage(FoundationMetadataPageMixin, Page):
...
 
+    def get_entries(self):
+        return self.get_children().live().order_by('-first_published_at')
+
     def get_context(self, request):
         context = super().get_context(request)
         context = set_main_site_nav_information(self, context, 'Homepage')
         context = get_page_tree_information(self, context)
-        context['entries'] = self.get_children().live().order_by('-first_published_at')
+        context['entries'] = self.get_entries()[0:1]
```
Pomax: realistically we'd probably want this to be `self.get_entries()[0: self.page_size]` for selecting a batch with a CMS-controlled size.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -556,13 +560,45 @@ class IndexPage(FoundationMetadataPageMixin, Page):
...
+        Get a set of entries, as rendered HTML
+        """
+        
+        page = 0
+        
+        if 'page' in request.GET:
+            page = int(request.GET['page'])
+
+        # Do we want to make this a CMS-controllable value?
+        page_size = 12
```
Pomax: this would be `page_size = self.page_size` as value in absence of a query argument

network-api/networkapi/wagtailpages/models.py
```diff
@@ -574,7 +610,6 @@ class BlogPageTag(TaggedItemBase):
 
 
 class BlogPage(FoundationMetadataPageMixin, Page):
-    template = 'wagtailpages/blog_page.html'
```
Pomax: this got landed at some point, but isn't necessary for CMS-controlled pages that have a template based on their model name (we do need explicit template variables for our static template pages)

network-api/networkapi/wagtailpages/models.py
mmmavis: It seems like on mobile we show two cards per row and three cards per row on bigger devices. To make new set of cards appear nicely we might wanna choose a number that's a common multiple of 2 and 3 so new cards can appear as full row(s).

source/js/main.js
```diff
@@ -520,6 +520,44 @@ let main = {
...
+
+      let page = 0;
+      let page_size = 1; // TODO: should we get this as data-attribute, set from IndexPage.page_size ?
+
+      const loadMoreResults = evt => {
+        // Construct our localized API url:
+        let url = `./entries/?page=${page++}&page_size=${page_size}`;
+
+        // And then fetch the results and render them into the page.
+        fetch(url)
```
mmmavis: Do we need to do anything for Percy here?

youriwims:
>Open questions: if we go with this approach, do we want to make the page_size value a field on the IndexPage, so that we can control the batch size through the CMS? It feels like we probably do, but always good to hear opinions.

🗣 Yes—I fully support this! But in reference to what Mavis said 👉  https://github.com/mozilla/foundation.mozilla.org/pull/3485#discussion_r311290981, we should give them predetermined batch sizes in the CMS to choose from (e.g. at desktop being able to choose between 3, 6, and 9) instead of letting the editor choose a number. If that's what you were already thinking, then ditto!

network-api/networkapi/wagtailpages/models.py
Pomax: I'm going to default it to 12, which gets us 2, 3, and 4 (and 6 but pff come on, 6? on a row? not everyone has an ultrawide monitor)

source/js/main.js
Pomax: Hm, this would be a good candidate for true interaction testing, rather than percy. Unfortunately, we still don't have that quite finished and landed...

Pomax:
I rebased and updated the code to perform progressive loading both for the "main" view as well as the tag-filtered view, @youriwims @mmmavis can you rereview the (fairly substantially changed) code?

@kristinashu this is ready for design review, too

kristinashu:
Load more is working great except for when there are several posts. In this PR the text says `4 results for iot` because 4 have loaded but there are actually 5 (and it continues to say 4 even after the 5th one has loaded). Could you please update it to say the total number of posts (example: `5 results for iot` even when only 4 have loaded)?

> filtered page: https://foundation-mofostaging-pr-3485.herokuapp.com/en/blog/tags/iot (5 total)

Screen cap from PR:
image

source/js/main.js
```diff
@@ -535,6 +535,49 @@ let main = {
...
+            return data.entries_html;
+          })
+          .then(entries_html => {
+            const div = document.createElement(`div`);
+            div.innerHTML = entries_html;
+
+            Array.from(div.children).forEach(child =>
+              entries.appendChild(child)
+            );
+          })
```
youriwims: In reference to Kristina's [comment](https://github.com/mozilla/foundation.mozilla.org/pull/3485#issuecomment-523062389): Maybe this would be a good place to update the results counter and get the most updated [entries length](https://github.com/mozilla/foundation.mozilla.org/pull/3485/files#diff-aa22ca63badaf6d5490c9347c4e8a92eR28) (?)

Pomax: yeah, forgot to update that span =S

Pomax:
That's a good question. @kristinashu I assume it's desirable to have the "filter" tag use the same casing and spacing as in the result card?

Pomax:
@youriwims the delay is caused by the fact that pressing the button starts the API call, so we'd have to pre-call the next set if we want "instant" results. If that's something we want to do, we should probably do that in follow-up.

youriwims:
> @youriwims the delay is caused by the fact that pressing the button starts the API call, so we'd have to pre-call the next set if we want "instant" results. If that's something we want to do, we should probably do that in follow-up.

Ah okay, makes complete sense.

kristinashu:
> I assume it's desirable to have the "filter" tag use the same casing and spacing as in the result card?

Yes, that would be great. Thank you for pointig that out @youriwims!

Pomax:
@youriwims @kristinashu PR updated with fixes for the flagged issues

Pomax:
@youriwims after talking with @kristinashu we've made the label simply say how many results there are total, now how many you're looking at right now.

kristinashu:
@youriwims I don't think it is that unusual to do it this way. The functionality you mentioned might be more common for paginated pages (example: page 3 of 12) but not necessarily as common for the "view more" type of loading where content load below. And sites that use lazy load (example: Medium) don't even show a total number at all. Here are the first few sites that I like which use the "load more" style and only show the total number of results:

https://thewirecutter.com/search/?s=canada
https://www.nytimes.com/search?query=canada
https://www.thisamericanlife.org/archive?keyword=canada

youriwims:
Thanks for explaining @Pomax & @kristinashu (and for the examples!) 😄

PR #3487, opened 2019-08-05
Replaced all /en Links 

### LINKED ISSUE #3475: Replace all /en links
Find all /en links in the repo and remove the locale part of the link

Patil2099:
Closes #3475
Removes all /en links except en-US

### DIFF
--- a/network-api/networkapi/buyersguide/templates/buyersguide_home.html
+++ b/network-api/networkapi/buyersguide/templates/buyersguide_home.html
@@ -62,7 +62,7 @@
           {% endif %}
         {% endif %}
 
-        <a class="product-image text-center mt-4 h-100 d-flex flex-column justify-content-between" href="privacynotincluded/products/{{ product.slug }}">
+        <a class="product-image text-center mt-4 h-100 d-flex flex-column justify-content-between" href="/privacynotincluded/products/{{ product.slug }}">
           {% if USE_CLOUDINARY %}
           <img
             class="product-thumbnail"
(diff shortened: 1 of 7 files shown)


Patil2099:
@Pomax  Can you Check this?

network-api/networkapi/buyersguide/templates/buyersguide_home.html
```diff
@@ -62,7 +62,7 @@ <h1 class="h1-heading text-center mt-4">Be Smart. Shop Safe.</h1>
           {% endif %}
         {% endif %}
 
-        <a class="product-image text-center mt-4 h-100 d-flex flex-column justify-content-between" href="/en/privacynotincluded/products/{{ product.slug }}">
+        <a class="product-image text-center mt-4 h-100 d-flex flex-column justify-content-between" href="privacynotincluded/products/{{ product.slug }}">
```
mmmavis: We should keep the leading `/` 

i.e., `href="/privacynotincluded/products/{{ product.slug }}"`

Pomax:
After investigating the original underlying error, it turns out that rather than removing /en/ we need to replace each instance with the user's current locale, based on their Django sessions, which is a fairly complicated bit of work.

As we don't currently have the capacity to walk contributors through the process to make that happen, we'll have to close this PR and do this work in-house instead.

Thank you so much for being willing to help out, and apologies that this turned out far more complicated than initially thought =(

network-api/networkapi/utility/middleware.py
```diff
@@ -37,7 +37,7 @@ def __call__(self, request):
                     redirect_url = '{protocol}://{hostname}{path}'.format(
                         protocol=protocol,
                         hostname=hostnames[0],
-                        path='/en/mozfest/'
+                        path='/mozfest/'
```
mmmavis: not 100% sure if we need to keep `/en` here

Issue #3474, opened 2019-07-31
Optimize Open Graph on localized pages


gideonthomas:
Update og:locale, or set og:locale:alternate on localized pages.

**Problem**
We’re currently using this everywhere: `<meta property='og:locale' content='en_US'>`

Pomax:
This was added in a PR I was working on, I'll see which one it is and close-cross-link.

Pomax:
Addressed in https://github.com/mozilla/foundation.mozilla.org/pull/3485/files#diff-466035fbdea07f4221da7ce361860856R6

Pomax:
extracted as separate PR to unblock other work.

PR #3501, opened 2019-08-07
set true locale as og:locale

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3474


source/js/main.js
```diff
@@ -38,6 +40,9 @@ let main = {
       csrfToken = document.querySelector(`meta[name="csrf-token"]`);
       csrfToken = csrfToken ? csrfToken.getAttribute(`content`) : false;
 
+      locale = document.querySelector(`meta[property="og:locale"]`);
+      locale = locale ? locale.getAttribute(`content`) : DEFAULT_LOCALE;
```
mmmavis: Where is `locale` used in this file?

source/js/main.js
```diff
@@ -21,8 +21,10 @@ import bindMozFestGA from "./mozfest-ga.js";
 
 const SHOW_MEMBER_NOTICE = false;
 
+const DEFAULT_LOCALE = `en`;
```
mmmavis: (I know little about localization) 

Our default locale used to be `en_US` (e.g., ```<meta property='og:locale' content='en_US'>```). Just wanna make sure we are intentionally changing it to just `en` now

source/js/main.js
Pomax: it's not, this provisions it so we can use it for the `/en/` hardcoded link work.

source/js/main.js
Pomax: We're changing it because it's based on the code used by Wagtail. We're using the short codes in wagtail, which are thankfully legal codes: https://kb.yoast.com/kb/changing-the-og-locale-output/#valid

Pomax:
@mmmavis @alanmoo so it looks like we need the `ll_UU` local code for og:locale, and Wagtail doesn't use those for its locale code... I can add a fairly trivial mapping for short code to long code, but if there are better ideas, I'd happily take them.

source/js/main.js
alanmoo: So, some vocabulary:

`en` is the language
`US` is the region
`en_US` is the locale

We've never used a full locale on the Foundation site, since it's not supported by default in Django/Wagtail. What's the concern here?

Pomax: opengraph needs locale, wagtail uses language. So we need to generate a locale, from a language, and there's nothing built in for that.

Pomax: I've updated the PR with a mapping so that we can generate the correct opengraph locale on the Django side of the fence.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/og_metadata.html
```diff
@@ -0,0 +1,9 @@
+{% load wagtailmetadata_tags i18n localization %}
+
+{% get_current_language as lang_code %}
+
+<meta property='og:type' content='website'>
+<meta property='og:locale' content='{{ lang_code|to_opengraph_locale }}'>
+<meta property='wagtail:language' content='{{ lang_code }}'>
```
Pomax: there is no `og:language`, unfortunately.

network-api/networkapi/wagtailpages/templatetags/localization.py
```diff
@@ -0,0 +1,22 @@
+from django import template
+
+register = template.Library()
+
+DEFAULT_LOCALE = 'en_US'
+
+mappings = {
```
alanmoo: @TheoChevalier While we don't have the ability to do specific locales right now, is this set what we generally translate to by default for foundation.mozilla.org right now?

TheoChevalier: We support the following locales `fr`, `es`, `de`, `pl`, `pt-BR`.
For French and Spanish in particular, it’s important not to specify a country code, we’re serving international (neutral) versions of the languages. If that’s not possible to use that (I think you’re saying that’s the case?), then yes, `fr-FR` and `es-ES` are the biggest markets we address.

However, I would change `pt-PT` to `pt-BR`

mmmavis: Does it matter if we use underscore or dash to connect language code and region code?

TheoChevalier: What’s the purpose of this list, is it only used to populate the og tags? Or do we have plans to use it for other things? If the former, then I assume Facebook dictates what format should be used.
However, for Mozilla purposes, I’d recommend to use a dash

mmmavis: Yea it seems like using underscore for `og:locale` is the standard.

Pomax:
Also it looks like there's a bunch of `{% block socialMetadata %}` definitions in several templates that need to be wrapped up into this.

network-api/networkapi/wagtailpages/templatetags/localization.py
```diff
@@ -0,0 +1,22 @@
...
+    'en': 'en_US',
+    'fr': 'fr_FR',
+    'es': 'es_ES',
+    'de': 'de_DE',
+    'pl': 'pl_PL',
+    'pt': 'pt_PT',
```
Pomax:
```suggestion
    'pt': 'pt_BR',
```

network-api/networkapi/wagtailpages/templatetags/localization.py
Pomax: this is purely for opengraph's `og:locale` metadata.

network-api/networkapi/wagtailpages/templatetags/localization.py
```diff
@@ -0,0 +1,22 @@
...
+
+DEFAULT_LOCALE = 'en_US'
+
+mappings = {
+    'en': 'en_US',
+    'fr': 'fr_FR',
+    'es': 'es_ES',
+    'de': 'de_DE',
+    'pl': 'pl_PL',
+    'pt': 'pt_BR',
```
Pomax:
```suggestion
    'pt': 'pt_BR',  # our main focus is Brazilian Portuguese
```

network-api/networkapi/wagtailpages/templatetags/localization.py
```diff
@@ -0,0 +1,22 @@
...
+    'en': 'en_US',
+    'fr': 'fr_FR',
+    'es': 'es_ES',
+    'de': 'de_DE',
+    'pl': 'pl_PL',
+    'pt': 'pt_BR',  # our main focus is Brazilian Portuguese
+}
+
+
+@register.filter
```
Pomax:
```suggestion
# This filter turns Wagtail language codes into OpenGraph locale strings
@register.filter
```

network-api/networkapi/mozfest/templates/mozfest/404.html
```diff
@@ -1,8 +1,7 @@
 {% extends "mozfest/mozfest-base.html" %}
 
 {% block socialMetadata %}
-<meta property='og:type' content='website'>
-<meta property='og:locale' content='en_US'>
+{% include "fragments/og_metadata.html" %}
```
alanmoo: Would it be better to pull this in at base.html and override in any templates that _don't_ need it?

Pomax: hm, maybe, although that feels like a followup rather than a refactor of what's in place now.

Pomax:
@TheoChevalier I'd like to do that as separate issue, probably in combination with Alan's suggestion

alanmoo:
Just file the follow up ticket before merging

Pomax:
filed https://github.com/mozilla/foundation.mozilla.org/issues/3540 for follow-up work.

Issue #3457, opened 2019-07-26
Add client-side validation for petition comment fields


alanmoo:
Petition comments are limited to 500 characters in Salesforce. We should do a check on the client side so that a signer knows if their comment is too long, instead of just slicing it quietly after submission.

PR #3503, opened 2019-08-07
add comment validation

Pomax:
To make sure we don't send comments with more than 500 characters to salesforce, comments have been character count validation.

If a comment exists, and its length exceeds 499, the comment field will be marked with a warning.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/3457

Tests:

- https://foundation-mofostaging-pr-3503.herokuapp.com/en/campaigns/personal-ok-image/ (no comment)
- https://foundation-mofostaging-pr-3503.herokuapp.com/en/campaigns/single-page/ (optional comment: no warning if left empty, warning if filled in with >499 chars)
- https://foundation-mofostaging-pr-3503.herokuapp.com/en/campaigns/multi-page/ (mandatory comment: warning if left empty, warning if filled in with >499 chars)


source/js/components/petition/petition.jsx
```diff
@@ -170,6 +170,10 @@ export default class Petition extends React.Component {
         return reject(new Error(`missing required comment`));
       }
 
+      if (comment && comment.length >= 500) {
```
mmmavis: @alanmoo from your comment

> Petition comments are limited to 500 characters in Salesforce.

Is 499 or 500 the max characters allowed??? If it's 500 this should be changed to `>`.

mmmavis: Since `500` is being used multiple times in this file, can we turn `500` into a variable with inline comment to describe why 500?

Pomax: honestly, I don't want to take that risk. If Salesforce says it's 500, I'd rather stick with 499 as limit.

Pomax: That said, using constants makes sense, but we might need more than one constant, even if the value is the same across multiple constants.

Pomax:
Added a `SALESFORCE_COMMENT_LIMIT` constant.

alanmoo:
Yeah, I believe review apps don't Salesforce well, unfortunately.

PR #3506, opened 2019-08-08
[WIP] Localize Django templates

TheoChevalier:
Alright folks, I was testing things up for donate-wagtail, and went down a rabbit hole…

So, I made strings in Django templates localizable and work with Pontoon. However I’m not sure how disruptive this approach is, so, totally open to changes and eager to know how you feel about it.

I’ve set up my fork with Pontoon staging, it seems to work fine https://mozilla-pontoon-staging.herokuapp.com/projects/mozilla-foundation-website/

Screenshot_2019-08-08 Mozilla Foundation website

### What’s missing:

- updating tasks.py? (hitting my Python knowledge limits…)
- building .mo files when running the server locally — couldn’t find where to do that
- updating documentation — will add if this approach feels right
- probably other things, as I’m not familiar with the code base

### Workflow overview:

- add or edit a string in a template file
- run `extract_strings.sh`
- commit and merge
- string is being picked up by Pontoon from the .pot file, translated, and committed back in individual .po files
- when deploying to production, running `release-steps.sh` creates the .mo files from the latest .po files

**Note:** Mozilla doesn’t support `pt` as a locale, so I’m working around by copying pt_BR translations into the pt folder when deploying. We should probably use pt-BR on the website, but that’s out of the scope of this PR.


Pomax:
Does this make use of Django's i18n `trans` template tag functionality?

Pomax:
Ah, it does. This is an amazing PR, @TheoChevalier, you're a superhero. One thing that might be worth looking at using is the `as ...` bit of syntax, which lets you define an operational key for the string rather than relying on an entire string matching.

Question about the translations themselves: given that this is Django rather than node+json files, text changes no longer need a completely new key, do they? I.e. we can change strings, rather than having to add strings?

Pomax:
I've told Heroku to create a review app for this so we can test it "live".

TheoChevalier:
> Ah, it does. This is an amazing PR, @TheoChevalier, you're a superhero. One thing that might be worth looking at using is the `as ...` bit of syntax, which lets you define an operational key for the string rather than relying on an entire string matching.

Hm, I somehow completely missed that syntax, and it doesn’t seem to be in [this doc](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/)
I can certainly have a look, do you have an example?

> 
> Question about the translations themselves: given that this is Django rather than node+json files, text changes no longer need a completely new key, do they? I.e. we can change strings, rather than having to add strings?

For this setup, that’s correct. I believe Pontoon considers the entire string as the ID.
Making a slight update to a string generally results in [`fuzzy` strings ](https://www.gnu.org/software/gettext/manual/html_node/Fuzzy-Entries.html)in the .po files (`makemessages` is running `msgmerge` from gettext in the background). But sometimes it creates more noise than actually helps localizers, so we may want to disable it at some point. And to do that, it will likely require customizing `makemessages`… but let’s not think about that for now :)

Pomax:
Aaah, no I misread the docs. https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#trans-template-tag talks about using `as` which I read as indicating a key, but instead it captures the translation into a variable that can then be used in the template, without printing it then and there.

extract_strings.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/bin/env bash
```
Pomax: Let's turn this into an `invoke` task instead, so that it works on every platform. If you look at tasks.py, you can see that most commands are instructions to either run "other, smaller tasks", or for small tasks are simply execution instructions for manage.py, copying a dir, etc.

Pomax:
I don't know how deep the rabbit hole went, but: is there a way to make django "find bare strings"? If so, we should totally add that to our `test` task, so that there's a report of where a `{% trans %}` or `{% blocktrans %}` is missing.

alanmoo:
@Pomax What's going on with this? Seems like it's at the root of blocking a few things.

Pomax:
superceded by https://github.com/mozilla/foundation.mozilla.org/pull/3529

PR #3511, opened 2019-08-08
[WIP] Removal of signup related code [Part 1 of 2]

### LINKED ISSUE #3296: Confirm deadness of and remove vestigial file

youriwims:
_Removes sign up template, instance, and README references_

Related PRs/Issues: #3296, #3515

Petition/Sign Up Test Page: https://foundation-mofostaging-pr-3511.herokuapp.com/en/bannered-campaign-signup-test/


network-api/networkapi/campaign/views.py
```diff
@@ -31,7 +31,7 @@ def send_message(self, QueueUrl, MessageBody):
         }
 
 
-# Basket/Salesforce SQS client
+# Basket/Salesforce SQS client  ##### UPDATE?
```
youriwims: Does this need to be updated at all? I can't tell 😬

network-api/networkapi/wagtailpages/README.md
```diff
@@ -74,7 +74,7 @@ The base CTA model has a title, header, description, and a newsletter identifier
 This is an empty subclass of CTA, as the base fields already cover the signup CTA case.
 
 #### 2. Petition
-
+<!-- UPDATE? -->
 This is a subclass of CTA that adds checkbox labels and Basket->Salesforce submission. See its definition in `models.py` for the full list of fields this CTA makes use of.
```
youriwims: I don't see our use of basket in the codebase...am I missing something?

mmmavis:
@Pomax can you review this PR? I know little about our setup for Basket and Salesforce so will need your help here.

network-api/networkapi/wagtailpages/README.md
mmmavis: Yea... I'm that sure either. @Pomax I know for example for petition we are posting to `https://foundation.mozilla.org/api/campaign/petitions/<id>`. Is this considered as posting to Basket or Saleforce directly? 🙄

network-api/networkapi/campaign/views.py
Pomax: no, we still need SQS for petitions.

network-api/networkapi/wagtailpages/README.md
Pomax: we don't use basket directly anywhere in the codebase (anymore).

network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta.html
```diff
@@ -3,7 +3,3 @@
 {% if cta_type == 'Petition' %}
 	{% include './cta/petition.html' %}
 {% endif %}
-
-{% if cta_type == 'Signup' %}
```
Pomax: I assume this also means removing the Signup CTA subclass, but in a different (follow-up) PR?

youriwims: This is for removing the `signup`. In that case, I'll remove basket in a separate pr and will also file a ticket to remove the signup cta subclass from `models.py`. 👍

Pomax:
@youriwims note that this does _not_ close the issue mentioned in the first comment, it's only part 1 of 3. We don't want #3296 to get closed just by landing this first PR =)

youriwims:
> @youriwims note that this does _not_ close the issue mentioned in the first comment, it's only part 1 of 3, and we don't want #3296 to get closed just by landing this first PR =)

@Pomax you're absolutely correct, I forgot to remove my initial comment. Thanks for catching that!

youriwims:
cc @alanmoo

network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta.html
alanmoo: Removing this code from this file but leaving the check for Petition here makes it hard to trace what's supposed to render at the end of the day. Instead of removing this chunk of code, can you replace the content of `signup.html` with [approximately these lines](https://github.com/mozilla/foundation.mozilla.org/blob/10d2a12881a237cee7baaaf9cbb319715e7b8363/network-api/networkapi/wagtailpages/templates/wagtailpages/bannered_campaign_page.html#L63-L73) and then point to this CTA template? It feels like that'll be a whole lot cleaner to trace in the future.

youriwims:
@alanmoo this is ready when you get a chance.

network-api/networkapi/wagtailpages/README.md
mmmavis: ^ That's super helpful. Thanks @Pomax

network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta.html
```diff
@@ -4,6 +4,6 @@
 	{% include './cta/petition.html' %}
 {% endif %}
 
-{% if cta_type == 'Signup' %}
+{% if page.signup %}
```
alanmoo: Why did this need to change? My instinct is that this and line 3 should be of the same format (though I'm not sure which.)

youriwims: Yeah, I agree, but for some reason it doesn't pull in `signup.html` when using `cta_type  == 'Signup'`. @Pomax would you happen to know why that is? I looked in the codebase and it appears that 'signup' is a `cta_type`. 🤷‍♀

Pomax: `cta_type` comes from https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templatetags/mini_site_tags.py#L19-L34, which needs to be loaded on the page by using `{% load mini_site_tags %}`, because the signup used on _only_ be for minisites so we never moved the cta tag out of that set of tags, and then you need to make sure to use the `cta` tag itself in the template.

Pomax:
A summary of how things were originally written up, hopefully explaining how it works "now":

### Mini Sites

Site content used to be "mini sites", which could either be campaign minisites, or opportunity minisites. Both were a subclass of the Minisite model, and so were known to have "a `.cta` field" but campaigns had `.cta` pointing to a `Petition` object, and opportunities had `.cta` pointing to a `Signup` object. 

In order for campaigns and opportunities to render correctly, the minisite template would load the `mini_site_tags` collection of Django templating tags, which includes the `cta` tag. This tag, when used in the same template that loads the mini_site_tags, takes a `page` object as argument, looks at its `.cta` field, checks which class the object tied to that field has, and then renders `wagtailpages/templates/tags/cta.html` with an enriched context that contains the template context variable `cta_type`. That variable then determines whether `tags/cta/petition.html` or `tags/cta/signup.html` gets used to render the CTA.

### Bannered Campaigns

At some point, we needed "things that had both a petition and a signup form", so we created the `BanneredCampaignPage`, a model that inherits from `CampaignPage` so that it would have the `.cta` field pointing to a petition, which was the most important thing, and then had a separate `signup` field that pointed to an optional Signup object. 

To make this work, the bannered campaign template would also load the `mini_site_tags` tag set, and would use `{% cta page %}` to get the petition rendered correctly, but because the signup was a diferent field, it couldn't use that same tag to get the signup rendered correctly. Also, because the template code for signups wasn't quite what we wanted for banneredcampaignpages, we went with "normal template code" instead.

So the bannered campaign page inherits the Petition `.cta` and can get that rendered "for free", but adds a secondary Signup fields called `signup`, without any special templating. It just gets straight-up rendes it into the bannered campaign template, if it exists.

### Today

And then everything exploded and we're where we are today, where we don't really use Opportunity pages anymore? Which means we don't render Signup objects through the cta tag anymore (but we DO still use Signup objects internally - they're tied to React `JoinUs` components now).

The reason `cta_type` didn't work for signups anymore, but `page.signup` did, is because the `cta` tag requires that its passed page has a `.cta` field, which is only true for Opportunity pages. 

So this might in fact be a more complex cleanup job, where we actually figure out what the information architecture _should_ be, and then have a look at whether, and if so, how, we can get transition the code to match that.

@alanmoo this probably deserves a meeting, although I don't know when you'd be available for that.

alanmoo:
Thanks for laying that all out, Pomax. There’s a few points I want to clarify:

We do still use Opportunity pages- existing ones are up, of course, and there could be a use for their navigation structure across the site. What we don’t generally do is use the signup component on these pages.

For BanneredCampaignPages, the idea is that we could use a signup OR a petition on them; if we had built this as XOR functionality, it would have been fine. But due to the nature of the models, it was easier just to allow both on a page (though in practice page editors shouldn’t do that).

That all being said, I agree it needs a meeting, but I think that’s even putting the cart before the horse. I’m starting to pretty strongly feel we need to get our testing in order before we undertake a refactor of any sort just for the sake of cleaning up code. Let’s talk about that after stand today.

youriwims:
See https://github.com/mozilla/foundation.mozilla.org/pull/3515#issuecomment-521422958

alanmoo:
This has gone on far too long to justify the effort. We'll probably end up removing it when we change from SQS to Basket API calls anyway.

Men in black brain eraser

PR #3516, opened 2019-08-09
Remove title casing enforcement of nav items

alanmoo:
In the primary site navigation, "MozFest House" becomes "Mozfest House" and that's not good. Giving the users a little more freedom, with a minor cost (but everything that's in production already will look fine, so I don't think it's a concern.)


alanmoo:
Yeah, they have been thus far. Plus if they don't, it'll be an obvious fix.

PR #3521, opened 2019-08-09
Petition comment 3517

Pomax:


source/js/components/petition/petition.jsx
```diff
@@ -246,8 +246,9 @@ export default class Petition extends React.Component {
       postalCode = !!this.postalCode.element.value;
     }
 
-    let commentValue =
-      this.comment && this.comment.element && this.comment.element.value;
+    let commentValue = this.comment
+      ? this.comment.element && this.comment.element.value
+      : ``;
```
Pomax: `commentValue` needs to always be string...

PR #3528, opened 2019-08-12
Move org schema to homepage only

### LINKED ISSUE #3369: Ensure Organization Schema markup is only on homepage
**Describe the bug**
From [this comment](https://github.com/mozilla/foundation.mozilla.org/issues/3174#issuecomment-508458669), we should only have the organization schema JSON on the homepage, not every page on the Foundation site. So, we should move that block from the base template to the homepage template.

youriwims:
Closes #3369


network-api/networkapi/templates/pages/base.html
```diff
@@ -16,15 +16,7 @@
...
-    <script type="application/ld+json">
-   {
-     "@context" : "http://schema.org",
-     "@type" : "Organization",
-     "legalName" : "Mozilla Foundation",
-     "url" : "https://foundation.mozilla.org",
-     "logo" : "https://foundation.mozilla.org/_images/mozlogo.jpg"
-    }
-   </script>
+    {% block orgSchema%}{% endblock %}
```
Pomax: We should decide on a naming convention and then make sure to update all of them. Python is a "no camelcase, ever" language, so that's probably worth following in Django templates, too. The choice between snake_case and kebab-case (omg those names) is a bit more arbitrary though, so I guess as long as we pick one, we can update everything to that.

network-api/networkapi/templates/pages/base.html
```diff
@@ -16,15 +16,7 @@
...
-    <script type="application/ld+json">
-   {
-     "@context" : "http://schema.org",
-     "@type" : "Organization",
-     "legalName" : "Mozilla Foundation",
-     "url" : "https://foundation.mozilla.org",
-     "logo" : "https://foundation.mozilla.org/_images/mozlogo.jpg"
-    }
-   </script>
+    {% block org-schema%}{% endblock %}
```
Pomax: Turns out we settled on snake_case =S

youriwims: haha yes, making the update now.

PR #3529, opened 2019-08-12
Template localization

Pomax:
This puts a few changes on top of https://github.com/mozilla/foundation.mozilla.org/pull/3506


Pomax:
@alanmoo @TheoChevalier I removed the pt_BR specific code, because we should probably be thinking about what the deficiencies are here that gave rise to special copy instructions for `pt_BR` from pontoon, but `pt` in wagtail.

network-api/locale/fr/LC_MESSAGES/django.po
```diff
@@ -0,0 +1,227 @@
...
+"PO-Revision-Date: YEAR-MO-DA HO:MI+ZONE\n"
+"Last-Translator: FULL NAME <EMAIL@ADDRESS>\n"
+"Language-Team: LANGUAGE <LL@li.org>\n"
+"Language: \n"
+"MIME-Version: 1.0\n"
+"Content-Type: text/plain; charset=UTF-8\n"
+"Content-Transfer-Encoding: 8bit\n"
+"Plural-Forms: nplurals=2; plural=(n > 1);\n"
+
+#: .\network-api\networkapi\settings.py:364
```
TheoChevalier: Hm, looks like the `--ignore=` params got… ignored, I assume because of the path separator (Win / vs Unix \). We will need to make that dynamic.

TheoChevalier:
`network-api/locale/templates/LC_MESSAGES/django.po` (`.po`, not `.pot`) got committed by mistake, I assume you ran `makemessages` at least once without the `-l` params? If so, let’s just remove the file before merging

> 
>  I removed the pt_BR specific code, because we should probably be thinking about what the deficiencies are here that gave rise to special copy instructions for pt_BR from pontoon, but pt in wagtail.
> 
> Ideally, we stop using pt on the wagtail size and use the real locale code... which the original PR notes is out of scope, but I'm not so sure that it is. This might actually be the kind of work that we should block on not having a proper locale code in wagtail.

Sounds reasonable, it was out of cope for me mostly because I don’t know Wagtail enough to do that change myself and didn’t want to dig too deep.

network-api/locale/fr/LC_MESSAGES/django.po
Pomax: Hm, is this a result of `makemessages`? because windows has supported `/` as path separate since literally day 1, so if we can tell Django to always use unix delimiters, we should be fine.

TheoChevalier: Yeah, and it seems they fixed that very problem in Django directly in 2014? https://code.djangoproject.com/ticket/23298

TheoChevalier:
@Pomax fyi, I’ve rebased against master, and added a few smaller changes. I’ve kept all commits, but feel free to squash some of these as needed.

What are the next steps for this PR, figuring out a way to support `pt-BR` instead of `pt` in Wagtail?

Pomax:
Interestingly, I'm now running into this when I try to docker-makemessages:
```
Starting foundationmozillaorg_postgres_1           ... done
Starting foundationmozillaorg_watch-static-files_1 ... done
Loading .env environment variables…
CommandError: Can't find msguniq. Make sure you have GNU gettext tools 0.15 or newer installed.
```

Will investigate more tomorrow =x

TheoChevalier:
> Interestingly, I'm now running into this when I try to docker-makemessages:
> 
> ```
> Starting foundationmozillaorg_postgres_1           ... done
> Starting foundationmozillaorg_watch-static-files_1 ... done
> Loading .env environment variables…
> CommandError: Can't find msguniq. Make sure you have GNU gettext tools 0.15 or newer installed.
> ```
> 
> Will investigate more tomorrow =x

I guess we should add gettext to requirements, I’ve installed it on macOS via Homebrew with `brew install gettext`. Not sure if there’s a pip equivalent to that

TheoChevalier:
@Pomax can you try again after running `pip install python-gettext`?

I notice you added `--no-location` to the makemessages command, but we should try to keep that information if possible. (Fine to do that in a followu-up) The location is showing up in Pontoon and it’s a convenient way for localizers to understand where a string is showing up, and if a string is used in multiple places.

Pomax:
Ah, I had added --no-location because it's supposed to stop the files from writing in paths with `\` rather than `/`.

Pomax:
@Theo even with `inv docker-pipenv "install get-text"` I still get that msguniq error... =S

TheoChevalier:
@Pomax  Alright, I’ve set up the project with Docker and was able to reproduce. Looks like gettext need to be installed at the image level, I managed to get it running with the following https://github.com/mozilla/foundation.mozilla.org/pull/3529/commits/ccd4953115e3c2281312285820d967d7168d56f6#diff-61b2823dc01ac137972c6d05716f43ca

TheoChevalier:
@Pomax , would you mind reviewing this bad boy so that we can try to land it soon?

Some notes on the latest commits:

- Imported Chris’ work to install gettext on heroku (assuming the same issue would happen on the staging server for this project)
- Removed `--no-location` from `makemessages`, because:
  - Pontoon will reformat those lines differently when committing anyway
  - Different path separators doesn’t break anything, it just makes bigger diffs if someone else is extracting strings from a different platform (But I think it will mostly be me doing it)
  - Localizers need this information
- Removed `pt` support altogether for now, in hope we can land this PR sooner

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -1,5 +1,6 @@
 from django.apps import apps
 from django.db.models import Count
+from django.utils.translation import gettext as _
```
Pomax:
```suggestion
from django.utils.translation import gettext
```

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -54,7 +55,7 @@ def get_descendants(node, list, authenticated=False, depth=0, max_depth=2):
         header = getattr(node.specific, 'header', None)
         if header:
             title = header
-        menu_title = title if depth > 0 else 'Overview'
+        menu_title = title if depth > 0 else _('Overview')
```
Pomax:
```suggestion
        menu_title = title if depth > 0 else gettext('Overview')
```

Pomax:
@cadecairos tagged you for review just to make sure the release script bits are good to go - we'd like to land this as soon as possible so we can finally tick off this box =)

TheoChevalier:
Moved all `compilemessages` logic from `release steps.sh` to `bin/post_compile` in https://github.com/mozilla/foundation.mozilla.org/pull/3529/commits/db070dc837aa67fc4c268da7354930112514cc56

Now that should be consistent with the solution on donate-wagtail, minus the pt-BR file copy that is specific to this project.

PR #3530, opened 2019-08-13
 wrap a couple more strings + move .pot file once generated

TheoChevalier:
@Pomax would you mind merging this into your branch? I can’t just push there.

We will need an equivalent of `os.rename("network-api/locale/django.pot", "network-api/locale/templates/LC_MESSAGES/django.pot")` in `docker_makemessages`, but I don’t know which function to use.

And I still don’t understand why the `--ignore=` didn’t work for you


TheoChevalier:
Nevermind, I can’t push, but can merge :)

TheoChevalier:
Actually, I should be able to push to the branch, I’ve updated my config to use ssh.
(I thought the repository was restricted to some people)

Issue #3535, opened 2019-08-14
Fix docker-setup so that it uses a clean db


Pomax:
Docker-setup uses a postgres image, but does not "clean" that image during setup, which causes load_fake_data to find all kinds of data already in the db, and causes createsuperuser to go "admin already exists, please input your own username".

PR #3536, opened 2019-08-14
make docker-setup fall through to docker-new-db

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3535  by making `inv docker-setup` fall through to the `docker_new_db` task, so that even though postgres is installed from an image, it gets a good wipe before bootstrapping the db (including fake data).

This hopefully also fixes converalls.

Test this by pulling master, running `inv docker-setup` to _verify things go wrong_ then checking out this branch and rerunning `inv docker-setup` and see if it makes it all the way to the end.


youriwims:
@Pomax is there a specific error that I'm looking for when 'things go wrong'?...or just any error _really_?

PR #3545, opened 2019-08-16
Remove /en from links and routes

### LINKED ISSUE #3475: Replace all /en links

mmmavis:
Closes #3475


network-api/networkapi/utility/middleware.py
```diff
@@ -37,6 +37,7 @@ def __call__(self, request):
                     redirect_url = '{protocol}://{hostname}{path}'.format(
                         protocol=protocol,
                         hostname=hostnames[0],
+                        # FIXME:TODO: what to do with this /en link
```
mmmavis: not 100% sure if we need to keep the `/en` here

network-api/networkapi/urls.py
```diff
@@ -40,6 +40,7 @@
         name='how-do-i-wagtail'
     ),
     url(r'^cms/', include(wagtailadmin_urls)),
+    # FIXME:TODO: what to do with this /en route
```
mmmavis: What to do with line 44? Do we keep the whole line? Remove the whole line? Replace `en` with the locale variable?

mmmavis:
@Pomax this is ready for review! (I left some questions regarding MozFest routes.)

network-api/networkapi/buyersguide/templates/category_page.html
```diff
@@ -19,7 +19,7 @@ <h1 class="h2-heading">{{ category.name }}</h1>
         <div class="row">
           {% for product in products %}
           <div class="category-item-container col-6 {% if product.draft %} draft-product{% endif %}{% if product.adult_content %} adult-content{% endif %}">
-            <a href="/en/privacynotincluded/products/{{ product.slug }}">
+            <a href="/privacynotincluded/products/{{ product.slug }}">
```
Pomax: if I understand the documentation correctly, we might be able to use `<a href="{{ product.url }}">` here, and wagtail will automatically use the correct relative url because it's internal... we should test that, because if it is a whole bunch of our urls can be turned into single `blah.url` instructions rather than partially hardcoding them.

network-api/networkapi/urls.py
Pomax: we want to keep it: the `/cms` is not part of the localised routes, but everyone in the engineering team keeps accidentally going to `localhost:8000/en/cms` etc, so we added this route to make sure that even though that's not a real route as far as wagtail is concerned, _we_ make it a real route and you can get to the cms rather than hitting a 404 =)

network-api/networkapi/utility/middleware.py
Pomax: yeah, this shouldn't need the `/en` prefix, it should automatically add that in.

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -59,7 +59,7 @@ <h2 class="h1-heading col-md-10">{{product.name}}</h2>
       <div id="product-research" class="mt-5">
         <div class="mb-5 d-flex flex-column flex-sm-row align-items-sm-center">
           <h3 class="h5-heading d-inline-block mb-2 mb-sm-0 mr-3">Our Research</h3>
-          <a href="/en/privacynotincluded/about" class="info-help">How to use this guide</a>
+          <a href="/privacynotincluded/about" class="info-help">How to use this guide</a>
```
Pomax: I wonder if there's a way to turn this into a `{{ pageurl about_page }}` instruction of some sort, so that we don't have to rely on this page actually being titled "About" in the cms...?

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -158,7 +158,7 @@ <h3 class="h2-heading mb-4">Compare products</h3>
           <div class="row">
             {% for product in product.related_products %}
             <div class="related-product col-6 col-md-3 mb-3 mb-md-0">
-              <a class="d-block{% if product.adult_content %} adult-content{% endif %}" href="/en/privacynotincluded/products/{{ product.slug }}">
+              <a class="d-block{% if product.adult_content %} adult-content{% endif %}" href="/privacynotincluded/products/{{ product.slug }}">
```
Pomax: `{{ pageurl product }}` probably works here, too.

network-api/networkapi/mozfest/templates/mozfest/404.html
```diff
@@ -21,6 +21,7 @@
                             <div class="row">
                                 <div class="col">
                                     <div class="nav-links pt-3">
+                                        <!-- FIXME:TODO: what to do with these en links? -->
                                         <div><a class="active" href="https://www.mozillafestival.org/en/">Home</a></div>
```
Pomax: we can remove the `/en` for both of these, as mozillafestival.org is served by the same wagtail instance, just over a different domain binding.

network-api/networkapi/buyersguide/templates/category_page.html
mmmavis: (If I understand the following line correctly) `Product` is not a Page object so I'm afraid we can't use pageurl on it.

```
class Product(models.Model):
```

mmmavis: Got errors when I tried it

```py
TemplateSyntaxError at /en/privacynotincluded/categories/toys-games/

Could not parse the remainder: ' product' from 'page_url product'
```

mmmavis:
```py
{% url 'product-view' product.slug %}
```
seems to work so I'll use this instead.

network-api/networkapi/buyersguide/templates/product_page.html
mmmavis: Using built-in template tag `url` works!

mmmavis:

network-api/networkapi/buyersguide/templates/category_page.html
Pomax: ah, good finds!

PR #3571, opened 2019-08-23
Signup form ga label

### LINKED ISSUE #3480: Newsletter sign-up component: include position info in GA event
Based on this [comment](https://github.com/mozilla/foundation.mozilla.org/issues/2994#issuecomment-517395655): 

(if possible) add position to the Label to identify where the form shows up.  For example `Signup submitted from Footer` OR form `from Header`.

youriwims:
Closes #3480

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -133,6 +133,20 @@ export default class JoinUs extends React.Component {
     });
   }
 
+  renderGAEventLabel() {
+    let formPosition = this.props.formPosition;
+    switch (formPosition) {
+      case `header`:
+        return `Signup submitted from Header`;
+        break;
+      case `footer`:
+        return `Signup submitted from Footer`;
+        break;
+      default:
+        return `Signup submitted`;
+    }
+  }
+
   /**
    * Process the form data, to make sure there is a valid
    * email address, and the consent checkbox has been checked,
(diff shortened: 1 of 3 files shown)


source/js/components/join/join.jsx
```diff
@@ -133,6 +133,20 @@ export default class JoinUs extends React.Component {
     });
   }
 
+  renderGAEventLabel() {
+    let formPosition = this.props.formPosition;
+    switch (formPosition) {
```
mmmavis: Since this is a simple if/else, we can simplify switch statement into

```jsx
let gaLabel = `Signup submitted`;

if (formPosition) {
  gaLabel = `${label} from ${formPosition}`;
}
```

(You can merge this further into one line but I find that code hard to read. That's why I separate the `if` statement out)

source/js/components/join/join.jsx
```diff
@@ -133,6 +133,20 @@ export default class JoinUs extends React.Component {
     });
   }
 
+  renderGAEventLabel() {
```
mmmavis: Since we are not actually "rendering" the label, let's change the method name to something else + make the name more specific as there are different GA events floating around in the file. Maybe something like `getGaSubmittedLabel`, `generateGaSubmittedLabel`, or `createGaSubmittedLabel`?

source/js/components/join/join.jsx
mmmavis: Note that

We don't have to worry too much on our end when it comes to letter case in event labels. GA does its magic to turn event labels into the letter case style it likes.
e.g.,


image

youriwims:
Matthew gave the go ahead to add the page title for the signups that are injected into the page via streamfield, so I updated that.

PR #3584, opened 2019-08-27
.has-danger form updates

### LINKED ISSUE #3508: "has-danger" class for the footer newsletter signup is invisible.
When the footer's newsletter signup input field has the `has-danger` class, which is supposed to highlight an error with a red border, there does not appear to be a visual cue:

image

As a design question: does this need a visual cue at all?
As dev question: if yes, what's the best SCSS to use here? If no, is it more work to remove the code that adds `has-danger` than it is to just leave it as is? (with a code comment that points to this issue)

youriwims:
Closes #3508 

Test Page: https://foundation-mofostaging-pr-3584.herokuapp.com/en/campaigns/test/


natalieworth:
hey @youriwims the desktop looks good, but the mobile version is to the right of the content. I don't mind it though/ am okay with it if it's too difficult to add it beside the second line.
Screen Shot 2019-08-28 at 10 33 22 PM

youriwims:
This is, when they're both in error form, it looks a it off to me...idk. I think it's fine at the end 🤷‍♀ 

Screen Shot 2019-08-29 at 2 23 31 PM

mmmavis:
@youriwims is this ready for code review or we are still waiting for @natalieworth to finish design review first?

youriwims:
No, this is ready for your review. I got design approval from Kristina. I'll remove Nat.

mmmavis:
@youriwims can you fix the base branch to it's `development` instead of `master`?

youriwims:
Most def, I have it in my to-dos in my initial comment. Can we do that after code-review?

mmmavis:
@youriwims um if the base branch is different we can't tell what the exact code changes we are adding the `development`. Is there something that's blocking you from changing the base branch now?

source/js/components/join/join.jsx
```diff
@@ -224,28 +224,48 @@ export default class JoinUs extends React.Component {
...
+            </p>
+          )}
         {this.state.userTriedSubmitting &&
           !this.state.apiSubmitted &&
-          !this.validatesAsEmail(this.email.value) && (
+          !this.validatesAsEmail(this.email.value) &&
+          (!this.email.value == "" || !this.email.value == null) && (
             <p className="body-small form-check form-control-feedback">
-              Please enter your email
+              Please enter a valid email address.
```
mmmavis: This block is quite confusing (Line 257-270)


```jsx

{this.state.userTriedSubmitting &&
  (this.email.value == "" || this.email.value == null) && (
    <p className="body-small form-check form-control-feedback">
      This is a required section.
    </p>
  )}
{this.state.userTriedSubmitting &&
  !this.state.apiSubmitted &&
  !this.validatesAsEmail(this.email.value) &&
  (!this.email.value == "" || !this.email.value == null) && (
    <p className="body-small form-check form-control-feedback">
      Please enter a valid email address.
    </p>
  )}
```

Why are we checking `this.state.apiSubmitted` just for the one case and not the other? (You might want to verify if checking `this.state.apiSubmitted` is really necessary.)

Since the two error messages is mutually exclusive we should be able to merge & simplify the above code into something like

```jsx
{this.state.userTriedSubmitting && (
  <p className="body-small form-check form-control-feedback">
    {figure out what error message to show here}
  </p>
)}
```

source/js/components/join/join.jsx
```diff
@@ -288,40 +308,47 @@ export default class JoinUs extends React.Component {
...
+              <p className="d-inline-block body-small my-0">
+                I'm okay with Mozilla handling my info as explained in this{" "}
+                <a href="https://www.mozilla.org/privacy/websites/">
+                  Privacy Notice
+                </a>
+              </p>
+            </label>
+          </div>
           {this.state.userTriedSubmitting &&
             !this.state.apiSubmitted &&
```
mmmavis: Verify if checking `this.state.apiSubmitted` is really necessary.

source/js/components/join/join.jsx
```diff
@@ -288,40 +308,47 @@ export default class JoinUs extends React.Component {
...
             !this.privacy.checked && (
-              <p className="body-small form-check form-control-feedback">
-                Please check this box if you want to proceed
-              </p>
+              <span class="form-error-glyph d-flex mt-1 mt-lg-0" />
             )}
-        </label>
+        </div>
+        {this.state.userTriedSubmitting &&
+          !this.state.apiSubmitted &&
```
mmmavis: Verify if checking `this.state.apiSubmitted` is really necessary.

source/js/components/join/join.jsx
```diff
@@ -224,28 +224,48 @@ export default class JoinUs extends React.Component {
...
+        <div className={classes}>
+          <input
+            type="email"
+            className="form-control align-items"
+            placeholder="Enter email address"
+            ref={el => (this.email = el)}
+            onFocus={evt => this.onInputFocus(evt)}
+          />
+          {this.state.userTriedSubmitting &&
+            !this.state.apiSubmitted &&
```
mmmavis: Verify if checking `this.state.apiSubmitted` is really necessary.

youriwims:
> have we decided to change where we wanna position the exclamation mark icon for email field (on non-desktop screen sizes)? What we have on the review app is different than what's in the Sketch file so I can't tell what's final.

@mmmavis the mobile mocks weren't initially included in the og ticket. After I finished building the error state, Nat pointed out the inconsistency & we talked about it and decided that the svg placement on mobile isn't a blocker and it's fine as is. She won't be able to update Sketch/Abstract though since she's out this week. 

cc @natalieworth

source/js/components/join/join.jsx
youriwims: The first error message is for when the input is empty vs the second error for the when there is text in the input but it's not a valid email address.

source/js/components/join/join.jsx
```diff
@@ -224,28 +224,48 @@ export default class JoinUs extends React.Component {
...
+            !this.validatesAsEmail(this.email.value) && (
+              <div className="glyph-container">
+                <span className="form-error-glyph" />
+              </div>
+            )}
+        </div>
+        {this.state.userTriedSubmitting &&
+          (this.email.value == "" || this.email.value == null) && (
```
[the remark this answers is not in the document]
youriwims: No, here I'm checking if the input is blank. Also, I didn't want a strict comparison here.

source/js/components/join/join.jsx
```diff
@@ -224,28 +224,48 @@ export default class JoinUs extends React.Component {
...
+            <p className="body-small form-check form-control-feedback">
+              This is a required section.
+            </p>
+          )}
         {this.state.userTriedSubmitting &&
           !this.state.apiSubmitted &&
-          !this.validatesAsEmail(this.email.value) && (
+          !this.validatesAsEmail(this.email.value) &&
+          (!this.email.value == "" || !this.email.value == null) && (
```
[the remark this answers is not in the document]
youriwims: Do you mean using `validatesAsEmail()`? I only used it here and for the error svg in `<input>`.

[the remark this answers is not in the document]
mmmavis: Yup 'cuz we are already checking if email is falsey (null or blank) in `validatesAsEmail()`, by doing `(!this.email.value == "" || !this.email.value == null)` we are doing the same check the second time.

Maybe we can modify `validatesAsEmail` to something like blow. This way we can better separate validation logic from render code.

```jsx
validatesAsEmail(email) {
  if (blank or null) {
    return {
      type: `blank`,
      errorMessage: `This is a required section.`
    }
  }

  if (invalid) {
    return {
      type: `invalid`,
      errorMessage: `Please enter a valid email address.`
    }
  }

  return {
    type: `valid`
  }
}
```

source/js/components/join/join.jsx
```diff
@@ -288,40 +308,47 @@ export default class JoinUs extends React.Component {
...
+        <div className="d-flex align-items-start">
+          <div className="mb-0 form-check d-flex align-items-start">
+            <input
+              type="checkbox"
+              className="form-check-input ml-0"
+              id="PrivacyCheckbox"
+              ref={el => (this.privacy = el)}
+              required
+            />
+            <label className="form-check-label">
```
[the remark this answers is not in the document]
youriwims: It works, but I can't line up the svgs [vertically] using bootstrap at medium devices.

source/js/components/join/join.jsx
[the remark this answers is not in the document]
youriwims: ah, good idea.

source/js/components/join/join.jsx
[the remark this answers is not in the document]
youriwims: Hey, this works but takes just as much to align the svgs as absolute positioning as the spacing can't be fixed with just bootstrap. So, if you're okay with it, can we leave the absolute positioning?

youriwims:
@mmmavis I tried changing this branch to master so ci could run, but it was 'in progress' for like an hour. But, it's ready for re-review! 😬

source/js/components/join/join.scss
```diff
@@ -2,27 +2,68 @@
...
+      border-color: $red;
+    }
+
+    .form-error-glyph {
+      @media (min-width: $bp-md) {
+        margin-left: 0;
+      }
+    }
+
+    // .form-error-glyph:nth-child(2) {
```
mmmavis: Don't forget to glance over code changes in the PR and remove irreverent code before requesting review!

source/sass/glyphs.scss
```diff
@@ -39,6 +39,14 @@
       top: 2px;
     }
   }
+
+  &.form-error-glyph {
+    &::before {
+      width: 13px;
+      height: 20px;
```
mmmavis: This glyph itself is 13x13px so let's make sure height is set to match that value.

source/js/components/join/join.scss
```diff
@@ -2,27 +2,68 @@
...
   }
 
   @at-root .dark-theme & {
-    form {
-      input {
-        border: none;
-      }
-    }
-
     .body-small.form-control-feedback {
```
mmmavis: Do we need to target `.body-small` here? Using just `.form-control-feedback` as selector should be enough.

source/js/components/join/join.scss
```diff
@@ -2,27 +2,68 @@
...
       color: $red;
     }
+
+    .has-danger .form-control {
+      border-color: $red;
+    }
+
+    .form-error-glyph {
+      @media (min-width: $bp-md) {
+        margin-left: 0;
```
mmmavis: What is this for? Why is this only set in `.dark-theme` mode?

source/js/components/join/join.scss
```diff
@@ -2,27 +2,68 @@
...
+      position: absolute;
+      top: 0;
+      bottom: 0;
+      right: 10px;
+      height: 13px;
+      width: 13px;
+      margin: auto;
+    }
+
+    .form-error-glyph:nth-child(2) {
```
mmmavis: Let's clean up how we set `margin-left` for this element. Currently I can see it being set three times but I can't understand why. Also I would suggest giving this targeted `.form-error-glyph` element a class name so we can just target it using that class name instead of `.form-error-glyph:nth-child(2)`.

image

package-lock.json
```diff
@@ -4606,8 +4606,7 @@
         "ansi-regex": {
           "version": "2.1.1",
           "resolved": false,
-          "integrity": "sha1-w7M6te42DYbg5ijwRorn7yfWVN8=",
```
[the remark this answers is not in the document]
youriwims: My node modules were empty and I ended up having to run a few `npm` commands when I was troubleshooting with pomax, so this may be from that.

source/js/components/join/join.jsx
```diff
@@ -224,28 +235,39 @@ export default class JoinUs extends React.Component {
    * Render the email field in signup CTA.
    */
   renderEmailField() {
-    let classes = classNames(`mb-2`, {
+    let wrapperClasses = classNames({
       "has-danger":
         (!this.state.apiSuccess &&
           this.state.userTriedSubmitting &&
-          !this.validatesAsEmail(this.email.value)) ||
+          this.validatesAsEmail(this.email.value).errorMessage) ||
```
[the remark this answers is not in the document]
youriwims: I tried storing this as a variable and each time, I get the error that `this.email` is undefined for some odd reason.

youriwims:
@mmmavis I added the padding fix similar to the one you did for the Pulse search to fix the overlap. I also added comments to address two of your comments. Let me know if you have any ideas for the storing the validation function in the variable.

source/js/components/join/join.jsx
```diff
@@ -83,9 +83,20 @@ export default class JoinUs extends React.Component {
    */
   validatesAsEmail(input) {
```
mmmavis: In `processFormData(event) { ... }`, there's this line `if (email && this.validatesAsEmail(email) && consent) {`. This means even when we have validation error `this.validatesAsEmail(email)` is still true as it returns a object. 

Because of this let's change `validatesAsEmail` to the following code to make sure the type of result this method returns is consistent.

```jsx
validatesAsEmail(input) {
  if (!input) {
    return {
      valid: false,
      errorMessage: `This is a required section.`
    };
  }

  let formatValid = input.match(/[^@]+@[^.@]+(\.[^.@]+)+$/) !== null;

  if (!formatValid) {
    return {
      valid: false,
      errorMessage: `Please enter a valid email address.`
    };
  }

  return {
    valid: true
  };
}
```

To check if email is valid, we do ` this.validatesAsEmail(email).valid` instead of ` this.validatesAsEmail(email)`.

source/js/components/join/join.jsx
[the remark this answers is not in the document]
mmmavis: Ah I see. This is because we are using `ref`. We should eventually rewrite this so we use `this.state` to store form values instead of binding fields using `ref` and access value using this.refName.value. However, this is out of the scope of this ticket.

Let's add this to `renderEmailField() {`,

```jsx
renderEmailField() {
  // During the first render, we bind the email field to this.email
  // using ref={el => (this.email = el)}
  // This means this.email is undefined until the second render.
  // To avoid TypeError we have do the following conditonal assignment.
  let emailValidation = this.email
    ? this.validatesAsEmail(this.email.value)
    : false;
```

(feel free to edit the inline comments so the wording sounds better for you)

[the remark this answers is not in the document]
mmmavis: To check if email value is valid let's do it more explicitly (e.g., Instead of checking `errorMessage` we check the `valid` property.)

So we do `emailValidation.valid` to check if email is valid. If email is invalid, we then render the error message using `emailValidation.errorMessage`.

mmmavis:
@youriwims code looks good but since `development` branch has been merged into `master`, you will have to rebase your branch (again 😬) off of `master`. I'll do a quick review after the rebase just to be safe.

youriwims:
ah yes!doing that now!

source/sass/glyphs.scss
```diff
@@ -60,6 +68,10 @@
   @include hoverGlyph("linkedin-hover");
 }
 
+.form-error-glyph {
+  @include compositeGlyph("form-control-error");
```
mmmavis: let's call this `form-error` so it matches the classname naming convension

source/images/glyphs/form-control-error-dark-theme.svg
```diff
@@ -0,0 +1 @@
+<svg height="13" viewBox="0 0 13 13" width="13" xmlns="http://www.w3.org/2000/svg"><path d="m6.5 13c-3.58985087 0-6.5-2.9101491-6.5-6.5 0-3.58985087 2.91014913-6.5 6.5-6.5 3.5898509 0 6.5 2.91014913 6.5 6.5 0 3.5898509-2.9101491 6.5-6.5 6.5zm.889-11.1h-1.778l.336 6.062h1.092zm-1.764 9.1h1.792v-1.778h-1.792z" fill="#ff4f5e" fill-rule="evenodd"/></svg>
```
mmmavis: change file name to `form-error-dark-theme.svg`

source/images/glyphs/form-control-error.svg
```diff
@@ -0,0 +1 @@
+<svg height="13" viewBox="0 0 13 13" width="13" xmlns="http://www.w3.org/2000/svg"><path d="m6.5 13c-3.58985087 0-6.5-2.9101491-6.5-6.5 0-3.58985087 2.91014913-6.5 6.5-6.5 3.5898509 0 6.5 2.91014913 6.5 6.5 0 3.5898509-2.9101491 6.5-6.5 6.5zm.889-11.1h-1.778l.336 6.062h1.092zm-1.764 9.1h1.792v-1.778h-1.792z" fill="#cc0011" fill-rule="evenodd"/></svg>
```
mmmavis: change file name to `form-error.svg`

Issue #3564, opened 2019-08-20
Blog: build Wagtail component


kristinashu:
Create a component (similar to the Pulse component) that can be embedded on any wagtail page to show recent blog posts within a category or a tag.

### Shows most recent posts from either
  - a category
  - or a tag

### Component content
- section title
- 6 most recent posts
- “More from our blog” button

### Thumbnail content
- OG image
- category (hidden if filtered by category since they will all be the same)
- author
- date posted
- OG title
- ~OG description~

### Comp
https://redpen.io/p/bha12cf7387de4f7e7

![image](https://user-images.githubusercontent.com/1068783/63386326-a5a2ed00-c357-11e9-84a7-f40531b606a8.png)

![](https://s3.amazonaws.com/redpen-prod/red-pen-fd362191-9a70-4af9-f876-c58b9ba183d4.jpg)

cc @Pomax

youriwims:
@kristinashu can you confirm if these will be the only tags for this component? 
- tags
  - 'mozilla'
  - 'privacy'
  - 'security'
  - 'internet health'
  - 'digital inclusion' 
  - 'advocacy'
  - 'policy' 

I took the list of tags from Pomax's blog-tag PRs. If not, can you send the list of pre-defined tags our users should be able to choose from?

kristinashu:
Categories are set to only 5 different ones https://github.com/mozilla/foundation.mozilla.org/issues/3560#issue-482979249 but tags could be anything. Staff can type anything into the tag field on a blog post detail page:
<img width="498" alt="image" src="https://user-images.githubusercontent.com/1068783/64054409-b1eb2f00-cb3b-11e9-8680-d76269d607ac.png">

I assumed filtering by a tag would be similar to the Pulse component where users could put in anything:
<img width="1256" alt="image" src="https://user-images.githubusercontent.com/1068783/64054564-65ecba00-cb3c-11e9-9b68-f8219a0f9f39.png">

youriwims:
Sure, I could do the tag filtering by search similar to the _Pulse Listing_ component. A few other questions:

1. I just want to confirm that we're not letting users choose how many entries to display? It will always be 6, right?
2. Will the “More from our blog” button take them to the tag-filtered blog page or should that component expand and continue loading results?
3. Will the title always be _Latest News_ or should that be editable?
4. What did you mean by OG **description** for the thumbnail?

Thank you! 😄

kristinashu:
Thank you for the questions!

[...]

PR #3597, opened 2019-08-30
[Component] Build recent blog entries wagtail component

youriwims:
Related PRs/issues #3564

## Checklist

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?


youriwims:
@mmmavis I created this branch from development and I see all of your commits 😬 this might get weird, hopefully not too weird🤞

mmmavis:
@youriwims I think once you change the branch your branch compares against from `master` to `development` we should be good 👍 .

network-api/networkapi/wagtailpages/customblocks/recent_blog_entries.py
```diff
@@ -0,0 +1,82 @@
...
+            ('all', 'All'),
+            ('Mozilla Festival', 'Mozilla Festival'),
+            ('Open Leadership & Events', 'Open Leadership & Events'),
+            ('Internet Health', 'Internet Health'),
+            ('Fellowships & Awards', 'Fellowships & Awards'),
+            ('Advocacy', 'Advocacy'),
+        ]
+    )
+
+    top_divider = blocks.BooleanBlock(
```
Pomax: If we really need these, we want to refactor this component and the pulse entry component to have a shared "parent" that does all this and we can get "for free" by extending, rather than duplicating it.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -586,25 +587,32 @@ def filter_entries_for_tag(self, entries, context):
...
-            for
-            entry in entries.specific()
-            if
-            hasattr(entry, 'tags')
-            and not
-            # Determine whether there is any overlap between 'all tags' and
-            # the tags specified. This effects ANY matching (rather than ALL).
-            set([tag.slug for tag in entry.tags.all()]).isdisjoint(terms)
-        ]
+        type = self.filtered.get('type')
```
Pomax: We probably need to revisit this change, as it's on the merged master now. It might not be correct anymore.

network-api/networkapi/wagtailpages/customblocks/recent_blog_entries.py
youriwims: ah yes, this makes complete sense 👍 ...but, what order would I do that in?

1. Create parent component
2. Extend class in recent/latest blog entry component
3. After 2 lands, refactor the other components that include the dividers and extend the class there?

network-api/networkapi/wagtailpages/models.py
youriwims: I updated this line to `terms = self.filtered.get('terms')` and removed lines 591-594 to match your category blog pr changes. Is that correct or should L590 be the only thing updated here?

network-api/networkapi/wagtailpages/customblocks/recent_blog_entries.py
Pomax: folding it into this PR should be fine, but if that would hold up the PR, we can land this first, with an immediate followup. (e.g. do the refactor while this PR is getting reviewed)

youriwims:
@kristinashu, I created a few test pages in the initial comment but of course, feel free to create others :)

youriwims:
@Pomax should we be add all new blocks to README?

Pomax:
ideally, yes

youriwims:
@kristinashu this is ready for re-review :)

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog-card.html
```diff
@@ -12,18 +12,16 @@
...
-      {% for tag in page.specific.tags.all %}
-        <li class="d-inline-block text-capitalize">
-          <a href="{% routablepageurl root.specific "entries_by_tag" tag.slug %}">{{ tag }}</a>{% if not forloop.last %}, {% endif %}
-        </li>
-      {% endfor %}
-      </ul>
-    {% else %}
-      {# Show the tag's own "main" tag (i.e. first tag in its tag list) as each blog card's visible tag #}
-
+  {% if filtered and root %}
```
Pomax: this needs that `{% if hide_classifiers != True %}.....{% endif %}` put back around it, as category pages shouldn't be showing the primary tag or any tags at all.

youriwims: just fixed the conflict.

Pomax:
there are 2 mozfest migrations in this PR that shouldn't be there.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic-card.html
```diff
@@ -8,7 +8,7 @@
   </div>
   {% block tags %}{% endblock %}
   <a class="h4-heading d-inline-block my-2" href={{ page.url }}>{{ page.title }}</a>
-  <p class="body-small mb-0">
+  <p class="body-small my-0">
```
Pomax: this also affects every other instance of cards: is this the intended spacing for index pages as well as "related cards" on individual blog pages?

youriwims: Yeah, that's a fix because in `wagtail.scss` there's a `margin-top: 1rem` on all `<p>` that are children of an element with the `.cms` class. Because of this, this component's blog cards  just have the extra margin-top removed and as for the other card instances, they're practically unaffected by this change.

Issue #3562, opened 2019-08-20
Blog: Build category overview page


kristinashu:
Blocked by - Add categories field https://github.com/mozilla/foundation.mozilla.org/issues/3560

We need one overview page for each of the ~6~ 5 categories

URL: foundation.mozilla.org/blog/[category-name]

### Page content: 
- title of category
- unique editable intro text 
- intro text can have url link in the copy
- most recent posts
- “See more” button
- future feature: ability to have featured posts at the top of the page

### Thumbnail content: 
- OG image
- category is hidden
- author
- date posted
- OG title

### Comp
https://redpen.io/bkd71cb03a53bdf596

![](https://s3.amazonaws.com/redpen-prod/red-pen-24dd6d42-9b68-49d3-e9ed-f44115a59231.jpg)

cc @alanmoo @Pomax

Pomax:
I see only fix categories in the codebase:

- Mozilla Festival
- Open Leadership & Events
- Internet Health
- Fellowships & Awards
- Advocacy

Which one are we missing?

Pomax:
Will categories only ever apply to Blog pages, or are the same categories going to end up getting used for things like campaigns and/or one-off pages?

kristinashu:
Oops only 5 categories, I've updated the first comment!

kristinashu:
There will be a Wagtail component https://github.com/mozilla/foundation.mozilla.org/issues/3564 that will pull post based on the category but I don't foresee using categories anywhere else.

Pomax:
Due to wagtail's routing model, category URLs will need to be `/blog/category/(catname)`. Hopefully that's not a problem?

kristinashu:
Yeah that url structure is ok with me.

Pomax:
In the comp, is the "Mozilla Festival Blog" heading a custom heading, or a compound of `{category name}` + `Blog`?

kristinashu:
I like the consistency of making it a compound `{category name}` + `Blog` (in Title Case).

Pomax:
Also, is the thumbnail content the standard blog card content that we also see on `/blog` and `/blog/tags/...` ? (or, even if not the same as exists right now, is it _supposed_ to all be the same cards, or do these somehow differ?)

kristinashu:
Right, it should be the same image, title, author, date as on `/blog`. Only difference is that the tag/category is hidden since they will all be the same (related: Blog: update tags to categories on /blog #3563).

[...]

PR #3604, opened 2019-09-03
Category pages (for Blog)

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3562 once finished

title is a compound of category name + base page title (e.g. for base page title "blog" and category "mozilla festival", this becomes "Mozilla Festival Blog")

filtered urls:
https://foundation-mofostaging-pr-3604.herokuapp.com/en/blog/category/advocacy/
https://foundation-mofostaging-pr-3604.herokuapp.com/en/blog/category/mozilla-festival/

Category intros are managed by the CMS, in https://foundation-mofostaging-pr-3604.herokuapp.com/cms/snippets/wagtailpages/blogpagecategory/

**After prod push**

Make sure that barely anyone has edit rights on categories, and only admins have add/delete rights (i.e. no regular user group policy allows add/delete)


Pomax:
@kristinashu this should cover the majority of work, can you give it a quick "test drive" to see if it feels reasonable?

Pomax:
- [ ] make sure to set the correct context variable so that dynamically loaded results obey the same `hide_classification` flag

kristinashu:
Amazing, it's super cool to see this!

Post page looks awesome! Nice to see links working. Is the intro text .body-large? It says it is but it looks a little small.
image

And as discussed in Slack, please just make sure only a few very select people can add or edit a category (category names shouldn't change and intro text should rarely change). 

image

image

Pomax:
Aha! the outer tag for the intro is the wrong tag, so you are absolutely correct: it's not using body-large properly.

Pomax:
@kristinashu fixed the two issues, can you confirm things look/work right now?

@mmmavis this is ready for code review

network-api/networkapi/wagtailpages/models.py
```diff
@@ -608,19 +656,29 @@ def generate_entries_set_html(self, request, *args, **kwargs):
         entries = self.get_entries()
         has_next = end < len(entries)
 
+        hide_classifiers = False
```
mmmavis: Wondering why we need to set `hide_classifiers` here. Aren't they already taken care of in `filter_entries_for_tag` / `filter_entries_for_category`?

network-api/networkapi/wagtailpages/models.py
```diff
@@ -560,29 +574,63 @@ def filter_entries_for_tag(self, entries, context):
...
+            # so we have to run through all categories =(
+            if bpc.slug == slug:
+                cat = bpc
+        
+        if cat is not None:
+            # make sure we bypass "x results for Y"
+            context['no_filter_ui'] = True
+
+            # and that we don't show the primary tag/category
+            context['hide_classifiers'] = True
```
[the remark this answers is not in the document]
Pomax: Whatever the card template decides to show.

network-api/networkapi/wagtailpages/models.py
Pomax: No: the template will show the first tag/all tags, and for categories we need to show "nothing". The filter functions only filter which entries to show, they don't do anything with the data for those entries. That's the template's job.

Pomax: specifically, when we get here from `generate_category_entries_set_html`, we get all entries using `entries = self.get_entries()`, which lacks a predefined context variable. Instead, we need to _build_ a context for use with the `render()` call below

network-api/networkapi/wagtailpages/models.py
```diff
@@ -560,29 +574,63 @@ def filter_entries_for_tag(self, entries, context):
...
+        cat = None
+        for bpc in BlogPageCategory.objects.all():
+            # We can't use .filter for @property fields,
+            # so we have to run through all categories =(
+            if bpc.slug == slug:
+                cat = bpc
+        
+        if cat is not None:
+            # make sure we bypass "x results for Y"
+            context['no_filter_ui'] = True
```
gideonthomas: I can't remember specifically, but should we set defaults for this? Does it not crash if you to access `no_filter_ui` in the template and it doesn't actually exist in the context dict? If not, ignore.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -560,29 +574,63 @@ def filter_entries_for_tag(self, entries, context):
...
+
+            # and that we don't show the primary tag/category
+            context['hide_classifiers'] = True
+
+            # explicitly set the index page title and intro
+            context['index_title'] = f'{cat.name} {self.title}'.title()
```
gideonthomas: `.title` isn't the best function because it can't handle apostrophes (see https://docs.python.org/3.7/library/stdtypes.html#str.title). You could use the workaround they suggest, but also confirm whether this is absolutely a requirement to force title case vs. making people set the title in the proper case manually.

network-api/networkapi/wagtailpages/models.py
```diff
@@ -548,9 +549,22 @@ def get_entries(self, context=dict()):
         """
         entries = self.get_all_entries()
         if hasattr(self, 'filtered'):
-            entries = self.filter_entries_for_tag(entries, context)
+            entries = self.filter_entries(entries, context)
         return entries
 
+    def filter_entries(self, entries, context):
+        type = self.filtered.get('type')
```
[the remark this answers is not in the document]
Pomax: ...look up 5 lines, where you will find the `hasattr`? =D

Good point about the reserved word, though. Then again, flake8 doesn't warn about it, so I'm also tempted to say that if this was a PEP policy, it would have flagged that?

network-api/networkapi/wagtailpages/models.py
```diff
@@ -548,9 +549,22 @@ def get_entries(self, context=dict()):
...
+        context['filtered'] = type
+
+        if type == 'tags':
+            entries = self.filter_entries_for_tag(entries, context)
+
+        if type == 'category':
+            entries = self.filter_entries_for_category(entries, context)
+
+        context['total_entries'] = len(entries)
```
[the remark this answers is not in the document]
Pomax: It cannot. It can be an empty queryset, but not None

network-api/networkapi/wagtailpages/models.py
Pomax: no: as per Django convention, non-existent variables are treated as false

network-api/networkapi/wagtailpages/models.py
```diff
@@ -560,29 +574,63 @@ def filter_entries_for_tag(self, entries, context):
         specific model before we can test for whether i) there are tags
         to work with and then ii) those tags match the specified ones.
         """
-        type = self.filtered.get('type')
-        context['filtered'] = type
-
-        if type == 'tags':
-            terms = self.filtered.get('terms')
+        terms = self.filtered.get('terms')
```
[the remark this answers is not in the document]
Pomax: The only thing I can think of to do if that is the case is to throw an exception, which is already the case, so I'd rather keep it crashy if called incorrectly?

network-api/networkapi/wagtailpages/models.py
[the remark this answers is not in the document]
gideonthomas: if `filter_entries` is only being called within this class and we know that the caller is always going to do the `hasattr` check, yep, this should be fine!

Issue #3609, opened 2019-09-05
`inv docker-setup` fails on a truly clean install 


Pomax:
I just got a new laptop, and after installing git, node, python3, invoke, and docker, the `inv docker-setup` command failed when it reached the database delete step, because there was nothing for it to delete:
```
MacBook-Pro:foundation.mozilla.org pomax$ invoke docker-setup
Setting default environment variables.
Keeping your existing .docker.env.
Building Docker images
postgres uses an image, skipping
Building watch-static-files
Step 1/4 : FROM node:8-stretch-slim
 ---> 80502021eebb
Step 2/4 : WORKDIR /app/
 ---> Using cache
 ---> 901741bf257a
Step 3/4 : COPY package.json package-lock.json ./
 ---> Using cache
 ---> 9d15d8525be1
Step 4/4 : RUN npm install
 ---> Using cache
 ---> 72865837c21d
Successfully built 72865837c21d
Successfully tagged foundationmozillaorg_watch-static-files:latest
Building backend
Step 1/7 : FROM python:3.7-slim
 ---> f96c28b7013f
Step 2/7 : ENV PYTHONUNBUFFERED 1
 ---> Using cache
 ---> a562193a96e4
Step 3/7 : ENV PYTHONDONTWRITEBYTECODE 1
 ---> Using cache
 ---> 8021d4a1eef0
Step 4/7 : WORKDIR /app/
 ---> Using cache
 ---> 9a9539222874
Step 5/7 : RUN pip install pipenv
 ---> Using cache
 ---> 1ea7a7559a05
Step 6/7 : COPY Pipfile Pipfile.lock ./
 ---> Using cache
 ---> 83f433045054
Step 7/7 : RUN pipenv install -d
 ---> Using cache
 ---> 0aee5d20f8a3
Successfully built 0aee5d20f8a3
Successfully tagged foundationmozillaorg_backend:latest
Setting up a clean database.
Stopping services first
Removing network foundationmozillaorg_default
Network foundationmozillaorg_default not found.
Deleting database
Error: No such volume: foundationmozillaorg_postgres_data
```
The solution is probably to try/except the delete step, so that the rest of the setup tasks continue to run even if there was no postgres volume to delete

Pomax:
superceded by https://github.com/mozilla/foundation.mozilla.org/pull/3673

PR #3610, opened 2019-09-05
fix delete-db not working on a truly clean OS

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3609

@patjouk I wasn't sure if there was a way to properly check for the volume so instead of an `if` I ended up using a `try` with a passing except... would you happen to know if this can be done cleaner?

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -203,12 +203,6 @@ def docker_test_node(ctx):
     ctx.run("docker-compose run --rm watch-static-files npm run test", **PLATFORM_ARG)
 
 
-@task
-def docker_test(ctx):
-    docker_test_python(ctx)
-    docker_test_node(ctx)
-
-
 def docker_shared_database_operations(ctx):
     print("Applying database migrations.")
     docker_migrate(ctx)
--- a/tasks.py
+++ b/tasks.py
@@ -203,6 +203,12 @@ def docker_test_node(ctx):
     ctx.run("docker-compose run --rm watch-static-files npm run test", **PLATFORM_ARG)
 
 
+@task
+def docker_test(ctx):
+    docker_test_python(ctx)
+    docker_test_node(ctx)
+
+
 def docker_shared_database_operations(ctx):
     print("Applying database migrations.")
     docker_migrate(ctx)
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -223,22 +215,31 @@ def docker_switching_branch(ctx):
...
-    print("Creating fake data")
-    docker_manage(ctx, "load_fake_data")
-    print("Updating localizable fields")
-    docker_l10n_sync(ctx)
-    docker_l10n_update(ctx)
-    print("Updating block information")
-    docker_manage(ctx, "block_inventory")
+    try:
+        ctx.run("docker volume rm foundationmozillaorg_postgres_data")
+    except:
```
Pomax: flake8 doesn't allow a bare except, but I don't know how to find out what the actual error type is: the log doesn't help in this case.

tasks.py
```diff
@@ -223,22 +215,30 @@ def docker_switching_branch(ctx):
...
-    print("Creating fake data")
-    docker_manage(ctx, "load_fake_data")
-    print("Updating localizable fields")
-    docker_l10n_sync(ctx)
-    docker_l10n_update(ctx)
-    print("Updating block information")
-    docker_manage(ctx, "block_inventory")
+    try:
+        ctx.run("docker volume rm foundationmozillaorg_postgres_data")
+    except:  # noqa: E722
```
Pomax: I don't know how to get the "real" exception so that we're not doing a bare except - the log doesn't tell us in this case.

Pomax:
@patjouk does this look sensible?

patjouk:
I took a look and can't find a quick solution. Can you add a code comment on why you're adding this change and a print so that the user know it's an expected error?

Pomax:
Good call - I've added both the code comment and print.

Pomax:
@patjouk is this good, or were there other things to address too?

patjouk:
There's one thing I would like to try tomorrow, if that's okay with you.

tasks.py
cadecairos: Read through the invoke source, I think it raises one of these exceptions: https://github.com/pyinvoke/invoke/blob/aa5a02bfd7e6a73ceed11f0a94cf30d4a422b8aa/invoke/runners.py#L276-L286

patjouk:
I think this can be closed because of #3673.

PR #3613, opened 2019-09-09
Add language switcher

### LINKED ISSUE #3580: Implement language switcher
Results of #2681

mmmavis:
Closes #3580 

@sabrinang I made some design adjustments to make the UI looks more balanced. 
For example,
- Changing the inner padding of the dropdown to have equal left/right padding (12px) instead of 20px/11px noted in Sketch
- Changing the alignment of the dropdown so it's closer to the Language label. Reason: If we want the text in the dropdown to align with footer links above, in most cases there will be a [big gap](https://user-images.githubusercontent.com/2896608/64649705-483c1200-d3d2-11e9-913f-6ac83a0fb594.png) between "Language" and the dropdown... which I think is probably not ideal.

I can explain the adjustments I made more clearly later today when we are both in the office 😉 .

### Main site

1. Go to https://foundation-mofostaging-pr-3613.herokuapp.com
2. Look for language dropdown near page bottom
3. Select a different language
4. See page URL is updated to have the current locale code. e.g., https://foundation-mofostaging-pr-3613.herokuapp.com/fr/initiatives/
5. See if page gets rendered in the selected language. On the CMS I have added [fake "localized" title](https://user-images.githubusercontent.com/2896608/64648366-9a2f6880-d3cf-11e9-8b6e-49c34811cd7c.png) for the About page. For example, if the page is set to render in French, on the nav bar you should see `About[FR]`.
image

### MozFest

1. go to https://foundation-mofostaging-pr-3613.herokuapp.com/en/mozilla-festival/
2. follow steps 2-5 from the above section


network-api/networkapi/templates/partials/footer.html
```diff
@@ -55,6 +57,26 @@ <h5 class="h5-heading">More about us</h5>
             </ul>
           </div>
         </div>
+        <div class="row mt-4 mb-4 mb-md-0">
+          <div class="col-12">
+            <form action="{% url 'set_language' %}" method="post">{% csrf_token %}
```
mmmavis: implementation based on https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#the-set-language-redirect-view

mmmavis: source code for `def set_language`: https://docs.djangoproject.com/en/2.2/_modules/django/views/i18n/

mmmavis:
@sabrinang this is ready for design review again!

mmmavis:
@Pomax this is ready for 👀

mmmavis:
Hey @TheoChevalier , thanks for testing this PR out. We just landed a PR to remove hardcoded `/en` from links on the site. So the issues you pointed out should be gone now. Do you mind giving it a try again?
---

Pomax:
@TheoChevalier note that privacynotincluded is actually an independent app that we host on the foundation domain - it is not localized (and is not wrapped through the Django i18n patterns approach)

TheoChevalier:
> @TheoChevalier note that privacynotincluded is actually an independent app that we host on the foundation domain - it is not localized (and is not wrapped through the Django i18n patterns approach). Any locale should get forced back to `/en/` for it.

Right — so I guess we will want to add a conditional here to exclude it from the footer? https://foundation-mofostaging-pr-3613.herokuapp.com/en/privacynotincluded/

network-api/networkapi/templates/partials/footer.html
```diff
@@ -45,19 +45,28 @@ <h5 class="h5-heading">More about us</h5>
...
+              <li><a href="https://www.mozilla.org/about/legal/terms/mozilla/" class="dark-theme">Legal</a></li>
+              <li><a href="https://creativecommons.org/licenses/by/4.0" class="dark-theme">License</a></li>
+              <li><a href="https://www.mozilla.org/about/governance/policies/participation/" class="dark-theme">Participation Guidelines</a></li>
+              <li><a href="https://mozilla.org/privacy/websites/" class="dark-theme">Privacy</a></li>
+              <li><a href="https://www.mozilla.org/privacy/websites/#cookies" class="dark-theme">Cookies</a></li>
               {% endblock %}
             </ul>
           </div>
         </div>
+        {% if exclude_language_switcher != True %}
```
Pomax: We should set this to False in the buyersguide `bg_base.html`, so that it does not include the switched (since PNI is english-only and we rigorously redirect for it no matter which locale you try to use)

Pomax: We should set this to False in the buyersguide bg_base.html, so that it does not include the switched (since PNI is english-only and we rigorously redirect for it no matter which locale you try to use)

Issue #3591, opened 2019-08-29
Blog: ability to backdate a blog post


kristinashu:
We will manually move over a few key blog posts from Medium over to our site so we need to add the ability to backdate these posts.

perrywilli:
You are very beautiful

kristinashu:
Kevin just inquired about this feature. Not sure if it's helpful but he pointed out that he is able to backdate News items. 

<img width="1437" alt="image" src="https://user-images.githubusercontent.com/1068783/64276819-ac0b9a00-cefd-11e9-92ae-e37eae45e64a.png">

Pomax:
News items exist "outside" of wagtail pages, so it has a dedicated date field. Normal pages are wagtail managed when it comes to publication date, so we can't change their dates quite as easily. How many posts would this concern?

Also, with an eye on republishing on the internet, a common practice is to byline republications as "This is a republication of "title here" found here [link to original] from [date]" and not backdate the new publication. Is that an option for us?

kristinashu:
Ah ok. I think it would been fine if he migrated older posts over and then started to publish new stuff. But I think it's tricky since we've already started to published stuff so the "older" posts will end up pushing the new stuff down.

But just let me know if you think it will be too tricky to do and I'll let Kevin know.

gideonthomas:
Is this a one time migration? If so, I suggest doing this manually in the db to make it easier. First, create the posts to be migrated in wagtail as new posts and publish them.

Then, in the db, just run queries like `UPDATE table_name SET publish_date_column = '2011-10-10' WHERE id = 123`. Using a database GUI we can simply queue all of these up and run them once.

kristinashu:
Yes, it's just a one time migration!

Pomax:
as a CMS, with a million tables that all interconnect, I am not inclined to believe it'll be quite that simple. I'll ask the wagtail folks if there is a way to backdate, and if not, how to manually achieve backdating (through python, by updating models and saving them, rather than through straight up DB updating, because the model update is likely to make sure everything that's related gets updated too, unlike a DB update)

gideonthomas:
cc @alanmoo

alanmoo:
I think we're going down the wrong path here. In the "settings" tab of posts, we can already set a date in the past or future. Can we update the blog template and index views to show "scheduled publish date, if in the past OR first publish date, if no scheduled date"?

Pomax:
good point, that should be possible

[...]

PR #3634, opened 2019-09-09
add the 'first published at' date as editable on the settings page

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3591 by adding the "first published at" date as an editable field on the settings tab for blog posts.

Test: fire up the /blog page on the review app and notice the ordering. Go into the CMS admin interface, and edit the most recent blog post by going to its settings tab and changing the 'first published at' date to january 1st, then publishing. Reload/revisit the /blog page, and that entry should now be the last one in the list, rather than the first one.


Pomax:
not for field panels, it'd have to be for the field itself, and it's buried deep inside Wagtail's own codebase

PR #3641, opened 2019-09-12
Add blogs to homepage

### LINKED ISSUE #3611: Blog: featured on homepage
We would like to update the homepage News feed to a featured Blog feed.

### Layout
- Keep the same layout as we have now but replace the news source with the blog category. The rest should stay the same.

### CMS
- Staff (usually Jen) should be able to go to the homepage in Wagtail and select 4 blog posts to feature.
- Most recent featured post should be on the left with an image. Oldest featured post should be bottom right.
- Thumbnail content should be pulled from the post and it's OG info.

### Comp
image

mmmavis:
Closes #3611


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
```diff
@@ -0,0 +1,64 @@
...
+{% if page.featured_blogs.all.count > 0 %}
+{% with page.featured_blogs.all|dictsortreversed:"blog.first_published_at" as featured_blogs %}
+<div class="row">
+  <div class="col mb-4">
+    <hr class="mb-4 hr-emphasis mt-0">
+    <h4 class="h4-heading">Latest news</h4>
+  </div>
+</div>
+<div class="row mb-3">
+  {% for blog in featured_blogs %}
```
Pomax: Are we using two lines here as a way to effect an "if there are featured blogs, use the first one"?

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
```diff
@@ -0,0 +1,64 @@
...
+  {% if forloop.counter is 1 %}
+  <div class="col-md-6 mb-0 pb-4">
+    <div>
+      {% image blog.blog.specific.get_meta_image width-540 %}
+    </div>
+    <div class="row">
```
Pomax: How much of this template is shared with the blog card template? Do we want to make those shared code, or keep them separate?

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
Pomax: Are we using two lines here as a way to effect an "if there are featured blogs, use the first one"?

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
```diff
@@ -0,0 +1,64 @@
...
+          <p class="body-small mb-2">{% if blog.blog.author %}by {{blog.blog.author}} on {% endif %} {% if blog.blog.first_published_at %}{{ blog.blog.first_published_at|date:"F j, Y" }}{% else %}not published yet{% endif %}</p>
+          <p>{{blog.blog.search_description}}</p>
+        </div>
+      </div>
+    </div>
+  </div>
+  {% endif %}
+  {% endfor %}
+
+  <div class="col-md-6 mb-0 pb-4">
```
Pomax: Should this col only be built if there's more than 1 featured blog item? E.g.
```
{% if featured_blogs|length > 1 %}
...
{% endif %}
```

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
mmmavis: They _could_ share the same code but it will complicate the logics in `blog-card.html` further as their DOM element types and class names are actually quite different. I'd keep them separate to keep things clean.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
mmmavis: Not sure if I understand your question 😬

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
mmmavis: Yes but I don't know how we can do this in the template. 😬 

We want the 2nd, 3rd, and 4th blog to all be wrapped in one "col" div. e.g.,

```html
<div class="col-md-6 mb-0 pb-4">
  2nd blog
  3rd blog
  4th blog
</div>
```

So if we wrap the "col" div in the following condition block
```
{% if featured_blogs|length > 1 %}
...
{% endif %}
```

then we will get the following HTML back, which isn't what we want

```html
<div class="col-md-6 mb-0 pb-4">
  2nd blog
</div>
<div class="col-md-6 mb-0 pb-4">
  3rd blog
</div>
<div class="col-md-6 mb-0 pb-4">
  4th blog
</div>
```

mmmavis:
@Pomax I left some comments. Please see above.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
Pomax: `{% for blog in featured_blogs %}` + `{% if forloop.counter is 1 %}` is the same as `{% with first = featured_blogs.first %}`, _iff_ there is always a `featured_blog`. If the list of featured blog entries can be empty, then using two separate lines makes more sense.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
Pomax: We can put the `{% if featured_blogs|length > 1 %}` outside of the div, and keep the code you already have, so:
```
  {% if featured_blogs|length > 1 %}
  <div class="col-md-6 mb-0 pb-4">
    {% for blog in featured_blogs %}
    {% if forloop.counter > 1 %}
    <div>
      <p class="h6-heading mb-2">
        {% for category in blog.blog.specific.category.all %}
        <a href="/blog/category/{{category.slug}}">{{ category }}</a>{% if not forloop.last %}, {% endif %}
        {% endfor %}
      </p>

     ...
  </div>
  {% endif %}
```

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
Pomax: That said, keeping this is more robust, so... marking as resolved =P

mmmavis:

PR #3656, opened 2019-09-16
Add 'review_date' to PNI Product model

### LINKED ISSUE #3572: Add "review date" to PNI product model
So that we can filter by products reviewed in different months/years in the guide, we should add a field to the model allowing us to indicate review month and year. I don't have a particular opinion on format (string vs an actual date), so I'll leave it to the implementer to decide.

mmmavis:
Closes #3572


mmmavis:
@Pomax the following approaches don't work 

#### [1] Use `auto_now_add` like what we used in mozilla/network-pulse-api@37b9f0d

This makes the `review_date` a non-editable field which isn't what we want.
Reference: https://docs.djangoproject.com/en/2.2/ref/models/fields/#django.db.models.DateField.auto_now_add

#### [2] Use `date.now`

The migration file alters the field value to `September 17, 2019` and not "today" as a variable.

---

My appraoch in this PR:

1. first migration: update the exisiting products so they all have `review_date` set to November 1, 2018
2. second migration: update the field so it doesn't have a default value

I tested it on CMS and it seems like the date widget jumps to today's date by default so I think UX wise this appraoch isn't that bad. Let me know if you have other suggestions!


review-date-product

network-api/networkapi/buyersguide/factory.py
```diff
@@ -33,6 +34,8 @@ class Meta:
 
     draft = Faker('boolean')
     adult_content = Faker('boolean')
+    review_date = Faker('date_time_between_dates',
+                        datetime_start=date(year=2018, month=11, day=1), datetime_end=None, tzinfo=timezone.utc)
```
mmmavis: Note this is why all the products on review app have random `review_date` values instead of November 1, 2018. Let me know if should update factory as a follow-up PR instead.

mmmavis:
@Pomax There are some errors in Travis. I will assign you for re-review after I fix those issues.

mmmavis:
@Pomax 👀 ?

mmmavis:
@Pomax Percy raised a bunch of flags but I couldn't tell why?

mmmavis:
Talked to Pomax on Slack. Percy flags are expected because  

> by adding a field, the random numbers that are generated no longer apply to the same elements

Issue #3659, opened 2019-09-17
querying blog entries for non-existent tag crashses the system


Pomax:
We're using a list comprehension for blog entry fetching based on ``, which will crash if the tag doesn't exist - since we cannot "ignore" throws, this needs to be rewritten as a standard for loop. Perf wise this should be irrelevant, but execution wise it means we won't be crashing the page process.

PR #3660, opened 2019-09-17
make nonexistent tags not cause a crash

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3659

Also adds a test for nonexistent tags as part of the percy run to ensure we have a screenshot in the future to warn us.

compare:
- https://foundation.mozilla.org/en/blog/tags/mozillawedwdwfwef/ (500 error)
- https://foundation-mofostaging-pr-3660.herokuapp.com/en/blog/tags/mozillawedwdwfwef/ (should be an actual page)


alanmoo:
Can the message "0 results for" on the now-existant page show the term you tried to use?

youriwims:
+1 to prev. comment.

Pomax:
Not without substantially more rewriting, since that string is pulled from the tags (the "real" tags), not from what's in the URL bar.

alanmoo:
What about removing the "for"?

Pomax:
also quite easy, let me add that in.

Issue #3658, opened 2019-09-17
https://foundation.mozilla.org/campaigns/ shouldn’t list private pages


TheoChevalier:
https://foundation.mofostaging.net/campaigns/ currently lists private pages (e.g. the Snapchat one), I think they shouldn’t appear here (or anywhere else). I’ve checked in private mode too, to make sure it was listed even if I’m not signed in.

Pomax:
good point, it looks like we're missing `public()` in addition to `live()`.

PR #3661, opened 2019-09-17
do not present private posts

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3658 by using `.live().public()` rather than merely `.live()`


TheoChevalier:
@Pomax Should it also honor the “Show in menus” boolean? That would allow us to hide the “Thank you for your response” page

Pomax:
No, the menu checkbox is purely for whether things show in menus/nav bars.

Issue #3662, opened 2019-09-17
Recent blog entries component's blog page lookup needs to be case-insensitive


Pomax:
The "most recent blog entries" component is looking for the blogpage based on `title='blog'`, but this should be `title__iexact='blog'` to make sure `blog` vs `Blog` doesn't cause an IndexPage.ObjectNotFound error.

PR #3663, opened 2019-09-17
fix recent blog component

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3662

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -92,7 +92,7 @@ APP_DIR = app()
 SECRET_KEY = env('DJANGO_SECRET_KEY')
 
 # SECURITY WARNING: don't run with debug turned on in production!
-DEBUG = False  # FILEBROWSER_DEBUG = env('DEBUG')
+DEBUG = FILEBROWSER_DEBUG = env('DEBUG')
 
 # Force permanent redirects to the domains specified in TARGET_DOMAINS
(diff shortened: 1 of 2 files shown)


network-api/networkapi/settings.py
```diff
@@ -269,6 +269,7 @@
             ])),
             'libraries': {
                 'bg_nav_tags': 'networkapi.buyersguide.templatetags.bg_nav_tags',
+                'blog_tags': 'networkapi.wagtailpages.templatetags.blog_tags',
```
youriwims: ...will never forget to register tem

network-api/networkapi/wagtailpages/customblocks/recent_blog_entries.py
```diff
@@ -46,7 +46,7 @@ class RecentBlogEntries(blocks.StructBlock):
     def get_context(self, value, parent_context=None):
         context = super().get_context(value, parent_context=parent_context)
         IndexPage = apps.get_model('wagtailpages.IndexPage')
-        blogpage = IndexPage.objects.get(title="blog")
+        blogpage = IndexPage.objects.get(title__iexact="blog")
```
youriwims: ....you learn something new every day.

network-api/networkapi/settings.py
Pomax: we may have to revisit whether this deserves its own file or whether it makes sense to fold it in somewhere else... (it's technically zero code related to blogs)

Pomax:
filed https://github.com/mozilla/foundation.mozilla.org/issues/3664 as a follow-up

PR #3668, opened 2019-09-18
Localize blog post author field 

### LINKED ISSUE #3667: Localize blog post author field
In some of the recent blog posts, I noticed the author field contains some parts that require localization, so after all, we should localize that field as well:

`Ashley Boyd, VP Advocacy at Mozilla`
`Becca Ricks and Kaili Lambe`

I’ve also increased the character limit to 70, as it seemed a little bit tight for some of the longer strings.

TheoChevalier:
Makes the author field localizable (fixes #3667)


Pomax:
I'm not sure I follow - why does someone's name need localization?

TheoChevalier:
> I'm not sure I follow - why does someone's name need localization?

That was my assumption too, until I see actual usage of this field:

Sometimes there are two authors, e.g. `Becca Ricks and Kaili Lambe` (so, need to localize the `and`) or a title is attached, e.g. `Ashley Boyd, VP Advocacy at Mozilla`

Issue #3643, opened 2019-09-12
Localize the Initiatives page


TheoChevalier:
Many fields are currently not localizable in Wagtail for this page, let’s fix that!

youriwims:
@TheoChevalier after speaking with @Pomax , localizing the headers and initiative blocks is the most that can be done right now and we'll need to localize the highlights in a follow-up ticket (ticket will be posted here).

TheoChevalier:
Sounds good @youriwims , thanks! I’ll go ahead and expose the strings in Pontoon, minus highlights. Having the initiative blocks alone should justify deploying translations of the page

youriwims:
Not a problem—sounds great! :)

Pomax:
Closing this issue due to having gone stale - if this is still something we want to do, let's file a new issue with all the information necessary for triaging.

PR #3671, opened 2019-09-18
Localize most fields of the Participate page

### LINKED ISSUE #3642: Localize the Participate page
Let’s make this page localizable!

There’s a start here: https://github.com/TheoChevalier/foundation.mozilla.org/commit/c24f1664cee8a2f95b726a804c668dda1155aa69

But it’s missing Highlights, and CTA group 4 (and I don’t know how to tackle that)

TheoChevalier:
This fixes #3642.

As mentioned in https://github.com/mozilla/foundation.mozilla.org/issues/3643#issuecomment-532411599 localizing the highlights section will require a separate effort, so this PR makes all other fields localizable.


Pomax:
Closing this for now, as recreating it will be a matter of adding the fields to translate and redoing the makemigrations, but right now this would make it _much_ harder to fix up the Participate page to what we want it to be.

PR #3673, opened 2019-09-19
Simplify docker invoke tasks

patjouk:
As pointed by Pomax in #3610 and #3536, if you run `inv docker-setup` multiple times, you will run into various unexpected issues. This is because the previous volumes are still there and need to be deleted before we can run the setup steps again.
To do that, I added a `docker-compose down --volumes` to delete all volumes (DB + nodes modules) and do a `docker-compose build --no-cache` to force rebuilding images in the setup step.

I also realized that switching branch and setting things up for the first time have a lot of similarities and constrains: I removed the `docker-switch-branch` command and renamed the `docker-first-setup` to `docker-new-env`. It should be more robust and less confusing that way.

The double env vars were another source of confusion: since the issue between pipenv and docker-compose around quotes is fixed, I reverted to having one `.env` file only. I'll make sure everyone knows on Slack but *you will need to manually update your .env or run the `inv docker-new-env`*: your `.env` won't be overridden, only the `DATABASE_URL` and `ALLOWED_HOSTS` will be modified.

Last but not least, I added a config file for invoke with `echo` at true. I think it will help people understand more how docker and pipenv work if I don't hide it in the back. Every command line run by invoke will know echo by default.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -19,10 +19,10 @@ else:
 
 
 def create_docker_env_file(env_file):
-    """Create a docker .env from an existing env file"""
+    """Create or update an .env to work with a docker environment"""
     with open(env_file, 'r') as f:
         env_vars = f.read()
-    # remove the double quote for docker-compose and python_dotenv compatibility reasons
+    # We need to strip the quotes because Docker-compose considers them as part of the env value.
     env_vars = env_vars.replace('"', '')
     # update the DATABASE_URL env
     new_db_url = "DATABASE_URL=postgres://postgres@postgres:5432/postgres"
@@ -34,7 +34,7 @@ def create_docker_env_file(env_file):
     env_vars = env_vars.replace(old_hosts.group(0), new_hosts)
 
     # create the new env file
-    with open('.docker.env', 'w') as f:
+    with open('.env', 'w') as f:
         f.write(env_vars)
 
 
@@ -82,9 +82,9 @@ def l10n_update(ctx):
 @task
 def test(ctx):
     """Run tests"""
-    print("Running flake8")
+    print("* Running flake8")
     ctx.run(f"pipenv run flake8 tasks.py network-api", **PLATFORM_ARG)
-    print("Running tests")
+    print("* Running tests")
     manage(ctx, "test")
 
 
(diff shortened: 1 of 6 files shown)


Pomax:
Let me find a machine without docker already set up (or one I can fully remove it from and then reinstall) but looking at the code it seems like the original issue I ran into (docker crashing on the database delete, because it tries to delete a volume that doesn't exist, rather than trying to do something and finding that already existing) is still there, so we probably still need a try/except around that specific operation.

tasks.py
```diff
@@ -190,103 +211,66 @@ def docker_l10n_update(ctx):
...
-def docker_setup(ctx):
-    """Prepare your dev environment after a fresh git clone"""
+def docker_new_env(ctx):
+    """Get a new dev environment and a new database with fake data"""
     with ctx.cd(ROOT):
-        print("Setting default environment variables.")
+        print("* Setting default environment variables")
         if os.path.isfile(".env"):
-            print("Copying your existing .env to .docker.env")
+            print("* Striping quotes and making sure your DATABASE_URL and ALLOWED_HOSTS are properly setup")
```
Pomax:
```suggestion
            print("* Stripping quotes and making sure your DATABASE_URL and ALLOWED_HOSTS are properly setup")
```

patjouk:
Which commands were you running and can you share logs?

tasks.py
```diff
@@ -190,103 +211,66 @@ def docker_l10n_update(ctx):
...
-    docker_l10n_update(ctx)
-    print("Updating block information")
-    docker_manage(ctx, "block_inventory")
+    docker_l10n_block_inventory(ctx)
 
 
 @task
-def docker_setup(ctx):
-    """Prepare your dev environment after a fresh git clone"""
+def docker_new_env(ctx):
```
Pomax: This is probably a crazy question, but: does docker have the ability to set some default named container? Or, is it easy to make a run task change the docker file so that it changes the name/id of the docker container to start?

Specifically, I was wondering whether it would it be possible to have an `inv docker-switch-env [name]` that either loads a named env, or falls through to the `new-env` task in order to build a new environment with its own database if it doesn't exist yet, so that if people are working on several issues at the same time with different migration chains, they can switch environments in a way that keeps the code and db in sync? 

(the previous `switch` task would build a new db by clearing the existing db - I'm wondering if there's a way to get "build new" if there's nothing, but "reuse" if there was something, with `new-env` always clearing if it exists, before building a new env, but `switch-env` only changing the dockerfile so that `docker-compose up` will load the correct container, rather than build anything, if that container exists?)

Pomax:
Looks like that's no longer an issue anymore - I did a full removal of docker (using Nektonic's App Cleaner & Uninstaller), deleted the foundation dir, recloned foundation, reinstalled docker, ran `inv docker-new-env`, and managed to make it to the end without issue, so: hurray!

tasks.py
patjouk: That's not something that I want to investigate. Maybe we could do something using image tags but I don't have the time to look into this.

Issue #3588, opened 2019-08-29
Blog: add RSS


kristinashu:
@Pomax suggested we add an RSS feature to our blog. @xmatthewx liked this because then we could get slack updates.

Pomax:
Possibly good resource: https://chrxr.com/adding-rss-feed-wagtail-site/

Issue #2797, opened 2019-03-12
Blog epic


kristinashu:
### Goal
- Use Foundation site as the primary destination for announcements etc to drive traffic to the site
- A place where we can start to tell deeper/richer stories about fellows, awardees, and the people in our network

### Links
Comps https://redpen.io/p/bha12cf7387de4f7e7
Deck https://docs.google.com/presentation/d/1NWfOotxMRcfE78CehF2uqZefEWEUm5siwrgOXzmbM_8/edit#slide=id.p

### Stage 1: Posting strategy
 - [x] Blog pros and cons [doc](https://docs.google.com/document/d/1oiO0LB_ej4FxhpbCbjdATd2hSjwTimdVF284kVNR5dc/edit#heading=h.py9jj56uzob)
- [x] Decide on name (aka url): `Blog` ~~vs Stories vs Articles~~
- [x] Consult with Kevin, Sydette, Jen, Mark, someone from IH team, someone from MozFest team, moco, and anyone else who currently posts blogs
- [x] Decide how we integrate with existing blogs: will we continue to also post to Medium and MoCo's blog [(doc)](https://docs.google.com/document/d/1FvzDd8zPL_rk3rRMzPaieZJEd7e5YmJrnNt4LtwAeec/edit)
- [x] Once build: onboard comms team and make sure they always link to our blog over Medium and MoCo's
- [ ] Document tutorial and hold a working session to onboard staff

### Stage 2: Blog post detail page 
- [x] Design https://github.com/mozilla/foundation.mozilla.org/issues/2911
- [x] Dev MVP https://github.com/mozilla/foundation.mozilla.org/issues/3029 
- [x] Dev share button https://github.com/mozilla/foundation.mozilla.org/issues/3157
- [x] Dev Related Posts https://github.com/mozilla/foundation.mozilla.org/issues/3162

### Stage 3: Blog overview pages
- [x] Design overview page, if someone clicks on a tag or authors #3165
- [x] Dev overview page for /blog #3370
- [x] Dev tag page for /blog?tag=[name] #3371
- [ ] Add to main nav (only later once we have content)
- [x] Explore adding RSS feed #3588

### Stage 4: Category pages
- [x] Dev: add categories to post page #3560
- [x] Dev: create category landing page #3562
- [x] Dev: update /blog tags to categories #3563 

### Stage 5: Integrate onto site 
- [x] Exploratory comps https://github.com/mozilla/foundation.mozilla.org/issues/3434
[...]

PR #3681, opened 2019-09-20
RSS/Atom for Blog

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3588
Related: https://github.com/mozilla/foundation.mozilla.org/issues/2797

This uses Django's `Feed` concept, with short descriptions coming from each blog's SEO description metadata field.

Feeds (both wrapped by i18npatterns):
- http://foundation-mofostaging-pr-3681.herokuapp.com/blog/rss
- http://foundation-mofostaging-pr-3681.herokuapp.com/blog/atom

(I used http://www.feedbucket.com to verify the feed output)


kristinashu:
I honestly don't use RSS so seeing if @alanmoo could please check?

alanmoo:
I'm not sure what the delay's supposed to be, but I'm not seeing a.) old posts in my feed, beyond 3 and b.) a new post I created.

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -0,0 +1,38 @@
...
+    Blog page RSS feed, using the content:encoded serializer above
+    """
+
+    title = 'Mozilla Foundation Blog'
+    link = 'https://foundation.mozilla.org/blog/'
+    feed_url = 'https://foundation.mozilla.org/blog/rss/'
+    description = 'The Mozilla Foundation Blog'
+
+    def items(self):
+        return BlogPage.objects.live().public()[:settings.FEED_LIMIT]
```
alanmoo: Is this going to catch all blog pages or just children of /blog? Because blog template get used in other areas of the site (like campaigns).

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -0,0 +1,38 @@
...
+
+    def item_title(self, item):
+        return item.title
+
+    def item_link(self, item):
+        return item.full_url
+
+    def item_description(self, item):
+        return item.get_meta_description()
```
alanmoo: If there's no meta description, I think we should return the first...140(?) characters of the content.

network-api/networkapi/wagtailpages/rss.py
Pomax: all BlogPage objects, so I can rewrite this to `reverse('/blog').children().public().live()` instead.

network-api/networkapi/wagtailpages/rss.py
Pomax: yeaaaaaaaah that's not something that's easy to do given that streamfield yields HTML code, and there's no guarantee the first 140 characters are even text, rather than an ``, which is why I went with meta description.

I can probably render out the body, run it through the python html library and then extract a starting snippet, but then we probably want to cache that somehow, so that we're not computing it constantly once we have a reasonable number of RSS readers hooked into this.

Pomax:
I'm making it pull 10 results, so if you're only seeing 3 in your reader, check what feedbucket gives you. We don't have 10 posts yet, so you should just see "however many public-and-live BlogPage entries we have".

Pomax:
@alanmoo updated the feed generator to render the blog's text (1000 chars and then [...]) and only tap into the `/en/blog` child pages.

http://www.feedbucket.com/?src=https%3A%2F%2Ffoundation-mofostaging-pr-3681.herokuapp.com%2Fen%2Fblog%2Frss%2F

Pomax:
taggin @gideonthomas mostly for a Python/Django sanity check

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -0,0 +1,48 @@
...
+
+    def item_link(self, item):
+        return item.full_url
+
+    def item_description(self, item):
+        page = item.specific
+        html = str(page.body)
+        parsed = BeautifulSoup(html, 'html.parser')
+        text = parsed.get_text()[:1000]
+        if len(text) >= 1000:
```
gideonthomas: not sure I understand this -- on the previous line, you're cutting it off at 1000, so wouldn't just `if len(text) == 1000` suffice?

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -0,0 +1,48 @@
...
+from .models import IndexPage
+
+
+class RSSFeed(Feed):
+    """
+    Blog page RSS feed, using the content:encoded serializer above
+    """
+
+    title = 'Mozilla Foundation Blog'
+    link = 'https://foundation.mozilla.org/blog/'
```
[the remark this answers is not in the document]
Pomax: because the Atom feed will turn that into `example.com/blog` for reasons that no one seems to be able to explain.

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -0,0 +1,48 @@
...
+    feed_url = 'https://foundation.mozilla.org/blog/rss/'
+    description = 'The Mozilla Foundation Blog'
+
+    def items(self):
+        blog_index = IndexPage.objects.get(title__iexact='blog')
+        blog_pages = blog_index.get_children().live().public()
+        return blog_pages[:settings.FEED_LIMIT]
```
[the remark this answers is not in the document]
Pomax: Very good point!

network-api/networkapi/wagtailpages/rss.py
Pomax:
```suggestion
        if len(text) == 1000:
```

network-api/networkapi/wagtailpages/rss.py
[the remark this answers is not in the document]
gideonthomas: It uses whatever is set in "Sites" as the domain for the site id. So if you have that set to `http://example.com`, it will use that as a domain.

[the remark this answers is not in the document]
Pomax: Right, for Django. But Wagtail, which often does the same thing, in this case doesn't: rather than `SITES`, Wagtail handles site/domain management in the database. So I'd rather keep things hardcoded in the rss file to make it immediately obvious what needs to be changed if we do need to change something, rather than adding a variable that effectively "bypasses" the Wagtail way of doing things, and will most likely not be found as being the reason for things not working when a future dev does some googling to figure out what to change.

PR #3696, opened 2019-09-23
remove donate slideup banner

### LINKED ISSUE #3682: Remove the timeout-based donate banner
We've gotten a fair bit of backlash against it on popular news sites, and it interferes with local dev work - Can we remove this without significantly impacting donations? (given that it seems to be burning goodwill... =S)

/cc @alanmoo @xmatthewx

Pomax:
Closes #3682


Pomax:
oh man, good catches! updated.

PR #3691, opened 2019-09-23
Add dropdown-fields to styleguide

youriwims:
Closes #3645 
Related Ticket #3628

This pr:

- Adds dropdowns to styleduide
- Overrides `select.form-control` styling [code]
- Overrides `select.form-control.form-control-lg` styling [code]
- Applies `.form-control` to language switcher 
- Removes `#lanugage-switcher` specific code

**Styleguide**: https://foundation-mofostaging-pr-3691.herokuapp.com/en/styleguide/

PR #3700, opened 2019-09-24
Localize signups + petitions

### LINKED ISSUE #3645: Add dropdown styles to style guide
Since we're using the dropdown in the language switcher (#2681, #3613) and now in localized petitions and signups (#3628), this would be a great time to unify styles and add to the style guide :)

cc @kristinashu @mmmavis @sabrinang

youriwims:
Closes #3645


network-api/networkapi/campaign/views.py
```diff
@@ -125,6 +125,7 @@ def petition_submission(request, petition):
         "email": request.data['email'],
         "email_subscription": request.data['newsletterSignup'],
         "source_url": request.data['source'],
+        "page_locale": request.data['locale'],
     }
 
     if petition.requires_country_code:
```
youriwims: @Pomax, does this send the selected country to the backend as long as the petition creator marks `requires_country_code` in `Petition` class `true`?

network-api/networkapi/campaign/views.py
```diff
@@ -125,6 +125,7 @@ def petition_submission(request, petition):
         "email": request.data['email'],
         "email_subscription": request.data['newsletterSignup'],
         "source_url": request.data['source'],
+        "page_locale": request.data['locale'],
```
youriwims: Is this the proper way to send locale data to backend? 😬 @Pomax

Pomax: According to https://basket.readthedocs.io/newsletter_api.html?highlight=country#news-subscribe this should be `lang`, not `page_locale`

network-api/networkapi/campaign/views.py
Pomax: Yes, if the petition is set to require a country code, then this block of code makes sure to either get the value from the form passed on, or generate an error about the face that this required field is missing.

network-api/networkapi/campaign/views.py
pmac: It doesn't look like basket is expecting a language at all for now 

https://github.com/mozmeao/basket/blob/23c8b7ce3ded8b82107b985d78d8827f2e9060b8/basket/news/tasks.py#L1057-L1112

pmac: You can see on line 1086 that we hard-code `en-US` if they've asked for a newsletter subscription

pmac: so `lang` would be fine, but I can add support for `page_locale` just as easily

pmac: Just file a basket issue for that when you're ready with all the necessary changes.

Pomax: thanks!

youriwims:
@kristinashu, 2 things:

1. The dropdown fields aren't styled in this pr, that's happening in #3691. Once everything is approved here, I can merge #3691 and request 'final' design review on this to make sure everything looks as it should and merge this right after. Does that sound good?
2. The petition dropdown said 'Your Country', so I made the other country dropdown match that. Did you want to update them all to just 'Country' instead? Let me know :)

kristinashu:
### 1. 
I know they aren't 'styled' but the country drop down seems different in the two and slightly cut off at narrow width. And the Language picker should be full width like the Country picker.

image

image

### 2.
I like that `Country` is shorter but I think we should stick with `Your country` for clarity and in case there were reasons for doing that in the first place.

### 3.
By spacing do you mean the spacing between each field/dropdown? If so I think you can just use the same spacing as the fields https://foundation.mozilla.org/en/fellowships/newsletter/
image

youriwims:
Sorry for the delay, @kristinashu this is ready for re-review.

youriwims:
@TheoChevalier can you provide translations for the sign up `Sign Up` button, please?

source/js/components/join/join.jsx
```diff
@@ -2,7 +2,9 @@ import React from "react";
 import ReactGA from "react-ga";
 import ReactDOM from "react-dom";
 import classNames from "classnames";
-import basketSignup from "../../basket-signup.js";
+import CountrySelect from "../petition/country-select.jsx";
+import get from "../petition/locales";
```
Pomax: This is too generic a name, let's change this to something that makes its role clearer in the code:
```suggestion
import getText from "../petition/locales";
```
(we can do this because `petition/locales.js` exports `get` as a default export, which means importing code is free to call it whatever it wants. It simply "gets something" as import result, and gets to store that by whatever name makes sense for the importer)

source/js/components/join/join.jsx
```diff
@@ -265,7 +283,7 @@ export default class JoinUs extends React.Component {
           <input
             type="email"
             className="form-control"
-            placeholder="Enter email address"
+            placeholder={get(`Please enter your email`)}
```
Pomax:
```suggestion
            placeholder={getText(`Please enter your email`)}
```

source/js/components/join/join.jsx
```diff
@@ -289,6 +307,29 @@ export default class JoinUs extends React.Component {
...
+    let header = this.props.formPosition == `header`;
+    let footer = this.props.formPosition == `footer`;
+    let classes = classNames(`form-l10n`, {
+      "d-none": footer || header
+    });
+
+    return (
+      <div className={classes}>
+        <div className="mb-2">
+          <CountrySelect label={get(`Your country`)} className="w-100" />
```
Pomax:
```suggestion
          <CountrySelect label={getText(`Your country`)} className="w-100" />
```

source/js/components/join/join.jsx
```diff
@@ -341,10 +382,9 @@ export default class JoinUs extends React.Component {
             />
             <label className="form-check-label d-flex align-items-start">
               <p className="d-inline-block body-small my-0 mr-1 mr-sm-5 mr-md-2 mr-lg-1">
-                I'm okay with Mozilla handling my info as explained in this{" "}
-                <a href="https://www.mozilla.org/privacy/websites/">
-                  Privacy Notice
-                </a>
+                {get(
```
Pomax:
```suggestion
                {getText(
```

source/js/components/petition/petition.jsx
```diff
@@ -152,6 +151,9 @@ export default class Petition extends React.Component {
       let postalCode = this.postalCode && this.postalCode.element.value;
       let comment = this.comment && this.comment.element.value;
       let newsletterSignup = false;
+      let locale = document
```
Pomax: This code was introduced several times, so instead, let's change `petition/locales.js` so that it exposes something like:
```
export {
  getCurrentLanguage, // this is currently getCurrentLocale, but we should rename this function
  getText: get
};
```
and then make getCurrentLocale tap into the meta tag if it exist with its current URL code as fallback code instead.

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
...
+          {languages[lang_code]}
+        </option>
+      );
+    });
+
+    return options;
+  }
+
+  render() {
+    let meta_lang = document
```
Pomax: we can turn this into a function call instead; see comment below

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
+import React from "react";
+
+export default class LanguageSelect extends React.Component {
+  constructor(props) {
+    super(props);
+    this.state = { value: "" };
+
+    this.handleChange = this.handleChange.bind(this);
```
Pomax: please don't do this. The constructor should not be responsible for reshaping the instance bindings.

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
...
+    let meta_lang = document
+      .querySelector(`meta[property="wagtail:language"]`)
+      .getAttribute("content");
+    let value = this.state.value;
+    let classes = this.props.className;
+
+    return (
+      <select
+        value={value ? value : meta_lang}
+        onChange={this.handleChange}
```
Pomax:
```suggestion
        onChange={evt => this.handleChange(evt)}
```

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
...
+  render() {
+    let meta_lang = document
+      .querySelector(`meta[property="wagtail:language"]`)
+      .getAttribute("content");
+    let value = this.state.value;
+    let classes = this.props.className;
+
+    return (
+      <select
+        value={value ? value : meta_lang}
```
Pomax:
```suggestion
        value={this.state.value || meta_lang}
```

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
...
+    });
+
+    return options;
+  }
+
+  render() {
+    let meta_lang = document
+      .querySelector(`meta[property="wagtail:language"]`)
+      .getAttribute("content");
+    let value = this.state.value;
```
Pomax: we don't need this variable.

source/js/components/join/language-select.jsx
```diff
@@ -0,0 +1,55 @@
...
+  constructor(props) {
+    super(props);
+    this.state = { value: "" };
+
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  handleChange(event) {
+    this.setState({ value: event.target.value });
+    console.log(`After: ${this.state.value}`);
```
Pomax: rogue console.log?

source/js/components/join/join.jsx
```diff
@@ -289,6 +307,29 @@ export default class JoinUs extends React.Component {
     );
   }
 
+  /**
+   * Render localization fields
+   */
+
+  renderLocalizationFields() {
+    let header = this.props.formPosition == `header`;
```
Pomax: `==` is a coercive compare, which I'm pretty sure we don't want?
```suggestion
    let header = this.props.formPosition === `header`;
```

source/js/components/join/join.jsx
```diff
@@ -289,6 +307,29 @@ export default class JoinUs extends React.Component {
...
+    let footer = this.props.formPosition == `footer`;
```
Pomax:
```suggestion
    let footer = this.props.formPosition === `footer`;
```

source/js/components/join/join.jsx
```diff
@@ -179,14 +181,30 @@ export default class JoinUs extends React.Component {
...
       action: `form focus`,
       label: `Signup form input focused`
     });
+
+    let emailFields = document.querySelectorAll(
+      '.fields-wrapper .form-control[type="email"]'
+    );
+
+    emailFields.forEach(emailField => {
+      let currentEmailField = emailField == event.target;
```
Pomax:
```suggestion
      let currentEmailField = event.target === emailField ;
```

network-api/networkapi/campaign/tests.py
```diff
@@ -25,6 +25,7 @@ def test_petition_post_route(self):
             'newsletterSignup': True,
             'comment': 'Test comment',
             'source': 'http://localhost/testing/petition',
+            'locale': 'en'
```
Pomax: We probably want to call this `lang` as well, to stay consistent with what we communicate to.

network-api/networkapi/campaign/views.py
```diff
@@ -125,6 +125,7 @@ def petition_submission(request, petition):
         "email": request.data['email'],
         "email_subscription": request.data['newsletterSignup'],
         "source_url": request.data['source'],
+        "lang": request.data['locale'],
```
Pomax:
```suggestion
        "lang": request.data['lang'],
```

source/js/components/petition/petition.jsx
Pomax: Let's also make sure to call this `lang`, since we're accessing the "wagtail:language" meta element

source/js/components/petition/petition.jsx
```diff
@@ -187,6 +189,7 @@ export default class Petition extends React.Component {
         checkbox2: this.props.checkbox2 ? !!this.refs.checkbox2.checked : null,
         newsletterSignup,
         country,
+        locale,
```
Pomax:
```suggestion
        lang,
```

source/js/components/join/language-select.jsx
youriwims: lol yes...sorry, thanks for pointing it out.

source/js/components/join/join.jsx
youriwims: > we can do this because petition/locales.js exports get as a default export, which means importing code is free to call it whatever it wants.

Oooo that's really good to know.

Pomax:
@kristinashu, does this something like "I'd like to receive the newsletter in" above the language picker? As a drop-down without a "placeholder" text the following is not super clear on what the selector is for:

Screen Shot 2019-10-03 at 3 29 48 PM

Pomax:
Also a question about the checkboxes: we're not asking people to fill in the same email address twice, and check a checkbox for "Receiving email updates" as well as a separate signup for "news and updates". Can we streamline that somehow?

kristinashu:
> @kristinashu, does this something like "I'd like to receive the newsletter in" above the language picker? As a drop-down without a "placeholder" text the following is not super clear on what the selector is for:

I think people will understand what the language is referring to. I checked mozilla.org and they seem to do the same so I think it's ok how it is.
image

> Also a question about the checkboxes: we're not asking people to fill in the same email address twice, and check a checkbox for "Receiving email updates" as well as a separate signup for "news and updates", with the same "I'm okay with [...]" checkbox on both. Can we streamline that somehow? This feels like it really wants to be a single form, instead.

The two signup snippets would never actually show up on the same page at the same page. I think Youri just did that for testing but staff wouldn't do that. We would only ever show either the petition snippet or the newsletter snippet, not both at the same time.

Pomax:
If showing both on the same page is purely for testing, then none of my concerns about email addresses and checkboxes applies!

Looking at the mozilla.org version, I guess using the same presentation makes sense, although from an accessibility standpoint, it'd be much nicer to have that drop-down explained (and I'm a bit surprised the moco version doesn't)

youriwims:
Yes, sorry for the confusion @Pomax, it was purely for testing. @TheoChevalier I'll go ahead and add those translations and fix the clickable area—thanks for pointing that out!

PR #3717, opened 2019-10-03
load more regrets until we're just overflowing

### LINKED ISSUE #3443: Add "load more" JS functionality to the listing page

Pomax:
Closes #3443

This uses inline code, which seemed much more sensible than loading this into main.js for a one-off page.


network-api/networkapi/wagtailpages/templates/wagtailpages/pages/load-more-regrets.html
```diff
@@ -0,0 +1,40 @@
+<div class="text-center mb-5 d-none">
```
Pomax: note: button is disabled unless JS is allowed to run.

network-api/networkapi/wagtailpages/templates/wagtailpages/pages/youtube_regrets_page.html
```diff
@@ -29,10 +29,18 @@ <h1>{{ page.specific.headline }}</h1>
   </div>
 
   <div class="row">
-    <div class="col-md-8 offset-md-2">
+    <style>
+      .regret-stories.show-1 .youtube-regret:nth-of-type(n+{% if page.specific.batch_size %}{{ page.specific.batch_size|add:"1" }}{% else %}11{% endif %}) {
+        display: none;
```
Pomax: note: this ensures that even if there's more than 10 regrets, only 10 are visible on initial page load so that the JS "load more" does not turn a page with 30 results into a page with 10 results, which would be super janky

PR #3754, opened 2019-10-08
YouTube Regrets - newsletter button and style tweaks

### LINKED ISSUE #3726: YouTube: add newsletter signup
From @mmmavis 
> I'm having a hard time positioning the newsletter button correctly in intro since there are too many moving elements in that intro. Can we make the button show:
> 1. not immediately below each intro text
> 2. or only show below the first intro text
> 3. or only show below the last intro text
> 4. or position the button elsewhere

What if it was fixed at this point? Enough room for 3 lines (sometimes there will be a gap but I think that's ok).

image

### LINKED ISSUE #3727: YouTube: small style fixes
### Main styling fixes:
- [ ] Story title should be 38px instead of 42px
- [x] End the background gradient higher up on the page (so only the first story has some pink background but the rest of the stories are on white background so they're easier to read)
- [ ] Increase spacing between stories to about 80px (see screen cap below)
- [ ] Make the spacing above and below the story title the same (maybe about 24px) (see screen cap below)
- [ ] Decrease the amount of opacity even further for the images that haven't "passed" the text yet. Sometimes the overlap makes it hard to read (see screen cap below).
- [x] Center the scroll icon (see screen cap below)

Screen cap from comp:
Screen Shot 2019-10-04 at 3 37 39 PM

Screen cap from comp:
Screen Shot 2019-10-04 at 3 39 15 PM

Screen cap from prod:
image

Screen cap from prod:
Screen Shot 2019-10-07 at 4 52 42 PM

### Nice-to-have circle updates:
- Would it be possible to increase the size of the photos? Currently they feel a bit small and makes the "tunnel" feel a little sparse
- Concentric circle 1: Can the size of the first circle be smaller, perhaps half the size it is now? 
- Concentric circle 2: The spacing between the concentric circles seem to be different for each circle. Could we make it the same and increase the spacing? I would guess perhaps 2-3x the spacing we have now. What https://www.onfaitquoipourleclimat.be/ has for its circles is a good benchmark?
- Would it also be possible to blur both sides of the circles and increase the blur? Currently it looks like only one side is blurred?

mmmavis:
Closes #3726 
Closes #3727

https://foundation-mofostaging-pr-3754.herokuapp.com/en/campaigns/youtube-regrets/

Added newsletter button (desktop only), MVP tweaks, and some nice-to-have changes.


kristinashu:
So this included:
- adding Newsletter to desktop 
- centering the scroll icon

If so those two things look good!

One nice-to-have if it's easy to do would be if the clicking the Newsletter signup button in the main nav could close the signup form.

mmmavis:
@kristinashu  👀 ?

kristinashu:
I've opened a follow up ticket - YouTube: add newsletter signup to mobile #3756

Pomax:
on desktop it looks like there's a very large "empty gap" between the end of the "tunnel" and the first actual content item. Is that because this is on staging, or would that still show up on staging/prod?

mmmavis:
> on desktop it looks like there's a very large "empty gap" between the end of the "tunnel" and the first actual content item. Is that because this is on staging, or would that still show up on staging/prod?

I should be able to tweak that by changing the object moving speed with some factor.

> On a perhaps untested desing point, computer with mouse-wheel based scrolling, rather than touchpad/magic mouse scrolling, the "join mozlla" button is almost always darker than the text above it, and so it feels like a constantly out-of-place element. Does it make sense to separate the button from the content a bit more? (e.g. moving it down to 3/4 view height rather than keeping it close to the text?)

Will need design feedback on this. Note that we'll need to leave sufficient room for the long text to show up properly ( e.g.,`that sent them down a rabbit hole they never meant to go down and didn’t know how to stop`).

> (On that note, does it make sense to make the button fade out with the last piece of text, too?)
I went with the simplest solution to get stuff landed first. Can this be a follow-up? (Need design to chime in)

PR #3764, opened 2019-10-10
Docker Compose doesn't use Postgres anymore

cadecairos:
Ensure the docker-compose stack makes use of the Postgres container by default


docker-compose.yml
```diff
@@ -19,7 +19,7 @@ services:
   postgres:
     image: postgres:9.6
     ports:
-      - "5432"
+      - "5432:5432"
```
cadecairos: expose the DB to the host, makes it easier to connect with external tools.

env.default
```diff
@@ -15,7 +15,7 @@ ASSET_DOMAIN=network.mofoprod.net
 CONTENT_TYPE_NO_SNIFF=True
 CORS_WHITELIST=*
 CORS_REGEX_WHITELIST=
-DATABASE_URL=
+DATABASE_URL=postgres://postgres@postgres:5432/postgres
```
cadecairos: I tried using environment variables in the docker-compose.yml file but had no luck there.

docker-compose.yml
patjouk: I would go for that change. @Pomax is it a problem?

env.default
cadecairos: What I mean is that in docker-compose I tried setting `POSTGRES_PASSWORD`, `POSTGRES_USER`, and `POSTGRES_DB` to override the default user/pass/db combo on container creation but it never seemed to work right.

cadecairos:
Looks like I skipped running the initial `inv docker-new-env` step and manually copied the default env file before running `inv docker-catchup`. This PR is not needed.

PR #3766, opened 2019-10-11
Fix newsletter forms

### LINKED ISSUE #3628: Localize newsletter forms
Ensure that newsletter signups are in the correct locale and country.

Task: Replicate a similar form & features from https://www.mozilla.org/fr/newsletter/mozilla/
(Critical) Select country + language in dropdown lists
(Good to have) Select default language in the list based on the page locale
As things stand, people subscribing from a localized page are being subscribed to the EN list.
Not just from the footer form, but also when checking “Yes, I want to receive email updates about Mozilla’s campaigns.” from localized petitions.

@alanmoo to add more info

youriwims:
Closes #3628

Fix that sends snippet `country` and `lang` to basket.

### DIFF
--- a/network-api/networkapi/campaign/views.py
+++ b/network-api/networkapi/campaign/views.py
@@ -86,7 +86,8 @@ def petition_submission_view(request, pk):
 def signup_submission(request, signup):
     data = {
         "email": request.data['email'],
-        "lang": "en",
+        "lang": request.data['lang'],
+        "country": request.data['country'],
         "format": "html",
         "source_url": request.data['source'],
         "newsletters": signup.newsletter,
@@ -126,12 +127,13 @@ def petition_submission(request, petition):
         "email_subscription": request.data['newsletterSignup'],
         "source_url": request.data['source'],
         "lang": request.data['lang'],
+        "country": request.data['country'],
     }
 
-    if petition.requires_country_code:
+    if petition:
         if 'country' in request.data:
             data["country"] = request.data['country']
-        else:
+        if petition.requires_country_code and 'country' not in request.data:
             return Response(
                 {'error': 'Required field "country" is missing'},
                 status=status.HTTP_400_BAD_REQUEST,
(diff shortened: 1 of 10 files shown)


network-api/networkapi/campaign/views.py
```diff
@@ -125,13 +126,13 @@ def petition_submission(request, petition):
         "email": request.data['email'],
         "email_subscription": request.data['newsletterSignup'],
         "source_url": request.data['source'],
-        "lang": request.data['lang'],
+        "lang": request.data['lang']
```
Pomax: this is python. Ending the last element on a comma is actually a good thing.

source/js/components/join/join.jsx
```diff
@@ -311,6 +317,14 @@ export default class JoinUs extends React.Component {
    * Render localization fields
    */
 
+  setCountry(country) {
+    this.setState({ country: country });
```
Pomax: not technically wrong or right, but `{ country: country }` and `{ country }` are equivalent in modern JS.

network-api/networkapi/campaign/views.py
youriwims: Sorry about that—I'll revert it.

youriwims: Should there have been errors raised for this when I ran `inv test`?

Pomax: nope, both are fine, but leaving that trailing comma on is "the finer of the two" as far as python is concerned.

network-api/networkapi/campaign/views.py
```diff
@@ -125,13 +126,13 @@ def petition_submission(request, petition):
...
     }
 
-    if petition.requires_country_code:
+    if petition:
         if 'country' in request.data:
             data["country"] = request.data['country']
-        else:
+        if petition.requires_country_code and 'country' not in request.data:
```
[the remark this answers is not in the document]
youriwims: ah true.

TheoChevalier:
Thank you both! I can confirm I can see both fields being sent in the request now 👍

There’s a small regression though, updating the country field sets the language to English.

STR: 
- open https://foundation-mofostaging-pr-3766.herokuapp.com/fr/
- click on the form in the footer
- Pick a language (German, for instance)
- Now pick a country

Current: Language is now set as English
Expected: Language remain German

PR #3784, opened 2019-10-15
Newsletter data updates

youriwims:
## **Issues Fixed**
- [x] Send page language as `lang` if none is chosen
- [x] Send nothing if a `country` isn't selected
---
### **Summary of Changes**
Instead of sending default `country` if none is chosen, we're sending an empty string. As for `lang`, instead of having a default language preset in `views.py`, we're updating `lang` in `<JoinUs />` component to pull from page locale for its initial state. 

Related #3700, #3771 , #3768, #3766

### DIFF
--- a/network-api/networkapi/campaign/views.py
+++ b/network-api/networkapi/campaign/views.py
@@ -4,7 +4,6 @@ from rest_framework.parsers import JSONParser
 from rest_framework import status, permissions
 from django.core.exceptions import ObjectDoesNotExist
 from django.conf import settings
-from django.utils import translation
 from datetime import datetime
 import boto3
 import logging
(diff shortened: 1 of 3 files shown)


network-api/networkapi/campaign/views.py
```diff
@@ -89,8 +89,8 @@ def signup_submission(request, signup):
         "format": "html",
         "source_url": request.data['source'],
         "newsletters": signup.newsletter,
-        "lang": request.data.get('lang', 'en'),
-        "country": request.data.get('country', 'US'),
+        "lang": request.data['lang'],
```
Pomax: let's file a general bug-fix issue here: if there are arguments that _need_ to be supplied, then we should make sure to check that they exist in the request, before forming a payload, and respond with an appropriate HTTP error code for missing parameters.

Pomax: Alternatively, let's tap into `lang` but fall back to the language wagtail sees in the request, or if there is none, fall back to `en`

youriwims: _Update_: decided to go with original method since pulling wagtail lang data takes you down a rabbit hole 🐇

PR #3811, opened 2019-10-21
Increase default number of fields upload limit

### LINKED ISSUE #3798: Bad Request when publishing es/pt YouTube stories
I’m hitting an issue with `es` and `pt` fields for the YouTube Regrets page. When publishing stories for these two locales (even just copying the English ones as in the GIF below), I get a Bad Request error. I managed to successfully publish stories in fr/de

STRs:
- Edit page, show ES fields
- Copy stories from EN
- Publish
bad-request

@cadecairos Are the logs showing anything?

TheoChevalier:
Fixes #3798


network-api/networkapi/settings.py
```diff
@@ -591,3 +591,6 @@
 
 # RSS / ATOM settings
 FEED_LIMIT = env('FEED_LIMIT')
+
+# Support pages with a large number of fields
+DATA_UPLOAD_MAX_NUMBER_FIELDS = int(env('DATA_UPLOAD_MAX_NUMBER_FIELDS'))
```
Pomax: `env` already does format conversion: we unpack the env all the way at the top, where you indicate which type to expect, and which fallback value to use when it's not defined. If you can update this line to just be `env('DATA_UPLOAD_MAX_NUMBER_FIELDS')` with the appropriate specification at the top of settings.py, then we can safely land this (otherwise, the moment we land this, the website will crash if heroku's missing that var)

env.default
```diff
@@ -75,3 +75,6 @@ USE_CLOUDINARY=False
 CLOUDINARY_CLOUD_NAME=
 CLOUDINARY_API_KEY=
 CLOUDINARY_API_SECRET=
+
+# Django config overrides
+DATA_UPLOAD_MAX_NUMBER_FIELDS=2500
```
Pomax: Let's leave this off. We don't want to make this a default thing - for testing, we can manually add this to our `.env` file during investigation instead.

Issue #3760, opened 2019-10-09
PNI homepage/categories: only show products with review date within one year


kristinashu:
We have added a review by date to each PNI product. Can we still show older products on our category product pages but hide them on the homepage?

Pomax:
On @caltrider's request, I'm turning this into hiding both on homepage and category pages

Pomax:
I'll fix the review cutoff date to June 1st 2019. That way we don't run into weird "sliding window" issues (because 1 year ago is a different cutoff date ever day), and should catch anything that's part fo the new batch.

PR #3815, opened 2019-10-24
only show products reviewed in the last year

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3760 by adding an `is_current` property to the Product model that checks whether the review date is less than 365 days old. That's not perfect maths because dates are hard, but we don't need true to-the-day checks here, either.

https://github.com/mozilla/foundation.mozilla.org/pull/3815/files?utf8=%E2%9C%93&diff=unified&w=1 for the sane file view (with indent changes hidden)

### DIFF
--- a/network-api/networkapi/buyersguide/models.py
+++ b/network-api/networkapi/buyersguide/models.py
@@ -1,3 +1,5 @@
+from datetime import datetime, timedelta
+
 from cloudinary import uploader
 from cloudinary.models import CloudinaryField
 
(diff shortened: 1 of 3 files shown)


network-api/networkapi/buyersguide/models.py
```diff
@@ -134,6 +136,13 @@ class Product(ClusterableModel):
         help_text='Review date of this product',
     )
 
+    @property
+    def is_current(self):
+        d = self.review_date
+        review = datetime(d.year, d.month, d.day)
+        cutoff = datetime(2019, 6, 1)
```
mmmavis: Is June 1st or July 1st our cutoff date? I think month for Python's `datetime` ranges from 1-12 instead of 0-11?

Pomax: I intended June here

PR #3829, opened 2019-10-31
Fix `Error: No such volume` while running docker-new-db

patjouk:
`inv docker-new-db` fails when the volume doesn't exist. I probably missed it last time I made changes in that file :(

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -227,7 +227,7 @@ def docker_test_node(ctx):
 @task
 def docker_new_db(ctx):
     """Delete your database and create a new one with fake data"""
-    print("* Stopping services first and deleting database")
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -227,10 +227,8 @@ def docker_test_node(ctx):
 @task
 def docker_new_db(ctx):
     """Delete your database and create a new one with fake data"""
-    print("* Stopping services first")
-    ctx.run("docker-compose down")
-    print("* Deleting database")
-    ctx.run("docker volume rm foundationmozillaorg_postgres_data")
+    print("* Stopping services first and deleting database")
```
cadecairos:
```suggestion
    print("* Stopping services and deleting volumes first")
```

PR #3849, opened 2019-11-05
Update IH category name to IHR

### LINKED ISSUE #3822: Blog category updates aren't reflected in Wagtail component
Changes when adding or editing a blog category do not get updated in the Wagtail component list of categories:

image

Can Wagtail component list automatically update or should adding/editing a category not be possible in the CMS?

cc @Pomax @youriwims

youriwims:
Closes #3822

## Checklist

**Changes in Models:**

---
### Test Pages

https://foundation-mofostaging-pr-3849.herokuapp.com/en/blog/
https://foundation-mofostaging-pr-3849.herokuapp.com/en/blog/category/internet-health-report/


Pomax:
The wagtailpages migration needs a bit that makes sure that the category we built over in https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/migrations/0075_auto_20190829_2317.py gets renamed as part of the migration, too.

gideonthomas:
@Pomax are you good to take this review on?

network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -40,6 +40,10 @@ def add_category(post):
     post.category.add(choice(categories))
     post.save()
 
+    internet_health = BlogPageCategory.objects.get(name='Internet Health')
```
youriwims: @Pomax would this be the correct way to rename IH in factory? Also, I updated the category name in the cms and looks like it may have fixed the [previous issue](https://github.com/mozilla/foundation.mozilla.org/pull/3849#pullrequestreview-312981442).

kristinashu:
Might be too late, but would it be easier/safer to keep 'Internet Health' and just add a new category for 'Internet Health Report'?

Pomax:
this is something we will need to do on occasion anyway, so running through this now is preferred (it gives us something to refer back to in the future =). We just need to make sure to catch all the places we've made category string assumptions.

Issue #3823, opened 2019-10-29
Changes to "It uses" section


beccaklam:
This is the implementation ticket for #3796

After seeing this section in action, @caltrider felt that it looked empty and uninformative with some of the answers being hidden. So, we're reverting back to how we laid out this section previously with some style changes (the section will still be sans icons). We also need to add a fourth variable to the answer selection.

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] Link to original design ticket
- [x] Link to Redpen: https://redpen.io/p/aj7f2e09f704b2e4ee
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included

## List of Requirements
- [x] Add fourth variable: "Unknown" to list of answers. So all possible answers for this section would be "_No, Yes, N/A and Unknown_"
- [x] **All** answers should be shown at all times
Add styling changes:
 - [x] Answers should be in italics
- [x] The diving line " | " should be #CCCCCC

## Mobile:
![Screen Shot 2019-10-29 at 2 04 55 PM](https://user-images.githubusercontent.com/20977376/67795821-44814f00-fa55-11e9-9df4-1d161bf48173.png)

## Desktop:
![Screen Shot 2019-10-29 at 2 05 38 PM](https://user-images.githubusercontent.com/20977376/67795831-48ad6c80-fa55-11e9-9fad-5cde7c4a4522.png)

PR #3855, opened 2019-11-06
Changes to "it uses" section.

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3823 by uplifting the device/app fields from NullBoolean to our new Yes/No/Unknown/NA field, and styling the infobox as per the comps.

Example product with all four values: https://foundation-mofostaging-pr-3855.herokuapp.com/en/privacynotincluded/products/percy-cypress/


network-api/networkapi/buyersguide/migrations/0032_uplift_app_device_fields.py
```diff
@@ -0,0 +1,133 @@
+# Generated by Django 2.2.5 on 2019-11-06 20:11
```
Pomax: similar to migration 0028, which did this for a whole bunch of other Boolean fields.

beccaklam:
@Pomax This is looking good! 
Only major bug I found was that at the iPad view (both vertical and horizontal) the words are at risk of pushing into one another:
Screen Shot 2019-11-07 at 3 37 11 PM
I think perhaps we can start the mobile media query earlier at the iPad width?

beccaklam:
> I would suggest removing `col-...` from the `.it-uses` text blocks and instead relying on flexbox a bit more for position + spacing. Can we try making the parent container `flex:column`, then `flex:row` + `justify-content:between` at `lg` screens?

That would be great! I did find the spacing a bit off but wasn't sure if there was a solution to it!

Pomax:
updated the PR - `lg` was actually too small (several `unknown` would still overlap) so I've set it to columns for `xl` (1200+). If we want `lg` (992+) we probably need to fiddle with font sizes/condensed typeface/something shorter than "unknown"... =S

When using `lg`, I was seeing things like this (and that's not even all-unknowns):

image

Issue #3867, opened 2019-11-07
Add testing of known wagtail URLS


Pomax:
the wagtailpages app should have a tests.py that tests for various urls "for real" (i.e. with wagtail actually running) in order to verify things like "does a known blog category page resolve?" and "Does a bad blog category page 302 to the base blog page" etc.

Pomax:
stale.

PR #3868, opened 2019-11-07
Redirect bad cat home

### LINKED ISSUE #3862: redirect to `/blog/` on unknown tags/categories
It is currently possible to hit up an unknown category on the blog page, and be presented with the same content as the unfiltered `/blog/`, but with a filter label, and a URL that suggests the results _were_ filtered...

So, based on conversations on Slack, we've decided to change the behaviour for unknown categories to simply explicitly redirect to `/blog/`, instead.

Pomax:
Closes #3862 by redirecting any unknown category to /blog (with locale preserved)

This is missing a test for good vs. bad category because I can't _for the life of me_ figure out how to test true URLs using Django's test approach. Every URL ends up being 404 because they're wagtail managed, so I filed https://github.com/mozilla/foundation.mozilla.org/issues/3867 as followup for the next sprint

test urls:
- https://foundation-mofostaging-pr-3868.herokuapp.com/de/blog/ (should show all test blog entries)
- https://foundation-mofostaging-pr-3868.herokuapp.com/de/blog/category/internet-health-report (should show filtered results)
- https://foundation-mofostaging-pr-3868.herokuapp.com/de/blog/category/lolcats (should redirect to `/de/blog/`)


network-api/networkapi/wagtailpages/models.py
```diff
@@ -618,43 +620,29 @@ def filter_entries_for_tag(self, entries, context):
...
-            entries = [
-                entry
-                for
-                entry in entries.specific()
-                if
-                hasattr(entry, 'category')
-                and
-                cat in entry.category.all()
-            ]
+        category = self.filtered.get('category')
```
Pomax: we can't get here anymore is the category doesn't exist, so the "if cat is not None" scoping became irrelevant

Pomax:
@kristina this feels like it's doing the right thing, can you confirm by checking the test urls?

Pomax:
cadecairos had a good link for trying to make tests work using https://docs.djangoproject.com/en/2.2/topics/testing/tools/#django.test.LiveServerTestCase so I'd like to get that working for this PR before we land it.

Pomax:
K, tests are still not working, so I'll land this as-is with follow-up https://github.com/mozilla/foundation.mozilla.org/issues/3867 to figure out how the heq to add live tests =)

PR #3541, opened 2019-08-14
Make all pages og:aware

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3540 by making our `base` template add in the `lang` attribute and og metadata, such that any child _can_ opt out by blanking the `social_metadata` block... although I have no idea why any page would do so.

PR #3887, opened 2019-11-12
"Dev help page" on review apps: fix 500

patjouk:
Review apps all have a `DEV HELP` page with a few links to different types of pages (`Multi page campaign`, `Single page opportunity`, etc). It suffered from the same issues as the [Mozfest 404 page](https://github.com/mozilla/foundation.mozilla.org/pull/3541) so this PR makes sure we don't load wagtail-managed metadata, and also update a few broken links.


Pomax:
agh! good catch, thanks!

PR #3892, opened 2019-11-13
Update pni sharecopy

youriwims:
Redoing this since somehow, this work was reverted.


youriwims:
Confirmed success!!! cc @Pomax 

Screen Shot 2019-11-13 at 1 15 30 PM

PR #3910, opened 2019-11-18
Fix #3900: comment out watch video button from mozfest homepage

### LINKED ISSUE #3900: Remove Watch Video button from mozfest homepage
Remove the button at the top of the site. Bonus: make it toggleable via the CMS so that we can easily add it back in later.

ragnarokatz:
Closes #3900

### DIFF
--- a/network-api/networkapi/mozfest/templates/partials/homepage_banner.html
+++ b/network-api/networkapi/mozfest/templates/partials/homepage_banner.html
@@ -3,7 +3,7 @@
     <div class="col-12 col-lg-8 dark-theme">
       <h1 class="h1-heading">{{ page.banner_heading }}</h1>
       <p class="body-large">{{ page.banner_guide_text }}</p>
-      {% if page.banner_video_url %}
+      {% if page.banner_video_url is not none %}
       <a class="btn btn-primary" href="{{ page.banner_video_url }}" target="_blank" id="mozfest-home-watch-video-button">Watch Video</a>
       {% endif %}
     </div>


gideonthomas:
@Pomax is this fix ok?

gideonthomas:
@ragnarokatz are you able to make the fix that @Pomax suggests here: https://github.com/mozilla/foundation.mozilla.org/issues/3900#issuecomment-555276136?

@mmmavis Feel free to take yourself off review if you're comfortable with me handling it.

ragnarokatz:
@gideonthomas yes. I will do so shortly.

Pomax:
@ragnarokatz it looks like maybe something went wrong with the commits: we don't want to remove the video url field, we just want to update the template so that if the field has no value, we don't show a video button. That way no models or admin panels need to change.

ragnarokatz:
@Pomax I see.

I think the existing code already does this in the following:

```
{% if page.banner_video_url %}
<a class="btn btn-primary" href="{{ page.banner_video_url }}" target="_blank" id="mozfest-home-watch-video-button">Watch Video</a>
{% endif %}
```

I'm not sure `if page.banner_video_url` will return true if a `none` object is passed in for `banner_video_url`. If that is the case, I can change it to either `{% if variable is defined %}` or `{% if variable is not none %}`

gideonthomas:
@ragnarokatz can you commit the changes you mention in your PR? I can test it and see if that works.

ragnarokatz:
@gideonthomas done. In order to remove the watch video button, simply do not pass in `page.banner_video_url` or pass in a null.

gideonthomas:
Sorry to get back to you so late @ragnarokatz. Since the fix for this ticket was changed to - if the url doesn't exist, don't show it -- I think the code that already exists in the repository is correct. You dug into the code to find out that we have a field that can be toggled and that's how we were able to identify what the right course was for fixing this ticket, so we really appreciate you digging into the code. With that said, it looks like code-wise, nothing needs to be fixed here.

ragnarokatz:
@gideonthomas I see. No worries.

PR #3913, opened 2019-11-18
React/JS internationalization

### LINKED ISSUE #3911: Enable localization of React/JS components
Tracking issue for an incoming PR (🍾)

TheoChevalier:
Fixes #3911

It turned out to be a pretty clean solution :)

The approach is described in https://medium.com/@hugosousa/hacking-djangos-makemessages-for-better-translations-matching-in-jsx-components-1174b57a13b1 and this was exactly what we needed! 

- `fr` .po is complete, also for testing purposes
- I’ve migrated existing translations from  `source/js/components/petition/locale-strings.jsx` to their respective .po files
- There’s a 24hrs cache set in `network-api/networkapi/urls.py` for the js catalog, copied over those lines from donate-wagtail. We will probably want to enable it on prod only?


README-FOR-WAGTAIL-LOCALISATION.md
```diff
@@ -12,7 +21,7 @@ While Wagtail forms a plan of attack around tackling this, we need a localisatio
 - `wagtail-modeltranslation` is used to enable content localisation based on a `LANGUAGE_CODE` default language identifier, and `LANGUAGES` list of tuples (of the form `('code', 'full name of locale')`).
 - a custom javascript library that improves usability by allowing users of the CMS to hide/reveal localisable fields based on the language they belong to.
 
-The modeltranslation library adds localised URLs to Wagtail, interjecting the locale code in the URL betewen the host domain and the content path, such that a url like:
+The modeltranslation library adds localised URLs to Wagtail, interjecting the locale code in the URL between the host domain and the content path, such that a url like:
```
Pomax: 😄

source/js/components/join/join.jsx
```diff
@@ -445,11 +450,17 @@ export default class JoinUs extends React.Component {
...
                 required
               />
-              <p className="d-inline-block body-small form-text mb-0">
-                {getText(
-                  `I'm okay with Mozilla handling my info as explained in this Privacy Notice`
-                )}
-              </p>
+              <p
+                className="d-inline-block body-small form-text mb-0"
+                dangerouslySetInnerHTML={{
```
Pomax: This part looks... questionable. If there's a react-appropriate gettext, it should understand "slotting in more JSX".

source/js/components/petition/petition.jsx
```diff
@@ -687,15 +689,21 @@ export default class Petition extends React.Component {
...
                   ref="privacy"
                 />
-                <span className="body-small form-text">
-                  {getText(
-                    `I'm okay with Mozilla handling my info as explained in this Privacy Notice`
-                  )}
-                </span>
+                <span
+                  className="body-small form-text"
+                  dangerouslySetInnerHTML={{
```
Pomax: see above

source/js/components/pulse-project-list/pulse-project-list.jsx
```diff
@@ -69,7 +69,9 @@ export default class PulseProjectList extends React.Component {
       let url;
 
       if (project.related_creators.length) {
-        byline = `By ${project.related_creators.map(rc => rc.name).join(`, `)}`;
+        byline =
+          gettext("By") +
```
Pomax: This makes no sense, template literals are explicitly to _avoid_ string concatenation =)
```
          byline = `${gettext("By")} ${project. ..... }`;
```

Pomax:
Please use template literal syntax for all strings. Ideally there are no `'` and `"` anywhere, except _inside_ strings because the actual text uses an apostrophe or double quote.

README-FOR-WAGTAIL-LOCALISATION.md
```diff
@@ -1,5 +1,14 @@
-# (Temporary) localised content in Wagtail
+# Wagtail localization
 
+## Localizing Python files and Django templates
+
+Strings should be wrapped for localization following the [Django localization documentation](https://docs.djangoproject.com/en/2.2/topics/i18n/translation/).
+
+## Localizing .js or .jsx files
+
+Strings should be wrapped for localization using `gettext("Your string")`
```
Pomax:
```suggestion
Strings should be wrapped for localization using `gettext(\`Your string\`)`
```
We don't want ' or " for strings.

Pomax:
Another worry: how does this behave with development, where webpack recompiles the bundle on jsx changes? Does this now require a python `gettext()` extraction during the webpack bundling?

TheoChevalier:
Thanks for the review! Will have a look at that later today :)

> Another worry: how does this behave with development, where webpack recompiles the bundle on jsx changes? Does this now require a python `gettext()` extraction during the webpack bundling?

It works well, meaning if you change ``gettext(`foo`)`` into ``gettext(`bar`)`` in your editor, `bar` will be displayed upon page reload.

If by Python extraction you mean running `makemessages`, no, that is not required to make things work. Like for Django templates, `gettext()` runs fine by itself, even when strings have not yet been extracted into .po

README-FOR-WAGTAIL-LOCALISATION.md
TheoChevalier: I think Atom fixed this for me :)

source/js/components/join/join.jsx
TheoChevalier: I must admit I have no idea how to do that…
Of course `<p>{gettext('I’m okay with Mozilla handling my info as explained in this <Link to="{link}">Privacy Notice</Link>')}</p>` won’t be interpreted as JSX

TheoChevalier:
``gettext(`foobar`)`` with backticks doesn’t work, unfortunately :/ Strings are not extracted at all

Pomax:
Hm, where is this `gettext` coming from? Do you have a link to the library that makes this happen? Because backticks are pretty required to conform to code conventions =(

TheoChevalier:
> Hm, where is this `gettext` coming from? Is this all the GNU xgettext? (If not, do you have a link to the library that makes this happen?) Because backticks are pretty required to conform to code conventions on the JS side. Changing that to make a tool that's still stuck in ES5 is probably not the right direction.

I think it’s coming from the `JavaScriptCatalog` served by Django https://docs.djangoproject.com/en/2.2/topics/i18n/translation/#using-the-javascript-translation-catalog

It looks like it’s indeed defined in `/jsi18n`
Capture d’écran 2019-11-20 à 17 02 42

For the backtick not recognized during extraction, it’s probably not supported by`xgettext` itself, as it’s the library called by `makemessages`

> 
> Have we looked at whether there other React localisation solutions (e.g. a context-based `gettext`) that know how to work with modern JS?

I couldn’t find a lot of examples, but just stumbled upon https://blog.sentry.io/2016/01/07/react-i18n
Which suggests we can have a look into Python Babel + https://github.com/getsentry/babel-gettext-extractor
Which

Pomax:
Oh nice! If babel-gettext isn't a crazy amount more work to set up,, that would certainly be a preferred solution.

TheoChevalier:
Also related, it looks indeed a gettext limitation https://savannah.gnu.org/bugs/?50920
They only allowed for extraction to complete when there are interpolated strings in the file, while it was ignoring the rest of the file before. Progress…

TheoChevalier:
So, Sentry’s approach if far from being trivial…

String extraction using `babel-gettext-extractor` + a bunch of stuff
https://github.com/getsentry/sentry/blob/4de0d59ca7308c753d5a1ff6c84e9e585e987e16/webpack.config.js#L55-L163

In-app .po to JSON conversion: https://github.com/getsentry/sentry/blob/4de0d59ca7308c753d5a1ff6c84e9e585e987e16/src/sentry/static/sentry/app/translations.tsx

[Jed](https://messageformat.github.io/Jed/) initialization, importing the converted files + redefining all the gettext functions 
https://github.com/getsentry/sentry/blob/4de0d59ca7308c753d5a1ff6c84e9e585e987e16/src/sentry/static/sentry/app/locale.tsx

Not sure we want to add that much code/complexity. What do you think @Pomax ?

TheoChevalier:
Let’s close this one too, I think we can have something better

PR #3915, opened 2019-11-18
Update wagtail-inventory to latest version from their repo

patjouk:
Closes MozillaFoundation/mofo-devops-private#265

We need to add `git` to our Python docker image to be able to install `wagtail-inventory` to the latest version since it's not yet released on PyPi.


Pomax:
It probably makes more sense to first ask them if they can publish to pypi, so we (and everyone else) won't need to resort to git links?

Pomax:
We didn't have to wait long - https://pypi.org/project/wagtail-inventory/ is now on 0.8 on pypi.

PR #3931, opened 2019-11-18
Testing eslint-plugin-jsx-a11y

youriwims:
Related PRs/issues #3934

Resources: 
https://reactjs.org/docs/accessibility.html#eslint-plugin-jsx-a11y
https://github.com/evcohen/eslint-plugin-jsx-a11y

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,10 +1,12 @@
 {
   "extends": [
-    "prettier"
+    "prettier",
+    "plugin:jsx-a11y/recommended"
   ],
   "plugins": [
     "prettier",
-    "react"
+    "react",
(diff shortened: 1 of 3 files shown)


youriwims:
@patjouk I think I'm going to need some help setting up `axe-core` tests. 😬 please & thank you 💟

youriwims:
I already set up `eslint-plugin-jsx-a11y` here...just wanted to set up `react-axe` and `axe-core` to see how they work together. Then, if everything goes well, maybe we can test `axe-cli` as well on the same branch?..what do you think?

patjouk:
I would say let's do two different PRs since it looks like axe will require more work.

Good work so far 👍

Issue #3163, opened 2019-05-16
Investigate adding .body-large style to Paragraph component


kristinashu:
Can we add [.body-large](https://foundation.mozilla.org/en/style-guide/) as an option to the Paragraph component? It would be quite useful sometimes.

cc @mmmavis @Pomax

mmmavis:
The Paragraph component auto wraps any non-block level elements in `<p>`. e.g., (bold text, anchor, etc). So adding a new ["feature" (option)](http://docs.wagtail.io/en/v2.0/advanced_topics/customisation/page_editing_interface.html#limiting-features-in-a-rich-text-field) for `.body-large` text to the list below doesn't seem to be a good way to achieve what we want.
<img width="856" alt="image" src="https://user-images.githubusercontent.com/2896608/57954411-b26efb00-78a7-11e9-9ecb-5a4755010197.png">

@Pomax correct me if I'm wrong - I think it's cleaner and easier if we create a new Wagtail component  (call it Large Paragraph???) that does basically the same thing as Paragraph component except by default has `.body-large` assigned to the wrapper `<p>` so everything in that component is in `.body-large` style without the need of extra SCSS overrides like [what we added for the blog page](https://github.com/mozilla/foundation.mozilla.org/blob/d5c61fa37e6e4be2ab3f40812839674256001561/source/sass/wagtail.scss#L78).

kristinashu:
This is a P2 for *Privacy Included piece https://github.com/mozilla/foundation.mozilla.org/issues/3801

Pomax:
creating a new component that does the exact same thing as paragraph but adds a `body-large` is not going to be something that people will remember to use, and will probably cause more confusion in the CMS view than it lets us solve. 

This might be one of those things that we can't really solve without hacks that work short term, but make things much larger long term, because "adding classes to only certain freeform paragraphs" is not a thing that fit in the the wagtail way of streamfield content editing.

We should investigate if there's something we can do here, which I would estimate as a 3, but note that only covers the investigation, not the implementation yet.

kristinashu:
Hmm ok, I'll put this on design. Maybe we need to increase the size of our regular body text... will explore.

kristinashu:
Maybe loosely related https://github.com/mozilla/foundation.mozilla.org/issues/3847?

Pomax:
[...]

Issue #3847, opened 2019-11-05
Bold doesn't work for italics in Paragraph component


kristinashu:
When `Bold` is used in addition to `italics` the text looks italic but not bold. It should be both bold and italic if both are selected.

Screen cap of selections in cms:

<img width="1238" alt="image" src="https://user-images.githubusercontent.com/1068783/68235874-e4405f00-ffb8-11e9-96f6-9b345c050125.png">

But doesn't look bold on prod:

<img width="1007" alt="image" src="https://user-images.githubusercontent.com/1068783/68235891-f02c2100-ffb8-11e9-99b6-55715f766aad.png">

Pomax:
It looks like this was a bug that got fixed in a wagtail update:

![image](https://user-images.githubusercontent.com/177243/69179379-4a9ca580-0ac0-11ea-81ac-df06b67a8402.png)

![image](https://user-images.githubusercontent.com/177243/69179393-4ec8c300-0ac0-11ea-9736-9c764af07f5b.png)

kristinashu:
Sweet! I don't see it effective yet, do you know when we will have the latest Wagtail version?

I'll close this ticket tho.

kristinashu:
Or maybe it is working... I think our bold is just quite subtle on my computer.

PR #3937, opened 2019-11-19
add a 'body-large' button to our rich text editor's button bar

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3163 by introducing a new button to the rich text editor for marking stretches of text as "body-large".


mmmavis:
Wow lovely! So excited that we are able to customize the rich text editor further to suit our needs. One question - do we plan to make this new "bodylarge" button available for all page types? Currently, I don't see it on blog pages (and I haven't checked other page types yet)

image

kristinashu:
Yay! Designers are going to be so happy with this! But we will warn people using the cms to make sure to preview the page before posting in case they have applied 'L' to one of the Headings.

> One question - do we plan to make this new "bodylarge" button available for all page types? Currently, I don't see it on blog pages (and I haven't checked other page types yet)

I _think_ it's showing up everywhere it should. Blog pages use body-large by default so we won't need to add this styling to it.

Only issue I see is that links aren't styled. Links should look like the section under `Paragraphs with links` in the [style guide](https://foundation.mozilla.org/en/style-guide/).

(I've also highlighted this issue https://github.com/mozilla/foundation.mozilla.org/issues/3847#issuecomment-555672631 but sounds like it's a Wagtail thing.)

image

Pomax:
@kristina so that's the limitation of the fact that this is applied "to a span": if we want it to kick in properly, we'll need to make doubly-sure that our SCSS is such that `.body-large` applies to spans and their content, not just paragraphs =S

Pomax:
As for the blog richtext editor: we have to specify this new button as one of the features to use, and things like the blog category and blog page use their own rich text feature list, so we'll need to make sure to update every separate one where we want this (or any =D) new feature to show up

mmmavis:
Haven't tested it out myself but I think something like this _should_ make sure `body-large` anchor created on CMS appears blue on the final page 🤞 

```scss
.cms {
  a .body-large {
    color: inherit;
  }
}
```

Pomax:
@mmmavis that solved it =) Does the rest of the code look fine to you?

@kristinashu if that was the only styling oddity this should now all work =)

kristinashu:
Hmmm view deployment isn't working for me

Pomax:
It looks like heroku cleared the review app. rebuilding it now.

Pomax:
@mmmavis this is ready for code review again

network-api/networkapi/wagtailpages/models.py
```diff
@@ -58,7 +58,7 @@ def titlecase(s):
 base_fields = [field for field in [
     ('paragraph', blocks.RichTextBlock(
         features=[
-            'bold', 'italic',
+            'bold', 'italic', 'bodylarge',
```
mmmavis: Should we rename this to `body_large` to align with our naming convention?

Pomax: I think it probably makes more sense to call this `large` rather than `bodylarge`, thinking about it some more

Pomax:
This was not supposed to land, it had CI errors. reverted through https://github.com/mozilla/foundation.mozilla.org/pull/3979

Issue #3934, opened 2019-11-19
[WIP] Master Accessibility Ticket


youriwims:
## CI Testing
1. Enable A11y CI 
    1.  Integrate `react-axe` / `cypress-axe` with Foundation repo for local testing [**#4058**]
    2.  Add `axe-cli` to Foundation site CI [ **#4059**]

## Pages in need of some TLC*
  1. Global
     1. Nav [**#4001**]
     2. Footer [**#4002**]
  2. Fo.Mo Blog
        1. Blog Post [**#4004**]
        2. Blog Homepage [**#4003**]
  3. PNI
       1. Home [**#4005**]
       2. Product [**#4006**]
       3. About [**#4007**]
  4. Fo.Mo
      1. About [**#4009**]
      2. Home [**#4008**]
      3. Initiatives
      4. Participate
      5. Internet Health
      6. Fellowships
---
**Key actions pp** **
      

  - Semantic Markup
    - Proper Outline
  - Use html tags
     - If unable, find proper aria-* or styling.
     - Each page should have a unique, descriptive h1 and page title
   - Colour contrast
     - Make sure works in b/w
     - Check in high contrast
   - Keyboard Navigation
      - Elements should be focusable
      - Other manual testing
   - Markup Validation
   - Keyboard events for newsletter signup forms 

_* Organized based on most visited/navigation summary from GA results._
_** In addition to/ an extension of standard browser extension testing._

**More details to come**

jamilasnell:
We should close this ticket and do a re-evaluation of this work so that we can be up to date!

PR #3938, opened 2019-11-19
Testing pa11y

youriwims:
Related PRs/issues #3934 

Resources:
https://github.com/pa11y/pa11y-ci
https://github.com/Princeton-CDH/cdh-web/commit/ecebc5fc9b88d4b086e3d9f9c6d1092b885d70cb
https://github.com/Princeton-CDH/ppa-django/commit/8c845e8dddbebbbdea0e60e17fa78dcc5c8ab7e4
https://marmelab.com/blog/2018/02/22/accessibililty-testing-e2e.html
https://andrewmee.com/posts/automated-accessibility-testing-node-travis-ci-pa11y/


youriwims:
Hmmm, @patjouk where does the `requirements.txt` file come in? 

Also, I want to set up tests in a way that it replaces the fo.mo urls in pa11y `json` file w/ `localhost:8000` when running a11y checks. This is the command I used to do it locally:
> ```pa11y-ci --sitemap https://foundation.mozilla.org/sitemap.xml --sitemap-find foundation.mozilla.org --sitemap-replace localhost:8000``` 

But it'd be better to automate it, right?...I tried to do that in `package.json` 😬

patjouk:
You're going in the right direction 🥇 

**Testing on `localhost:8000`:**
Right now, your tests are running against the URLs defined in `.pa11yci` (no need for the `.json` 😉 ), which are all foundation site prod URLs. To change this, you can either edit this file to `localhost:8000/{page you want to test}` or pass a sitemap ([Pa11y doc on how to do that](https://github.com/pa11y/pa11y-ci#sitemaps)). I saw you tried to use the sitemap from the Foundation site to replace every URL present in the xml by `localhost:8000`. The issue is that `https://foundation.mozilla.org/sitemap.xml` doesn't exist since [it's only created when `DEBUG` is at `True`](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/urls.py#L47). If you run the tests locally, you will see that it's never calling any `localhost:8000` URLs, just silently failing and going to the default configuration from `.pa11yci` 😞 
The good thing is that we can pass the sitemap directly to pa11y-ci that way: `pa11y-ci --sitemap http://localhost:8000/sitemap.xml`.

Keep in mind that the content on the foundation site and on review apps are different. Running pa11y locally, you will see that we don't get the same number of errors for the same tests:
Screenshot 2019-11-20 at 14 30 01

**Running the tests and django server on Travis:**
The example I sent you are using pip with requirements files. We don't use that because we're using pipenv and have Pipefile instead. Take a look at how Percy tests are set up: you will need to [replicate this setting](https://github.com/mozilla/foundation.mozilla.org/blob/master/.travis.yml#L65-L85), without the part for percy. You won't need L68 either, but you will need to [add the Chrome addon](https://docs.travis-ci.com/user/chrome) so that our code on Travis can talk to a Chrome browser.

**A suggestion:**
Since it's catching a lot of errors, confirming that the work you're doing is important, it might require a lot of work and time to fix. We might want to look if it would be possible to make the tests return warnings instead of errors: that way, Travis stays green for now, while we work on fixing those issues. That way, it's already set up in the CI but not blocking. It's a double-edged sword: warning means green ci, means no one checking what's actually happening there. Plus, we might never switch it back to errors. But at the same time, I don't want people to start merging PRs with tests failing on Travis. Again, only a suggestion, but something we should keep in mind.

youriwims:
_Update_: Pa11y CI works locally + within Travis. It still needs to be configured to error `0` instead of `1` or `2`. 

_Feedback_: Travis output isn't very nice to look at...too many lines and formatting is overly dependent on bullets and minor indentation.

PR #3940, opened 2019-11-20
Localize Blog Page Category

### LINKED ISSUE #3939: Blog page categories can’t be localized
As per Slack convo, it turns out those fields are not enabled for localization

TheoChevalier:
Fixes #3939

Looks like we will need to input English name/intro again, but otherwise it seem to work fine


TheoChevalier:
Let’s close this for now, as it’s way more complicated than it looked. I’ll keep #3939 open

Issue #3941, opened 2019-11-20
Load more button on blog category pages is broken


kristinashu:
`Load more results` button at the bottom of blog category pages doesn't do anything, it should load more blog posts.

Example: https://foundation.mozilla.org/en/blog/category/internet-health-report/

cc @Pomax

Pomax:
Found out why, PR filed.

PR #3944, opened 2019-11-20
make both category subroutes resolve category objects in the same way

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/3941 by making sure both '/blog/category/.../` and the JSON fetch request deeper in that route resolve category slugs to category objects in the exact same way.


network-api/networkapi/wagtailpages/models.py
```diff
@@ -735,8 +735,21 @@ def entries_by_tag(self, request, tag, *args, **kwargs):
     category routes
     """
 
+    # helper function to resolve category slugs to actual objects
```
Pomax: before, this code only existed in the `category/somecategory` route handler, so it never kicked in for the "load more" request (which uses a different route) and so the actual result fetching completely failed.

kristinashu:
Yes, made a test page here https://foundation-mofostaging-pr-3944.herokuapp.com/en/blog/category/advocacy/ and it seems to work!

mmmavis:
@Pomax redirecting bad category url doesn't seem to work. https://foundation-mofostaging-pr-3944.herokuapp.com/en/blog/category/foo/ gives me `500`

Pomax:
Good catch @mmmavis, fixed!

PR #3946, opened 2019-11-20
Pull subscription status from pni home url

youriwims:
Previously we were searching if `subscribed=1` was coming form a product page url and this pr fixes that issue and instead looks at the pni home url `subscribed` param.

https://foundation-mofostaging-pr-3946.herokuapp.com/en/privacynotincluded/
https://foundation-mofostaging-pr-3946.herokuapp.com/en/privacynotincluded/?subscribed=1


Pomax:
So... weird thing: "home" in the nav bar is not underlined in the `/?subscribe=1` link O_o

youriwims:
@Pomax filed a follow up here 👉 #3949.

PR #3932, opened 2019-11-19
About content updates cont.

youriwims:
Closes #3896
Related PRs/issues #3800

Issue #3800, opened 2019-10-18
PNI: Implement About Page additions 


kristinashu:
Design ticket https://github.com/mozilla/foundation.mozilla.org/issues/3619

## List of Requirements
- Update and add new about pages
- Keep fonts, layout, and general styling the same as our current about pages
- remove accordions (mainly for _How to Use this Guide_)

## Copy

- [x] [How to Use This Guide](https://docs.google.com/document/d/12QuPQ_RXkmLlAlhCcPTX50-Nup2JskAIwC5ntcn8R4Q/edit)
- [x] [Why We Made This Guide](https://docs.google.com/document/d/1rMo-TNMcHkgYOs2JSP8UxpW6AHdrx3N2odbL-uRP1Xs/edit)
- [x] [Minimum Security Guidelines Explained](https://docs.google.com/document/d/1Y49QM9Yb-BWI7IhtwQUdXMWS6CQoIy8WN6j8Ox_rZSE/edit#)
- [x] [About our Methodology](https://docs.google.com/document/d/1oaPPgFRlA1_zfpd5tPPe8THC20iFnmSYBvSvFppkmjA/edit)
- [x] [Press](https://docs.google.com/document/d/153YTZ2edXYxMUtFAXyX34d7-WlKCW4xC9mYQd7i36JA/edit#heading=h.2ws9vo67f5s)

kristinashu:
Might not have final copy until Thursday Nov 7 so might not get this done for launch.

kristinashu:
@Pomax @youriwims Jen is swapped but is trying really hard to finalize the copy for tomorrow morning. If she gets it done by noon do you think this can get done for a prod push on Tuesday Nov 12? I see you estimated it as a 5 but is that still the case with these designs? There are no long any accordions, just copy. Lots and lots of copy.

youriwims:
I can take some time to do this tomorrow once Jen has the copy. Since it's just copy, this is more so a 2-3...I'm just going to put 3 since I skimmed the comps lol

kristinashu:
Note: we won't be using the accordions anymore. We think people who like to read this type of content would prefer to see everything all at once.

caltrider:
Sorry @beccaklam, I haven't had a chance to think on that. We can put a link in anywhere you see fit.

beccaklam:
@kristinashu @youriwims  Can we do this in the cms or is this something a dev has to do?

kristinashu:
[...]

Issue #3619, opened 2019-09-09
PNI: About Page additions


kristinashu:
Talk to Jen and get copy from her before starting.

Add simple pages to our About section:
- Add minimum security standards explainer section
- Add press section

beccaklam:
@caltrider Do you have content ready for this?

caltrider:
I don't yet. I do have some place holder content that will work for two sections.

Here's the link for the ["Methodology](https://docs.google.com/document/d/1oaPPgFRlA1_zfpd5tPPe8THC20iFnmSYBvSvFppkmjA/edit?userstoinvite=lotta@mozillafoundation.org&ts=5d9caea0&actionButton=1)" section.
Here's the link for the ["Our Minimum Security Standards Explained"](https://docs.google.com/document/d/1Y49QM9Yb-BWI7IhtwQUdXMWS6CQoIy8WN6j8Ox_rZSE/edit#heading=h.25auqwqk2y01) section.

I haven't yet reviewed or edited that content, but it can work as placeholder for now. 

I'll have to find time to go in and edit the existing "How to Use this Guide" and "Why We Made This Guide" but that might not happen before MozFest.

beccaklam:
@caltrider @kristinashu I've mocked up the About Page with the new placeholder content Jen provided. Jen, I assumed that the links you provided ^ were replacing "How to Use this Guide" and "Why We Made This Guide" but I wasn't 100% sure. Let me know how to reorganize things if that's not what you were envisioning. https://redpen.io/p/xed29db03743c5348b

caltrider:
Heya @beccaklam, thanks!
For the About page, these two sections will be new section, in addition to the already existing sections. So, the order of the sections on the left hand side of the About Page should go:

- How to Use This Guide (which I will update, after Mozfest)
- Why We Made This Guide (I'll review and make any needed tweaks after MozFest)
- Our Minimum Security Standards Explained (Becca will finalize after Mozfest)
- Methodology (Becca will finalize after MozFest)
- Contact Us
- Press Kit (this is an optional section we can add if we want to/have time based on Kevin's input)

beccaklam:
Updated redpens to reflect the order of the sections: https://redpen.io/p/xed29db03743c5348b

beccaklam:
@KevZawacki Any thoughts on if we should add the Press section to our About page and if yes, what content we should have there?

KevZawacki:
[...]

PR #3972, opened 2019-11-25
Added monetization section to pni about

### LINKED ISSUE #3974: Add copy to PNI about page
Design ticket https://github.com/MozillaFoundation/Design/issues/436

Please add copy about Web Monetization (highlighted in yellow) to the [How to Use This Guide](https://foundation.mozilla.org/en/privacynotincluded/about/) page.

Copy doc https://docs.google.com/document/d/12QuPQ_RXkmLlAlhCcPTX50-Nup2JskAIwC5ntcn8R4Q/edit

youriwims:
Closes #3974
Related PRs/Issues: 
#3619 
#3932 
#3800

Updates 👉  https://foundation-mofostaging-pr-3972.herokuapp.com/en/privacynotincluded/about/

### DIFF
--- a/network-api/networkapi/buyersguide/templates/about/how_to_use.html
+++ b/network-api/networkapi/buyersguide/templates/about/how_to_use.html
@@ -187,7 +187,7 @@
     <p id="guide-monetization"><b>Monetization of the Guide</b></p>
     <p class="body">
       Mozilla Foundation does not receive any money from the companies featured in this guide or from linking to sites where users purchase products. We are currently experimenting with Web Monetization on a small number of our properties. Web Monetization is an open standard that enables publishers and platforms to compensate creators directly, outside of the dominant advertising model. Only users with a Web Monetization enabled extension in their browsers will be aware of the implementation. Read more about our Web Monetization experiment <a
-        href="https://foundation.mozilla.org/en/blog/introducing-our-experiment-web-monetization/">here</a>.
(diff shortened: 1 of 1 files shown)


kristinashu:
Running this past Anil now but could you please update the link to the blog post so it doesn't have `/en`? It's good practice to not force a language on a user.

youriwims:
Good point—on it.

kristinashu:
Anil says it looks good!

network-api/networkapi/buyersguide/templates/about/how_to_use.html
```diff
@@ -183,11 +183,19 @@ <h1 class="h2-heading">
     </p>
   </section>
 
+  <section aria-labelledby="guide-monetization">
+    <p id="guide-monetization"><b>Monetization of the Guide</b></p>
```
Pomax: it really feels like somewhere along the line we forgot to make sure we made these headings, rather than paragraphs with `<b>` tags - @kristinashu is there an appropriate heading class that we use if we turn these into `<h2>` or `<h3>` instead?

(then we can file a separate issue to do some replacing here, but `<b>`, `<i>`, `<s>`, etc shouldn't really be in HTML that also extensively uses CSS for its styling =D)

kristinashu: Because there are so many titles on this page and each one only has a paragraph of text under it, we wanted to use the more subtle <b> instead of a heading style that would stand out too much.

PR #3984, opened 2019-11-29
React/JS i18n using LinguiJS

### LINKED ISSUE #3911: Enable localization of React/JS components

TheoChevalier:
This is an alternative internationalization solution to https://github.com/mozilla/foundation.mozilla.org/pull/3913, based on [LinguiJS](https://lingui.js.org/index.html).

The ultimate goal is to close #3911

I’ve just wrapped two strings here, and translated them in the French .po:

Screenshot_2019-11-29 Homepage

Localizing links is cleaner with this solution, as using `<a>` tags in `<Trans>` just works™. It even make it abstract in the .po be replacing actual tags with `<0>`

**For the sake of the demo, I’ve committed files that should be only generated locally, so that you have a better idea of the solution right away.** This includes `locale/*/messages.js` (which acts like a `.mo` for the Django strings) and `network-api/locale/_build/source/js/components/join/join.jsx.json` which is used by LinguiJS to generate the .po files when running `npm run extract`

Not all dependencies should actually be required, I’ve played with more things while setting this up. Like dynamic loading, but we don’t really need that with this project setup.

To test:

- Wrap a string with `<Trans>` (or [any other component](https://lingui.js.org/ref/react))
- `npm run extract` to update the .po files
- Translate a .po with PoEdit
- `npm run compile` (Generates the .js files)
- `docker-compose up`


source/js/nav-newsletter.js
```diff
@@ -33,7 +37,7 @@ class NavNewsletter {
         "transitionend",
         handleTransitionend
       );
-      elements.buttonDismiss.textContent = "No thanks";
+      elements.buttonDismiss.textContent = i18n._(t`No thanks`);
```
TheoChevalier: Example with JS strings

source/js/components/join/join.jsx
```diff
@@ -446,9 +450,7 @@ export default class JoinUs extends React.Component {
                 required
               />
               <p className="d-inline-block body-small form-text mb-0">
-                {getText(
-                  `I'm okay with Mozilla handling my info as explained in this Privacy Notice`
-                )}
+                <Trans>I’m okay with Mozilla handling my info as explained in <a href="https://www.mozilla.org/privacy/websites/">this Privacy Notice</a></Trans>
```
TheoChevalier: Example with React strings

source/js/nav-newsletter.js
```diff
@@ -132,7 +136,7 @@ class NavNewsletter {
       0}/`;
     props.csrfToken = props.csrfToken || csrfToken;
     props.isHidden = false;
-    this.form = ReactDOM.render(<JoinUs {...props} />, elements.joinUs);
+    this.form = ReactDOM.render(<I18nProvider language={getCurrentLanguage()} catalogs={catalogs}><JoinUs {...props} /></I18nProvider>, elements.joinUs);
```
TheoChevalier: Note: this makes `this.form.reset()` fail, and I’m not sure how to work around it

source/js/components/join/join.jsx
```diff
@@ -53,7 +57,7 @@ export default class JoinUs extends React.Component {
       navWrapper.querySelector(".form-dismiss") &&
       this.state.apiSuccess
     ) {
-      navWrapper.querySelector(".form-dismiss").textContent = "Back to menu";
+      navWrapper.querySelector(".form-dismiss").textContent = i18n._(t`Back to menu`);
```
Pomax: is there an equivalent function that isn't `_` ? For Django, `_` might be the convention, but in JS land that's the "underscore" library, which we don't use, nor want to create the impression we are.

Pomax: Also, does this _need_ the `t` template tag? It seems odd to need a function as well as an implicit function for the only argument.

source/js/components/join/join.jsx
Pomax: Nice, I like this

source/js/main.js
```diff
@@ -275,7 +278,7 @@ let main = {
       apps.push(
         new Promise(resolve => {
           ReactDOM.render(
-            <JoinUs {...props} whenLoaded={() => resolve()} />,
+            <I18nProvider language={getCurrentLanguage()} catalogs={catalogs}><JoinUs {...props} whenLoaded={() => resolve()} /></I18nProvider>,
```
Pomax: It might make sense for us to create a wrapper that does this part automatically, to replace direct calls to `ReactDOM.render`

source/js/nav-newsletter.js
Pomax: What if we move this into the main `render()` output of the `JoinUs` component instead?

source/js/components/i18n/catalogs-loader.jsx
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import enMessages from '../../../../network-api/locale/en/messages.js';
```
Pomax: Does lingui have the ability to write an `index.js` that does this for us, so we can say something like `import { en, de, es, fr, pl, pt } from "../../../../network-api/locale/index.js"`?

source/js/components/join/join.jsx
TheoChevalier: I don’t think we can change `i18n._`, unless we can override `@lingui/react`

`t` is a macro, c.f. https://lingui.js.org/ref/macro.html & https://github.com/stereobooster/react-lingui-example#translate-text-without-trans-macro

So we could definitely choose not to use the macro, and write `i18n._(/*i18n*/{id: "Back to menu"})` instead. But it’s arguably less elegant…

source/js/components/i18n/catalogs-loader.jsx
TheoChevalier: I don’t think we can tell `lingui extract` to create a single bundle. However, we should be able to do something like this and dynamically load only the locale we need https://github.com/stereobooster/react-lingui-example/blob/master/src/i18n.js

source/js/nav-newsletter.js
TheoChevalier: I’m still hitting that weird issue, unfortunately. Even if I move `<I18nProvider>` inside the `<form>` element.

source/js/components/join/join.jsx
TheoChevalier: Oh, actually, it looks like ``i18n.t`Back to menu` `` works just fine!

TheoChevalier: Such a roller-coaster. This has been deprecated in v3: https://lingui.js.org/releases/migration-3.html

image

TheoChevalier:
I don’t think we should move forward with this, I’ll have another proposal up soon

PR #3993, opened 2019-12-02
Fix `ObjectDoesNotExist` when visiting the IHR blog

patjouk:
Today, Kasia reported that the IHR blog was down. After investigating, it looks like we were passing a `BlogPageCategory` object to the `extract_category_information` function when it expects a `string` (slug of the category).


network-api/networkapi/wagtailpages/models.py
```diff
@@ -738,13 +738,15 @@ def entries_by_tag(self, request, tag, *args, **kwargs):
...
     def get_category_object_for_slug(self, category_slug):
-        category_object = None
 
         # We can't use .filter for @property fields,
         # so we have to run through all categories =(
         for bpc in BlogPageCategory.objects.all():
             if bpc.slug == category_slug:
                 category_object = bpc
+                break
+        else:
```
Pomax: cool, I didn't know this existed

cadecairos: for-else FTW! won't forget this handy little language feature.

patjouk: I learned about this last week, and now I see it everywhere :D

PR #3997, opened 2019-12-02
Hide nav's newsletter sign up button if there's a sign up form in main

### LINKED ISSUE #3925: [Header/body newsletter list conflicts] Hide the “newsletter” top button in pages that have form in their body
### Issue:
If a page has a newsletter form in its body, we should hide the top form button (“newsletter”)so the audience is more focused on the body form, reducing the chances of confusion. 

### Screenshot:
image

### Mockup:
https://mozilla.invisionapp.com/share/8TUW2XEGAF3#/393964297_Final-Proposal-Page

mmmavis:
Closes #3925 

- Bannered campaign page that has a sign up form in "main-content": https://foundation-mofostaging-pr-3997.herokuapp.com/en/campaigns/test-bannered-campaign/
- Page that doesn't have form in "main-content": https://foundation-mofostaging-pr-3997.herokuapp.com/en/about/


source/js/main.js
```diff
@@ -52,6 +53,8 @@ let main = {
       this.bindGlobalHandlers();
       this.bindGAEventTrackers();
 
+      utility.hideNavNewsletterButtonConditionally();
```
mmmavis: 🙈 Long function name. Any naming suggestion?

network-api/networkapi/templates/pages/base.html
```diff
@@ -65,7 +65,9 @@
       </div>
       {% block hero_guts %}{% endblock %}
 
+      <div class="main-content">
```
Pomax: Why a `<div>` instead of a `<main>`? Unless this isn't actually the main content, in which case we probably want to steer clear of using the word `main` in the class name.

source/js/main.js
Pomax: hm, `toggleNewsletterButtons()`?

source/js/utility.js
```diff
@@ -27,6 +27,18 @@ const Utility = {
     });
 
     return allFound ? elements : false;
+  },
+
+  /**
+   * Hide newsletter button on nav if the page's main content section has a sign up form already.
+   */
+  hideNavNewsletterButtonConditionally() {
```
Pomax: Did we look into whether we can do this at the Django level, rather than the JS level?

network-api/networkapi/templates/pages/base.html
mmmavis: Ah I asked in the [original ticket](https://github.com/mozilla/foundation.mozilla.org/issues/3925#issuecomment-557756135) about using a class on `<div>` vs `<main>` and it seemed like both [Youri](https://github.com/mozilla/foundation.mozilla.org/issues/3925#issuecomment-558228965) and [you](https://github.com/mozilla/foundation.mozilla.org/issues/3925#issuecomment-558881492) agreed that we should go with the former case?

Pomax: I recall the issue there being we had `<main>` somewhere else, and it would have conflicted, but in that case let's not call this `main-content` but just `content`, especially if the block itself uses that name too.

source/js/utility.js
mmmavis: I'll look into it.

mmmavis: With Django approach, I can only test if a page has a signup snippet associated with it to determine if the nav button should be rendered or not. For example, in the future, if dev adds  `<div class="join-us"></div>` somewhere in the template level, a signup form will get injected via React, and Django won't know. This probably isn't a huge issue but I'll ask in the original ticket to double check.

mmmavis:
@Pomax I updated this PR to use a Django-level approach. Let me know what you think!

(For ref: here's Alan's [reply](https://github.com/mozilla/foundation.mozilla.org/issues/3925#issuecomment-562639113) to my "JS vs Django fix" question)

PR #3998, opened 2019-12-03
Updating creep vote newsletter frequency + timing

### LINKED ISSUE #3943: Change Newsletter Signup Timing an Frequency
## Problem
Someone contacted Jen and said the email interstitial in the creep-o-meter was annoying them. Jen herself feels that the ask three times in a row is too much. After some discussion among the design and dev team we've decided to change the frequency of the interstitial. The ask will now only show once, occurring after the second time a user has voted on the creep-o-meter. 

## List of requirements
- [x] Change timing of the newsletter signup to after a user has voted twice
- [x] Change frequency of newsletter signup in creep-o-meter to only ~once~ twice

cc: @kristinashu @caltrider

youriwims:
Closes #3943


source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -52,9 +52,8 @@ export default class CreepVote extends React.Component {
         creepiness: creepinessId,
         confidence: confidence[0] > confidence[1] ? 0 : 1
       },
-      submitAttempted: false,
       subscribed,
-      nextView: subscribed ? "DidVote" : "SignUp",
+      showNewsletter: undefined,
```
mmmavis: Normally we don't assign `undefined` to a variable as it's already `underfined` by default. Let's set this to `false`.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -66,12 +65,21 @@ export default class CreepVote extends React.Component {
...
-        view = "DidVote";
+    const {
+      creepinessSubmitted,
+      confidenceSubmitted,
+      voteCount,
+      subscribed
+    } = this.state;
+
+    if (creepinessSubmitted && confidenceSubmitted) {
+      if (voteCount == 2 || 3) {
```
mmmavis: Not sure what this condition check is about. Are we checking if `voteCount` is equal to `2` or `3`? According to the originally ticket I thought we only needed to check it against `2`?

Anyways I think we can simply line 76 - 82 to

```jsx
this.setState({
  showNewsletter: voteCount === 2,
  didVote: true
});
```

And just a note to `if (voteCount == 2 || 3)` (for future usage)

1. always use `===` instead `==`
2. `voteCount == 2 || 3`: this actually gives you the number `3` back. If you wanna compare `voteCount` with the number 2 and 3. You should do 

```js
if (voteCount === 2 || voteCount === 3) { ... }
```

youriwims: Ah you're right, it is evaluating to 3. I'll fix that, but no, we want to make sure the newsletter [shows only after the 2nd and 3rd vote](https://github.com/mozilla/foundation.mozilla.org/issues/3943#issuecomment-560533493).

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
youriwims: Sorry about that, I meant to update it `false` 🙈

youriwims:
> Also I'm getting errors when I click on the email field.

Thanks for pointing that out—should be fixed now and ready for re-review.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
mmmavis: Ah. I def missed that comment. Thanks for clarifying.

Issue #4004, opened 2019-12-03
[blog] Blog Post Accessibility Testing


youriwims:
---
## Step 1️ : Automated Testing*

  
- [ ] Use `eslint-plugin-jsx-a11y` to fix a11y issues
   > Run  `npm test` and fix any issues related to the blog post template(s)
  

- [ ] Use `pa11y` to fix a11y issues
   > Run  `pa11y ./path/to/your/file.html`against file(s) to fix any issues related to the blog post template(s)
   > **OR**
   > Run `npm start` then `pa11y http://localhost:8000/blog/eample-blog-post-path` to fix any issues related to a blog post

## Step 2️ : Web Testing*

- [ ] Use [axe](https://www.deque.com/axe/axe-for-web/) browser extension to fix any remaining issues related to a blog post page
  > Available for Chrome + Firefox

## Step 3️: Manual Testing

- [ ] Use [Accessibility Insights for Web](https://accessibilityinsights.io/docs/en/web/overview) browser extension _Assessment_ feature to guide you through manual testing issues related to a blog post page
  > *Currently only Available in Chrome + Edge
  > *Assessments 2-23 (_as applicable_)

- [ ] Export + Upload Completed Assessments 
  > `insert blog post assessment here` 👈

  
- [ ] Follow up ticket for further action/discussion (_if applicable_)
  > `insert follow-up ticket here` 👈 

 
---
Related Issue: #3934 
_* Updates should cover the overarching key actions listed at the bottom of the [master a11y ticket](https://github.com/mozilla/foundation.mozilla.org/issues/3934)._

Pomax:
time passed this work by.

PR #4018, opened 2019-12-04
Blog page accessibility fixes

youriwims:
Related PRs/issues #3934 #4004 

Primarily based fixes on Firefox devtools a11y label issue checker since a11y cli + ci is ongoing.


source/js/components/join/join.jsx
```diff
@@ -327,14 +327,20 @@ export default class JoinUs extends React.Component {
...
           )}
-          <input
-            name="userEmail"
-            type="email"
-            className="form-control"
-            placeholder={getText(`Please enter your email`)}
-            ref={el => (this.email = el)}
-            onFocus={evt => this.onInputFocus(evt)}
-          />
+          <div
```
alanmoo: Why did you need to add a div here?

source/js/components/join/language-select.jsx
```diff
@@ -36,13 +36,19 @@ export default class LanguageSelect extends React.Component {
     let classes = classNames(`form-control`, this.props.className);
 
     return (
-      <select
-        value={this.props.selectedLang}
-        onChange={evt => this.handleChange(evt)}
-        className={classes}
+      <div
+        id={`userLanguage-${this.props.formPosition}`}
```
alanmoo: What's this prop about and where does it get passed in?

source/js/components/join/join.jsx
youriwims: Great question—so this `<input>` needs a label. I could've added `<label>`; however, our `<label>` have `display: inline-block` styling and I didn't want to have to do any style overrides to remove that without having a clear picture of what else that would effect. Simplest way to rememdy this without introducing any style changes was by using a `<div>` instead and adding `aria-label` attr. and `id`.

Screen Shot 2019-12-05 at 3 43 57 PM

source/js/components/join/language-select.jsx
youriwims: It gets passed in [here](https://github.com/mozilla/foundation.mozilla.org/pull/4018/files#diff-bcd043d2cebc25df2ecfd6167302e3c8R398); I added that prop to the `id` since multiple `JoinUs` components can appear on a page and don't want any `id` errors 😬 .

Screen Shot 2019-12-05 at 3 55 55 PM

source/js/components/join/join.jsx
alanmoo: But why couldn't you just add the "aria-label" directly to the `input` instead of the redirect through `aria-labeledby`?

youriwims: Because I've rarely seen that done 😛 But, I found a use-case for it [here](https://www.w3.org/TR/WCAG20-TECHS/ARIA14.html) so, I'll update the PR 👍

source/js/components/join/join.jsx
```diff
@@ -334,6 +334,8 @@ export default class JoinUs extends React.Component {
             placeholder={getText(`Please enter your email`)}
             ref={el => (this.email = el)}
             onFocus={evt => this.onInputFocus(evt)}
+            aria-label={!this.isFlowForm() ? "Email" : ""}
+            id={this.isFlowForm() ? "userEmail-flow" : ""}
```
alanmoo: A little weird but I think I get it.

PR #4025, opened 2019-12-06
update the devcontainer configuration for better dev experience in VS Code

cadecairos:
uses the docker-compose syntax so that we can provide some QoL overrides via docker-compose.extend.yml

- git installed
- ssh folder mounted in read-only state
- invoke installed
- configure line endings

@alanmoo I know you use VS Code, so can you take this for a spin if you have a chance?

### DIFF
--- a/.devcontainer/docker-compose.extend.yml
+++ b/.devcontainer/docker-compose.extend.yml
@@ -1,0 +1,14 @@
+version: '3'
+
+services:
+
+  watch-static-files:
+    security_opt:
+      - seccomp:unconfined
+
+  backend:
+    security_opt:
+      - seccomp:unconfined
+    volumes:
+      - ~/.ssh:/root/.ssh-localhost:ro
+    command: /bin/sh -c "while sleep 1000; do :; done"
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,7 +1,11 @@
 {
 	"name": "Mozilla Foundation CMS",
-	"dockerFile": "../dockerfiles/Dockerfile.python",
+	"dockerComposeFile": ["../docker-compose.yml", "./docker-compose.extend.yml"],
 	"context": "..",
+	"service": "backend",
+	"workspaceFolder": "/app",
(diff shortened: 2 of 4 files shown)


.devcontainer/docker-compose.extend.yml
```diff
@@ -0,0 +1,14 @@
...
+  watch-static-files:
+    security_opt:
+      - seccomp:unconfined
+
+  backend:
+    security_opt:
+      - seccomp:unconfined
+    volumes:
+      - ~/.ssh:/root/.ssh-localhost:ro
+    command: /bin/sh -c "while sleep 1000; do :; done"
```
alanmoo: ???

alanmoo:
I can't actually seem to get the code editor and the project up and running at the same time; if I open the folder in a VS Code container, I can't hit localhost:8000 (expected behavior), but then when I run `docker-compose up` in iTerm, it disconnects VSCode from the container.

.devcontainer/docker-compose.extend.yml
cadecairos: [According to the docs](https://code.visualstudio.com/docs/remote/containers#_using-an-image-or-dockerfile):

By default, when VS Code starts a container, it will override the container's default command to be `/bin/sh -c "while sleep 1000; do :; done"`. This is done because the container will stop if the default command fails or exits

I pulled this out of a sample docker-compose override in the advanced-containers setup guide.

cadecairos:
> I can't actually seem to get the code editor and the project up and running at the same time; if I open the folder in a VS Code container, I can't hit localhost:8000 (expected behavior), but then when I run `docker-compose up` in iTerm, it disconnects VSCode from the container.

So, in this setup, you'd actually just run `inv runserver` inside the VSCode terminal, so that the server process itself starts up in the container you're working out of, rather than restart the docker-compose stack in another terminal.

alanmoo:
`npm` doesn't seem to be installed in the container, and the `inv setup` command fails as a result.

cadecairos:
Nope

alanmoo:
😢

PR #4038, opened 2019-12-09
Wrap accessibility strings for l10n

youriwims:
For some reason, CI wouldn't complete after I created the app on #4035, so moving that work to this PR.

cc: @TheoChevalier


youriwims:
_Update_: #4035 finally rebuilt so, closing this PR.

PR #4044, opened 2019-12-10
pin django-admin-sortable

patjouk:
Related PR #4032
Pinning `django-admin-sortable`, as [recommended in their doc](https://github.com/alsoicode/django-admin-sortable#supported-django-versions):

> For Django 3 use the latest version
> For Django 1.8.x < 3.0, use 2.1.8.

Also update boto3 since it wasn't updated for at least 8 hours...


cadecairos:
> Also update boto3 since it wasn't updated for at least 8 hours...

😆

PR #4045, opened 2019-12-10
Make sure dev help page doesn't show up on prod

patjouk:
We added dyno metada on Heroku to give more context to errors caught by Sentry, which added `HEROKU_APP_NAME` on prod. Updating the if in the template to make sure that the dev help page doesn't show up on prod.

### DIFF
--- a/network-api/networkapi/context_processor.py
+++ b/network-api/networkapi/context_processor.py
@@ -3,7 +3,7 @@ from django.conf import settings
 
 # Used to export env variable to Django templates
 def review_app(request):
-    return {"REVIEW_APP": settings.REVIEW_APP}
+    return {'REVIEW_APP': settings.REVIEW_APP}
 
 
 # Used in BuyersGuide templates to check if we're using cloudinary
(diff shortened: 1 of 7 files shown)


network-api/networkapi/context_processor.py
```diff
@@ -3,7 +3,7 @@
 
 # Used to export env variable to Django templates
 def review_app(request):
-    return {'HEROKU_APP_NAME': settings.HEROKU_APP_NAME}
+    return {"REVIEW_APP": settings.REVIEW_APP}
```
cadecairos:
```suggestion
    return {'REVIEW_APP': settings.REVIEW_APP}
```

network-api/networkapi/settings.py
```diff
@@ -100,6 +101,9 @@
         release=HEROKU_RELEASE_VERSION
     )
 
+# At True when running on a review app
+REVIEW_APP = env("REVIEW_APP", default=False)
```
cadecairos:
```suggestion
REVIEW_APP = env('REVIEW_APP', default=False)
```

PR #4047, opened 2019-12-10
Configure the environment option with Sentry

cadecairos:
The Environment key in Sentry will let us better organize our projects on the Sentry interface. At the moment we have a project per environment, and with this change we can organize errors by environment in a single project.


webpack.config.js
```diff
@@ -36,7 +36,8 @@ let main = {
       __SENTRY_DSN__: JSON.stringify(process.env.SENTRY_DSN),
       __HEROKU_RELEASE_VERSION__: JSON.stringify(
         process.env.HEROKU_RELEASE_VERSION
-      )
+      ),
+      __SENTRY__ENVIRONMENT__: JSON.stringify(process.env.SENTRY_ENVIRONMENT)
```
patjouk:
```suggestion
      __SENTRY_ENVIRONMENT__: JSON.stringify(process.env.SENTRY_ENVIRONMENT)
```

Issue #4046, opened 2019-12-10
TypeError: '<' not supported between instances of 'datetime.datetime' and 'NoneType'


patjouk:
https://sentry.prod.mozaws.net/operations/foundation-site-prod/issues/6835288/

```
TypeError: '<' not supported between instances of 'datetime.datetime' and 'NoneType'
(3 additional frame(s) were not displayed)
...
  File "contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "wagtail/core/views.py", line 24, in serve
    return page.serve(request, *args, **kwargs)
  File "wagtail/core/models.py", line 721, in serve
    self.get_context(request, *args, **kwargs)
  File "networkapi/wagtailpages/models.py", line 902, in get_context
    context['related_posts'] = get_content_related_by_tag(self)
  File "networkapi/wagtailpages/utils.py", line 163, in get_content_related_by_tag
    reverse=True
```

Pomax:
Debug code landed so that instead of a server error 500, we get the error sent to Sentry while the page itself renders, just without running the sort call.

One thought is that this might be from finding a related post that's still in draft, without ever having been published. In that case, there might not be a `first_published_at` / `last_published_at` value for the sorting lambda to hook into.

Pomax:
This error has not shown up for a week, so this feels like it got fixed and if it does happen again, sentry will tell us.

Issue #3173, opened 2019-05-17
Add tagging to more page types


alanmoo:
Right now, Blog pages support tags. Let's add that functionality to Campaign, BanneredCampaign, and Opportunity pages as well.

This furthers the work to be done in #3362

kristinashu:
IHR team is asking for something related to this https://github.com/mozilla/foundation.mozilla.org/issues/3801... they have a series of 3 stories which we've posted as blog posts (ex https://foundation.mozilla.org/en/blog/how-smart-homes-could-be-wiser/). They also have a landing page introducing these stories https://foundation.mozilla.org/privacy-included/. They would like the Related Blog posts at the bottom of a post to show the 2 related stories and also the landing page. So ideally, we tag the 3 blog posts and the landing page with 'Privacy Included' and then 3 of those things show up a the bottom of those 3 blog posts.

alanmoo:
@Pomax You wanna take the lead on this? Let's figure out what the right solution is before building anything.

Pomax:
Can we clarify what the initial work here is to be? The original comment and what the IHR team would like to do sound like two fairly different issues - the original comment is about adding tags to more page  types, while the IHR team wants the related content block on blog pages changed to do something it currently doesn't do.

alanmoo:
IIRC, the idea with the IHR team is to leverage tagging with some unique tags to force specific items to appear on those pages.

This is a bit of a drive-by comment though, part of the work I'd like you to do is to actually dig in and figure out what they're looking for/how we can achieve it without overextending ourselves/writing in super specific logic for this page.

kristinashu:
Also related to https://github.com/mozilla/foundation.mozilla.org/issues/3362

Pomax:
Do we envision us wanting to do related content cross-content-type? E.g. a campaign with tag "X" showing related content that is a blog post that also uses tag "X"? Or do we want related content to be "the same type" so that campaigns only show related campaigns, blog posts only show related blog posts, etc?

Pomax:
I've filed a PR based on some investigatory work on adding tagging to bannered campaign pages, implementing what I think is a reasonable way to make all of this happen, but it might be good to see if this feels right, or whether it feels wrong, before I build it out further.

(It take the "only show related same-page-types" which might be too restrictive... for example, should bannered campaigns not show "regular" campaigns if we add tagging?)

alanmoo:
[...]

PR #4049, opened 2019-12-10
debug branch for #4046

Pomax:
Adds a try/except for us to add some Sentry logging.

See https://github.com/mozilla/foundation.mozilla.org/issues/4046
Related to https://github.com/mozilla/foundation.mozilla.org/issues/3173


Pomax:
@cadecairos can you add what feels like an appropriate Sentry log call in the `except`?

PR #4053, opened 2019-12-10
Remove hamburger menu lg-screen+

youriwims:
Related PRs/issues #4018

The hamburger code was refactored [here](https://github.com/mozilla/foundation.mozilla.org/pull/4018/files#diff-231f11884173d6383dd6469548e82a4dR37) and should've included logic for if `page.zen_nav` is undefined.

### DIFF
--- a/network-api/networkapi/templates/partials/primary_nav.html
+++ b/network-api/networkapi/templates/partials/primary_nav.html
@@ -34,7 +34,7 @@
                 <div class="d-flex align-items-center flex-wrap">
 
                   {% block burger_and_logo %}
-                    <button class="burger {% if page.zen_nav == False or page.zen_nav == null %} hidden-lg-up ml-md-0 {% endif %}" aria-label="{% trans 'Open menu' %}">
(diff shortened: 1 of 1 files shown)


network-api/networkapi/templates/partials/primary_nav.html
```diff
@@ -34,7 +34,7 @@
                 <div class="d-flex align-items-center flex-wrap">
 
                   {% block burger_and_logo %}
-                    <button class="burger {% if page.zen_nav == False %} hidden-lg-up ml-md-0 {% endif %}" aria-label="{% trans 'Open menu' %}">
+                    <button class="burger {% if page.zen_nav == False or page.zen_nav == null %} hidden-lg-up ml-md-0 {% endif %}" aria-label="{% trans 'Open menu' %}">
```
alanmoo: Can you check for "not True" in Python instead of the double condition?

youriwims: sure can

PR #4087, opened 2019-12-18
Source client side Sentry configuration from environment.json

cadecairos:
This will make pipeline promotions work once again, now that environment specific variables aren't built into the webpack build.


cadecairos:
> R+ with a little `if` check

`if` check where?

cadecairos:
@Pomax hey can you explain that ancient review comment? :)

Pomax:
I think it was about the fact that you only test for one value, then assign from three.

cadecairos:
I see what you mean. If the DSN was set, we should expect the other values as well, but if they're unset it doesn't break the initialization.

Issue #4104, opened 2020-01-07
[Bug on PNI staging] Some images are not showing


mmmavis:
**Describe the bug**
Some images on PNI staging aren't showing.

**To Reproduce**
1. Go to https://foundation.mofostaging.net/en/privacynotincluded/ and browse around

**Expected behavior**
All images should show.

**Screenshots**
<img src="https://user-images.githubusercontent.com/2896608/71862685-8959df80-30af-11ea-9ae8-1acd9f5eb40d.png" width=500>

**Additional context**
Review apps and production site don't have this issue.

mmmavis:
Screencap from Slack convo with @cadecairos 

<img width="606" alt="image" src="https://user-images.githubusercontent.com/2896608/71862921-2ae13100-30b0-11ea-9643-75c068998f33.png">

patjouk:
Looks like something changed and the images on prod are not uploaded to `foundationsite/buyersguide` anymore, but just in the home dir. The issue is that the auto mapping feature for staging is only activated for that directory. I'll check tomorrow what's happening but it's probably a change in the cloudinary python lib (can be reproduced on staging).

patjouk:
Transfer done: images are back on staging 🎉

mmmavis:
Awesome! Thanks @patjouk :D

PR #4126, opened 2020-01-13
Use CloudinaryField directly since we can now pass options

patjouk:
Related PRs/issues #4104 

This can be safely deployed to prod since we're not changing anything for images already in place. To fix the issue we have on staging, we will need to upload PNI prod images again, so that they go in the good directory instead of `home`, and wait or manually trigger the `copy_db` script. I'll take care of this.

I also need to remove the old `CloudinaryImageField` and it's going to be "fun" (squashing migrations is requiring manual work). This will happen as a separate PR.

## Checklist

**Changes in Models:**
~- [ ] Did I squash my migration?~ Will come as a separate PR because there's a bit of work to do to make that happen
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?


patjouk:
Opened a follow-up PR :)

Issue #4145, opened 2020-01-20
Replace PNI banner and OG graphic


kristinashu:
Design ticket https://github.com/mozilla/foundation.mozilla.org/issues/4012

Now that the holidays are over, we should update to the evergreen version.

[Graphics are here](https://drive.google.com/drive/folders/1RGBHgkb923dZo0rEs43EKs-SAT68Mc-r?usp=sharing).

- [x] Update homepage banner
- [x] Update OG graphic

cc @natalieworth

kristinashu:
@Pomax since Mavis is out all week, do you mind doing this?

PR #4172, opened 2020-02-04
update PNI graphics

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4145


kristinashu:
I checked the homepage graphic at https://foundation-s-update-pni-ikkejq.herokuapp.com/en/privacynotincluded/ and it looks good. 

Checked the OG graphic in slack and it also looks good!

image

Does do these look good to you @natalieworth?

PR #4176, opened 2020-02-06
Add signup CTAs to mozfest pages

### LINKED ISSUE #4173: MozFest: add newsletter signup snippet to page template
Based on https://github.com/mozilla/foundation.mozilla.org/issues/4130 

- [ ] Add an optional newsletter signup snippet to the right side of the mozfest site primary pages

The layout, design, and functionality should match that of the foundation site https://foundation.mozilla.org/en/mozfest-save-the-date/ (screen cap below).
image

cc @gideonthomas

Pomax:
test url: https://foundation-s-mozfest-pa-1epw7d.herokuapp.com/en/page-signup/

Closes #4173


Pomax:
WIP for https://github.com/mozilla/foundation.mozilla.org/issues/4173

Pomax:
@mmmavis the sticky cta doesn't seem to act sticky, would you happen to know what I'm missing in these changes? =x

network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
```diff
@@ -27,18 +31,16 @@
...
-      <div class="pb-5 {% block bootstrap_width %}col-lg-8 offset-lg-2{% endblock %}">
-        {% for block in page.body %}
-          {% if block.block_type == 'heading' %}
-            <h1>{{ block.value }}</h1>
-          {% else %}
-            {% include_block block %}
-          {% endif %}
-        {% endfor %}
-      </div>
+    <div class="row my-1 d-flex align-items-start">
```
mmmavis: @Pomax it's the `align-items-start` that's breaking the sticky CTA. We can actually remove both `d-flex` and `align-items-start`

Pomax: schweet

kristinashu:
This is working well both as someone using the CMS and also as a user filling out the form.

But I've noticed two styling issues
- The grid is narrower than on the foundation site [example](https://foundation.mozilla.org/en/mozfest-save-the-date/) but they should be the same
- The Privacy Policy text is lower than on the foundation site example but they should be the same (more inline with the checkbox)

Here's a screen cap of the foundation site example vs the PR.
Screen-Shot-2020-02-10-at-2 22 04-PM

Pomax:
oh, weird! yeah, let's fix that...

Pomax:
@kistrina should be fixed now.

image

kristinashu:
Nice, those two things look good! But the space between the main content and the snippet should be bigger like on the foundation site.

(Screen cap of the foundation site with a screen cap from PR below)
image

Also, the footer should be full width.

(Screen cap from PR)
image

Pomax:
ah, looks like a bootstrap number that's off by one, let me fix that too!

Pomax:
@kristinashu updated. Though, could I ask you use the review option to make sure to either approve or reject the PR? (even if you leave a detailed comment, without that signal on github the PR doesn't end up in my list of PRs that have received a review and so should be looked at again)

Pomax:
It looks like percy caught a valid problem, with the streamfield update also affecting the homepage. Investigating why that is at the moment...

Pomax:
Turns out there was a `</div>` missing, which was really messing with the alignments... as far as I can tell, that's now been fixed, with both the homepage and primary pages rendering at the correct widths again. App should be rebuilding atm

network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
```diff
@@ -27,18 +31,24 @@
 
 {% block content %}
   {% block secondary_nav %}{% endblock %}
+  <div class="container">
+    <div class="row my-1">
```
mmmavis: nitpicking: if we wanna match the layout we have on [this page](https://foundation.mozilla.org/en/mozfest-save-the-date/), then we should remove `my-1` from here as it adds some extra vertical margin.

network-api/networkapi/mozfest/templates/partials/signup.html
```diff
@@ -0,0 +1,22 @@
+<!--
```
mmmavis: If this is something that might be used in the future let's we add a comment here. If not, let's remove this chunk of code.

network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
Pomax:
```suggestion
    <div class="row">
```

Pomax:
@mavis hmm, I haven't touched any JS, let's see what's going on with that =S

Pomax:
So, it turns out this is caused by there not being a `data-form-position` attribute on the join-us HTML, but it turns out that there's only two places where that attribute is included in the HTML, and several places where it already wasn't. Like the mozfest footer. So this error has been in the code for _a while_ now, and we should fix it, but not part of this PR.

Pomax:
In it goes~

PR #4179, opened 2020-02-07
Add a redirect decorator to redirects to the main CMS site

patjouk:
We want PNI to only be accessible on the main CMS site (foundation.mozilla.org). I'm adding a small decorator that we can reuse if we need to build another django app in the future.

## Checklist
**Tests**


Pomax:
Is there a way to do this as part of urls.py, so that the PNI routes only work when accessed as part of the main site, without having to decorate each view?

patjouk:
I thought about this but I would prefer to keep the logic in the view like we did with the `enforce_en_locale` decorator.

cadecairos:
Yeah, decorating specific url patterns in urls.py is probably possible with a clever hack, but the Django way seem to revolve more around decorating the view methods.

network-api/networkapi/utility/redirects.py
```diff
@@ -0,0 +1,22 @@
+from django.shortcuts import redirect
+from wagtail.core.models import Site
+
+
+# Redirect to the same page, but using the default CMS site.
+# Ex: PNI should only be accessible on foundation.mozilla.org
+def redirect_to_default_cms_site_decorator(func):
```
cadecairos: Lets drop the trailing `decorator` in the method name.

network-api/networkapi/utility/redirects.py
```diff
@@ -0,0 +1,22 @@
...
+    def wrapper(request, *args, **kwargs):
+        wagtail_sites = Site.objects.values("hostname", "is_default_site")
```
cadecairos: Is there any upside or downside to caching this site record vs fetching it from the db on each request? Could we leverage an in memory/redis cache for this lookup?

cadecairos: Or maybe cache the whole result of sorting the sites we do just below this?

patjouk: Can't we rely on postgresql cache? https://devcenter.heroku.com/articles/understanding-postgres-data-caching

cadecairos: The first sentence of that article doesn't inspire much confidence 

> Data caching in Postgres cannot be preallocated or guaranteed

We already define a key/value cache in [settings.py](https://github.com/mozilla/foundation.mozilla.org/blob/d93517e86b09241051bb18d7e1edd25c42bf6f8e/network-api/networkapi/settings.py#L315-L338), which has example usage in the buyers guide views:

* [import the cache interface](https://github.com/mozilla/foundation.mozilla.org/blob/d93517e86b09241051bb18d7e1edd25c42bf6f8e/network-api/networkapi/settings.py#L315-L338)
* [Fetch a key](https://github.com/mozilla/foundation.mozilla.org/blob/d93517e86b09241051bb18d7e1edd25c42bf6f8e/network-api/networkapi/buyersguide/views.py#L90-L91)
* [Set a key if we hit the database](https://github.com/mozilla/foundation.mozilla.org/blob/d93517e86b09241051bb18d7e1edd25c42bf6f8e/network-api/networkapi/buyersguide/views.py#L99-L101)

[Django Cache interface docs](https://docs.djangoproject.com/en/2.2/topics/cache/)

patjouk: So I went and checked our current postgresql usage for foundation prod: our Table Cache Hit Rate is at 100.0%, which means that all requests can be served from cache (our DB usage is 124.7 MB and the standard 0 offers 4GB of ram). I'm confident we don't need extra optimization on top of what postgresql already provides.

PR #4181, opened 2020-02-10
React i18n with Fluent

TheoChevalier:
Third time’s the charm? :)

Proposal PR for fixing #3911, this one is based on https://projectfluent.org, developed by our MoCo colleagues. @Pomax what do you think of the overall solution?

### Relevant documentation:
https://github.com/projectfluent/fluent.js/wiki/React-Bindings
https://github.com/projectfluent/fluent.js/wiki/Localized
https://github.com/projectfluent/fluent.js/wiki/LocalizationProvider

### Handy comparisons:
[Fluent vs gettext](https://github.com/projectfluent/fluent/wiki/Fluent-vs-gettext)
[Fluent vs MessageFormat](https://github.com/projectfluent/fluent/wiki/Fluent-and-ICU-MessageFormat) (note, MessageFormat is not supported by Pontoon [here’s why](https://github.com/projectfluent/fluent/wiki/Fluent-and-ICU-MessageFormat))

### Notes:
- There’s this string I could not localize, because it’s outside the React app. https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/nav-newsletter.js#L36 But Stas mentioned there may be a way to work around this. We could also hide/show elements instead of editing the text here.
- We will need to load `.ftl` files differently in `l10n.js`, especially because l10n files will soon be moved to an external repo. Instead, we should scan /locale and import all .ftl files, if any.
- Unlike with multilingual formats, devs need to update the string ID when updating an existing string.


TheoChevalier:
Also CC @stasm in case I made obvious mistakes

Pomax:
Is this still a draft or is this ready for review?

source/js/components/join/join.jsx
```diff
@@ -53,7 +54,9 @@ export default class JoinUs extends React.Component {
       navWrapper.querySelector(".form-dismiss") &&
       this.state.apiSuccess
     ) {
-      navWrapper.querySelector(".form-dismiss").textContent = "Back to menu";
+      navWrapper.querySelector(".form-dismiss").textContent = (
+        <Localized id="back-to-menu">{`Back to menu`}</Localized>
```
stasm: I'll take a closer look tomorrow; for now I wanted to quickly ask about this: wouldn't this assign `[object Object]` to `textContent`? AFAIK, the return value of `React.createElement`, even for function components`, is a plain object representing the component instance?

TheoChevalier: I thought I tested that, but you’re absolutely right

TheoChevalier:
> Is this still a draft or is this ready for review?

It’s ready for you to have a look at the approach, but it’s not code complete. If we’re okay to move this way, I would first:

- Localize all the React strings for PNI (I can do it)
- Remove the current `getText` function and `locale-strings.jsx` (I can do it)
- Load .ftl files differently (will need help for that)

package.json
```diff
@@ -45,13 +45,18 @@
...
     "@babel/preset-react": "^7.7.4",
+    "@fluent/bundle": "^0.15.0",
+    "@fluent/langneg": "^0.3.0",
+    "@fluent/react": "^0.11.1",
     "@sentry/browser": "^5.10.2",
     "autoprefixer": "^9.7.3",
     "babel-core": "^6.26.3",
     "babel-loader": "^8.0.6",
     "classnames": "2.2.6",
+    "core-js": "^3.6.4",
```
Pomax: Do we know why this is a direct dependency?

package.json
```diff
@@ -63,6 +68,7 @@
     "react": "^16.12.0",
     "react-dom": "^16.12.0",
     "react-ga": "2.7.0",
+    "regenerator-runtime": "^0.13.3",
```
Pomax: 0.x version, but also: where did this come from?

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -244,16 +244,16 @@ export default class CreepVote extends React.Component {
...
-        formPosition="flow"
-        flowHeading={getText(`You Voted! You Rock!`)}
-        flowText={getText(
-          `Now that you’re on a roll, why not join Mozilla? We’re not creepy (we promise). We actually fight back against creepy. And we need more people like you.`
-        )}
-        csrfToken={this.props.joinUsCSRF}
-        apiUrl={this.props.joinUsApiUrl}
-        handleSignUp={successState => this.handleSignUp(successState)}
-      />
+      <Localized id="join-us" attrs={{ flowHeading: true, flowText: true }}>
```
Pomax: What is this `id` for?

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -270,7 +270,12 @@ export default class CreepVote extends React.Component {
         <div className="mb-5">
           <div className="col-12 text-center">
             <h3 className="h2-heading mb-1">
-              {this.state.totalVotes + 1} Votes — invite your friends!
+              <Localized
+                id="voted-header"
```
Pomax: what is this `id` for?

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -270,7 +270,12 @@ export default class CreepVote extends React.Component {
...
+                $voteCount={this.state.totalVotes + 1}
```
Pomax: This looks like the kind of non-standard magic I'd love to keep out of a codebase. Is there a way to achieve the same thing without relying on a completely non-standard method and syntax for variable declarations?

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -270,7 +270,12 @@ export default class CreepVote extends React.Component {
...
+              >
+                {`{$voteCount} Vote — invite your friends!`}
```
Pomax: especially this is _incredibly_ bad - template strings already use `${thing}` to get values into the string, if we also introduce `{$thing}` as valid syntax, we've just guaranteed that reviewers are going to miss errors around using one where the other should have been used.

source/js/components/join/join.jsx
Pomax: what is this `id` for?

source/js/components/join/join.jsx
```diff
@@ -449,11 +495,14 @@ export default class JoinUs extends React.Component {
...
                 required
               />
-              <p className="d-inline-block body-small form-text mb-0">
-                {getText(
-                  `I'm okay with Mozilla handling my info as explained in this Privacy Notice`
-                )}
-              </p>
+              <Localized
+                id="privacy-policy"
+                pp={<a href="https://www.mozilla.org/privacy/websites/"></a>}
```
Pomax: what is `pp`, is this an attribute we get to name, or something baked in? (if the former, let's not call this `pp` but use something indicative of what it is)

source/js/components/join/join.jsx
```diff
@@ -490,7 +543,7 @@ export default class JoinUs extends React.Component {
   renderFormContent() {
     if (this.state.apiSuccess) return null;
 
-    let formClass = `d-flex flex-column`;
+    let formClass = `d-flex flex-column join-form`;
```
Pomax: why do we need a new class here?

source/js/components/petition/petition.jsx
```diff
@@ -670,8 +688,12 @@ export default class Petition extends React.Component {
                   this.comment.element.value.length >=
                     SALESFORCE_COMMENT_LIMIT && (
                     <small className="form-check form-control-feedback">
-                      Comments cannot be longer than {SALESFORCE_COMMENT_LIMIT}{" "}
-                      characters
+                      <Localized
+                        id="comments-limit"
+                        $limit={SALESFORCE_COMMENT_LIMIT}
```
Pomax: see previous comment about this.

source/js/components/petition/petition.jsx
```diff
@@ -687,17 +709,19 @@ export default class Petition extends React.Component {
...
-                </span>
-                {this.state.userTriedSubmitting &&
-                  !this.refs.privacy.checked && (
-                    <small className="has-danger">
-                      {getText(`Please check this box if you want to proceed`)}
-                    </small>
-                  )}
+                <Localized
+                  id="privacy-policy"
+                  pp={<a href="https://www.mozilla.org/privacy/websites/"></a>}
```
Pomax: see previous comment about this

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -93,31 +100,51 @@ export default class ShareButtonGroup extends React.Component {
...
-      >
-        {label}
-      </a>
+    let link = this.state.linkCopied ? (
+      <Localized id="link-copied" attrs={{ title: true }}>
+        <a
+          class={classes}
+          href="#"
+          onClick={event => this.handleLinkButtonClick(event)}
+          title="Copied"
```
Pomax: If this is the only difference, why do we need to fully duplicate this block? Surely we can set this property to the correct locale string without the entire block wrapper.

source/js/l10n.js
```diff
@@ -0,0 +1,38 @@
...
+const MESSAGES_ALL = {
+  de: deMessages,
+  en: sourceMessages,
+  es: esMessages,
+  fr: frMessages,
+  pl: plMessages,
+  pt: ptMessages
+};
+
+export function* generateBundles() {
```
Pomax: what is the benefit of a generator here?

webpack.config.js
```diff
@@ -19,6 +19,14 @@ let rules = [
         [`@babel/preset-react`]
       ]
     }
+  },
+  {
+    test: /\.ftl$/,
+    use: [
+      {
+        loader: `raw-loader`
```
Pomax: is there no dedicated ftl loader that ensures that .ftl files are actually valid, rather than arbitrary string data? This should not silently pass if the ftl file is malformed...

package.json
```diff
@@ -45,13 +45,18 @@
     "@babel/core": "^7.8.0",
     "@babel/preset-env": "^7.8.3",
     "@babel/preset-react": "^7.7.4",
+    "@fluent/bundle": "^0.15.0",
```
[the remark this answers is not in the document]
stasm: Fluent is not yet ready to release 1.0 versions of our packages, but I'm committed to making our versioning scheme easy for developers to rely on. The "patch" version number in all `@fluent/` packages is only bumped when changes don't affect the public API. This follows the [behavior of npm](https://docs.npmjs.com/misc/semver#caret-ranges-123-025-004) for handling caret ranges:

> Allows changes that do not modify the left-most non-zero digit in the [major, minor, patch] tuple. In other words, this allows patch and minor updates for versions 1.0.0 and above, patch updates for versions 0.X >=0.1.0, and no updates for versions 0.0.X.

source/js/buyers-guide/bg-main.js
```diff
@@ -8,6 +8,8 @@ import navNewsletter from "../nav-newsletter.js";
 import CreepVote from "./components/creep-vote/creep-vote.jsx";
 import Creepometer from "./components/creepometer/creepometer.jsx";
 import JoinUs from "../components/join/join.jsx";
+import { LocalizationProvider } from "@fluent/react/compat";
```
stasm: Depending on your build setup, you might not need the `compat` build. It might be a good idea to try importing from `@fluent/react`.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
stasm: This is a Fluent-ism: props to `<Localized>` which start with `$` are passed into translations as arguments.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
stasm: The string here is intended as a fallback for localization. The `{$voteCount}` is Fluent syntax right there. Technically, the string is not required, because `@fluent/react` will insert the translation in its place anyways. But it helps when grepping the code for copy, and in the future we might offer tools to autoextract translations directly from JSX. There's one such tool written by the community: https://github.com/udacity/fluent-react-utils/tree/master/packages/extraction

source/js/components/join/join.jsx
stasm: This is a feature of Fluent: see https://github.com/projectfluent/fluent.js/wiki/React-Overlays.

source/js/l10n.js
stasm: This is a Fluent-ism, too: https://github.com/projectfluent/fluent.js/wiki/LocalizationProvider

It's not strictly required, and a simple array would work here as well. The benefit of using the generator is that it yields bundles lazily, so if all translations are found in the first bundle (the user's most preferred language), you won't pay the cost of parsing the translations in other languages.

package.json
TheoChevalier: [@babel/polyfill has been deprecated](https://babeljs.io/docs/en/babel-polyfill) in favor of `core-js` + `regenerator-runtime`.

If I remove them, I get a `regeneratorRuntime is not defined` error. Looking closely at the polyfill documentation, it may only be needed for the generator function in l10n.js? Let’s see it replacing it with a simple array allows us to drop these two dependencies

package.json
TheoChevalier: See previous comment

source/js/buyers-guide/bg-main.js
TheoChevalier: `"@fluent/react"` seem to work just fine, I thought this was needed to support older browsers

stasm: It all depends on your build setup and compatibility story. If you don't transpile `node_modules`, you might want to keep using `compat` here. Also, keep in mind that even `compat` targets browsers capable of supporting ES2017.

package.json
TheoChevalier: Confirmed, let’s remove them!

source/js/l10n.js
TheoChevalier: Thanks, replaced this with a array since it allows us to drop two dependencies and we’re not really doing language negotiation anyway, we just have to follow the Django locale. We’re only generating bundles for "en" + the current page locale, so there wasn’t much of a performance gain in the first place.

source/js/components/join/join.jsx
TheoChevalier: This is due to the changes in nav-newsletter.js, since we’re wrapping `<JoinUs {...props} />` with `<LocalizationProvider>`, we can’t just call `this.form.reset();` directly anymore.

https://github.com/mozilla/foundation.mozilla.org/pull/4181/files#diff-29b89fce3fc1f43c28b636b6a3e9d246R32

source/js/components/share-button-group/share-button-group.jsx
TheoChevalier: I’m not sure we can do this differently, unless we can store a translation in a variable. @stasm do you know?

webpack.config.js
TheoChevalier: I don’t think there is, but I also don’t think it matters, because unlike gettext, Fluent is [resilient and has a strong recovery logic](https://github.com/projectfluent/fluent/wiki/Fluent-vs-gettext) so things shouldn’t break if a file is malformed. In our case, Fluent would just fall back to the strings in the "en" ftl file. Right @stasm?

TheoChevalier: Also I think Pontoon cannot commit a malformed .ftl file

source/js/components/share-button-group/share-button-group.jsx
stasm: I guess you could do `<Localized id={this.state.linkCopied ? "link-copied" : "link-copy"}>` and then the same for the `title` attribute (or skip it in JSX as Fluent will add it anyways). It probably comes down to a personal preference.

webpack.config.js
stasm: Yes, that's correct.

source/js/l10n.js
```diff
@@ -0,0 +1,38 @@
...
+import { FluentBundle, FluentResource } from "@fluent/bundle/compat";
+import { negotiateLanguages } from "@fluent/langneg";
+import { getCurrentLanguage } from "./components/petition/locales";
+
+import deMessages from "../../network-api/locale/de/messages.ftl";
+import esMessages from "../../network-api/locale/es/messages.ftl";
+import frMessages from "../../network-api/locale/fr/messages.ftl";
+import plMessages from "../../network-api/locale/pl/messages.ftl";
+import ptMessages from "../../network-api/locale/pt_BR/messages.ftl";
+import sourceMessages from "../../network-api/locale/templates/messages.ftl";
```
stasm: So this code includes all translations in the final bundle, right? This might be fine for now since there aren't many locales nor messages to translate, but it might be a good idea to have a plan for when bundling all translations will increase the bundle size by too much. Then, you'll likely want to `fetch` the FTL files just for the current locale and for English (for fallback) on the runtime.

webpack.config.js
stasm: To expand on this: Fluent is designed such that a syntax error can only break the messages around it. Then, parsing recovers from the error and resumes at the next well-formed message.

source/js/l10n.js
TheoChevalier: We’re importing files for all locales, but only creating bundles for English and the current locale. Since the import is done at build time, we don’t know yet what the page locale will be.
`fetch` seem indeed to do what I was looking for. Let’s see if I manage to get something working :)

stasm: Right, sorry, I wasn't clean with my usage of the word `bundle`. In my previous comment I meant `bundle` as in `the JS file that has all your code and is sent to the users`, not the `FluentBundle` :)

source/js/components/share-button-group/share-button-group.jsx
TheoChevalier: Oh, that’s smarter :) Thanks, updated!

source/js/components/join/join.jsx
TheoChevalier: renamed this to `privacyLink` for better clarity

source/js/components/petition/petition.jsx
TheoChevalier: renamed this one to `privacyLink` too

Pomax:
@TheoChevalier @stasm I have some strong reservations about the current syntax that I'd like us to address before we move forward with adopting this approach.

The mean two are that `id` is already an HTML attribute and should not be overloaded (any HTML attribute supported by HTML-generating JSX should not be used for other purposes), and the `<Localized $thing="text">...{$thing}...` syntax is too close to standard JS string templating, which uses `${...}` rather than `{$...}` and this is guaranteed going to cause problems not just for contributors or new devs during onboarding, but also as persistent bug that someone will keep reading over and miss as having been done wrong.

Is it possible to change `id` to `localeId` (or something similar) and at the very least remove the dollar symbols from the inline-style variables so that it's "just more JSX templating"?

stasm:
> The mean two are that `id` is already an HTML attribute and should not be overloaded (any HTML attribute supported by HTML-generating JSX should not be used for other purposes), and the `<Localized $thing="text">...{$thing}...` syntax is too close to standard JS string templating, which uses `${...}` rather than `{$...}` and this is guaranteed going to cause problems not just for contributors or new devs during onboarding, but also as persistent bug that someone will keep reading over and miss as having been done wrong.

There are two things about `<Localized $thing="text">...{$thing}...` which I'd like to unpack:

- We're getting rid of `$`-prefix props to `Localized`. I'm almost done porting `@fluent/react` to TypeScript in https://github.com/projectfluent/fluent.js/pull/458, which made it clear that a more explicit approach would have been a better choice, e.g. `<Localized vars={{thing: "text"}}>` or `<Localized args={{thing: "text"}}>` (prop name TBD).

- `$thing` is how we write [variables in Fluent](https://projectfluent.org/fluent/guide/variables.html). There's no need to write `{$thing}` inside `<Localized>`, although experience shows that it's useful. Let me explain.

There are two modes of using `<Localized>`:

- Rendering the wrapped component, and possibly setting attributes on it, e.g. `<Localized id="foo"><p>abc</p></Localized>`.
- Rendering a string, e.g. `<Localized id="foo">abc</Localized>`.

In both of these modes, the contents (`abc`) will be completely _removed and replaced_ by the translation, unless the translation is missing, in which the contents will be used as a fallback. Absent any contents, we fall back to the `id`. I wouldn't recommend doing any JSX templating inside the contents, because it's just wasted cycles. `Localized` won't use it anyways—unless in an error scenario.

What I usually recommend is to put a simple string in the contents instead: `<Localized id="foo"><p>{"abc"}</p></Localized>`. This helps track down a piece of code responsible for something you see in the app (with tools like `grep`). It's not required, though, and you can also search through the `*.ftl` files first, and then grep the rest of the source for the id for the same effect.

If you do put a simple string as contents, and the messages contains [Fluent variables](https://projectfluent.org/fluent/guide/variables.html), then you end up with this (I'm using the new `vars` API from the unreleased version of `@fluent/react`):

```jsx
<Localized id="hello" vars={{userName: profile.user}}>
    {"Hello, {$userName}!"}
</Localized>
```

I understand that this can be confusing, since both JSX and Fluent use `{..}` for interpolation. A simple solution would be to decide to not put any contents inside `<Localized>`, but I see how this isn't an ideal solution either. Note that in the future we might have an official tool for extracting such contents from source into `*.ftl` files.

> Is it possible to change `id` to `localeId` (or something similar)

`<Localized>` doesn't produce any HTML of its own, so I'm not sure I agree with your reservation about the `id` overloading an existing HTML attribute. `<Localized>` runs its logic and "evaporates" from the the final rendered DOM. I don't think I'd like to rename the `id` to anything else at this point in `@fluent/react`. If you prefer `l10n-id` or something similar, however, that's understandable and easy to do with a wrapper on top of `<Localized>` :)

Pomax:
It's good to hear that we don't need that `$`, that effectively solves half my reservations.

However, the argument that <Localized> doesn't produce any HTML of its own is basically irrelevant, as devs won't know that unless and until they're already familiar with Fluent (which right now they're not, and onboarding contributors/contracts will make this a constant necessity). Unlike a HoC, it is impossible to tell whether `<Localized>` generates HTML or not, which makes using properties that already have meaning for HTML elements a guaranteed source of bugs, confusion, but likely both.

Pomax:
@TheoChevalier I'm worried about writing wrappers that may break in the future, but I also don't want to land things with `id=...` because I know our devs, and I know this is going to make their lives harder - do you want to see if you can come up with minimal wrapper code that solves the `id` issue for us?

TheoChevalier:
> @TheoChevalier I'm worried about writing wrappers that may break in the future, but I also don't want to land things with `id=...` because I know our devs, and I know this is going to make their lives harder - do you want to see if you can come up with minimal wrapper code that solves the `id` issue for us?

Sure, do you know any recommended documentation / code samples I can read to learn how to do that? Right now, I barely know what a wrapper is :)

source/js/components/join/join.jsx
```diff
@@ -262,23 +266,61 @@ export default class JoinUs extends React.Component {
...
+          id="cta-manage-subscriptions"
+          elems={{
+            preferencesLink: (
+              <a
+                href="https://www.mozilla.org/newsletter/recovery/"
+                target="_blank"
+              ></a>
+            )
+          }}
+        >
```
TheoChevalier, line 294: @stasm It turns out eslint is not a big fan of the new format, do you think there’s a workaround for that?
https://travis-ci.org/github/mozilla/foundation.mozilla.org/jobs/677437756#L295

stasm: I don't think that's related to the `@fluent/react` 0.12 update. The lint complains about the `<a>` element being empty; it's empty because it expects content to be inserted from the translation. This behavior hasn't changed in 0.12. You can put some fallback content inside the `<a>`.

Pomax:
@TheoChevalier sorry for not getting to this for a while again - in case case it would look something like this, as `source/js/components/localized.js`:

```
import { Localized } from "@fluent/react";

const Customized extends Localized {
  constructor(props) {
    if (props.stringId) {
      const { stringId, ...newprops } = props;
      newprops.id = stringId;
    }
    super(newprops);
  }
}

export { Customized as Localized };
```

and then everywhere else, rather than `import { Localized } from "@fluent/react";` we'd use `import { Localized } from "source/js/components/localized";`, and then we can use the same code as before but with the `id` attribute changed to `stringId` instead:

```
...
    let howLikely = (
      <Localized stringId="how-likely-buy">{`How likely are you to buy it?`}</Localized>
    );
...
```

TheoChevalier:
@Pomax Nice! I was just looking into this after a relatively heavy rebase, and was totally failing… until you saved the day :)
Going to play with this!

TheoChevalier:
@Pomax I couldn’t make that code run, so I tried to make some changes since it had to be a class somewhere, because we’re defining a constructor. But even though, it keeps crashing with `Uncaught TypeError: Super expression must either be null or a function`

I’ll try to have another look tomorrow

```js
import { Localized } from "@fluent/react";

const localizedHOC = (Localized) => {
  class Customized extends Localized {
    constructor(props) {
      if (props.stringId) {
        const { stringId, ...newprops } = props;
        newprops.id = stringId;
        super(newprops);
      } else {
        super(props);
      }
    }
  }
};

export { localizedHOC as Localized };

```

Pomax:
my bad, I accidentally wrote `const` where it should have been `class`.

TheoChevalier:
@Pomax I tried that too, but I’m getting `Uncaught TypeError: Cannot call a class as a function`
Don’t we need to instantiate that class somewhere?

Pomax:
That sounds like it's getting invoked incorrectly now, do you have that code in your branch? If so, I can have a look at it.

TheoChevalier:
Thanks @Pomax, just pushed

TheoChevalier:
@Pomax That works wonderfully, thank you!

There’s just one lint error from your code, that I can’t figure out what it wants us to change, but otherwise the custom function works like a charm. ✨

```
/home/travis/build/mozilla/foundation.mozilla.org/source/js/components/localized.js
 5:23  error  Parsing error: Unexpected token ..
```
Which happens here: 
```js
const { stringId, ...newprops } = props;
```


I’ve updated all the code to use the new Localized component, and performed the usual rebase.

I think the only thing we need now are stuff in https://github.com/mozilla/foundation.mozilla.org/issues/4297 
Stas recently updated the examples in Fluent-react, and I think we can leverage this function, as long as we can convert that from TS https://github.com/projectfluent/fluent.js/blob/master/fluent-react/example/src/l10n.tsx#L16

TheoChevalier:
> home/travis/build/mozilla/foundation.mozilla.org/source/js/components/localized.js
5:23  error  Parsing error: Unexpected token ..


I believe the linting issue on the spread syntax is related to this https://stackoverflow.com/questions/51979699/eslint-parsing-error-unexpected-token-on-only-one-rest-operator

> The spread syntax for object literals is part of the language since ES8. It is separate from the spread syntax for array literals and function calls, which was added in ES6.

Pomax:
filed https://github.com/mozilla/foundation.mozilla.org/pull/5229, which should unblock this PR

Pomax:
I'm going to close this PR, given our new localization direction: if we think this will be worth doing still, let's file new work to tackle this once the wagtail localization work has wrapped up.

PR #4188, opened 2020-02-11
Localize petition loading string

TheoChevalier:
Very simple patch to localize a string + fix copy


network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta/petition.html
```diff
@@ -33,7 +35,7 @@
         data-share-text="{{ cta.share_link_text }}"
         data-thank-you="{{ cta.thank_you }}"
         data-modals="{{ modals_json }}"
-	>loading CTA</div>
+	>{% trans "Loading…" %}</div>
```
Pomax: Hm, let's still mention what is actually loading?

TheoChevalier: What about `Loading form…`? Despite the file name, this isn’t always used as an actual petition, and I think `CTA` is super cryptic

Pomax: it is, it's a pretty horrible acronym. "Loading form..." would be good.

network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta/petition.html
```diff
@@ -33,7 +35,7 @@
         data-share-text="{{ cta.share_link_text }}"
         data-thank-you="{{ cta.thank_you }}"
         data-modals="{{ modals_json }}"
-	>loading CTA</div>
+	>{% trans "Loading form…" %}</div>
```
Pomax: can we make this the literal `...` rather than the single unicode glyph `…`?

TheoChevalier: I’m curious why we would want to do that, but I’ll have to insist we use the Unicode character. This is a standardized practice across Mozilla projects, and part of our [l10n best practices](https://developer.mozilla.org/en-US/docs/Mozilla/Localization/Localization_content_best_practices)

`U+2026 (\u2026) is the horizontal ellipsis character. This character should be used in place of three consecutive periods.`

Pomax: if this is a moz-wide policy then that's fine. It's just a !@$# to actually type on "not macos" =P

Issue #4143, opened 2020-01-17
[Header/body newsletter list conflicts] Implementation ticket: updated Thank You page copy


taisdesouzalessa:
Implementation ticket for #3928

**[Mockup here](https://mozilla.invisionapp.com/share/8TUW2XEGAF3#/401284807_Screenshot_2020-01-17_Open_Leadership_Events-4-)**

**Copy:**

Thanks!

One more thing. To protect you from unwanted emails, we ask you to confirm your opt-in to our email subscriptions. 

If you haven’t previously confirmed your opt-in to a Mozilla-related email subscription you may have to do so now. **Please check your inbox or spam filter for an email from us to click and confirm your subscription.** 

If you have already confirmed your opt-in to receive Mozilla-related emails, you can now [manage your subscriptions](https://www.mozilla.org/en-US/newsletter/recovery/) and update your email preferences.

**Update:**
We removed the first sentence:  "One more thing. To protect you from unwanted emails, we ask you to confirm your opt-in to our email subscriptions." to: 1. make the copy shorter so it adjusts to the required space and 2. it is easier for users to scan.

mmmavis:
@taisdesouzalessa Can you link me to the Thank You page this ticket refers to?

taisdesouzalessa:
@mmmavis this should apply to all "thank you" pages that ask for newsletter sign up: header of the site, footer of the site and all pages that have our newsletter widget such as "https://foundation.mozilla.org/en/initiatives/open-leadership-events/" - please let me know if you run into any issues as implementing this copy since it is a longer copy than the one we currently have.

mmmavis:
This wasn't the troublesome ticket I thought it was. Changing estimate from 3 to 1.

mmmavis:
PR has been merged!

PR #4192, opened 2020-02-11
Update default signup form description & thank you copy

### LINKED ISSUE #4142: [Header/body newsletter list conflicts] Implementation ticket: updated footer copy
Implementation ticket for #3927

**Screenshot (mockup):**
image

**Copy:**
Join our **Mozilla News** email list to stay informed and take action.

mmmavis:
Closes #4142 and #4143

Pretty straight-forward fixes. I think those weren't the troublesome tickets I thought they were 😂 

Note: @taisdesouzalessa and I worked together to tweak 'thank you' copy and design (removed the first paragraph to make the total copy shorter.

https://foundation-s-issue-4142-juuwq9.herokuapp.com/en/campaigns/late-possible-record-unit-season-organization/


source/js/components/join/join.jsx
```diff
@@ -536,8 +557,7 @@ export default class JoinUs extends React.Component {
 
 JoinUs.defaultProps = {
   ctaHeader: `Protect the internet as a global public resource`,
-  ctaDescription: `Join our email list to take action and stay updated!`,
-  thankYouMessage: `If you haven’t previously confirmed a subscription to a Mozilla-related newsletter you may have to do so. <strong>Please check your inbox or your spam filter for an email from us.</strong>`,
```
mmmavis: Removed as none of the signup form has its own customized thank you message (e.g., none of the `.join-us` has `data-thank-...` attribute)

taisdesouzalessa:
@mmmavis the new Thank You message works great on:
- the footer
- the side form
in both desktop and mobile.

We have an issue for the one that appears after signing up for newsletter using the top nav button: the message is cut off. 
image

mmmavis:
PR is ready for re-review!

source/js/components/join/join.jsx
```diff
@@ -265,17 +267,34 @@ export default class JoinUs extends React.Component {
...
             }}
           />
         ) : (
-          <p
-            dangerouslySetInnerHTML={{
-              __html: this.props.thankYouMessage
-            }}
-          />
+          <React.Fragment>
+            <p>
```
Pomax: We'll need to wrap this in `getText` so that it can be localised.

source/js/components/join/join.jsx
```diff
@@ -265,17 +267,34 @@ export default class JoinUs extends React.Component {
...
+              If you haven’t previously confirmed your opt-in to a
+              Mozilla-related email subscription you may have to do so now.{" "}
```
Pomax: Does `&nbsp;` work rather than `{" "}`? If so, that'll make future localization a lot easier.

mmmavis: Moved the strings to `locale-strings.jsx` and it seems like strings there are using `{" "}` already so I'm gonna keep the new strings that way.

source/js/components/petition/locale-strings.jsx
```diff
@@ -122,7 +188,29 @@ export default {
     "Please check this box if you want to proceed": `Veuillez cocher cette case si vous désirez poursuivre`,
     "Yes, I want to receive email updates about Mozilla's campaigns.": `J’accepte de recevoir des informations par courriel au sujet des campagnes de Mozilla.`,
     "Add my name": `Ajouter mon nom`,
-    "Sign up": `Je m’inscris`
+    "Sign up": `Je m’inscris`,
+    "If you haven’t previously confirmed your opt-in to a Mozilla-related email subscription you may have to do so now.": (
```
mmmavis: @Pomax do you know where we can get the actual localized strings from? Do we have a system in place that auto updates strings here?

Pomax: No, these were all temporary work that we had localisers do for us because it was far less work to do the set of at-the-time fixed strings than it was to figure out a true localisation strategy. The work to do this properly is currently being investigated by @TheoChevalier

Pomax:
In principle but we should get those strings translated before we land this. I don't think it's wise to block this on the current work around adding proper localisation for React content, but we'll definitely need @TheoChevalier's help in making sure those few new strings have the correct localised versions set.

Pomax:
@TheoChevalier what would be required here to make sure we capture the string changes in this PR while we figure out how to do React content localization properly?

TheoChevalier:
> @TheoChevalier what would be required here to make sure we capture the string changes in this PR while we figure out how to do React content localization properly?

Not much from your side :) I’ve just exposed the strings to Pontoon, and I’m going to update this PR with the translations as they come in https://github.com/mozilla-l10n/engagement-l10n/commit/c718b5cb7be60c5f535a457b6c6cae1edb77633d

Let’s check back the translation coverage on Monday and decide if we’re good to go with that? Then I can open a follow-up PR if we get other translations later. Would that be okay?

source/js/components/join/join.jsx
```diff
@@ -261,17 +263,21 @@ export default class JoinUs extends React.Component {
...
         ) : (
-          <p
-            dangerouslySetInnerHTML={{
-              __html: this.props.thankYouMessage
-            }}
-          />
+          <React.Fragment>
+            <p>
+              {getText(`If you haven’t previously confirmed your opt-in to a
+              Mozilla-related email subscription you may have to do so now.`)}
```
TheoChevalier: This doesn’t work, because the leading spaces are considered part of the key here. So that string can’t be found.
Let’s probably just use a shorter key so that it fits in one line?

TheoChevalier:
Alright, we already have de, fr, pl, pt. We can merge this anytime now, with the key/identifier change I mentioned in my review.
If we ever get Spanish, I’ll make a follow up PR.

source/js/components/join/join.jsx
mmmavis: Good point! Will update!

mmmavis:
@Pomax Please see my [last commit](https://github.com/mozilla/foundation.mozilla.org/pull/4192/commits/f3d3a6b64175d2deffaec1dbd953f7c2291a32a1). I've updated `CTAFactory` so that its `description` is more realistic (2 paragraphs instead of 6) and it's up to date with the DOM tag changes that I introduced in this PR. I've also updated `petition.jsx`  so it matches what we have for the signup component.

mmmavis:
This _only_ took 2 weeks to land 😂 😂 😂 

🎉  Merging now! Thanks, all!

PR #4199, opened 2020-02-13
Add helper command to extract strings to an external repo

TheoChevalier:
This is related to https://github.com/MozillaFoundation/mofo-devops/issues/735

Once the new workflow is live, translation files will live in an external repo. The string extraction will still happen in this repo, so this PR adds a script facilitating the string extraction.

It knows where to find the external repo via an env var. Mine looks like `LOCAL_PATH_TO_L10N_REPO=/Users/theochevalier/Documents/GitHub/fomo-l10n/`

What the script does:

- Import the latest translation files from your local copy of the `fomo-l10n` repo to the right places in your local `foundation.mozilla.org` repo
- Run `makemessages` to update all .po files
- Export the updated .po files to the `fomo-l10n` repo

Then if everything looks right, I can easily commit & push the updates to `fomo-l10n`


TheoChevalier:
Thanks! Fixed the tests, and added some doc in the readme

PR #4203, opened 2020-02-14
Localize MozFest homepage

TheoChevalier:
@Pomax Let’s chat about this one when you’re around


Pomax:
Sure thing, what are the concerns in this PR?

Pomax:
R+, but this needs a rebase and then get that translation migration rebuilt (e.g. delete the migration this introduced, rebase on master, then rerun `makemigrations`), because a different migration got landed for the mozfest app since this PR got opened.

PR #4224, opened 2020-02-19
Update post_compile step to use the translations archive instead of compilemessages

patjouk:
Replace running `compilemessages` by downloading an archive from S3 with already processed translations.

Related PRs/issues MozillaFoundation/mofo-devops#735


patjouk:
Closed in favor of #4199

Issue #4230, opened 2020-02-19
Newsletter signup pulldown menu is white text on white background.


Pomax:
![image](https://user-images.githubusercontent.com/177243/74866950-9f8ec880-5308-11ea-9e3c-27c2148b9f0b.png)

It looks like this affects both drop downs, so we probably just need a `.dark-theme select.form-control.form-control option { color: $black; }`

Pomax:
This turns out to only be applicable to Chrome, so the solution here is more than just a one liner SASS fix... we'll have to explicitly set a style for the dropdown that works in FF, Chrome, and Safari/Edge.

Do we have style guidance for what dropdowns should be colored as? /cc @sabrinang @kristinashu

sabrinang:
I've bumped into styling issues on a previous footer drop down ticket so we've decided to use the browser default styling for the footer drop down so we don't bump into issues across browsers or on mobile: https://github.com/mozilla/foundation.mozilla.org/issues/2681#issuecomment-528610094  so similarly I'd advise this newsletter drop down to also be using the default to use no styling/colour.

It has been appearing like the default for me, on Chrome: 
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/2492510/74886486-2c4b7d80-532d-11ea-959f-04f9b2b19b48.png">

as well as Firefox:
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/2492510/74886557-513ff080-532d-11ea-9dfa-d723ead75d01.png">

Pomax:
Hm, this might be Chrome on Windows, then. Will get someone else on windows to confirm.

mmmavis:
I was able to reproduce the issue using BrowserStack 😬

PR #4231, opened 2020-02-19
fix white-on-white for signup pulldowns

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4230


Pomax:
This turns out to only be applicable to Chrome, so the solution here is more than just a one liner SASS fix... we'll have to explicitly set a style for the dropdown that works in FF, Chrome, and Safari/Edge.

PR #4250, opened 2020-02-24
turn off devtool sourcemapping with eval!

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4245by turning off the frckn Webpack "devtool" option explicitly. It's silently generating `eval` based source maps.

See https://webpack.js.org/configuration/devtool/ for what this value does. It's _supposed_ to do nothing by default, but that is absolutely not what it's doing.


Pomax:
alright, let me add in the axe-core and react-axe version bumps, too.

mmmavis:
@Pomax it seems to work if you do `devtool: ""` instead of `devtool: "none"`.

Pomax:
kk, updated, with the bumps for the axe libraries too (since there was talk on the axe repos that some older versions had eval as well...)

PR #4259, opened 2020-02-25
Do not strip off query params for newsletter signup source

### LINKED ISSUE #4102: Newsletter page doesn't pass on tracking information to SFMC
**Describe the bug**
When you try to sign up for the newsletter through https://foundation.mozilla.org/newsletter/, if you landed on it with a query string including utm codes (like https://foundation.mozilla.org/newsletter/?src=acqtestdec2019&utm_source=acqtestdec2019&utm_medium=cpc&utm_campaign=pro&utm_content=daily-kos), none of the querystring is passed on to Salesforce. We should pass this on so that we can properly report in SF.

Not sure if this is also the case with the footer newsletter signup, etc.

**To Reproduce**
Steps to reproduce the behavior (any browser):
1. Go to https://foundation.mozilla.org/en/newsletter/?utm_source=gideon&utm_medium=gideon&utm_campaign=gideon&utm_content=gideon and sign up for the newsletter.
2. Check your customer record's Signup Source URL in SF. It will only show https://foundation.mozilla.org/en/newsletter/ instead of the whole url.

**Expected behavior**
The entire landing page url should be recorded in SalesForce when signing up for the newsletter.

cc @stephaniemcv who can help with verifying + @WillatMozFdn who first discovered this with me.

mmmavis:
Closes #4102

### DIFF
--- a/source/js/components/join/join.jsx
+++ b/source/js/components/join/join.jsx
@@ -118,8 +118,8 @@ export default class JoinUs extends React.Component {
     return new Promise((resolve, reject) => {
       let payload = {
         email: this.email.value,
-        /* Strip query params in source url for newsletter signups: https://github.com/mozilla/foundation.mozilla.org/issues/2994#issuecomment-516997473 */
-        source: window.location.href.split(`?`)[0]
+        /* keeping query params in source url for newsletter signups: https://github.com/mozilla/foundation.mozilla.org/issues/4102#issuecomment-590973606 */
+        source: window.location.href
       };
 
       if (this.givenNames) {


source/js/components/join/join.jsx
```diff
@@ -118,8 +118,8 @@ export default class JoinUs extends React.Component {
     return new Promise((resolve, reject) => {
       let payload = {
         email: this.email.value,
-        /* Strip query params in source url for newsletter signups: https://github.com/mozilla/foundation.mozilla.org/issues/2994#issuecomment-516997473 */
-        source: window.location.href.split(`?`)[0]
+        /* keeping query params in source url for newsletter signups: https://github.com/mozilla/foundation.mozilla.org/issues/4102#issuecomment-590973606 */
```
Pomax: lol

mmmavis: 😂

PR #4262, opened 2020-02-27
Remove Array.from where possible

### LINKED ISSUE #3109: Clean up "Array.from" in codebase
From https://github.com/mozilla/foundation.mozilla.org/pull/3023#discussion_r281295700

> It turns out that `querySelectorAll` returns a static NodeList, so we don't need the `Array.from` anymore.

Pomax:
Port of https://github.com/mozilla/foundation.mozilla.org/pull/3710 onto the current codebase

Closes #3109


source/js/main.js
```diff
@@ -493,19 +483,22 @@ let main = {
...
-      // Refetch the profile cards, because they'll have gone stale.
-      profileCards = document.querySelectorAll(`.profiles .person-card`);
-      bindProfileCardAnalytics(profileCards);
-    });
+
+    document.addEventListener(`profiles:list-updated`, () =>
+      updateProfileList()
+    );
+
+    updateProfileList();
```
mmmavis: I simplied this chunk of code further + added back the inline comments
```suggestion
function updateProfileList() {
  let profileCards = document.querySelectorAll(`.profiles .person-card`);
  bindProfileCardAnalytics(profileCards);
}

// Checks for profile cards in the initial page load
updateProfileList();

// And start listening for profile filter events,
// in case profile cards get updated.
document.addEventListener(`profiles:list-updated`, () =>
  updateProfileList()
);
```

Pomax: sweet

Issue #4284, opened 2020-03-03
Update Site / Page Title of MozFest Website


marcwalsh:
Currently all page titles start with `Mozilla Festival 2019 -`

As part of the city announcement updates are we able to remove 2019 so page titles start with just `Mozilla Festival -`

kristinashu:
@mmmavis so just to clarify, we don't want to mention a year.

imskr:
Can I work on this?

mmmavis:
Hi @imskr , thanks for offering help 😃 . This ticket is part of the bulk tasks I'm currently working on and I have a fix underway already. Our dev team is quite small now so we don't have the capacity to help contributors get started with the project. If you are familiar with this codebase and would like to contribute your awesome code, I encourage you to look for tickets that have no assignees and are labelled with "enhancement".

mmmavis:
PR merged. Fix will be live today.

Issue #4270, opened 2020-03-02
MozFest - make signup form's button sticky on mobile


mmmavis:
Follow what we did for the petition form. Example: https://foundation.mozilla.org/en/campaigns/tell-amazon-require-two-factor-authentication-all-ring-accounts/

mmmavis:
PR merged. Fix will be live today.

Issue #4249, opened 2020-02-24
MozFest: make the nav CTA button an optional field


kristinashu:
Related to https://github.com/mozilla/foundation.mozilla.org/issues/4243

Currently the CTA button in the main nav has two required fields in the CMS. The mozfest team wouldn't have a CTA for a little while. So please make those fields option.

Screen cap from prod
<img width="894" alt="image" src="https://user-images.githubusercontent.com/1068783/75201093-61384580-571c-11ea-971b-4aead8d5a78b.png">

mmmavis:
PR merged. Fix will be live today.

PR #4286, opened 2020-03-03
Various MozFest updates

mmmavis:
(will assign people for code review after this passes design review)

Closes

- [x] Add sticky 'SIGN UP' button for MozFest homepage's mobile view #4270 (Should behaviour like what we have on campaign pages. [Example](https://foundation.mozilla.org/en/campaigns/amazon-echo-kids/))
- [x] Remove hardcoded 'year' from `<title>` #4284
- [x] Make the nav CTA button an optional field #4249

---

- CMS Edit page: https://foundation-s-various-mo-hkrw2u.herokuapp.com/cms/pages/53/edit/
- Review app page: https://foundation-s-various-mo-hkrw2u.herokuapp.com/en/mozilla-festival/


mmmavis:
@beccaklam thanaks for the note.

- I will address the spacing issue as a follow-up
- Sticky button issue will have to be a follow-up as well since it will be an enhancement to the existing behaviour (the sticky button script is used on both Campaign pages and MozFes homepage)
- > Oh and is this bug (where content is flowing into the snippet portion of the page) captured in another ticket?

Yup, this is covered in another PR. This PR only covers the 3 items I put in the PR description 😉

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
```diff
@@ -39,7 +39,9 @@
                   </div>
                 </div>
               </div>
+              {% if menu_root.specific.cta_button_destination and menu_root.specific.cta_button_label %}
```
Pomax: this might be better to turn into a fragment then, so we can say `{% include "fragments/primary_cta.html" with root=menu_root.specific %}` and then we put these three lines in that fragment so that it does nothing if there's no destination/label

network-api/networkapi/mozfest/templates/partials/signup.html
```diff
@@ -1,4 +1,9 @@
+{% load i18n %}
+
 <div class="cta col-lg-4 offset-lg-1">
+  <div class="narrow-sticky-button-container hidden">
+    <a href="#cta-anchor" class="btn btn-primary d-block">{% trans "SIGN UP" context "Sticky button in mobile view on MozFest homepage" %}</a>
```
Pomax: is this a real href?

mmmavis: Oh yea we want the button to scroll to the `<div id="cta-anchor">` node when user clicks it.

network-api/networkapi/mozfest/templates/partials/primary_mozfest_nav.html
mmmavis: Love this suggestion!

mmmavis: Done!

Issue #4288, opened 2020-03-03
Move IndexPage into its own file


Pomax:
As part of the work for https://github.com/mozilla/foundation.mozilla.org/issues/4232 we should move the `IndexPage` model into its own file so that we're not further polluting `models.py` with multiple new models.

PR #4289, opened 2020-03-03
Moved IndexPage into its own file

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4288

`inv docker-test-python` should pass as usual.


network-api/networkapi/wagtailpages/models.py
```diff
@@ -82,58 +66,6 @@ def titlecase(s):
 ] if field is not None]
 
 
-# Override the MetadataPageMixin to allow for a default
-# description and image in page metadata for all Pages on the site
-class FoundationMetadataPageMixin(MetadataPageMixin):
```
Pomax: moved to `wagtailpages/pagemodels/mixin/foundation_metadata.py`

network-api/networkapi/wagtailpages/models.py
```diff
@@ -530,282 +462,6 @@ def get_context(self, request):
         return get_page_tree_information(self, context)
 
 
-class IndexPage(FoundationMetadataPageMixin, RoutablePageMixin, Page):
```
Pomax: moved to `wagtailpages/pagemodels/index_page.py`

network-api/networkapi/wagtailpages/models.py
```diff
@@ -39,17 +34,6 @@
 from .donation_modal import DonationModals  # noqa: F401
 
 
-# See https://docs.python.org/3.7/library/stdtypes.html#str.title
```
Pomax: moved into `wagtailpages/utils.py`

network-api/networkapi/wagtailpages/pagemodels/index_page.py
```diff
@@ -0,0 +1,300 @@
...
+        return IndexPage.serve(self, request, *args, **kwargs)
+
+    """
+    category routes
+    """
+
+    # helper function to resolve category slugs to actual objects
+    def get_category_object_for_slug(self, category_slug):
+        # FIXME: this should not need to be an app consult
+        BlogPageCategory = apps.get_model('networkapi.wagtailpages', 'BlogPageCategory')
```
Pomax: This is the only "new" code, because we can't import `BlogPageCategory` from `wagtailpages.models` (that would create a circular import dependency)

network-api/networkapi/wagtailpages/pagemodels/index_page.py
```diff
@@ -0,0 +1,300 @@
...
+        return IndexPage.serve(self, request, *args, **kwargs)
+
+    """
+    category routes
+    """
+
+    # helper function to resolve category slugs to actual objects
+    def get_category_object_for_slug(self, category_slug):
+        # FIXME: this should not need to be an app consult
+        BlogPageCategory = apps.get_model('wagtailpages', 'BlogPageCategory')
```
Pomax: This is the only line of "new" code, because we can't have an `from ..models import BlogPageCategory` at the top of this file, since this file gets imported _by_ `wagtailpages.models`, so that would lead to a circular import dependency. Instead we get that model "whenever this def gets called" by asking the Django app for it by name.

Pomax:
Hm, I had a closer look at how that code works, and it feels like it's better to keep the import from wagtailpages.models rather than its refactored location, since the mozfest app "extends" wagtailpages, and shouldn't need to know anything about the directory structure inside that app. It should be able to trust that it can import from `models` and `utils` without needing to know anything about specifics.

(so if the files move around inside wagtailpages, keeping it the way it is means that mozfest files won't need to be updated)

PR #4293, opened 2020-03-04
Enable the cached template loader when DEBUG=False

### LINKED ISSUE #4292: Implement template caching
I'd like us to attempt to implent cached template loaders in our staging and production to try and improve page load times.

https://docs.wagtail.io/en/v2.6.3/advanced_topics/performance.html#templates

cadecairos:
Closes #4292

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -312,6 +312,16 @@ TEMPLATES = [
     },
 ]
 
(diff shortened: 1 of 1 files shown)


cadecairos:
"This loader is automatically enabled if OPTIONS['loaders'] isn’t specified and OPTIONS['debug'] is False (the latter option defaults to the value of DEBUG)."

Issue #4269, opened 2020-03-02
expose postgresql so that it can be accessed outside of Docker


Pomax:
This is especially important for database issues relating to real data, such as deleting user accounts from the prod/staging instances.

cadecairos:
The root of the problem here is that we're defaulting to using SQLite.

We need to update a few things. Firstly, we need to update docker-compose.yml:

In the `postgres` service, update `ports` to be: `"5432:5431" (changes the port exposed on the host in case it's already running a postgres DB). In the same service, add:
```yaml
environment:
  - POSTGRES_USER=foundationsite
  - POSTGRES_PASSWORD=devmode
  - POSTGRES_DB=foundationsite
```
This will, upon DB creation, add `foundationsite` as a super user, set their password to `devmode` and create a default database named `foundationsite`. This will allow us to connect to the database without needing to use trust authentication, which is the default.

We'll need to update the default connection string in `env.default` to be `postgres://foundationsite:devmode@localhost:5431/foundationsite`. There *might* be some other places we need to set this, but I don't think we need to do anything to support this change on our CI, since it's not using Docker.

cadecairos:
helpful referece: https://hub.docker.com/_/postgres

Scroll down to "Environment Variables"

Pomax:
turns out the ports are `host:docker`, rather than `docker:host`. Things now work except that postgres does not know the 'foundationsite' role, so I'm simply going to make this `postgres` for db, user, and database and see if that works.

Pomax:
don't seem to be making much headway here.  My docker-compose.yml is currently:
```
  postgres:
    image: postgres:9.6
    ports:
      - "5678:5432"
    environment:
      - POSTGRES_DB=wagtail
      - POSTGRES_USER=foundation
      - POSTGRES_PASSWORD=mozilla
    volumes:
      - postgres_data:/var/lib/postgresql/data/:delegated
```

with me .env containing:

```
DATABASE_URL=postgresql://foundation:mozilla@postgres:5432/wagtail
```

When I do this, I can get into postgres from the host using:
```
> psql postgresql://foundation:mozilla@localhost:5678/wagtail
psql (10.6, server 9.6.17)
WARNING: Console code page (850) differs from Windows code page (1252)
         8-bit characters might not work correctly. See psql reference
         page "Notes for Windows users" for details.
Type "help" for help.

wagtail=# \dt
Did not find any relations.
wagtail=#
```
[...]

PR #4294, opened 2020-03-04
exposing postgres to hosts

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4269 by adding explicit credentials and a host mapping to the docker compose yaml config.

This will keep everything running on 5432 as far as anything inside Docker knows, with `docker-new-env` updating the `.env` file with the user/pass/dbname from `docker-compose.yml` (if set; If not set, they default to the string `postgres`).

However, postgres is now also accessible on port 5678 from the host OS so that `psql postgresql://username:password@localhost:5678/dbname` gets you into the Docker-hosted postgres instance, to allow for things like importing a copy of the staging/production db to debug problems associated with real data.


Pomax:
Of course the one challenge now is to allow for dropdb/createdb, because during a normal docker-compose run, the db is locked due to wagtail being connected to it...

Pomax:
sorted:
```
SET ROLE_NAME=......
pg_dump -F c prodOrStageDATABASE_URL > foundation.db.archive
SET PGPASSWORD=mozilla
dropdb -h localhost -p 5678 -U foundation wagtail
createdb -h localhost -p 5678 -U foundation wagtail
createuser -h localhost -p 5678 -U foundation -s $ROLE_NAME
createuser -h localhost -p 5678 -U foundation -s datastudio
pg_restore -d "postgresql://foundation:mozilla@localhost:5678/wagtail" database.dump
```
Where  `ROLE_NAME` is the username from the DATABASE_URL

(I have this in a .bat file that is on a completely different drive, in a non-git-managed dir so I can't ever accidentally check it in, with some automated prompts for whether to redownload a db dump, and if so, whether to use staging or prod)

Issue #4271, opened 2020-03-02
MozFest - Homepage - main body text section should wrap and not go full width


mmmavis:
- [x] hide the following fields from CMS if `banner_type` flag is set to `video`
  - 'banner', 'header', 'intro', 'banner_guide_text', 'banner_video_url'
- [x] Make sure no content crosses over the signup form (i.e., text in body section should wraps within the left column instead of going full width).
- [x] Update fake data / Percy tests where necessary so changes are reflected there as well.

mmmavis:
Done!

PR #4296, opened 2020-03-05
MozFest homepage - banner and field changes for "video" version of homepage

### LINKED ISSUE #4264: MozFest site: video background
The MozFest site design updates have been finalized in #4243. The main changes will be a video background header with minimal text and a pause button for accessibility purposes. Other smaller updates can be done in https://github.com/mozilla/foundation.mozilla.org/issues/4242. 

cc: @kristinashu 

## List of Requirements
- [ ] Create video background
- [ ] Text on video background should be limited to one line, and preferable up to only 8 columns on desktop/tablet, can be 10 column on mobile
- [ ] Text should be localizable
- [ ] Include pause button, assets here: [pause-circle-assets.zip](https://github.com/mozilla/foundation.mozilla.org/files/4295257/pause-circle-assets.zip)
- [ ] Include play button when video is paused, assets here: [play-circle-assets.zip](https://github.com/mozilla/foundation.mozilla.org/files/4295258/play-circle-assets.zip)

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] Link to original design ticket: #4243
- [x] Link to related ticket: #4242
- [x] Link to Redpen: https://redpen.io/p/rhbb6e67da007091f2
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Assets are included
- [x] Hover state for elements are designed

## Desktop
mozilla-festival v3@1 5x

## Hover State
mozilla-festival-hover-v3@1 5x

## Play Button (when video is paused)
Group 13

## Play Button Hover State
Group 11

## Mobile
mozilla-festival  M  Copy 2@2x

mmmavis:
(To code reviewer: it's a lot easier to review code changes in this PR without whitespace changes)

Closes #4264, #4271

https://foundation-s-mozfest-vi-mnnj5z.herokuapp.com/en/mozilla-festival/

**Note**
- Browser support for muted autoplay video is inconsistent. On a browser that does not fully support it, `play` button will show instead and users will have to click that button in order for the video to play.
- final video is to be finalized and shrunk to reduce file size


source/js/mozfest-event-handlers.js
```diff
@@ -0,0 +1,45 @@
...
+
+  if (homeVideoControl && homeVideo) {
+    let showVideoControl = () => {
+      homeVideoControl.classList.remove(`invisible`);
+    };
+
+    homeVideoControl.addEventListener(`click`, () => {
+      if (!homeVideo.paused) {
+        homeVideo.pause();
+        homeVideoControl.innerText = `Resume`;
```
TheoChevalier: I wonder if it would be possible to hide/show HTML elements instead of having copy in JS? We won’t be able to localize that otherwise

mmmavis: Thanks @TheoChevalier . Yea I planned on doing that. I'm still waiting on designers to let me know what we wanna do to the button after users click on it.  Was just playing around to see what it would look like if button text changes 😄

TheoChevalier: You rock 👊 Sorry for jumping the gun!

Pomax:
As someone acutely affected by this: is it possible to make the video pause by clicking anywhere on the banner, rather than just the pause button?

kristinashu:
> As someone acutely affected by this: is it possible to make the video pause by clicking anywhere on the banner, rather than just the pause button?

We could also ask Natalie to make the video less dynamic at the start.

Otherwise, nice! It looks so good and is playing really well in my FF.

It isn't playing on my iPhone tho. I think it's _ok_ if it doesn't play on mobile but we should just hide the play button.

The play/pause button does something a little odd on my desktop safari Version 13.0.5 (15608.5.11). See GIF:
Large GIF (592x224)

And a small, non-blocking thing is that the text becomes misaligned in Chrome Version 79.0.3945.130 at narrow width: 
Screen Shot 2020-03-05 at 4 01 55 PM

Pomax:
Win10 Edge, and Android Chrome and Firefox work great, too. I did notice it's possible for the video to start playing before the play/pause icon has finished loading, which we probably don't want? Also, does it make sense to transfer the first frame as image, first, so that rather than looking at a purple/blue gradient, people will see what looks like the video, even though it's not actually loaded to play yet?

network-api/networkapi/mozfest/models.py
```diff
@@ -98,6 +98,8 @@ def get_context(self, request, bypass_menu_buildstep=False):
 
 
 class MozfestHomepage(MozfestPrimaryPage):
+    banner_type = "video"
```
Pomax: Let's add some comments here that explains why this field exists, because it's sort of a sneaky "not django, not wagtail" field. Probably a little more than a one liner, using python's block-of-string comment syntax:

```python
class MozfestHomepage(MozfestPrimaryPage):
    """
    ...description of this class and the banner_type value in particular
    """
    banner_type = "...."
```

network-api/networkapi/mozfest/templates/partials/homepage_banner.html
```diff
@@ -2,11 +2,13 @@
...
   <div class="row">
-    <div class="col-12 col-lg-8 dark-theme">
+    <div class="col-10 col-lg-8 dark-theme">
       <h1 class="h1-heading">{{ page.banner_heading }}</h1>
-      <p class="body-large">{{ page.banner_guide_text }}</p>
-      {% if page.banner_video_url %}
-      <a class="btn btn-primary" href="{{ page.banner_video_url }}" target="_blank" id="mozfest-home-watch-video-button">{% trans "Watch Video" %}</a>
+      {% if banner_type != "video" %}
+        <p class="body-large">{{ page.banner_guide_text }}</p>
+        {% if page.banner_video_url %}
```
Pomax: if the banner type isn't video, why would there be a banner video url?

mmmavis:
@kristinashu  I just pushed some code to fix the mobile "video not playing" bug. I’ll need to address the Safari UI issue and Chrome mobile misalign bugs as follow-up.

@pomax 
>  I did notice it's possible for the video to start playing before the play/pause icon has finished loading, which we probably don't want?

I can’t find a perfect way to detect when autoplay actually starts. So the best I can do for today is to have controls loaded *after* video has stared playing. I can file an enhancement ticket and get to it tomorrow.

> Also, does it make sense to transfer the first frame as image, first, so that rather than looking at a purple/blue gradient, people will see what looks like the video, even though it's not actually loaded to play yet?

Yea. the video is still being finalized. Once that's ready I will drop in a static image for it to show while video is being loaded.

kristinashu:
Sweeeet! Nice it's working on my phone now!

source/js/mozfest-event-handlers.js
```diff
@@ -0,0 +1,59 @@
...
+    });
+
+    showVideoControls();
+
+    video.addEventListener(
+      `canplay`,
+      () => {
+        showVideoControls();
+      },
+      false
```
Pomax: `useCapture` is basically an irrelevant option in modern browsing, and defaults to `false` anyway, so there shouldn't be a reason to specify it.

network-api/networkapi/mozfest/templates/fragments/hero/large.html
```diff
@@ -1,23 +1,47 @@
-{% load wagtailcore_tags wagtailimages_tags %}
+{% load wagtailcore_tags wagtailimages_tags i18n %}
 
 <div class="background-wrapper">
   <div class="overlay"></div>
-  {% if page.banner %}
+  {% if banner_type == "video" %}
+    <video class="banner-video" playsinline autoplay muted loop preload poster="" >
```
Pomax: we may want to take off `autoplay` and instead trigger a `video.play()` (and label reveal) only after we know that the play/pause icons have fully loaded, otherwise the video can start playing before the controls are ready.

network-api/networkapi/mozfest/templates/partials/homepage_banner.html
mmmavis: ahhhhhh i should rename the variable to avoid confusion

network-api/networkapi/mozfest/models.py
```diff
@@ -98,6 +98,15 @@ def get_context(self, request, bypass_menu_buildstep=False):
...
 
 class MozfestHomepage(MozfestPrimaryPage):
+    """
+    MozFest Homepage
+
+    'banner_video_type' determines what version of banner design the page should load
+    """
+
+    #  this tells the templates to load a hardcoded, pre-defined video in the banner background
+    banner_video_type = "hardcoded"
```
mmmavis: i can't think of a better value now...

Pomax: this is perfect, thanks

network-api/networkapi/mozfest/templates/fragments/hero/large.html
mmmavis: Follow-up filed: https://github.com/mozilla/foundation.mozilla.org/issues/4309

mmmavis:
I will file more follow-up tickets to address issues mentioned above.

Merging this PR now.

PR #4298, opened 2020-03-05
Add robots.txt to exclude the thimble artifact page

patjouk:
Bots seem to love the thimble artifact page. Adding a `robots.txt` to the `mofo-redirector` cut some of the traffic, but some google bots are hitting this page directly. This should help.
I didn't want to disallow the whole site, just that page. I used a regex to cover the different locales we have (yeah, [it's possible](https://developers.google.com/search/reference/robots_txt#url-matching-based-on-path-values)).


network-api/networkapi/urls.py
```diff
@@ -19,6 +20,13 @@
 admin.autodiscover()
 
 urlpatterns = list(filter(None, [
+    # Add robots.txt to exclude the thimble artifact page
+    path('robots.txt', lambda x: HttpResponse(
+        "User-Agent: *\nDisallow: /*artifacts/thimble",
+        content_type="text/plain"),
```
cadecairos:
```suggestion
        content_type="text/plain; charset=utf-8"),
```

PR #4300, opened 2020-03-05
Work around unsupported POSIX locales by project-config

TheoChevalier:
Like we needed to add more complexity here, I discovered that project-config doesn’t support POSIX locales.
Meaning that Pontoon cannot commit to the `pt_BR` folder when the project is setup with project-config, while it was possible when we were using Pontoon’s own file detection algorithm.

So, translation files now need to
1/ be stored in `pt-BR` in `fomo-l10n`
2/ copied into `pt_BR` in `foundation.mozilla.org` project for Django to extract and merge strings
3/ copied back into `pt-BR`

1/ has been covered by https://github.com/mozilla-l10n/fomo-l10n/commit/f66774bb4a7e7f1a079f16b38567a0ce6735e80a and I updated archive.sh in the process to pick files from `pt-BR`.

This PR takes care of 2/ and 3/


tasks.py
```diff
@@ -22,7 +22,6 @@
 # so we build it here rather so that we don't clutter up the tasks.
 locale_abstraction_instructions = " ".join([
     "makemessages",
-    "-l de -l es -l fr -l nl -l pl -l pt_BR",
```
TheoChevalier: This is not 100% related, it’s just we don’t need to explicitly to do that anymore since files are stored in an external repo

PR #4316, opened 2020-03-09
Support ISO 3166 locale codes

### LINKED ISSUE #4314: Support ISO 3166 locale codes
Like on donate-wagtail, it would be great to support `ab-CD` locales to be consistent with the format we support at Mozilla.

I’m going to backport here the fix Torchbox made

TheoChevalier:
Closes #4314

+ Backporting the work done on donate-wagtail to support `ab-CD` locale codes. 
+ Not touching anything related to Portuguese here, but that should allow us to migrate to `pt-BR` later on (if there’s a way to migrate).
+ Adding support for `fy-NL` in order to test things easily, and because we would like to support it on the MozFest site.
+ I’ve fixed the language switcher, as it was relying on two-letter codes. [Explanation of the fix](https://stackoverflow.com/questions/54262273/how-do-i-include-the-region-in-language-code)
+ ~~I’ve made my `translation_management.sh` script dynamic to support a list of locales that need to be copied from `ab-CD` into `ab_CD` during Django string extraction process.~~
UPDATE: I managed to leverage relative symlinks in the `fomo-1l0n` repo to entirely get rid of moving files between `ab-CD` and `ab_CD` folders. Or even to `pt` for Portuguese.

Most of the code comes from these two commits:
https://github.com/mozilla/donate-wagtail/commit/f582d53a606fc37d22cd513e1581b5631fa84cff
https://github.com/mozilla/donate-wagtail/commit/d3bfd0086b1ba775812816d480785958216ac90f

Adding Pomax as reviewer, since he reviewed the original fix


network-api/networkapi/wagtailpages/__init__.py
```diff
@@ -16,3 +20,54 @@ def register_howto_menu_item():
...
         name='howdoIwagtail', classnames='icon icon-help', order=900
     )
+
+
+# WARNING: this is not necessarily a good idea, but is the only way to override
+# Django's default behaviour of requiring language codes to be lowercased.
+# We have to modify the core Django method, because we have no way to replace
+# all the core functionality that relies on this - e.g., url resolvers that
+# the Django admin and third party apps use.
+# This is pretty ugly, and ideally this should be fixed upstream.
```
Pomax, line 30: Do we have a link to the issue on the Django side for this? If there isn't one, we should probably file one, especially if MoCo's run into this problem too.

network-api/networkapi/wagtailpages/__init__.py
```diff
@@ -16,3 +20,54 @@ def register_howto_menu_item():
...
+
+
+def language_code_to_iso_3166(language):
+    """Turn a language name (en-us) into an ISO 3166 format (en-US)."""
+    language, _, country = language.lower().partition('-')
+    if country:
+        return language + '-' + country.upper()
```
Pomax, line 37: let's use formatted strings here. It's a minor change, but avoiding string concatenation in favour of string templating is always a good idea

```suggestion
        return f'{language}-{country.upper()}'
```

Pomax:
@TheoChevalier we can probably land this soon, I rebased and fixed the migrations (because we did a significant squashing since this PR got opened, so the migration dependencies had completely changed), but there's only the two comments.

network-api/networkapi/wagtailpages/__init__.py
TheoChevalier: Thanks, do you want me to update this on donate too?

network-api/networkapi/wagtailpages/__init__.py
TheoChevalier: I don’t think we have one, unless Torchbox filed one when they wrote this. I can check with MoCo folks too.
I often stumbled upon Django tickets when looking for i18n-related stuff I needed, managed to find the exact ticket covering my needs, just to see it closed as WONTFIX :S But we can try!

TheoChevalier:
@Pomax Thank you! And sorry I haven’t had much time to look into this today

network-api/networkapi/wagtailpages/__init__.py
Pomax: that's fine, as long as we have something to point to

Pomax:
@TheoChevalier let's get this landed before the end of this week? =)

TheoChevalier:
@Pomax Sorry for the wait, finally took the time to file https://code.djangoproject.com/ticket/31795
Updated the comment with that ticket

PR #4331, opened 2020-03-11
Add 'poster' for MozFest video

mmmavis:
Show a still image while MozFest video is loading.


network-api/networkapi/mozfest/templates/fragments/hero/large.html
```diff
@@ -3,7 +3,8 @@
 <div class="background-wrapper">
   <div class="overlay"></div>
   {% if banner_video_type == "hardcoded" %}
-    <video class="banner-video" playsinline muted loop preload poster="" >
+    <video class="banner-video" playsinline muted loop preload
+           poster="https://s3.amazonaws.com/mofo-assets/foundation/video/mozfest-banner-video-poster.png" >
```
mmmavis: I put the file on s3 as welll so if we ever have to update the video, we can just update both video and the still image without a new deploy.

Pomax: nice!

Pomax:
As "natural" data, we should probably use JPG compression rather than PNG - right now it's a 1.7MB image, which will take a long time to transfer as well, whereas because the kind of image data it is, even a q=3 JPG will still look pretty good at only 155kb:

mozfest-banner-video-poster

mmmavis:
Good call. PR & images have been updated!

Issue #4233, opened 2020-02-19
Create a BlogIndex subclass of the IndexPage


Pomax:
Implement a BlogPage specific subclass of the index page, with code updates that make everything work specifically for BlogPage content.

gabbyprecious:
Hello, can you please me more comprehensive, I'll love to work on this project.

PR #4333, opened 2020-03-11
dedicated blog index page extension on index page

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4233 by creating a new `BlogIndexPage(IndexPage)` that uses the same template as before, but has all the "specific to blog posts" code for category routes, with the `sub_pages` list set exclusively to `BlogPage` .

Note that tag filtering is still handled by the base `IndexPage` model, so that the next-up `CampaignIndexPage` can inherit those for free.

**NOTE** when landing this, we will need to make a new `/_blog` parent page of this type, move all blog pages over to that new parent, then remove the current `/blog`  from the CMS and rename the slug for the new parent to `/blog`.

URLS:
- homepage with BlogIndexPage for `/blog`: https://foundation-s-blogindex-san5t2h.herokuapp.com/cms/pages/3/
- actual BlogIndexPage: https://foundation-s-blogindex-san5t2h.herokuapp.com/cms/pages/10/edit/ with live page https://foundation-s-blogindex-san5t2h.herokuapp.com/en/blog/


network-api/networkapi/wagtailpages/models.py
```diff
@@ -54,30 +58,31 @@
...
-    NewsPage,
-    InitiativesPage,
-    ParticipatePage2,
-    Styleguide,
-    Homepage,
-    RedirectingPage,
     BanneredCampaignPage,
-    CampaignPage,
-    OpportunityPage,
+    BlogIndexPage,
```
Pomax: alphabetical sort. Import order is _very important_ but export ordering is irrelevant, so sorting is nice.

network-api/networkapi/wagtailpages/pagemodels/blog_index.py
```diff
@@ -0,0 +1,111 @@
...
+    """
+    The blog index is specifically for blog pages,
+    with additional logic to explore categories.
+    """
+
+    subpage_types = [
+        'BlogPage',
+    ]
+
+    template = 'wagtailpages/index_page.html'
```
Pomax: super important new code.

network-api/networkapi/wagtailpages/pagemodels/blog_index.py
```diff
@@ -0,0 +1,111 @@
...
+            hasattr(entry, 'category')
+            and
+            category in entry.category.all()
+        ]
+
+        return entries
+
+    # helper function to resolve category slugs to actual objects
+    def get_category_object_for_slug(self, category_slug):
+        # We can't use .filter for @property fields,
```
Pomax: removed the line that got the BlogPageCategory model from the database rather than as direct import, because in the old code that would have been a circular dependency, but with these refactors: that's no longer the case and we can import it at the top of this file and it's very nice yes.

PR #4338, opened 2020-03-12
Add 'eslint-plugin-jsx-a11y' and set violation to raise 'warning' for now

mmmavis:
Related to https://github.com/mozilla/foundation.mozilla.org/issues/3934


.eslintrc.json
```diff
@@ -17,6 +19,13 @@
   },
   "rules": {
     "prettier/prettier": "error",
-    "react/jsx-uses-vars": "error"
+    "react/jsx-uses-vars": "error",
+    "jsx-a11y/click-events-have-key-events": "warn",
```
mmmavis: We currently have 13 jsx-ally errors. It's too big for one single PR to cover all of them so I set the related rules to raise warning instead. That way they won't break our tests and we can tackle them in several PRs.

@Pomax if this approach sounds good, I will file a PR to make sure we remove these temporary jsx-ally config from this file once we fix all jsx-ally issues.
Screen Shot 2020-03-11 at 6 20 14 PM

Pomax:
Let's make this a separate config, with a separate task, so like a `.eslintrc.a11y.json` file and then a task specifically only for .jsx files, like `"test:eslint:a11y": "eslint --config ./.eslintrc.a11y.json \"source/js/**/*.jsx\"",

PR #4353, opened 2020-03-16
Replace whitenoise_root by staticfiles_dirs + fix templates

patjouk:
Ref MozillaFoundation/mofo-devops#749

- I'm adding a missing svg file
- Replacing `[whitenoise_root](http://whitenoise.evans.io/en/stable/django.html#WHITENOISE_ROOT)` because it's not what we want by `[staticfiles_dirs](https://docs.djangoproject.com/en/3.0/ref/settings/#staticfiles-dirs)`
- Removing the staticfile utility since we don't need it anymore and replace it with ` STATICFILES_STORAGE = 'whitenoise.storage.CompressedManifestStaticFilesStorage'` for [compression and caching support](http://whitenoise.evans.io/en/stable/django.html#add-compression-and-caching-support).

Now, collectstatic will actually collect the frontend files :)


patjouk:
DO NOT MERGE PLS XD

network-api/networkapi/settings.py
```diff
@@ -415,12 +415,12 @@
 # https://docs.djangoproject.com/en/1.10/howto/static-files/
 
 
-WHITENOISE_ROOT = app('frontend')
 WHITENOISE_INDEX_FILE = True
 
 STATIC_URL = '/static/'
 STATIC_ROOT = root('staticfiles')
-STATICFILES_STORAGE = 'networkapi.utility.staticfiles.NonStrictCompressedManifestStaticFilesStorage'
```
cadecairos: Do we need to stop using this overridden class to fix the issue? I think this change is why the review app isn't working.

patjouk:
Whitenoise was not configured properly, I'm working on a fix for it that requires to edit all templates.

source/js/buyers-guide/components/creepometer/creepometer.jsx
```diff
@@ -6,7 +6,7 @@ export default class Creepometer extends React.Component {
 
     this.faceCount = 40; // Number of face frames
     this.faceHeight = 70; // pixel height for one frame
-    this.framePath = `/_images/buyers-guide/faces/`;
+    this.framePath = `../../../../static/_images/buyers-guide/faces/`;
```
patjouk: @Pomax @mmmavis @cadecairos Any idea on how to make that less ugly? The url for that file is `root_url/static/_images/...`. If I don't do that, it resolves to `root_url/products/NAME/_images/...` which of course does a 404.

cadecairos:
Also noticed that `banner-black-white-marble.jpg` isn't loading on many pages.

patjouk:
Yeah, it's probably because I haven't go through them yet. I [made a list](https://github.com/MozillaFoundation/mofo-devops/issues/749#issuecomment-600637183) of all templates I need to check. I'll also have to check the CSS since many links broke.

patjouk:
Okay it's nearly ready for review. I'm just hitting an issue with CSP and don't have the energy to look at it now. Will resume on Monday :)

patjouk:
PR is ready for review 🎉 (OMG everything is finally green, including Percy!)

I checked with staging DB but will do another run of manual checks when it hits staging.

The reasoning behind some changes:
- `WHITENOISE_ROOT` should [not be used](http://whitenoise.evans.io/en/stable/django.html#WHITENOISE_ROOT) for all the static files. We want to use [`staticfiles_dirs`](https://docs.djangoproject.com/en/3.0/ref/settings/#staticfiles-dirs) instead.
- Using static tags [like we're supposed to](http://whitenoise.evans.io/en/stable/django.html#make-sure-staticfiles-is-configured-correctly).
- Moving `whitenoise.middleware.WhiteNoiseMiddleware` above the gzip middleware [like recommended](https://github.com/evansd/whitenoise/issues/145).
- Moving `whitenoise.runserver_nostatic` at the top of the `INSTALLED_APPS` so that it [also works in local dev](http://whitenoise.evans.io/en/stable/django.html#using-whitenoise-in-development).
- Added CloudFront support, but I could use extra help (cf Slack DMs). Also, do we want CDN support for review app or not?

Todos:
- [ ] prepare CDN for staging and prod
- [ ] domain name for CDN
- [ ] Deactivate CDN support for review app?
- [ ] Update CSP for staging and prod (script + style)
- [ ] Coms to dev

cadecairos:
>     * do we want CDN support for review app or not?

I don't think we need this, it would be too much work for too little gain.

patjouk:
> > ```
> > * do we want CDN support for review app or not?
> > ```
> 
> I don't think we need this, it would be too much work for too little gain.

I deactivated it in my last commit ;)

PR #4354, opened 2020-03-16
Enable preload and max-requests for gunicorn

cadecairos:
Based on [the documentation on Heroku](https://devcenter.heroku.com/articles/python-gunicorn) there's a couple of options we can try to help alleviate issues with our gunicorn threads.

The first is to enable `preload`, which loads application data into the parent thread and forks it for worker threads, potentially improving startup times. The second recreates workers after they handle a certain number of requests, in order to keep them healthy over time.

let's try running with these for a little while to see if it has any effect on the gunicorn worker issues.


cadecairos:
> K8s setup for running load tests?

nope, but it shouldn't be difficult to recreate.

PR #4356, opened 2020-03-17
Adjust sticky CTA position for MozFest & tiny SCSS code refactor

### LINKED ISSUE #4304: MozFest Homepage - add spacing above signup form
Follow-up from https://github.com/mozilla/foundation.mozilla.org/pull/4286#pullrequestreview-369907594

- [ ] Can we add some spacing between the sign up snippet and the nav? I think 24px should do it:
Screen Shot 2020-03-05 at 3 51 03 PM

/cc @kristinashu to triage.

mmmavis:
Closes #4304

https://foundation-s-issue-4304-s1nzcf.herokuapp.com/en/mozilla-festival/

Ended up going with a SCSS-only approach as I thought using JS to detect the `sticky` state and dynamically adjust the position of the CTA/form section just for MozFest homepage would be an overkill.

### DIFF
--- a/source/sass/mixins.scss
+++ b/source/sass/mixins.scss
@@ -89,7 +89,7 @@
   }
 }
 
-@mixin cta-anchor-position($extra-vertical-spacing: 0) {
+@mixin cta-anchor-position($extra-vertical-spacing: 0px) {
   $offset: 80px;
 
   margin-top: -$offset;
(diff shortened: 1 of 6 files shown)


source/sass/mixins.scss
```diff
@@ -88,3 +88,10 @@
     }
   }
 }
+
+@mixin cta-anchor-position($extra-vertical-spacing: 0px) {
```
mmmavis: Have to specify unit as it will be using in interpolation on line 96.

PR #4360, opened 2020-03-17
Wrap 'card' description with appropriate HTML tags

### LINKED ISSUE #4287: Highlight component's decription should be wrapped in <p></p>
**Context:**
See #https://github.com/mozilla/foundation.mozilla.org/pull/4263#issuecomment-594102345

Highlight component's `description` is a `TextField` but currently rendered as rich text. We should make sure the `description` is wrapped in  `<p></p>` for semantic purposes and for the right style to kick in.

**Relevant code to look at**
https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/card.html#L14 
Since this template is used by different components, we shouldn't assume `description` is always in rich text format.

mmmavis:
Closes #4287 

Notes: some vertical spacing around the `descrition` also got fixed in this PR as before we had redundant empty `<p>` nodes hanging around. e.g., 
image

Go to the following pages and look for section that looks like
image

## Hompage

https://foundation-s-issue-4287-fjnyxw.herokuapp.com/en/

## Initiatives

https://foundation-s-issue-4287-fjnyxw.herokuapp.com/en/initiatives/

## Participate

https://foundation-s-issue-4287-fjnyxw.herokuapp.com/en/participate/

### DIFF
--- a/network-api/networkapi/wagtailpages/templatetags/card_tags.py
+++ b/network-api/networkapi/wagtailpages/templatetags/card_tags.py
@@ -1,4 +1,5 @@
 from django import template
+from bs4 import BeautifulSoup
 
 register = template.Library()
 
@@ -9,6 +10,7 @@ def card(image, title, description, link_url, link_label, commitment=None):
         'image': image,
         'title': title,
         'description': description,
+        'description_is_rich_text': BeautifulSoup(description, 'html.parser').p is not None,
         'link_url': link_url,
         'link_label': link_label,
         'commitment': commitment,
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/templatetags/card_tags.py
```diff
@@ -9,6 +10,7 @@ def card(image, title, description, link_url, link_label, commitment=None):
         'image': image,
         'title': title,
         'description': description,
+        'description_is_rich_text': BeautifulSoup(description, 'html.parser').p is not None,
```
Pomax: I'm not sure this is a safe way to check this: where is `description` coming from? Because rich text could also be `<ul>...</ul>` etc. so we don't want to hardcode that only things with `<p>` are rich text.

mmmavis: Ah agreed that this is not very future proof. I only took CTA's `subhead` into consideration (passed to template as `description`).

```py
subhead = RichTextField(
        features=[
            'bold', 'italic', 'link',
        ],
        blank=True,
    )
```

Will update PR.

PR #4367, opened 2020-03-18
Turn all href='#' anchors into <button> & fix bad React attribute

### LINKED ISSUE #3366: clean up <a href="#">
Under HTML5 rules, the `<a>` element is a purely navigational element, and adding `href="#"` tells the browser to navigatet to the top of the document, which is almost never the intended behaviour when `<a href="#">` is used, so let's find all instances of `<a href="#" ...>` and make sure we replace them with correctly semantically marked up code instead (which will almost always be `<button>`)

Marked as small effort, but this might turn out to be an invalid assessment.

mmmavis:
Closes #3366

I'm not sure why Percy gives false positive flags though. Things look fine when I visit Review app pages.


network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -38,7 +38,7 @@
       <div class="row intro d-flex justify-content-between align-items-center">
         <a id="product-company-url" class="product-price small-link col-md-8 mb-3" href="{{product.url}}" target="_blank">{{product.company}} ${{product.price}}</a>
         <div class="copy-link-wrapper col-md-4 mb-3 text-md-right">
-          <a id="product-copy-link-button" class="copy-link small-link" href="#" data-url={{product.url}} data-success-text="link copied!">copy & share link</a>
+          <button id="product-copy-link-button" class="copy-link small-link" data-url={{product.url}} data-success-text="link copied!">copy & share link</button>
```
mmmavis: `copy & share link` button on PNI product page.

e.g., 
https://foundation-s-issue-3366-qgqe9x.herokuapp.com/en/privacynotincluded/products/decision-long/
image

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -20,7 +20,7 @@ export default class ShareButtonGroup extends React.Component {
   renderFacebookButton() {
     let label =
       this.props.version === `mini` ? (
-        <span class="sr-only">Share on Facebook</span>
+        <span className="sr-only">Share on Facebook</span>
```
mmmavis: React uses `className`

source/js/components/share-button-group/share-button-group.jsx
```diff
@@ -109,14 +109,13 @@ export default class ShareButtonGroup extends React.Component {
     });
 
     return (
-      <a
-        class={classes}
-        href="#"
+      <button
```
mmmavis: e.g., 

https://foundation-s-issue-3366-qgqe9x.herokuapp.com/en/blog/initial-test-blog-post-with-fixed-title/
image
and further down the page
image

source/js/components/share-button-group/share-button-group.jsx
Pomax: omg we still had some of those left? =_=

PR #4372, opened 2020-03-19
add mergify for dependabot PRs only for now

patjouk:
Closes MozillaFoundation/mofo-devops#746

Add Mergify for Dependabot PRs only. I'm adding `strict` as a rule to make sure that the PR is always up to date before merging.
Rule for Dependabot: https://doc.mergify.io/examples.html#dependabot

Mergify has a "simulator" to validate your rules
PR #4365 would be merged
Screenshot 2020-03-19 at 13 46 58

PR #4358 wouldn't be merged
Screenshot 2020-03-19 at 13 57 53

----

If we're happy with the results, we could also add it to all PRs, using a label `ready-to-merge` to prevent [unwanted merge](https://doc.mergify.io/examples.html#using-labels-to-enable-disable-merge).


patjouk:
Since we had a random person approving a PR on `wagtail-donate`, I wanted to make sure that it would not automatically merged. Tried with #4391: it wouldn't be merged :)

PR #4381, opened 2020-03-20
Remove forced redirect to English for PNI

### LINKED ISSUE #4379: PNI: expose translations
Now that https://github.com/mozilla/foundation.mozilla.org/commit/6f070b1fcedf46984abb36def9b4edcc07d7f92a has landed, let’s expose this work to the world.

The approach will be to progressively localize new areas (React, then Wagtail). Jen has greenlit the approach, so we can remove the redirect to /en and start updating the (many) hardcoded links to /en

TheoChevalier:
Closes #4379

Removing forced /en/ redirect for PNI and adding back language switcher in footer.

https://foundation-s-pni-lang-s-i6oqgg.herokuapp.com/fr/privacynotincluded/

(This time, with the correct branch)


network-api/networkapi/buyersguide/tests.py
```diff
@@ -333,30 +333,6 @@ def test_localised_homepage(self):
         response = self.client.get('/privacynotincluded/')
         self.assertEqual(response.status_code, 302, 'simple locale gets redirected')
 
-    def test_only_en(self):
```
Pomax: We can't just remove these, we'll have to replace them with new tests that verify the various locales now work as expected.

network-api/networkapi/buyersguide/templates/bg_base.html
```diff
@@ -93,7 +93,7 @@
       {% block guts %}{% endblock %}
     </div>
 
-    {% include "partials/footer.html" with exclude_language_switcher=True %}
```
Pomax: do we still use `exclude_language_switcher` anywhere else? If not, we should clean up the footer partial, too.

TheoChevalier: Yes, it’s used on 404 pages

network-api/networkapi/settings.py
```diff
@@ -408,6 +408,7 @@
 
 LOCALE_PATHS = (
     os.path.join(BASE_DIR, 'locale'),
+    os.path.join(BASE_DIR, 'networkapi/buyersguide/templates/about/locale'),
```
TheoChevalier: Just realized that files in this folder were correctly extracted, but translations were NOT picked up by Django.
Looks like we need to specify that path, as it’s not the default `[app name]/locale`

network-api/networkapi/buyersguide/tests.py
TheoChevalier: Updated the tests, hopefully that’s covering most cases?

PR #4382, opened 2020-03-21
'opportunity' minisite cleanup

### LINKED ISSUE #2971: Move /opportunity pages to /initiatives
At some point we decided to no longer use /opportunity and use /initiatives instead. Now we have projects arbitrarily split between both which is more confusing and complicated for staff and users.

Next steps:
- [x] confirm that we want to sunset /opportunity
- [x] move all /opportunity pages to /initiative
- [x] redirects those pages (one PR)
- [x] archive /opportunity (follow-up PR after the redirect PR lands)

mmmavis:
Closes #2971

- updated fake data so factory doesn't create an 'opportunity' minisite anymore
- updated tests
- updated docs

### DIFF
--- a/cypress/integration/endpoint-tests.js
+++ b/cypress/integration/endpoint-tests.js
@@ -92,27 +92,29 @@ describe(`Visual regression testing for foundation.mozilla.org`, () => {
   // Opportunity page tests (single and multi-page)
 
   it(`Single-page opportunity should redirect to /intiatives`, function() {
-    cy.request({
-      url: `/en/opportunity/single-page-opportunity/`,
-      followRedirect: false
-    }).then(res => {
-      expect(res.status).to.eq(302);
-      expect(res.redirectedToUrl).to.eq(
-        `${Cypress.config().baseUrl}/en/initiatives/single-page-opportunity/`
-      );
-    });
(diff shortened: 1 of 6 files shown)


network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -76,7 +76,7 @@ def generate(seed):
     home_page = get_homepage()
 
     try:
-        blog_namespace = WagtailPage.objects.get(title='blog')
+        blog_namespace = WagtailPage.objects.get(title='Blog')
```
mmmavis: to match line 85

cypress/integration/endpoint-tests.js
```diff
@@ -92,27 +92,29 @@ describe(`Visual regression testing for foundation.mozilla.org`, () => {
...
+    );
+    cy.window()
+      .its(`main-js:react:finished`)
+      .should(`equal`, true);
+    cy.wait(500);
+    cy.percySnapshot();
   });
 
   it(`Multi-page opportunity should redirect to /intiatives`, function() {
-    cy.request({
```
mmmavis: `/en/opportunity/multi-page-opportunity/` doesn't exist now so this would throw errors

PR #4402, opened 2020-03-24
Update default list <ol> & <ul> styling & update styleguide to include them

### LINKED ISSUE #4373: Increased body size should be applied everywhere
We decided to increase the font size of our body text and it looks like it was applied to regular paragraph text but not all the special cases. For example, bullet point text is too small and should match the rest of the text. See https://foundation.mozilla.org/en/docs/process/ 

image

I think the "happening now" cards on our homepage should also have the larger text.
image

cc @sabrinang @mmmavis

mmmavis:
Closes #4373

Go to https://foundation-s-issue-4373-zpdxbk.herokuapp.com/en/styleguide/ and search for "Lists" to review the updated list style.

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/static/styleguide.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/static/styleguide.html
@@ -148,6 +148,23 @@
         <p class="body-small">.body-small with <a href="">link</a></p>
       </div>
 
+      <h3 class="h2-heading mt-5">Lists</h3>
+      <hr />
+      <div class="py-2">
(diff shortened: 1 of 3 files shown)


mmmavis:
> Unordered List: by default, can the listed items be marked with bullets (list-style-type:disc) and indented as well (like the ordered list)?

@sabrinang we'll have to file a different ticket for this since the only changes I made to list were font size changes and colour changes (to make them white on dark theme).

mmmavis:
@sabrinang I've filed a ticket to address the unordered list styling update: https://github.com/mozilla/foundation.mozilla.org/issues/4413 Can you R+ this PR? 😄

Issue #4408, opened 2020-03-24
The RSS feed pulls content from the wrong model


Pomax:
The RSS feed is still pulling blog posts from `IndexPage` rather than `BlogIndexPage`.

PR #4409, opened 2020-03-24
blog page namespace is no longer an IndexPage

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4408

### DIFF
--- a/network-api/networkapi/wagtailpages/rss.py
+++ b/network-api/networkapi/wagtailpages/rss.py
@@ -4,7 +4,7 @@ from django.utils.feedgenerator import Atom1Feed
 
 from bs4 import BeautifulSoup
 
-from .models import IndexPage
+from .models import BlogIndexPage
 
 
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailpages/rss.py
```diff
@@ -18,7 +18,7 @@ class RSSFeed(Feed):
     description = 'The Mozilla Foundation Blog'
 
     def items(self):
-        blog_index = IndexPage.objects.get(title__iexact='blog')
+        blog_index = BlogIndexPage.objects.get(title__iexact='Blog')
```
mmmavis: does this need to be locale-aware (e.g., `title_en__iexact`)?

Pomax: hm, yeah probably. We have a "get _en, if that doesn't exist, try plain" somewhere else, let me put that same logic here.

Issue #4406, opened 2020-03-24
`inv docker-test-python` no longer passes on master for local dev


Pomax:
It looks like the whitenoise PR made it so that master no longer passes its tests for local dev work. 

STR:
- check out master branch
- (optional `docker system prune --all`, probably not needed)
- `inv docker-new env`
- `inv docker-test-python`

This has five failing tests due to static content, it looks like:

```bash
======================================================================
ERROR: test_category_view (network-api.networkapi.buyersguide.tests.BuyersGuideViewTest)
Test that the category view returns a 200 for both slug and name URLs
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/app/network-api/networkapi/buyersguide/tests.py", line 397, in test_category_view
    response = self.client.get(f'/en/privacynotincluded/categories/Smart%20Home/')
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/test/client.py", line 535, in get
    response = super().get(path, data=data, secure=secure, **extra)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/test/client.py", line 347, in get
    **extra,
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/test/client.py", line 422, in generic
    return self.request(**r)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/test/client.py", line 503, in request
    raise exc_value
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/core/handlers/exception.py", line 34, in inner
    response = get_response(request)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/core/handlers/base.py", line 115, in _get_response
    response = self.process_exception_by_middleware(e, request)
  File "/root/.local/share/virtualenvs/app-4PlAip0Q/lib/python3.7/site-packages/django/core/handlers/base.py", line 113, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.7/contextlib.py", line 74, in inner
    return func(*args, **kwds)
  File "/app/network-api/networkapi/utility/redirects.py", line 21, in wrapper
    return func(request, *args, **kwargs)
[...]

PR #4414, opened 2020-03-25
fix errors that failed due to 'missing' static assets

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4406 by ensuring that for the tests that need to rely on static assets, we force `DEBUG=True` so that static assets are "hosted" by Django rather than requiring `collectstatic` or a whitenoise server etc.


network-api/networkapi/buyersguide/tests.py
```diff
@@ -393,6 +395,7 @@ def test_category_view(self):
         self.assertEqual(response.status_code, 200, 'The category "Smarth Home" should work by slug')
 
 
+@override_settings(DEBUG=True)
```
patjouk: `DEBUG=True` is too broad. Use `@override_settings(STATICFILES_STORAGE="django.contrib.staticfiles.storage.StaticFilesStorage")` instead.

cadecairos: yeah, this works for me too. Is there no way to use the Whitenoise backend and have it build the static file manifest on request?

Pomax: I'll leave this un-fixed for a little bit because if I update the code this comment thread will disappear. If there is no way to use whitenoise for local dev then I'll update it to use the contrib.staticfiles.storage backend.

PR #4424, opened 2020-03-26
Basic accessibility fixes

Pomax:
This addresses all the accessibility issues that get flagged by the a11y Eslint rules.

@kristinashu this also includes key handling for the creepometer: if you go to https://foundation-s-accessibil-zmiqpz.herokuapp.com/en/privacynotincluded/products/model-their/ you can now tab your way to the creepometer, and when focussed on it, the left/right cursor keys move the emoji aorund.


Pomax:
This currently still has two errors:
```
...\Git\mozilla\foundation.mozilla.org\source\js\components\multipage-nav-mobile\multipage-nav-mobile.jsx
  33:11  error  The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid
  39:11  error  The href attribute is required for an anchor to be keyboard accessible. Provide a valid, navigable address as the href value. If you cannot provide an href, but still need the element to resemble a link, use a button and change it with appropriate styles. Learn more: https://github.com/evcohen/eslint-plugin-jsx-a11y/blob/master/docs/rules/anchor-is-valid.md  jsx-a11y/anchor-is-valid
```

Pomax:
@mmmavis I think I might need your help in figuring out how we want to update this code so it doesn't trigger that particular error.

source/js/buyers-guide/components/creepometer/creepometer.jsx
```diff
@@ -73,7 +73,18 @@ export default class Creepometer extends React.Component {
...
+    let p = this.state.percentage;
+    if (k === `ArrowLeft`) {
+      this.saveTrackHead(p - 5);
+    }
+    if (k === `ArrowRight`) {
+      this.saveTrackHead(p + 5);
+    }
+  }
+
+  slideFromClick(e) {
```
Pomax: @mmmavis this is new code that slides the creepiness on cursor keys. Thank fully, it's pretty simple =)

Pomax:
The only other alternative is to still tab to it, but clicking on the page just before the slider. As an accessibility solution there's isn't really a shortcut to testing it other than "doing the thing" =D

source/js/buyers-guide/components/creepometer/creepometer.jsx
```diff
@@ -153,7 +169,14 @@ export default class Creepometer extends React.Component {
           <div
             className="slider"
             ref={e => (this.sliderElement = e)}
-            onClick={evt => this.slideClick(evt)}
+            tabIndex="0"
+            role="slider"
+            onClick={evt => this.slideFromClick(evt)}
+            onKeyDown={evt => this.slideFromKey(evt)}
+            aria-valuemax={100}
+            aria-valuemin={0}
```
mmmavis: Should this number match the mininum value we set for `percentage` (line 123)?

source/js/buyers-guide/components/creepometer/creepometer.jsx
mmmavis: Nice! Works pretty well too!

source/js/components/join/language-select.jsx
```diff
@@ -38,7 +38,7 @@ export default class LanguageSelect extends React.Component {
     return (
       <select
         value={this.props.selectedLang}
-        onChange={evt => this.handleChange(evt)}
+        onBlur={evt => this.handleChange(evt)}
```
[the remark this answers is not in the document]
Pomax: You know what, I read up the specs again, and I think the ESLint rule here is plain wrong: `onChange` *always* fires, there is literally no reason to _replace_ it with `onBlur`, because blurring already triggers the `change` event as part of the blur.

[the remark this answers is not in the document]
Pomax: turns out this was entirely the wrong place to encode `value`

Pomax:
@mmmavis I had to switch some code around, so this will need a real re-review ^^;

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -131,7 +131,8 @@ export default class CreepVote extends React.Component {
     this.setState({ creepiness });
   }
 
-  setConfidence(confidence) {
+  setConfidence(confidence, key) {
+    if (key && key === "Tab") return;
```
Pomax: obviously, skip over the tab key when tabbing through the page >_>;;

source/js/buyers-guide/components/creepometer/creepometer.jsx
```diff
@@ -10,8 +10,7 @@ export default class Creepometer extends React.Component {
 
     this.state = {
       dragging: false,
-      percentage: 50,
-      value: 50
```
Pomax: I don't remember why `value` was there, but it literally did nothing, it was always the same as `percentage`.

source/js/components/join/language-select.jsx
```diff
@@ -22,11 +22,14 @@ export default class LanguageSelect extends React.Component {
...
-        </option>
-      );
+      const optionAttributes = {
+        key: lang_code,
+        value: lang_code
+      };
+      if (this.props.selectedLang === lang_code) {
+        optionsAttributes.selected = "selected";
+      }
+      return <option {...optionAttributes}>{languages[lang_code]}</option>;
```
Pomax: this is where the value selection should be recorded, not on the `<select>` element.

mmmavis: Good catch!

PR #4426, opened 2020-03-27
Add missing closing div

patjouk:
The `index_page` template is missing a closing div.


Pomax:
I don't see a missing `</div>` in the current index_page, which line is the unclosed `<div>` on?

patjouk:
It's to close the `<div class="container">` on line 7

patjouk:
Argh it's the if that is messing up with pycharm >.<

Issue #4362, opened 2020-03-17
replace `manage createsuperuser` with a direct python code call?


Pomax:
Docker on windows is great, but the one last thing that doesn't work is `createsuperuser`, due to TTY requirements that don't really work.

However, this does work, creating an admin/admin account:

```
inv docker-manage "shell -c \"from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin')\""
```

Would it make sense to update the tasks so that rather than calling `manage createsuperuser`, we simply directly call the django code that builds a superuser?

/cc @patjouk

patjouk:
Wroks for me!

PR #4431, opened 2020-03-30
automatically build an admin/admin superuser in the docker workflow

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4362 by updating the `docker_create_super_user` task to run the user creation as a direct python one-liner, rather than using the TTY `manage createsuperuser` command.

This means there will now always be an `admin`/`admin` superuser when running `inv docker-new-env` and `inv docker-new-db` without needing to specify your own.


tasks.py
```diff
@@ -189,18 +189,8 @@ def docker_l10n_block_inventory(ctx):
...
-              "To create an admin user:\n"
-              "docker-compose run --rm backend pipenv run python network-api/manage.py createsuperuser\n")
-    else:
-        print("* Creating superuser.")
-        ctx.run(
-            "docker-compose run --rm backend pipenv run python network-api/manage.py createsuperuser",
-            pty=True
-        )
-
+    docker_manage(ctx, "shell -c \"from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin')\"")
```
cadecairos: Flake8 sez: "line 2 loooong"

tasks.py
```diff
@@ -189,18 +189,8 @@ def docker_l10n_block_inventory(ctx):
...
+    print("\nCreated superuser `admin` with password `admin`.")
```
cadecairos: need an additional line break here.

tasks.py
Pomax: damnit flake8

tasks.py
Pomax: Where do you want it, before or after?

PR #4432, opened 2020-03-31
Fix Safari icon resizing bug

### LINKED ISSUE #4311: MozFest Homepage - icon for video control gets resized on hover (Safari)
Follow-up for https://github.com/mozilla/foundation.mozilla.org/pull/4296#issuecomment-595510114

> The play/pause button does something a little odd on my desktop safari Version 13.0.5 (15608.5.11). See GIF:
Large GIF (592x224)

mmmavis:
Closes #4311

https://foundation-s-issue-4311-4o8fr1.herokuapp.com/en/mozilla-festival/


source/sass/mozfest.scss
```diff
@@ -42,6 +42,9 @@ body.mozfest {
           height: $icon-size;
           background: center left/$icon-size $icon-size no-repeat transparent;
           padding-left: calc(#{$icon-size} + 0.5rem);
+          // due to a Safari bug, we have to remove transition for these buttons
```
Pomax: do we have an official webkit issue we can point to for this, so that future-us (or future-someone-else) can follow the clues?

mmmavis: This seems to be a desktop Safari specific issue. I couldn't reproduce it on other browsers. Can't find any official post/ticket about this either but many people on Stackoverflow were having the same issue 😂 

Another way to fix this while keeping transition effect is to update the SVG source code to make sure `height` and `width` are both set to `50px` (the size we want them to appear, see lin 39) though I think we should probably refrain from touching SVG source code in case the same files are used elsewhere and need to be in a different size/color. etc.

Pomax: nah, if this is the solution, that's fine.

Issue #4444, opened 2020-04-01
ReactGA is not imported everywhere


Pomax:
We have a few files that try to trigger ReactGA.event, without having imported ReactGa.

PR #4445, opened 2020-04-01
make sure ReactGA gets imported everywhere we use it

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4444


source/js/common/react-ga-proxy.js
```diff
@@ -1,4 +1,4 @@
-import ReactGAModule from "react-ga";
+import ReactGA  from "react-ga";
```
mmmavis, line 1: oops I forgot to revert this back in my previous PR

mmmavis:
```suggestion
import ReactGA from "react-ga";
```

fix linting issue

Issue #4449, opened 2020-04-02
Make global components tests pass for a11y


Pomax:
Part of https://github.com/mozilla/foundation.mozilla.org/issues/3934, make all tests in the `global_component_assessment.js` integration test set work.

PR #4450, opened 2020-04-02
Make a11y testing pass the "global components" integration tests.

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4449

DISCOVERED FAILURES: the text field placeholder text is too low contrast, and so `checka11y()` won't pass until we address that. For the purposes of this PR, the two problematic checks have been moved into an inert code path, but we probably want to look at what we can do to fix this.


Pomax:
@kristinashu I'm not sure how, but it might be worth looking into what styling we can use for placeholder text in order to achieve a high enough contrast ratio:

image

kristinashu:
cc @sabrinang is this case something you've considered previously?

kristinashu:
@sabrinang is going to test out our Gray 40 ($gray-40) #999999 from [our Style Guide](https://foundation.mozilla.org/en/style-guide/) to see if that will be dark enough.

sabrinang:
Forms currently don't use a standard gray variable and uses `#aeaeb1` and petition forms also have a `color: transparent;` included with the use of #aeaeb1 too. I’m not sure what is happening there as well. 

forms-comparison


**Recommendation**
I suggest we use $gray-60 `#666666` which will have contrast ratio 5.74:1 which passes WCAG AA contrast checks and for our different form filed sizes.

This contrast passes both sizes:
- Petition and Pulse forms have placeholder text at 20px.
- Newsletter form on page, nav dropdown and footer have placeholder text at 16px.

- To improve accessibility it is also suggested that we use floating labels which move out of the field and stay visible when there is a value entered in the input _(this style is seen on our petition pages)_ or always have a label on top _(this is seen on adding pulse entry)_ but since email newsletter is only one field it may not be necessary to apply _(we just aren't very consistent with labels)_

**Dark Theme**
If we need a placeholder grey for a dark-theme version we can use $gray-40 `#999999` which has contrast ratio 7.37:1 to pass.

_Note: I tried this $gray-40 on white but has contrast ratio of 2.84:1 which fails the checks._

mmmavis:
The "placeholder" text we are seeing in petition form seems to be a regular form label but positioned inside of form field, which makes the label looks like a placeholder text. Currently there's an opacity `0.43` applied to it (see [code](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/js/components/petition/petition.scss#L35)). I wonder for this PR if we can just remove the opacity and change the label colour to `$gray-60`? If there are further a11y / form field related issues we wanna address we should create a new ticket.

Pomax:
@mavis your CSS updates indeed removed the contrast warning, so this should be good to go.

PR #4498, opened 2020-04-13
PNI product split

Pomax:
Does:

- creates a new BaseProduct, GeneralProduct, and SoftwareProduct
- duplicates the old product to new GeneralProduct entries
- uses BaseProduct for homepage / category page / product page generation
- the SoftwareProduct model has the software-only fields from the google sheet
- duplicated votes over from old to new products, with all-updated vote-related models, utility functions, management commands, etc...

Does not:

- remove the old Product model because I'm way too scared to do that in the same PR.

Basic testing:
- `inv docker-test-python`

Testing with stage data:
- copy the staging db into the `postgres` docker container (`docker-compose up postgres` in one terminal, then dump staging into localhost:5678 to the local instance - I'm using `pg_restore -d "postgresql://foundation:mozilla@localhost:5678/wagtail" foundation.db.archive ` after dumping the staging db to a file called `foundation.db.archive`)
- run `inv docker-manage "shell -c \"from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin')\"" ` to create the admin/admin user
- `inv docker-migrate` to uplift the db (0039 will take _forever_ because we have so. many. votes.)
- `docker-compose up`, then see below

Once running:
- test that http://localhost:8000/privacynotincluded works
- test that http://localhost:8000/cms works, with updated buyers guide options: instead of a single product, there should now be "product: old", "product: general", and "product: software".
- try to edit some general products, and verify they're the same as the corresponding old product (the ids will not match, but the names should).
- create a "Video Call Apps" buyers guide product category, then
  - create a software product
  - assign it the new product category
  - fill in the mandatory values, make sure to uncheck "draft" and save
- load the home page:
  - you should see a video call category
  - you should see the software listed on the homepage
  - you should see it listed in the Video Call Apps category page
  - you should be able to load the product as dedicated product page

You will notice that voting on your new product will say `NaN votes` - this is because of how we do stats. After voting on it once, run `inv docker-manage aggregate_product_votes`, and then try voting again: it should now have "real" values.

TODO:

We need to verify with Jen that we have all the fields, and in the right place. We can fix it afterwards, but it'd be nicer not to have to.

CAUTION: 

Percy caught something that's genuinely distressing: the data for the percy test product seems to no longer contain the "how does it handle privacy" data... even though it should.


cadecairos:
>     * the duplication migration (0036) does not seem to correctly copy over related products, which is very curious. I see the Apple Watch 5 as related product for the Garmin vivio in the original Product, not in the duplicated GeneralProduct.

I'm seeing products duplicated properly in GeneralProduct entries locally!

Pomax:
added @patjouk just to make sure there's maximum eyes on this one.

network-api/networkapi/buyersguide/pagemodels/products/base.py
```diff
@@ -0,0 +1,370 @@
...
+        """
+        Rather than rendering products based on the instance object,
+        we serialize the product to a dictionary, and instead render
+        the template based on that.
+
+        NOTE: if you add indirect fields like Foreign/ParentalKey or
+              @property definitions, those needs to be added!
+        """
+        model_dict = model_to_dict(self)
+        model_dict['product_type'] = self.product_type
```
Pomax, line 344: this doesn't seem to take. Should this be `BaseProduct.product_type` and then overwritten in General/Software?

PR #4526, opened 2020-04-16
Prototyping a Typeform block

cadecairos:
Kinda works, but also kinda janky.

Height as a block attribute is a poor way to control the element, but I borrowed that from the airtable block.

When in responsive design mode, emulating an S9+, the Typeform frame goes full-screen until completion, which was not quite what I expected, but probably makes sense. However, the height attribute looks pretty silly on mobile, so maybe a responsive CSS rule could make it less janky?

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
@@ -1,0 +1,21 @@
+<script
+  src="https://embed.typeform.com/embed.js"
+  type="text/javascript"
+></script>
+
+<div
+  id="typeform-embed"
+  class="typeform-embed"
+  style="height: {{ value.height }}px;"
+></div>
+
+<script type="text/javascript">
+  window.addEventListener("DOMContentLoaded", () => {
+    let el = document.querySelector("div#typeform-embed");
+    window.typeformEmbed.makeWidget(el, "{{ value.url }}", {
+      hideFooter: true,
+      hideHeader: true,
+      opacity: 0
+    });
+  });
+</script>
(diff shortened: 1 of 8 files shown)


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
```diff
@@ -0,0 +1,21 @@
+<script
```
Pomax, line 1: Would this do "fun things(tm)" if we have multiples of this block on a single page?

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
```diff
@@ -0,0 +1,21 @@
...
+  src="https://embed.typeform.com/embed.js"
+  type="text/javascript"
```
Pomax, line 3: no longer needed in HTML5 - any `<script>` without a type is a JS script tag.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
```diff
@@ -0,0 +1,21 @@
...
+></script>
+
+<div
+  id="typeform-embed"
+  class="typeform-embed"
+  style="height: {{ value.height }}px;"
+></div>
+
+<script type="text/javascript">
```
Pomax, line 12: Let's make this code that gets invoked in `source/js/main.js` (but probably housed in its own file) - we already do this for a fair number of other things, so there should be some examples to copy from, and also means we're not unsafe-inlining our own code.

cadecairos: How would we get the form URL passed in? look for a meta tag?

mmmavis:
Closing. This is being taking care of by the new PR https://github.com/mozilla/foundation.mozilla.org/pull/4687 . Thanks for doing the bulk of the work @cadecairos

PR #4537, opened 2020-04-20
Let petition form's sticky CTA button stay on page a bit longer

### LINKED ISSUE #4305: Enhance mobile sticky CTA button behaviour
Suggested by @beccaklam 

>The sticky button is working but it's disappearing just before the form shows up. Is there anyway for it stay a little longer until the top of the form shows so we know the user has seen it?

Sticky CTA button can be tested on (shown on small screen sizes)
- Campaign page that has a petition form. ([Example](https://foundation.mozilla.org/en/campaigns/tell-amazon-require-two-factor-authentication-all-ring-accounts/))
- MozFest homepage if it has a signup form. ([Example](https://www.mozillafestival.org/))

mmmavis:
Closes #4305 

https://foundation-s-issue-4305-rwaou8.herokuapp.com/en/campaigns/single-page/

Note: Sticky CTA button only shows up on small screen.


source/sass/views/shared/sticky-cta.scss
```diff
@@ -3,11 +3,13 @@
 }
 
 .cta {
+  $sticky-cta-z-index: $zindex-sticky;
```
Pomax, line 6: it feels like we need a comment that points to https://getbootstrap.com/docs/4.0/layout/overview/#z-index here, even if bootstrap offers a built-in var - zindices are tricky as heck, so explaining that we need something that's lower than our header (which uses 1030 because it's sticky-fixed) is a good idea.

mmmavis: Done. Added notes & reference pointing to https://v4-alpha.getbootstrap.com/layout/overview/#z-index

mmmavis:
Ah I didn't scroll far down enough so didn't catch that bug. PR fixed now!

PR #4555, opened 2020-04-23
Allow any authenticated user to clear the PNI product cache

Pomax:
Right now only admins have this power, but really any authenticated user should be able to call this route. Ideally, any authenticated user that, according to their session, is on a PNI product page, but that's more work than worth it, really.


cadecairos:
@Mergifyio refresh

Issue #4546, opened 2020-04-21
Clean up Bootstrap overrides


mmmavis:
~Blocked by #4545~ Unblocked

Clean up overrides from `mofo-bootstrap` and existing overrides for the new brand so there's only one set of overrides in Foundation site repo.

PR #4559, opened 2020-04-24
Clean up Bootstrap overrides (mostly around colours and type)

mmmavis:
Related to #4546 (More clean-up work to come in order to close that ticket)

Cleaned up Bootstrap overrides (mostly around redundant colours and type). Compiled CSS is down by another ~100KB woohoo.

Three worth noting changes

- **[This particular change introduced UI changes]** We have decided to change placeholder colour from `mofo-bootstrap`'s `#aeaeb1` to our current brand's `$gray-60` (`#666666`). You can see this change reflected on Percy test. I've checked with @sabrinang on this. (Note that this should resolve the a11y low contrast issue for placeholder text as well)
- Removed Open Sans from codebase / stopped loading it from Google Font as we don't use Open Sans anymore
- Updated base font family variable to reflect our current brand & removed extra type overrides (this shouldn't introduce any visual changes)
image

---

@sabrinang tagging you for review so you get a chance to look at the placeholder colour changes before we merge this PR


Pomax:
Looks like there might be a button styling issue:

image

mmmavis:
Ah will fix. Also interesting that Percy didn't pick up the background colour change on the disabled `bt-primary` button 🤔

mmmavis:
PR updated - cleaned up SCSS code for buttons.

Pomax:
That... is a good point, why _didn't_ it catch that different color? O_o

Issue #4560, opened 2020-04-27
[PNI] allow product updates to be featured


Pomax:
Apparently we never had one, which is kind of amazing.

- [x] add a featured field
- [x] update product view to order updates such that featured is on top

PR #4564, opened 2020-04-27
allow product updates to be featured

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4560


network-api/networkapi/buyersguide/pagemodels/product_update.py
```diff
@@ -0,0 +1,32 @@
...
+class Update(models.Model):
+    source = models.URLField(
+        max_length=2048,
+        help_text='Link to source',
+        blank=True,
+    )
+
+    title = models.CharField(
+        max_length=256,
+        blank=True,
```
mmmavis, line 13: This model doesn't have any required fields. Can we make `title` a required field?

network-api/networkapi/buyersguide/pagemodels/product_update.py
```diff
@@ -0,0 +1,32 @@
...
+    )
```
mmmavis, line 14: This model doesn't have any required fields. Can we make `title` a required field?

mmmavis:
Would be awesome if we could add a column for `featured` on Updates' admin panel so at a glance it's easy to spot featured updates.

Pomax:
good calls, added in!

Pomax:
filed https://github.com/mozilla/foundation.mozilla.org/issues/4565 to investigate that seed value oddity

PR #4579, opened 2020-04-29
Percy color test case [do not merge]

Pomax:
This PR exists to test whether percy catches something that it really should be catching: a background color difference on one of our button styles.


Pomax:
percy changed their fuzzing threshold so hopefully we don't run into this again... but we probably will.

Issue #4512, opened 2020-04-15
[PNI] post merge: update the panels so they're in the right spots.


Pomax:
Right now the general and software fieldpanels have been tacked onto the end of the base fieldpanels, which isn't as tidy as before.

This is a cosmetic, admin-only piece of work.

PR #4581, opened 2020-04-29
Reorder the PNI product fields in the CMS to match the on-page presentation of those fields.

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4512 by introducing a helper function that lets us inject (lists of) panels based on "the previous section's heading", with the general and software products both using that to inject their fields in the right spots.

### DIFF
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -190,7 +190,7 @@ def get_content_related_by_tag(page, result_count=3):
     return result_list[:result_count]
 
 
-def insert_panels_after(panels, more, after_label):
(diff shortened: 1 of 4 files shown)


network-api/networkapi/wagtailpages/utils.py
```diff
@@ -188,3 +188,22 @@ def get_content_related_by_tag(page, result_count=3):
         capture_exception(err)
 
     return result_list[:result_count]
+
+
+def insert_panels_after(panels, more, after_label):
```
Pomax, line 193: I'm not married to this signature, would `insert_panels_after(panels, label, more)` make more sense?

Issue #4558, opened 2020-04-24
[PNI] sort FieldPanel for related products differently from the default queryset ordering


Pomax:
PNI products are sorted on `id` (oldest first), but for FieldPanels we want this ordering to be alphabetical. Can we do this? Because right now I don't know how.

Related SO question for this: https://stackoverflow.com/questions/61414398/specifying-custom-ordering-of-m2m-relations-in-a-fieldpanel-for-snippet-register

Pomax:
Looks like the SO answer may need some more work, because applying the code (with some reordering to prevent a NameError) generates magic django errors.

PR #4604, opened 2020-05-05
Pni related fields

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4558

This sorts the `updates` and `related products` multi-select boxes in the admin edit view for products, without affecting the ordering of the normal products queryset (e.g. the thing that determines the order of products on the category pages, and in the general "pick one to edit" list in the CMS)


network-api/networkapi/buyersguide/admin.py
```diff
@@ -1,15 +1,43 @@
...
     Update,
     Product,
+    BaseProduct,
     GeneralProduct,
     SoftwareProduct,
     BuyersGuideProductCategory
 )
 
 
+# Effect customer ordering (in the admin only) for
```
Pomax, line 20:
```suggestion
# Effect custom ordering (in the admin only) for
```

Issue #4543, opened 2020-04-21
[PNI] remove old Product model and  voting code


Pomax:
Follow-up on https://github.com/mozilla/foundation.mozilla.org/issues/4486

- [x] unreference the old `Product` type
  - [x] remove the menu entry
  - [x] update the cache clear button to redirect to the page we were on, not /products
  - [x] remove any mention of `Product` in factory code
- [x] remove passthrough export for `Product` in the bg app `models.py`
- [x] remove the old voting code
- [x] remove the old type (deduplicates data)

And then possibly:

- [x] rename the new type to the old type
  - [x] rename all simple classes
  - [x] prefactor `BaseProduct` to have an alias `Product`, and have everything use that in preparation  for the last step
  - [x] rename `BaseProduct` to `Product`

Note that due to the inheritance we have set up, renaming `BaseProduct` to `Product` is going to care _a lot_ of care, because need to migration all `GeneralProduct` and `SoftwareProduct` such that their `baseproduct_ptr`, pointing to the superclass record, becomes a `product_ptr` that points to the renamed superclass record.

Pomax:
changed the estimate to 13 because the last step in particular was much harder than expected, because it's not something Django can do "on its own" right now (there's a PR open to teach Django how to rename superclass models, but it's from May 2019 and has not seen any activity recently)

PR #4612, opened 2020-05-07
Pni remove old product 4

Pomax:
Partial fulfillment of #4543

Aliases `BaseProduct` as `Product`, but stops short of the final model rename because https://stackoverflow.com/questions/61665607/renaming-a-django-superclass-model-and-updating-the-subclass-pointers-correctly


network-api/networkapi/management/commands/image_migration.py
```diff
@@ -1,15 +1,15 @@
 from django.core.files.uploadedfile import UploadedFile
 from django.core.management.base import BaseCommand
 
-from networkapi.buyersguide.models import Product
+from networkapi.buyersguide.models import BaseProduct
```
cadecairos, line 4: Should we also update this reference?

network-api/networkapi/management/commands/flush_models.py
```diff
@@ -16,7 +16,7 @@
     InternetHealthIssue,
 )
 from networkapi.wagtailpages.models import CTA
-from networkapi.buyersguide.models import Product
+from networkapi.buyersguide.models import BaseProduct
```
cadecairos, line 19: Should we also update this reference to BaseProduct?

cadecairos:
I misread the diff previously. R+

PR #4614, opened 2020-05-08
Add styling for button active state

### LINKED ISSUE #3608: Implement Button Active State
Design issue #3367 

**Primary active:** $light-blue with $white text
**Secondary active:** $blue with $white text
**Pop active:** $black with $white text

**Inverse Primary active:** $dark-blue with $black text
**Inverse Secondary active:** $blue with $black text
**Inverse Pop active:** $white with $black text

Screen Shot 2019-09-04 at 12 45 44 PM
on style guide sketch file

mmmavis:
Closes #3608

Implemented button active style & SCSS code refactor for buttons

To review this PR, you can use Firefox's devtool to toggle pseudo-classes styling. See https://developer.mozilla.org/en-US/docs/Tools/Page_Inspector/How_to/Examine_and_edit_CSS#Viewing_common_pseudo-classes


source/sass/buttons.scss
```diff
@@ -62,81 +77,86 @@ button,
...
-
-  $default-hover: (
-    "color": $white,
-    "background-color": $dark-blue,
-    "border-color": none
+  $regular-mode: (
+    "default": (
+      "color": $white,
+      "background-color": $blue,
+      "border-color": transparent
```
Pomax, line 84: Do we really want `border-color: transparent`, rather than `border: none`? (do we still rely on the border's width to affect positioning?)

source/sass/mixins.scss
```diff
@@ -118,76 +133,50 @@
...
   }
 
-  &.disabled,
-  &:disabled,
-  &[disabled] {
+  @include disabled {
+    @include button-base-colors($default-colors);
+  }
+
+  &:not(:disabled):not(.disabled) {
```
Pomax, line 161: Does `&:not(:disabled, .disabled)` work? Or does each thing need to be wrapped in its own `not()`?

source/sass/buttons.scss
mmmavis: `none` is not a valid value for `border-color`. We can also update `border-color` to the same color variable `background-color` is set to, but I find setting it to`transparent` is more easy to read (and parse in my head).

source/sass/mixins.scss
mmmavis: I think `:not()` only allow one argument.

source/sass/buttons.scss
mmmavis: I removed the `border-color` rules for `.btn-primary` as it by default is border-less (line 78). 

For `btn-secondary`:
`none` is not a valid value for `border-color`. We can also update `border-color` to the same colour variable `background-color` is set to, but I find setting it to `transparent` is easier to read (and parse in my head).

Pomax: okay, so we do intentionally want to keep the border there, just acting as another way to force a margin?

mmmavis: Yes, some button states have border so I thought keeping them consistent is easier to manage in code.

Screen Shot 2019-09-04 at 12 45 44 PM

mmmavis: Yes, default state for btn-secondary has border. If we remove border for the hover and active state, button position will shift as its state change from default to hover/active.

Screen Shot 2019-09-04 at 12 45 44 PM

PR #4618, opened 2020-05-11
rename BaseProduct back to Product

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4543 by renaming `BaseProduct` back to `Product`, using a custom Django migration that first "undoes" the model hierarchy as far as Django knows _while migrating_, then renaming the superclass and relinking the subclasses to the new model name.

### DIFF
--- a/network-api/networkapi/buyersguide/models.py
+++ b/network-api/networkapi/buyersguide/models.py
@@ -3,15 +3,15 @@ from .pagemodels.get_product_image_upload_path import (
 )
 
 from .pagemodels.base_voting import (
+    BooleanProductVote,
+    BooleanVote,
+    BooleanVoteBreakdown,
     ProductVote,
     RangeProductVote,
-    BooleanProductVote,
-    VoteBreakdown,
-    BooleanVoteBreakdown,
+    RangeVote,
     RangeVoteBreakdown,
     Vote,
-    BooleanVote,
-    RangeVote,
+    VoteBreakdown,
 )
 
 from .pagemodels.cloudinary_image_field import (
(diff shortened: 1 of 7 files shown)


network-api/networkapi/buyersguide/models.py
```diff
@@ -3,15 +3,15 @@
 )
 
 from .pagemodels.base_voting import (
-    ProductVote,
-    RangeProductVote,
     BooleanProductVote,
-    VoteBreakdown,
+    BooleanVote,
```
Pomax, line 7: sort-lines-alphabetically

network-api/networkapi/buyersguide/pagemodels/base_voting.py
```diff
@@ -6,34 +6,47 @@
 from .products.base import Product
 
 
-class ProductVote(models.Model):
-    votes = models.IntegerField(
-        default=0
-    )
+class Vote(models.Model):
```
Pomax, line 9: The only thing I did to this file is move the classes around so that they're always "X", then "BooleanX", then "RangeX", because I had to look things up in this file and the nonuniform ordering made that harder than it had to be.

network-api/networkapi/buyersguide/utils.py
```diff
@@ -12,3 +15,29 @@ def quad_to_tri(input):
     if input == 'No':
         return False
     return None
+
+
+class AlterModelBases(ModelOperation):
```
Pomax, line 20: I added this as a util function rather than putting this in the migration itself, *just in case* we need it in the future, at which point we can move it into the utils file for "everything" rather than just "the buyers guide"

PR #4620, opened 2020-05-12
SCSS breakpoint improvements

### LINKED ISSUE #4471: SCSS cleanup: assign breakpoint variables to a common variable
Similar to what we did in https://github.com/mozilla/foundation.mozilla.org/pull/4467/files#diff-a7e457c11566b243d750e6a6acd87ad7

```scss
.parent-class {
  @media screen and (min-width: $bp-lg) {
    // blah blah
  }

  .child-class {
    // blah blah

    @media screen and (min-width: $bp-lg) {
      // blah blah
    }
  }

  .child-class-2 {
    // blah blah

    @media screen and (min-width: $bp-lg) {
      // blah blah
    }
  }

  ...
  ...
  ...
}
```

Can be simplified to

```scss
.parent-class {
  $breakpoint: $bp-lg;

  @media screen and (min-width: $breakpoint) {
    // blah blah
  }

  .child-class {
    // blah blah

    @media screen and (min-width: $breakpoint) {
      // blah blah
    }
  }

  .child-class-2 {
    // blah blah

    @media screen and (min-width: $breakpoint) {
      // blah blah
    }
  }

  ...
  ...
  ...
}
```

This way if we ever need to update breakpoint to something else (for example, `$bp-lg` to `$bg-md`) we only have to update the `$breakpoint` variable assignment instead of find-and-replace multiple times in the same parent class ruleset scope.

mmmavis:
Closes #4471

Assign "shared" breakpoint used in media query to SCSS variable. Note that I did this SCSS improvement only* to breakpoints that are shared by multiple DOM elements **in a meaningful way**.  There are still many instances of `@media (min-width: $bp-blah)` in the codebase because I couldn't find any correlation between them in order to store `$bp-blah` to a shared SCSS variable.

### DIFF
--- a/network-api/networkapi/buyersguide/templates/primary_nav.html
+++ b/network-api/networkapi/buyersguide/templates/primary_nav.html
@@ -4,7 +4,7 @@
 
 {% block narrow_screen_menu %}
   <div class="narrow-screen-menu-background">
-    <div class="narrow-screen-menu-container">
+    <div class="narrow-screen-menu-container d-md-none">
 
       <div class="container text-center">
         <div class="row">
--- a/source/js/buyers-guide/components/primary-nav/primary-nav.scss
+++ b/source/js/buyers-guide/components/primary-nav/primary-nav.scss
@@ -1,3 +1,5 @@
+$pni-primary-nav-breakpoint: $bp-sm;
+
 .moz-logo {
   background-image: url("../_images/mozilla-m.svg");
   flex-shrink: 0;
@@ -36,7 +38,7 @@
     font-weight: 600;
     max-width: 85px;
 
-    @media (min-width: $bp-sm) {
+    @media (min-width: $pni-primary-nav-breakpoint) {
       font-size: 22px;
       max-width: none;
     }
@@ -52,7 +54,7 @@
   position: absolute;
   padding-top: 62px;
 
-  @media (min-width: $bp-sm) {
+  @media (min-width: $pni-primary-nav-breakpoint) {
     padding-top: 72px;
   }
 
@@ -117,10 +119,6 @@
     background: $white;
     height: 100%;
   }
-
-  @media (min-width: $bp-md) {
-    display: none;
-  }
 }
 
 .btn-donate {
@@ -141,7 +139,7 @@
     display: none;
   }
 
-  @media (min-width: $bp-sm) {
+  @media (min-width: $pni-primary-nav-breakpoint) {
     &.menu-open {
       display: initial;
     }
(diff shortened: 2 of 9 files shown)


network-api/networkapi/buyersguide/templates/primary_nav.html
```diff
@@ -4,7 +4,7 @@
 
 {% block narrow_screen_menu %}
   <div class="narrow-screen-menu-background">
-    <div class="narrow-screen-menu-container">
+    <div class="narrow-screen-menu-container d-md-none">
```
mmmavis, line 7: removed SCSS code in favour of this

source/sass/buyers-guide/views/product.scss
```diff
@@ -1,7 +1,4 @@
-$product-detail-padding-x: (
```
mmmavis, line 1: not needed as a variable since it's only used in one ruleset in this file (`.product-section-padding-x { }`)

network-api/networkapi/buyersguide/templates/primary_nav.html
Pomax: nice, taking advantage of bootstrap <3

source/js/buyers-guide/components/primary-nav/primary-nav.scss
```diff
@@ -1,3 +1,5 @@
+$pni-primary-nav-breakpoint-larger: $bp-sm;
```
Pomax, line 1: it feels odd to see "breakpoint-larger" but then have a value "sm" (a bit like seeing `$bold: 100` =), can we think of a better name for this breakpoint?

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -1,10 +1,12 @@
 #view-home {
+  $breakpoint-larger: $bp-md;
```
Pomax, line 2: probably here too, "larger" is being set to "medium", we probably want a clearer name

source/js/buyers-guide/components/primary-nav/primary-nav.scss
mmmavis: I know but this is tricky. Maybe we can just call it `$pni-primary-nav-breakpoint`

source/sass/mozfest.scss
```diff
@@ -19,6 +19,7 @@ body.mozfest {
 
   &#view-mozfest-home {
     #hero {
+      $breakpoint-larger: $bp-md;
```
Pomax, line 22: are these technically all the "same" breakpoint? There's a lot of declarations for this same-named var, I'm wondering if we can just add this to our list of global variables instead.

mmmavis: Good point. Each page / component were designed individually. It's coincidental that some breakpoints happen to be the same. That's why I think these variables should stay in local scope instead being declared globally.

PR #4625, opened 2020-05-12
I4330

### LINKED ISSUE #4330: Add meta tag to remove staging websites from Google search
Update from Gideon:
We've noticed that several staging websites popup at some points in Google searches (donate-wagtail, fo'mo, and pulse).

To fix this, let's add `<meta name="robots" content="noindex">` to all pages unless we're in a production environment.
Alternatively, if it is easier to do this with a header (via CDN or code), we can also just set `X-Robots-Tag: noindex` on those pages.

---

I was searching for different MozFest terms on Google this morning and found that some of the top results for some terms were for mozillafestival.mofostaging.net, which is likely because the production site hasn't been re-crawled yet.

cadecairos:
Closes #4330 
## Checklist

**Tests**

**Documentation:**
- [x] Is my code documented?


network-api/networkapi/settings.py
```diff
@@ -28,62 +28,60 @@
...
     SSL_REDIRECT=bool,
-    DOMAIN_REDIRECT_MIDDLEWARE_ENABLED=(bool, False),
-    MOZFEST_DOMAIN_REDIRECT_ENABLED=(bool, False),
+    STATIC_HOST=(str, ''),
     TARGET_DOMAINS=(list, []),
+    USE_CLOUDINARY=(bool, False),
     USE_S3=(bool, True),
     USE_X_FORWARDED_HOST=(bool, False),
+    XROBOTSTAG_ENABLED=(bool, False),
     XSS_PROTECTION=bool,
```
cadecairos, line 84: Alphabetic sorting. Only added `XROBOTSTAG_ENABLED` on line 83

network-api/networkapi/utility/middleware.py
```diff
@@ -22,6 +22,18 @@ def __call__(self, request):
         return response
 
 
+class XRobotsTagMiddleware:
```
Pomax, line 25: 🤖

PR #4630, opened 2020-05-13
Remove PNI extra code for share buttons & tweak share button styling

### LINKED ISSUE #4629: Button visual regressions
We’re having some visual issues with buttons, which I think are due to https://github.com/mozilla/foundation.mozilla.org/commit/6d9ffd442f14f041c0e4c6f5573d4b8d9e064e2c
@mmmavis any chance we could add this line back, or find a different solution?

Creep-o-meter post-vote screen:

image

Footer (French), where this used to fit in one line

image

Footer (Portuguese)

image

mmmavis:
Closes #4629

Go to any PNI product page's voting section to test. E.g., https://foundation-s-issue-4629-wc70yn.herokuapp.com/en/privacynotincluded/products/weight-join/

Note:
Percy flags are legit - share buttons weren't in the same size and this PR fixes it.

---

@beccaklam I fixed the issue Theo pointed out in #4629. I also discovered other issues we had been having since we added those buttons (buttons weren't always the same size and on certain screen they looked [funky](https://user-images.githubusercontent.com/2896608/81872524-be2ee500-952e-11ea-92f4-ced1814ba820.png)). I added annotation below to explain the issues. If we want the four buttons to always be the same width, we 

1. Either have to shrink button's default padding/margin just for this use case
2. OR make buttons stacked on non-extra large screens just for this use case

Let me know which one you prefer! :D

The following screenshots are taken from this PR's review app (not on production).

screen-1201px(extra large)
screen-1199px(large)
screen-991px(medium)
screen-767px(small)


source/js/buyers-guide/components/creep-vote/creep-vote.scss
```diff
@@ -106,20 +106,4 @@ $btn-shadow-width: 3px;
 
     background-color: $gray-40;
   }
-
-  .share-button-group.rectangle {
```
mmmavis, line 110: We shouldn't need these overrides.

source/js/components/share-button-group/share-button-group.scss
```diff
@@ -1,13 +1,12 @@
 .share-button-group {
```
mmmavis, line 1: `marign-right` didn't get added to the right elements. Buttons weren't always in the same size as a result.

beccaklam:
Hey @mmmavis thanks for your great annotation. I think I prefer the stacked button solution, how about you?

mmmavis:
@beccaklam Yea I prefer stacked version as well. Will update this PR and ask for your design feedback!

source/js/components/share-button-group/share-button-group.scss
```diff
@@ -23,24 +22,35 @@
...
+
       &:last-child {
         .btn:last-child {
           margin-right: 0;
         }
       }
     }
 
-    @media (max-width: $bp-md) {
+    @include media-breakpoint-down("sm") {
```
mmmavis, line 36: I find using this Boostrap mixin less confusing for `max-width` media query:

`@media (max-width: $bp-md) {` to target screen sizes of `sm`  and smaller
vs
`@include media-breakpoint-down("sm") {`

mmmavis:
Note:
Percy flags are legit - those share buttons weren't in same size and this PR fixes it.

source/js/components/share-button-group/share-button-group.scss
Pomax: Hm, does it behave the same though? max-width: medium is not the same as "small and below", there's a gap in between those two values.

mmmavis: it's the same here. `$bp-md` is the minium width of Bootstrap defined "medium device". So `(max-width: $bp-md)` essentially means "small and below"

mmmavis:
@beccaklam 320px wide screen is too narrow for our stacked share buttons to show. If we'd like to have them stacked one per row we should file a new ticket and update our style guide to reflect this new style for very small screen.

image

beccaklam:
> @beccaklam 320px wide screen is too narrow for our stacked share buttons to show. If we'd like to have them stacked one per row we should file a new ticket and update our style guide to reflect this new style for very small screen.
> 
> image

Ah I see. Okay, would you like me to file that ticket and merge this for now?

mmmavis:
@beccaklam if you could file the ticket and R+ this PR that would be awesome! ✨

mmmavis:
Thanks @beccaklam !

Issue #4613, opened 2020-05-07
[Blog page] Add Pulse component 


kristinashu:
Kevin has requested we add the Pulse component as an option to the blog templates.

PR #4632, opened 2020-05-13
add pulselisting as valid blog content block

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4613 by adding the `PulseProjectList` component as one of the selectable streamfield blocks for blog pages.


mmmavis:
Once this gets merged, we will have one more streamfield to deal with for blog post print mode heh

PR #4634, opened 2020-05-13
Added nowrap rule back from buttons

### LINKED ISSUE #4629: Button visual regressions

mmmavis:
Related to #4629 (fixes the sign up button issue)

Confirmed with Sabrina that we are okay with adding nowrap rule back.


Pomax:
looks like it might be affecting more than we intend?

image

mmmavis:
@Pomax I'm afraid we can't do anything about loooong button text exceeding viewport. We can't conditionally wrap button text based on it's container size and we can't predict how long a button text is going to be. We always discourage long button text so I guess this best practice helps a bit?

PR #4658, opened 2020-05-18
pni homepage: un-nest figcaption

### LINKED ISSUE #4642: [pni] homepage: un-nest figcaption
The `<figcaption>` that we use in `<figure>` elements for each product cannot be nested inside another element and needs to be a direct child of `<figure>`.

berliners1:
Closes #4642

I made `<figcaption>` a direct child of `<figure>` by changing the `<a>` tag to be the parent of `<figure>` instead of its child. I then adjusted homepage.scss to account for this change to keep functionality the same.

### DIFF
--- a/network-api/networkapi/buyersguide/templates/buyersguide_home.html
+++ b/network-api/networkapi/buyersguide/templates/buyersguide_home.html
@@ -56,36 +56,36 @@
       {% for product in products %}
         {% if product.is_current %}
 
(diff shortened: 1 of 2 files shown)


Pomax:
closing in favour of https://github.com/mozilla/foundation.mozilla.org/pull/4681, which does not require scss changes.

berliners1:
Makes sense, I'll keep that in mind. Thanks!

Issue #4645, opened 2020-05-14
[pni] product: thumbnail is missing "alt" attribute


Pomax:
Our product thumbnail does not have an alt attribute O_o

Issue #4644, opened 2020-05-14
[pni] category: thumbnails are missing "alt" attribute


Pomax:
Our product thumbnails do not have an alt attribute, I don't even know how that happened O_o

PR #4670, opened 2020-05-19
fix missing alt attribute and collapse cloudinary vs direct serving

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4645, closes https://github.com/mozilla/foundation.mozilla.org/issues/4644

This change adds `alt` attributes where they were missing, primarily by collapsing "cloudinary" vs "direct from cms" image serving in a way that ensures there is only one `` tag, but with conditionally different `src` attributes.

Viewing both the PNI category view and product view should not generate any AXE errors (locally as well as in the review app)


Pomax:
Hmm, it looks like the image widths on the product page changed ever so slightly...

PR #4678, opened 2020-05-20
Remove PNI old mobile nav code

### LINKED ISSUE #4619: PNI: remove unused "narrow menu" code
Just spotted some [potential dead code](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/templates/primary_nav.html#L5-L24) for PNI's nav. 

I can't recall if PNI ever had a toggleable nav menu for mobile and I can't seem to find it on PNI's Sketch file. (By toggleable mobile nav I mean something like Foundation site's 🍔 button + black overlay mobile nav menu). If the above code is not needed, we should remove it + its related SCSS.

mmmavis:
Closes #4619


source/js/common/initialize-primary-nav.js
```diff
@@ -7,6 +7,8 @@ import navNewsletter from "../nav-newsletter.js";
  * @param {Object} primaryNavModule primary nav module to initiate
  */
 export const initializePrimaryNav = (siteUrl, csrfToken, primaryNavModule) => {
-  primaryNavModule.init();
+  if (primaryNavModule && typeof primaryNavModule.init === `function`) {
```
Pomax, line 10: We should be able to assume that if `primaryNavModule` exists, it's the correct shape and `primaryNavModule.init()` will work. If not, we probably have a problem elsewhere (as in, we're calling initializePrimaryNav with something in the spot for `primaryNavModule` that isn't actuallly a `primaryNavModule`...)

mmmavis: Alright. Will update!

PR #4687, opened 2020-05-23
Typeform block

### LINKED ISSUE #4652: Embed Typeform on Pages
@cadecairos is in progress of making a prototype of embedding the form on wagtail pages: https://github.com/mozilla/foundation.mozilla.org/pull/4526 and I have left some feedback on it https://github.com/MozillaFoundation/Design/issues/512#issuecomment-616817785

Using this issue to track progress of integrating typeforms on our site and refining front-end issues the user may experience.

Design issue here: https://github.com/MozillaFoundation/Design/issues/512

mmmavis:
Based on https://github.com/mozilla/foundation.mozilla.org/pull/4526
Closes #4652

**Test pages** (feel free to test this streamfield on other pages) 
- https://foundation-s-mavis-type-atio4y.herokuapp.com/en/campaigns/typeform-test/
- https://foundation-s-mavis-type-atio4y.herokuapp.com/en/about/

---
Note:
Need to whitelist `https://embed.typeform.com` from CSP script source & `https://devopstypeform.typeform.com` from CPS frame source to make this work


source/js/embed-tf.js
```diff
@@ -0,0 +1,29 @@
...
+    document.addEventListener("DOMContentLoaded", () => EmbedTF.setup());
+  },
+  setup: function() {
+    let formMeta = document.querySelector("meta[name=typeform-formurl]");
+
+    if (!formMeta) {
+      return;
+    }
+
+    let popup = window.typeformEmbed.makePopup(
```
mmmavis, line 12: https://developer.typeform.com/embed/modes/#popup-mode

source/js/embed-tf.js
```diff
@@ -0,0 +1,29 @@
+let EmbedTF = {
+  init: function() {
+    document.addEventListener("DOMContentLoaded", () => EmbedTF.setup());
+  },
+  setup: function() {
+    let formMeta = document.querySelector("meta[name=typeform-formurl]");
```
Pomax, line 6:
```suggestion
    const formMeta = document.querySelector("meta[name=typeform-formurl]");
```

source/js/embed-tf.js
```diff
@@ -0,0 +1,29 @@
+let EmbedTF = {
```
Pomax, line 1:
```suggestion
const EmbedTF = {
```

source/js/embed-tf.js
Pomax, line 12: We definitely want to put this behind `if (window.typeformEmbed) { ... }` because that property may not exist (bad networking, script blockers, etc)

source/js/embed-tf.js
```diff
@@ -0,0 +1,29 @@
...
+        hideHeaders: true,
+        hideFooter: true
+      }
+    );
+
+    document
+      .querySelector("#btn-typeform-popup")
+      .addEventListener("click", () => {
+        popup.open();
+      });
```
Pomax, line 25:
```suggestion
      .addEventListener("click", () => popup.open());
```

source/js/embed-tf.js
```diff
@@ -0,0 +1,29 @@
+let EmbedTF = {
+  init: function() {
+    document.addEventListener("DOMContentLoaded", () => EmbedTF.setup());
```
Pomax, line 3: This will not fire if `DOMContentLoaded` has already happened. As per https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event#Checking_whether_loading_is_already_complete, let's change this to:
```suggestion
    if (document.readyState === 'loading') {
      document.addEventListener('DOMContentLoaded', () => EmbedTF.setup());
    } else {
      EmbedTF.setup();
    }
```

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
```diff
@@ -0,0 +1,5 @@
+<script src="https://embed.typeform.com/embed.js" ></script>
```
Pomax, line 1: Turns out this is not how we should be loading css/js for streamfields: https://stackoverflow.com/a/43561880/740553

network-api/networkapi/wagtailpages/pagemodels/customblocks/typeform_block.py
```diff
@@ -0,0 +1,24 @@
...
+            ('btn-secondary', 'Secondary button'),
+        ],
+        default='btn-primary',
+    )
+
+    button_text = blocks.CharBlock(
+        required=True
+    )
+
+    class Meta:
```
Pomax, line 21: As per https://stackoverflow.com/a/43561880/740553:

```suggestion
    @property
    def media(self):
        return forms.Media(
            js=["https://embed.typeform.com/embed.js"],
        )

    class Meta:
```

source/js/embed-tf.js
mmmavis: Nice catch!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
mmmavis: @Pomax the link you included seems to be the way to inject JS to Wagtail **admin** but not the published page. I'm gonna keep this `<script>` line.

mmmavis:
@Pomax I've fixed most things you suggested. Didn't change how this PR included an extra Typeform script because the [Stackoverflow post](https://stackoverflow.com/questions/43558572/wagtail-streamfield-block-injecting-jquery-into-the-wagtail-admin/43561880#43561880) you mentioned seems to be the way to inject script to Wagtail "admin" but not the published page.

I've also updated file name and variable name from `embed-tf / EmbedTF` to `embed-typeform / EmbedTypeform` so they are more self-explanatory.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/typeform_block.html
Pomax: ahhhhh damn that is a very good point. I swear there was a way to inject a script "only once" this way too.... O_o

Issue #4624, opened 2020-05-12
Front-end: Add featured post on /blog


kristinashu:
Follow up from: Featured post on /blog #3693

Add the ability to feature two blog posts at the top of /blog.

Comp:
![blog w featured copy@2x](https://user-images.githubusercontent.com/1068783/65466960-12a51900-de15-11e9-86ec-22d306e39165.jpg)

Style of the thumbnail should be the same as on https://foundation.mozilla.org/
<img width="1431" alt="image" src="https://user-images.githubusercontent.com/1068783/65467034-4aac5c00-de15-11e9-9f4e-e780c67abaa0.png">

Which blogs to feature should be selected on the /blog cms page:
<img width="1435" alt="image" src="https://user-images.githubusercontent.com/1068783/65467236-d0c8a280-de15-11e9-81bb-5821d6cf1851.png">

Pomax:
Tasks
- [x] add featured blog posts to BlogIndexPage
- [x] extend the indexpage.html template to blogindexpage.html
- [x] set up a two-colum layout above the main content listing
- [x] reuse the homepage "featured blog post" template to show the feature blog posts
- [ ] ensure that featured blog posts are excluded from the general set so there's no duplication?

Issue #3714, opened 2019-10-02
Selection order of blog posts on homepage


kristinashu:
Jen has pointed out that it would be better if the most recent posts were at the top.

Screen cap from prod:
<img width="1434" alt="image" src="https://user-images.githubusercontent.com/1068783/66071078-567bda80-e507-11e9-84cd-84620e0e7595.png">

Pomax:
I've filed a Stackoverflow question to hopefully find out whether this is possible at all - https://stackoverflow.com/questions/61353431/sorting-pages-in-a-wagtail-pagechooserpanel-dialog

PR #4689, opened 2020-05-25
Featured blog posts

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4624 by introducing "featured blog pages" for the blog index page, with template updates to make sure they show up with the same styling as the "featured posts"/"latest news" on the homepage

- /blog should show 2 featured posts, with those not showing the subsequent set of blog posts
- /category pages should not show featured posts
- /tags pages technically shouldn't either, but we don't really expose those so... win?

Test URL: https://foundation-s-featured-b-oyndlu.herokuapp.com/en/blog/


network-api/networkapi/wagtailpages/templates/wagtailpages/blog_index_page.html
```diff
@@ -0,0 +1,15 @@
+{% extends "./index_page.html" %}
+
+{% load wagtailcore_tags wagtailimages_tags i18n %}
+
+{% block featured %}
```
Pomax, line 5: the new part, which injected featured posts as long as we're not on a /category or /tag view

network-api/networkapi/wagtailpages/templates/wagtailpages/blog_index_page.html
```diff
@@ -0,0 +1,15 @@
...
+  {% if not filtered %}
+    <div class="row">
+      {%  for feature in page.specific.featured_pages.all %}
+      <div class="col-6 mb-5">
+        {% include "./fragments/featured_blog.html" with blog=feature %}
```
Pomax, line 10: refactored-to-stand-alone featured post as found on the homepage

Pomax:
@kristinashu using the standard template for latest news/featured blog as we do on the homepage can lead to what feels like oddly misaligned hbars:

image

We don't notice this on the homepage because we only feature one thing per section, but do we want to some kind of "height based on lines" with a `text-overflow: ellipsis` if the SEO search text runs longer than that, so that the hbars are guaranteed height-aligned?

kristinashu:
Thank you @Pomax! 

## hbar alignment 
The misaligned hbars issue seems to have been fix on our homepage's "Featured stories" section. Can you you do something like that? (screen cap from prod)
image


## Narrow widths
The two posts will need to stack on top of each other narrower widths. (see issue in this screen cap from PR)
image

Could it stack similar to the homepage? (screen cap from prod)
Screen Shot 2020-05-25 at 4 44 09 PM


## Selection order
For selection which posts to feature, can the selection order show the most recent at the top? Related to https://github.com/mozilla/foundation.mozilla.org/issues/3714.

Pomax:
The entries can be ordered in the CMS itself, so it feels like that doesn't need explicit hardcoding:

image

kristinashu:
That feature is perfect. What I mean when staff choose which blog post to feature there. Can the most recent blog post show up at the top? I think you updated this on the homepage recently. 
(screen cap from PR)
image

kristinashu:
Or actually, maybe it is showing the most recent at the top now.

kristinashu:
Confirmed, it's working correctly ( I swear it wasn't yesterday lol). 

image

Pomax:
I think it shows the most recently edited page, rather than published - we can change that, but let's file a separate issue for that, as that's code in completely different files =P

Pomax:
@kristinashu ready for re-review

Pomax:
@mmmavis this is ready for code review

Pomax:
Looks like we don't control that - I've filed https://github.com/wagtail/wagtail/issues/6080 so that the wagtail folks are aware of this issue.

Issue #4641, opened 2020-05-14
[pni] cross page: ensure landmark elements exist


Pomax:
We should have `<header>`, `<main>` and `<footer>` at the least, but we currently only have that last one.

PR #4700, opened 2020-05-27
Pni landmarks

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4641 by adding `<header>`, `<nav>`, `<main>`, and `<footer>` elements where missing

Also fixes the heading-skip issue in the prefooter ("we all love...")


network-api/networkapi/buyersguide/templates/bg_base.html
```diff
@@ -71,10 +71,12 @@
 
 {% block body_wrapped %}
 
+  <header>
   {% include "./primary_nav.html" %}
+  </header>
 
   <div class="underneath-screen-overlay">
-    <div class="category-nav">
+    <nav class="category-nav" title="category navigation">
```
mmmavis, line 79: PNI nav includes not only links for categories but also `Home` and `About`. Can we remove this `title` and add `aria-label` to category related nav links so they can be more descriptive for assistive technology.

Pomax: we cannot remove the title: it is required to inform whatever accessibility tool is look at this what the purpose of this `<nav>` is, because you can have a million of them on a single page.

mmmavis: just "site navigation"? 🤔

Pomax: Sure - it kind of sucks that PNI is just a bunch of pages instead of its own domain, making this way weirder than it has to be, but "site navigation" is good enough until we revisit.

Pomax, line 79:
```suggestion
    <nav class="category-nav" title="site navigation">
```

Issue #4647, opened 2020-05-14
[pni] about page: don't skip heading levels


Pomax:
Headings should not skip levels - particularly the methodology page has a lot of `<h3>` that should be `<h2>`

PR #4702, opened 2020-05-27
fixes headings on some about pages

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4647


network-api/networkapi/buyersguide/templates/about/methodology.html
```diff
@@ -33,10 +33,10 @@ <h1 class="h2-heading mb-4">
     </p>
   </div>
 
-  <article class="mt-5 mb-4 about-methodology">
-    <h3>
+  <div class="mt-5 mb-4 about-methodology">
```
mmmavis, line 36: Since we are editing this file, instead of `div` we should use `section` so we have a better HTML outline for this page. Note that `<section>` should have a heading element inside to help identify the section. This also means all the existing `<section>` on this file should be updated to include a heading element.

This is a great tool to check out HTML outline: https://validator.w3.org/nu/ 
(choose "show outline") 
image

Pomax: That is definitely more work than this issue requires: I changed it from article to div because these are _very much not_ articles - if we want to change this to section, let's do that in a different issue.

mmmavis: Ticket filed: https://github.com/mozilla/foundation.mozilla.org/issues/4706

Issue #4677, opened 2020-05-20
Removing time commitment labels from "Participate" page


taisdesouzalessa:
**Issue:**
In the past we decided to have time commitment labels on our [Participate page](https://foundation.mozilla.org/en/participate/). 
• Those aren't relevant to us anymore
• They complicate decision-making due to lack of clarity on the label system  (like: "should we add a 10 minutes time commitment because it is the time it takes people to apply for it or should we add 1 month because this is the duration of the program once people are accepted?")
• They add  noise to the layout.

<img width="1105" alt="image" src="https://user-images.githubusercontent.com/22157921/82493142-4fa4d680-9a9c-11ea-8198-b9aafe013e64.png">

**Solution:**
Could we remove those labels from this page? Ping me if you have any questions or need help to figure out solving any design consequences this decision may have. I envision that is just a label removal but it could be more complicated than that.

psjishnu:
@taisdesouzalessa Can you please mention the part of code that needs to be modified .

Pomax:
@taisdesouzalessa Do we want to keep this estimate in the CTA model, or should those be thrown away, too?

taisdesouzalessa:
by CTA model you mean this section here, @Pomax?

Pomax:
Ah, no I mean the field in the CMS:

![image](https://user-images.githubusercontent.com/177243/83068139-ad956900-a01c-11ea-909e-5d0d7d98711e.png)

taisdesouzalessa:
Ahh I see thanks for adding clarity here :). We should remove it the backend field as well! I don't think we use that feature anywhere else and there is nothing like that in the updated IA project we are working on.

Pomax:
gotcha - to the code-deleting machine! 🌀

PR #4704, opened 2020-05-27
remove all commitment code

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4677 by removing the model fields for commitment, as well as removing the commitment presentation from the template code, `card` template tag, and stylesheets


Pomax:
good catch - removed!

Pomax:
migration "rebased" to play as 101, on top of 100 =)

Issue #4220, opened 2020-02-18
Improve template names


kristinashu:
Based on discussion in https://github.com/mozilla/foundation.mozilla.org/issues/4064

#### 1) Rename `opportunity` template to `default`
- this template is used for _short_ lived info
- any staff can make anytime
- allows for a lot of subpages with longer names

#### 2) Rename `bannered campaign` template to `banner`
- layout is used a lot and not only for campaigns
- used when a page is more important and longer lasting
- receives some design attention
- sub nav is limited
- optional petition or newsletter signup snippet

cc @xmatthewx

Pomax:
I would like to suggest considering the word `default` reserved - it means too many things in the various programming language we use, better not to overload it more than out tech stack already overloads it.

Could we name it based on that short-lived-ness? Would `shortterm` be an option?

kristinashu:
Hmm what about `standard` or `basic` or `simple`? The idea we're trying to get across to staff is that this is the simplest template and should be the go-to template if they aren't sure which one to use.

Pomax:
`basic` sounds appropriate here.

Pomax:
@kristinashu just to confirm, we still want to do this?
- `OpportunityPage ` => `Basic`
- `BanneredCampaignPage` => `Banner`

kristinashu:
Yes, still want to go ahead with those two renames. 

I don't fully understand your comment in parenthesis. Is it a concern about the names?

Pomax:
It was, the original request is to rename them to "basic" and "banner", but in order for the names to stay in line with all the code we've written, they'll have to be "basic page" and "banner page".

kristinashu:
Ok, "basic page" and "banner page" work for me.

Pomax:
Immediate tasks:
- [ ] set Meta name for OpportunityPage to read "Basic page"
- [ ] set Meta name for BanneredCampaignPage to read "Banner page"

Possible follow-up:
- [ ] rewrite the OpportunityPage model to actually be BasicPage in the code
- [ ] rewrite the BanneredCampaignPage model to actually be BannerPage in the code

PR #4705, opened 2020-05-27
rename the opportunity and banneredcampaign pages, using meta-only

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4220 by setting the verbose names for these models to their suggested new names


Pomax:
@kristina I didn't add in any sorting, mostly because that would be _really hard_, but the review app should now show "Basic Page" and "Banner Page" under campaigns, rather than "OpportunityPage" and "BanneredCampaignPage".

Pomax:
@kristinashu I didn't add in any sorting, mostly because that would be _really hard_, but the review app should now show "Basic Page" and "Banner Page" under campaigns, rather than "OpportunityPage" and "BanneredCampaignPage".

Issue #4638, opened 2020-05-14
[pni] cross page: "Join us" content exists on page is invisible state


Pomax:
The "Join us" content is invisible instead of display:none, which means tabbing to the main content will "get stuck" cycling through everything in the "join us" newsletter slide-out panel, without any indication that that's where focus is.

We can either make that slide-out be `display:none` with extra work to make sure the slide out/slide in transitions work properly, or we can make it so that if there is focus anywhere inside the slide-out, the slide-out itself gets revealed (so that tabbing into it reveals it, and tabbing out of it hides it again)

Pomax:
@kristinashu  @mmmavis my preference would be to make the newsletter content be `display:none` until it's needed, but if you think we should just auto-reveal when people tab through the content I can make that happen instead.

kristinashu:
Sorry, what is the Join Us content you're talking about? Is it the Donate prompt that used to pop up?

mmmavis:
@kristinashu I think Pomax was talking about the slide-down nav newsletter form 
<img width="1103" alt="image" src="https://user-images.githubusercontent.com/2896608/82858645-53d05a00-9ec9-11ea-9cbc-1f2d4779ebe5.png">

@Pomax our newsletter form (join us component) and nav code are very intertwined. This is because we use the same component and script for different scenarios (mobile, desktop, nav, footer, PNI, YouTube campaign, etc). In my opinion, the code has become pretty troublesome to deal with 😓 and since we don't have interaction tests in places for them, it's hard to spot regression when we introduce a new code change. I would say we should choose the most straightforward solution so we don't have to modify the existing code much.

> my preference would be to make the newsletter content be display:none until it's needed, but if you think we should just auto-reveal when people tab through the content I can make that happen instead.

By "display:none until it's needed", did you mean when users tab through the page, we show the nav newsletter form without any slide down effect? How is it different than the "auto-reveal" approach?

Pomax:
The two options would be:

1. `display:none` so that the content does not exist on the page. The only way to reveal it is to click the "newsletter" button, which will flip the display property to `display:block` and then perform the slide-out animation in the same way it does right now. Dismissing it would also preserve the current animation, but set a `display:none` as final step so that content is gone, as far as the DOM is concerned.
2. We keep the current behaviour with `opacity: 0` but the moment useraction puts focus in its content, we immediately trigger the reveal animation, and when focus leaves the content, we trigger the hide animation

[...]

PR #4708, opened 2020-05-28
make newsletter wrapped d-none until needed

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4638 such that the newsletter container is set to `d-none` by default, meaning it is not part of the DOM as far as tabbing through indexed elements is concerned. This means that tabbing will now go from the "newsletter" button in the header to the nav bar, rather than invisible tabbing into the newsletter content.

### DIFF
--- a/source/js/nav-newsletter.js
+++ b/source/js/nav-newsletter.js
@@ -42,8 +42,13 @@ class NavNewsletter {
   // remove the global 'closeFormClickHandler' click event handler
   // and reset the form
   closeDesktopNewsletter(event) {
-    elements.container.classList.remove("expanded");
+    const wrapper = elements.container;
+    wrapper.classList.remove("expanded");
     elements.buttonDesktop.classList.remove("active");
+    setTimeout(() => {
+      if (wrapper.classList.contains("expanded")) return;
+      wrapper.classList.add("d-none");
+    }, 500);
     // Make sure we don't leak event listeners
     document.removeEventListener("click", this.closeFormClickHandler);
     document.removeEventListener("scroll", this.closeFormClickHandler);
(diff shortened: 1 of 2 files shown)


source/js/nav-newsletter.js
```diff
@@ -42,8 +42,13 @@ class NavNewsletter {
   // remove the global 'closeFormClickHandler' click event handler
   // and reset the form
   closeDesktopNewsletter(event) {
-    elements.container.classList.remove("expanded");
+    const wrapper = elements.container;
+    wrapper.classList.remove("expanded");
     elements.buttonDesktop.classList.remove("active");
+    setTimeout(() => {
```
mmmavis, line 48: Assuming this is to wait for the sliding animiation to finish? Can we add inline comment here for future reference?

source/js/nav-newsletter.js
```diff
@@ -69,17 +76,24 @@ class NavNewsletter {
   // remove the global 'closeFormClickHandler' click event handler,
   // and reset the form
   closeMobileNewsletter() {
+    const wrapper = elements.container;
     elements.narrowMenuContainer.classList.remove("d-none");
-    elements.container.classList.remove("faded-in");
+    wrapper.classList.remove("faded-in");
+    setTimeout(() => {
```
mmmavis, line 82: Can we please have some inline comments here as well?

source/js/nav-newsletter.js
Pomax: can do

source/js/nav-newsletter.js
Pomax: yep!

source/js/nav-newsletter.js
Pomax:
```suggestion
    // Schedule a "display:none" to happen after the `expanded` animation finishes.
    // See `#nav-newsletter-form-wrapper` transition in ./source/sass/components/primary-nav.scss 
    setTimeout(() => {
```

source/js/nav-newsletter.js
Pomax:
```suggestion
    // Schedule a "display:none" to happen after the `expanded` animation finishes.
    // See `#nav-newsletter-form-wrapper` transition in ./source/sass/components/primary-nav.scss 
    setTimeout(() => {
```

Pomax:
comments added, pointing to where in the SASS code the relevant rule can be found

Issue #4722, opened 2020-06-02
Last bits of PNI a11y cleanup


Pomax:
Last fixes to PNI a1yy after every a11y PR landed. Closes #4006:

## homepage

- [x] Text for `The 😮 below shows how creepy users find these products. ` should be `gray-60`
- [x] Footer text "Protect the internet as a global public resource" should be `<p`> with h5-heading CSS, no an actual `<h5>`
- [x] Footer text "More about us" should be a `<p>` as above

## Category page

- [x] company + name should be `<p>`, not `<h3>`

## About page

- [x] side nav should use `gray-60`
- [x] inner `<main>` should be `<div>`

PR #4723, opened 2020-06-02
Pni a11y revisits

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4722

The only failing cypress:a11y result I now see is the mainjs react:finished event being `undefined` rather than `true` on the homepage, for mobile size only, so that's 100% not related to the PNI work, and can be safely ignored.

## homepage

- [x] Text for `The 😮 below shows how creepy users find these products. ` should be `gray-60`
- [x] Footer text "Protect the internet as a global public resource" should be `<p`> with h5-heading CSS, no an actual `<h5>`
- [x] Footer text "More about us" should be a `<p>` as above

## Category page

- [x] company + name should be `<p>`, not `<h3>`

## About page

- [x] side nav should use `gray-60`
- [x] inner `<main>` should be `<div>`


source/sass/components/multipage-nav.scss
```diff
@@ -30,6 +30,9 @@
   &.vertical-nav {
     .multipage-link-wrapper {
       margin-bottom: 24px;
+      .multipage-link:not(:active):not(.active) {
+        color: $gray-60;
```
mmmavis, line 34: This override affects the hover state styling (see "poor employee ..." link)

image

If our goal is to update the default `.multipage-link` colour, we can just update line 5 in `nav.scss` to `$gray-60`.

Pomax:
@mmmavis PR updated

Issue #4729, opened 2020-06-03
Add an npm task for running prettier --write


Pomax:
Whenever the Prettier dependency updates, it's a good bet some linting rules change and we need to run prettier with `--write` to ensure the code passes linting. Having to run that manually is very silly, let's automate that.

patjouk:
As a side note and when the PR is merged here, can you open the same one on pulse front and doante-wagtail? Thanks!

Pomax:
Good call, let me make issues for those so I don't forget!

PR #4730, opened 2020-06-03
add 'fix' passes for js(x) and (s)css

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4729 by adding an `npm run fix` command, with a note on its use in the documentation around testing.


package.json
```diff
@@ -23,6 +23,9 @@
     "cypress:test": "npm run cypress:clean && wait-on tcp:8000 && cypress run --spec ./cypress/integration/endpoint-tests.js",
     "cypress:test:a11y": "npm run cypress:clean && wait-on tcp:8000 && cypress run --spec ./cypress/integration/a11y-tests.js",
     "cypress:test:manual": "cypress open",
+    "fix": "run-s fix:*",
+    "fix:js": "prettier --write \"cypress/integration/*.js\" \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
```
mmmavis, line 27: Is there an advantage of using `prettier --write` over `eslint --fix` and `stylelint --fix`? If not, I think we should just copy the `test:eslint:js` task and `test:scss` task and add the `--fix` option to them for these new `fix` tasks. This way we can make sure we use the same config to lint and to fix.

patjouk: I'm also wondering if fixing linting issues should be a separate step: what would be the issue with adding `--fix` to the existing tests commands?

Pomax: the prettier folks warn to never use `--fix` by default because it _can_ break things. So I really want to keep this a separate task.

README.md
```diff
@@ -66,6 +66,10 @@ When relevant, we encourage you to write tests. You can run the tests using `inv
 - Run Node tests: `inv test-node`
 - Run Python tests: `inv test-python`
 
+#### Fixing linting errors
+
+If `inv test-node` shows linting errors for either JS/JSX or CSS/SCSS, you can run the `npm run fix` command to have `prettier` automatically fix any linting errors it knows how to fix. This will almost always be the only step required to ensure the linting phase of testing passes.
```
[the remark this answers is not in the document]
Pomax: good call.

Pomax:
@mmmavis @patjouk the prettier folks warn to never use `--fix` by default because it _can_ break things. So I really want to keep this a separate, intentional task. I will be running highly infrequently, and any sane code editor (sublime, vs code, jetbrains, etc) can be told to format code both "on request" and on save, so during normal dev work, the linting should never find anything, this command mostly exists for when prettier updates

package.json
```diff
@@ -23,6 +23,9 @@
     "cypress:test": "npm run cypress:clean && wait-on tcp:8000 && cypress run --spec ./cypress/integration/endpoint-tests.js",
     "cypress:test:a11y": "npm run cypress:clean && wait-on tcp:8000 && cypress run --spec ./cypress/integration/a11y-tests.js",
     "cypress:test:manual": "cypress open",
+    "fix": "run-s fix:*",
+    "fix:js": "npm run test:eslint:js -- --fix",
+    "fix:scss": "npm run test:scss -- --fix",
```
mmmavis, line 28: Lovely!

Issue #4656, opened 2020-05-15
[PNI] Make Product Headings appear on Mobile


natalieworth:
Currently, the PNI product headings don't appear on mobile or tablet screen sizes. To improve accessibility, we want to have a type override to show the product headings (it'll need to be outside of our normal style-guide sizes). 

**To Reproduce**
-  go to https://foundation.mozilla.org/en/privacynotincluded/
- view on viewport < 768px

<img width="450" alt="79252572-b51afd00-7e36-11ea-9bcb-ee9e2a6a8090" src="https://user-images.githubusercontent.com/14876567/82099097-4b9f4000-96bb-11ea-818c-12bc27da320f.png">

---

**🗒  Design -> Dev Handoff Checklist**
(Feel free to remove items that are not applicable.)

- [x] Design has been finalized
- [x] [Link to original design ticket](https://github.com/mozilla/foundation.mozilla.org/issues/4510)
- [x] [Link to Redpen](https://redpen.io/p/toa5350e17d1b0efec)
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included (*only a mobile change)

File is in Abstract:
<img width="937" alt="Screen Shot 2020-05-15 at 2 34 31 PM" src="https://user-images.githubusercontent.com/14876567/82101036-2cef7800-96c0-11ea-9539-273aadb6c9a6.png">

cc @kristinashu

Issue #4510, opened 2020-04-14
[PNI] Product Headings don't appear on Mobile


sabrinang:
**This is probably an enhancement rather than a bug?**

**To Reproduce**
- go to https://foundation.mozilla.org/en/privacynotincluded/
- view on viewport < 768px

**Expected behavior**
- product name and company headings are seen (similarly to viewports >768px)

**Screenshots**

<img width="450" alt="image" src="https://user-images.githubusercontent.com/2492510/79252572-b51afd00-7e36-11ea-9bcb-ee9e2a6a8090.png">

**Desktop (please complete the following information):**
 - OS: iOS 10.15.4
 - Browser: Firefox 74.0.1 & Chrome 80.0.3987.163

**Smartphone (please complete the following information):**
 - Device: Pixel 2
 - OS: android 10
 - Browse: Firefox 68.7.0 & Chrome 81.0.4044.96

**Additional context**
- This will help from an accessibility perspective too when people do not recognize the look of the product or logo

/cc @natalieworth @caltrider

kristinashu:
Note that we if we add titles to the mobile homepage then we should also add them to the related products at the bottom of product pages.

We had left the titles out on mobile because it gets pretty busy looking. Here's a comp with the company and product title. I think the product title will require an override to reduce the font size and line height. We will also need to reduce the size of the product images.

<img width="318" alt="image" src="https://user-images.githubusercontent.com/1068783/80036262-3cefaf80-84a6-11ea-9ea4-08f45ee8b5e8.png">

cc @mmmavis

natalieworth:
I've mocked up an option (with type overrides needed) [in redpen here](https://redpen.io/p/toa5350e17d1b0efec).

cc @kristinashu @mmmavis @sabrinang @caltrider

natalieworth:
If there are no issues with this, I'm going to close and we can make a dev ticket? Not sure if there's dev capacity yet cc @kristinashu ?

Update to this is in Abstract Sketch document.

![[M] Home Copy 2@2x.jpg](https://images.zenhubusercontent.com/57a10963e40e5714b16cdb3b/4fac7fa2-1e70-4e30-bf92-ec76840fff6e)

natalieworth:
I'm going to close this, but if there are issues or need changes.

Let me know if I should make the dev ticket @kristinashu ? :)

kristinashu:
[...]

PR #4731, opened 2020-06-03
always show company and product name

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4656


mmmavis:
Nat is off. Reassigning review to @sabrinang

Pomax:
given that "mobile" isn't a thing, I assume that just means "below breakpoint md"?

Pomax:
I've updated the code to reflect those sizes but that is some _very small_ text I have to say. I can barely read it O_o

mmmavis:
@Pomax yea I agree that 9px text is small. @kristinashu pinged me [about this](https://github.com/mozilla/foundation.mozilla.org/issues/4510#issuecomment-628061441) ahead of time and I was fine since this text style [passed the a11y tests](https://github.com/mozilla/foundation.mozilla.org/issues/4510#issuecomment-628938731).

I'd say let's wait for @kristinashu to come back from PTO so she can chime in on this?

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -121,6 +117,26 @@
...
+        .product-links {
+          @include hover-focus-active {
+            text-decoration: none;
+          }
+
+          @media screen and (max-width: $breakpoint) {
+            .product-company {
+              color: $gray-80;
+              font-size: 9px;
+              line-height: 14px;
```
mmmavis, line 129: We can use the `scaleText` mixin for settting font-size and line-height
```scss
@include scaleText(9px, 14px);
```

source/sass/buyers-guide/views/homepage.scss
```diff
@@ -121,6 +117,26 @@
...
+            }
+
+            .product-name {
+              color: $black;
+              font-size: 11px;
+              line-height: 14px;
```
mmmavis, line 135:
```scss
@include scaleText(11px, 14px);
```

source/sass/buyers-guide/views/homepage.scss
Pomax: ah, great

Pomax: Tangential question: if it sets text size as well as lineheight, could we rename this mixin to `setTextSize` rather than suggesting it relies on ``transform: scale`?

Pomax: I've filed https://github.com/mozilla/foundation.mozilla.org/issues/4739 for that discussion/possible work.

mmmavis: Yea we can rename it. This is an old mixin that I meant to rename but kept forgetting 😂 (But I was thinking about just fixing the letter casing so it's `scale-text` instead)

I'm not sure if we wanna use`transform: scale` for this because mobile/desktop font types aren't always upsized/downsized in scale. I also prefer explicitly putting down font size and line height for mobile and desktop so they correspond with how the design files and brand guide were documented (in fixed `px` values). This also makes it easy to tell what font style we are using without going through math calculation lol. 

I can work with the designers for that ticket when the time comes.

kristinashu:
I think 9px is fine for this specific case. It's important but not critical information, it passes accessibility, and as a general rule text can be small on "mobile" since people tend to hold it closer to their eyes than a laptop or monitor.

Issue #4706, opened 2020-05-28
Update PNI's methodology.html to improve HTML outline


mmmavis:
Update `network-api/networkapi/buyersguide/templates/about/methodology.html` to improve HTML outline.

This requires updating some DOM nodes with the right tag. (Mostly around `<section>` and heading)

Seeing context: https://github.com/mozilla/foundation.mozilla.org/pull/4702#discussion_r431465927

Pomax:
@mmmavis there's a lot of `<p id="product-selection"><b>{% trans "Product Selection" %}</b></p>` in this file, too, do we want to change that to something more sensible, like `<h3 id="product-selection" class="h3-heading">{% trans "Product Selection" %}</h3>`?

PR #4734, opened 2020-06-03
Pni update html outline

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4706 by properly using `section` markup, and by making the section headings true headings with a new `p-heading` class, instead of `<p><b>...</b></p>`


source/sass/type.scss
```diff
@@ -58,6 +58,11 @@ h6,
   text-transform: uppercase;
 }
 
+.p-heading {
```
mmmavis, line 61: We should limit type styles to only the ones that are specified in the [style guide](https://foundation.mozilla.org/style-guide/).

- Page on production: https://foundation.mozilla.org/en/privacynotincluded/about/methodology
- Page on this review app: https://foundation-s-pni-update-p94fcd.herokuapp.com/en/privacynotincluded/about/methodology

@sabrinang can you suggest what existing text style works best for the third tier headings on the PNI's methodology page? (e.g., `Product Selection`, `Encryption`, `Security updates`, ... etc)

Pomax:
The problem with `<p><b>` is that it's not accessible: if it's a heading, it _needs_ to use heading markup, so maybe @caltrider can pitch in here, too: are you okay with us turning the headings into semantic headings, with the heading markup as found in [our styleguide](https://foundation.mozilla.org/en/style-guide/)? It'll make the page look a little different, but it'd make the code and accessibility work a lot easier.

sabrinang:
@Pomax 
> The problem with `<p><b>` is that it's not accessible: if it's a heading, it needs to use heading markup for accessibility tools to recognize them, so at the very least that markup needs to get replaced.

Sorry for any confusion but +1 I agree to avoid adding a new heading style that visually looks the same as our bold paragraph and that we should not use `<p><b>` tag and use a type of `<h#>` markup for accessibility best practice

> I don't know who the original designer for this set of pages was, what is the best way to initiate a round design on this page?

I'm not quite sure who as well, but I think filing a design issue in design repo will bring awareness that this page could be revised with accessibility and style guide in mind to be triaged for future sprint. I'll help file one now.

sabrinang:
related design issue here: https://github.com/MozillaFoundation/Design/issues/585

mmmavis:
@Pomax let's avoid adding new type style while we are waiting for a conclusion for MozillaFoundation/Design#585 to happen. To help this PR move along, let's just move the heading text SCSS overrides to PNI part of .scss file (instead of introducing a new style to `_type.scss`).

Pomax:
I've moved the `p-heading` to the SCSS file specifically only for the PNI about page, with a large comment block explaining that it's only temporary while the design issue around how best to update the markup to reflect our styleguide is open, while at least making sure heading markup, rather than paragraph markup, gets used for the headings.

As far as regular users go, they'll never know, but until we know the exact classes and heading depth we want, accessibility tools will see real headings at least.

Issue #4586, opened 2020-04-30
it looks like we have a horizontal scrollbar at medium and smaller breakpoints


Pomax:
Chrome:

![image1](https://user-images.githubusercontent.com/177243/80755354-8151ff80-8ae5-11ea-8323-c0ee4a4f67e1.png)

Firefox:

![image](https://user-images.githubusercontent.com/177243/80755363-84e58680-8ae5-11ea-831a-e633362712c8.png)

mmmavis:
Looks like it's something to do with `narrow-screen-menu`. I'm gonna use this as an opporutunity to clean up `primary-nav.scss` code for both Foundation site and PNI. They seem to be sharing quite a bit of code.

mmmavis:
Aside from `narrow-screen-menu`, our`.full-bleed` related rules are the main cause of this. I took a stab at it but ended up spending too much time without any success. Initially, I thought this would be a quick fix that's why I added it to this sprint last Thursday. Re-accessing the ticket now I don't think I have bandwidth for this anymore, I'm gonna remove this ticket from this sprint.

PR #4741, opened 2020-06-04
fix our mystery h-scrollbar at small width

Pomax:
Closes  https://github.com/mozilla/foundation.mozilla.org/issues/4586 by updating our `full-bleed` rule to use a different way to center content, setting widths based on `100%` but boosted by the size that `$grid-gutter-width` says Bootstrap is using.


network-api/networkapi/wagtailpages/templates/wagtailpages/static/initiatives_page.html
```diff
@@ -7,7 +7,7 @@
 {% image page.primaryHero fill-2400x700-c100 format-jpeg as big_hero %}
 <div style="background-image: url({{ big_hero.url }}); background-repeat:no-repeat; background-position: top center; background-size: cover; padding-top: 100px;">
 
-  <div class="container p-3 p-sm-5 mx-4 mx-sm-auto mb-4 bg-white">
+  <div class="container p-3 p-sm-5 mx-sm-auto mb-4 bg-white">
```
Pomax, line 10: This one was _amazing_

mmmavis: This change fixes the overflow issue but breaks the design in a different way. 😬 
 
Mockup: (notice the "cutout" effect for the intro section)
image

Review app:
image

Pomax: Removed the margins on the block itself, and instead turned them into padding on the containing div. That way we don't need to worry about "width + margin".

Pomax:
Had to rebase this because we split up the mixins, so marking for re-review-request just to be safe =D

PR #4743, opened 2020-06-04
Remove 'color-hex-length' stylelint rule

### LINKED ISSUE #4733: Remove "long hex colors" SCSS rule?
Our `.stylelintrc` currently reads:
```
{
  "plugins": ["stylelint-prettier"],
  "rules": {
    "prettier/prettier": true,
    "color-hex-length": "long"
  }
}
```

But for most of our CSS we only allow named color variables, so we might be able to remove that exception?

mmmavis:
Closes #4733

Remove 'color-hex-length' stylelint rule since we don't allow color code in all files other than _colors.scss

### DIFF
--- a/.stylelintrc
+++ b/.stylelintrc
@@ -1,7 +1,6 @@
 {
   "plugins": ["stylelint-prettier"],
   "rules": {
-    "prettier/prettier": true,
-    "color-hex-length": "long"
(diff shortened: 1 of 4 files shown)


mmmavis:
Uh oh the commit history doesn't look right. I'll open a new PR.

Issue #2828, opened 2019-03-14
Remove old Participate page model


alanmoo:
It's been replaced by Participate2 and that's now being mocked in fake data generation. Let's get rid of the old one entirely.

abhinavsagar:
Can i work on this?

alanmoo:
Sure, feel free to open a PR and we'll take a look at it (just keep in mind this team mainly works daytime hours in North America)

abhinavsagar:
Ok I am working on this. I will open a PR as fast as I can.

Pomax:
Speed is less important than doing it right - make sure to read through https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#how-to-prepare-and-deploy-your-backward-incompatible-migrations because removing models is _not_ a simple matter of "deleting the code" and generating a new migration file =)

youriwims:
Removing this model is blocked by removing its usage in `/network-api/networkapi/wagtailpages/translation.py`. We're working on establishing a process to fully remove translation bindings in order to remove the model's import in `translation.py` thereby making it okay to completely remove the model in `/network-api/networkapi/wagtailpages/models.py`. [This is also an issue](https://github.com/mozilla/foundation.mozilla.org/pull/2904#pullrequestreview-221278351) with `PeoplePage` (Reference #2904 , #2898).

youriwims:
Removing this model is blocked by removing its usage in `/network-api/networkapi/wagtailpages/translation.py`. We're working on establishing a process to fully remove translation bindings in order to remove the model's import in `translation.py` thereby making it okay to completely remove the model in `/network-api/networkapi/wagtailpages/models.py`. [This is also an issue](https://github.com/mozilla/foundation.mozilla.org/pull/2904#pullrequestreview-221278351) with `PeoplePage` (Reference #2904 , #2898).
cc: @Pomax @cadecairos @patjouk

PR #4745, opened 2020-06-04
Remove old participate page

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2828 by removing the old `ParticipatePage` from models.py as associated classes, as well as translation.py


Pomax:
what is going on with these commits

PR #4746, opened 2020-06-04
remove old Participate page

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2828 by removing the old ParticipatePage from all models and the translation.py file, with the associated migrations.

Tested using `inv test`.


Pomax:
gah! yes, I keep forgetting to delete those orphaned assets =_=

Pomax:
@mmmavis one more review just to make sure the new migration makes sense~

Issue #2964, opened 2019-04-04
Enhance landing page inheritance for easier author experience


xmatthewx:
In authoring a campaign that uses the new landing page template, it's a wee bit of a pain to select the banner image on every single child page. Would it possible for child pages to inherit their parents banner image if nothing is specified for that specific page? And only fall back to generic b&w if neither have an image?

Let's discuss and then file a real ticket @alanmoo @kristinashu

kristinashu:
Yes, this would be amazing!

I've also wondered if the child pages could inherit the OG content from parent page too?

alanmoo:
The original request should be possible. 

@kristinashu - It looks like OG Text does inherit currently, but not images. I'm not sure if that's an oversight in the implementation of #2347 or a bug. @cadecairos Can you comment on that part?

kristinashu:
@Pomax any thoughts/insights on this?

Pomax:
It should be possible to fall back to the parent page's data in absence of "local" data. I would estimate a 2 for figuring out how much work the implementation would be, and then have a separate issue to do the implementation work? I recall we already had fallthrough for _something_ in our existing models, but I can't remember where off the top of my head.

Pomax:
@kristinashu is this something we still want to do?

kristinashu:
@Pomax yes, this would still be a very valuable feature to have!

PR #4781, opened 2020-06-10
walk up the chain to find a banner

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2964 by making primary pages serve their own banner if they have one set, or "a parent's banner if there is one", walking up the parent chain until either there is a PrimaryPage in the parent chain with a banner set (and then we use that), or we hit the homepage (and then we use the default banner)

Test urls:
- top banner page with explicit banner set: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/en/campaigns/whats-this/ (Edit link: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/cms/pages/30/edit/)
- child banner page without explicit banner set: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/en/campaigns/success-drop-deep-general-run-break/child-without-banner/ (edit link: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/cms/pages/54/edit/)
- tertiary child, which I don't think we even use but it works anway: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/en/campaigns/success-drop-deep-general-run-break/child-without-banner/tetriary-page/ (edit link: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/cms/pages/55/edit/)

Also test url:
- top primary page without a banner set should use the default banner: https://foundation-s-inherit-pa-f5gtxv.herokuapp.com/en/campaigns/red-yet-cup-house-country-another/


Pomax:
So, fun fact: this technically applies to all `PrimaryPages`, so it _might_ already work for mozfest!

Pomax:
Ah, just checked, mozfest uses `class MozfestPrimaryPage(FoundationMetadataPageMixin, Page)`, so it doesn't directly extend `PrimaryPage`, so no free lunch... want to file the issue to get this working for Mozfest, too? Should be relatively easily achieved.

network-api/networkapi/wagtailpages/pagemodels/primary.py
```diff
@@ -37,6 +37,21 @@ class PrimaryPage(FoundationMetadataPageMixin, Page):
         help_text='Choose an image that\'s bigger than 4032px x 1152px with aspect ratio 3.5:1',
     )
 
+    def get_banner(self):
+        if self.banner is None:
+            # note: we need to reverse get_ancestors, because by default
```
mmmavis, line 42: nice

kristinashu:
Opened a follow up ticket for mozfest site https://github.com/mozilla/foundation.mozilla.org/issues/4789

Issue #4827, opened 2020-06-30
sanitize directory listing js


Pomax:
the directory listing code currently sets `innerHTML` based on profile information, and while that information is pretty trustworthy (given that we sanitize it on the pulse side), it'll be good to add an addtitional layer of sanitization just to make sure that the data we're templating in doesn't contain active code.

PR #4828, opened 2020-06-30
because let's be safe

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4827

Note that this does not touch the wagtail block template, because Django templating already performs sanitization unless `{{ ... | safe }}` is used, which we do not use for the `person-card` data.

### DIFF
--- a/source/js/foundation/pages/directory-listing-filters.js
+++ b/source/js/foundation/pages/directory-listing-filters.js
@@ -13,6 +13,45 @@
     .apiEndpoint;
 
   /**
+   * We're reasonably sure that profiles are sane already, but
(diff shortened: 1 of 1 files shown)


source/js/foundation/pages/directory-listing-filters.js
```diff
@@ -12,6 +12,45 @@
   const API_ENDPOINT = document.querySelector(`[data-api-endpoint]`).dataset
     .apiEndpoint;
 
+  /**
+   * We're reasonably sure that profiles are sane already, but
+   * let's make sure we're _absolutely_ sure because we're relying
```
mmmavis, line 17: 😄

Issue #4775, opened 2020-06-09
Rename ParticipatePage2 (back) to ParticipatePage


Pomax:
With all the care that comes with because it's almost guaranteed going to need some finessing.

Pomax:
changing estimate due to the magic of "things go horribly wrong". Django creates a rename migration:
```
from django.conf import settings
from django.db import migrations


class Migration(migrations.Migration):

    dependencies = [
        ('contenttypes', '0002_remove_content_type_name'),
        ('wagtailimages', '0022_uploadedimage'),
        migrations.swappable_dependency(settings.AUTH_USER_MODEL),
        ('wagtailcore', '0045_assign_unlock_grouppagepermission'),
        ('wagtailpages', '0003_delete_homepagefeaturednews'),
    ]

    operations = [
        migrations.RenameModel(
            old_name='ParticipatePage2',
            new_name='ParticipatePage',
        ),
    ]
```
That migration works fine on local faked data, but running it on the staging DB leads to the page turning into this:

![image](https://user-images.githubusercontent.com/177243/86390908-bdbceb80-bc4d-11ea-883a-734d8bdcd91c.png)

Looking at the page type to figure out why it's not using the correct template reveals that Wagtail has _no idea what the page type is_ anymore:

![image](https://user-images.githubusercontent.com/177243/86390957-d2997f00-bc4d-11ea-84ac-3a7dbd8fb61a.png)

Not super great.

KalobTaulien:
I was able to confirm the problem from the staging database, but my outcome was different. 

I still had the template render properly, but all the page fields disappeared, the name of the page disappeared, and the page could no longer be saved. 

When I looked at the database, there was a row for `participatepage` in the `django_content_type` table, and that shouldn't be there anymore. I think there was historical data left behind somehow and that needs to be cleaned out but that row might be protected, I didn't look into _that_ much. 

I tried to create a new model named ParticipatePage, migrate, delete, and migrate again but that didn't cascade through the db the way I was hoping. 

We could consider renaming the ParticipatePage2 model to something less verbose, like `class Participate(Page): ..` and use the `verbose_name` Meta option to call it "Particpate Page".

Pomax:
It'd be better if we could figure out what's actually wrong in the db and fix that (as a one-time action, this is something we should be able to issue directly on the live dbs, provided we make doubly sure to test the actual SQL we'll issue on prod and staging copies)

Pomax:
[...]

PR #4833, opened 2020-07-02
[WIP] renaming ParticipatePage2 to ParticipatePage

Pomax:
This is supposed to close https://github.com/mozilla/foundation.mozilla.org/issues/4775 but instead it's doing all the weirdness. See issue for the description of said weirdness.


network-api/networkapi/wagtailpages/pagemodels/base.py
```diff
@@ -126,7 +126,7 @@ class InitiativesPage(PrimaryPage):
     ]
 
 
-class ParticipatePage2(PrimaryPage):
+class ParticipatePage(PrimaryPage):
```
Pomax, line 129: This is the only "true" change, everything else is just changes to make sure references to the old name are now references to the new name.

network-api/networkapi/wagtailpages/pagemodels/base.py
```diff
@@ -126,7 +126,7 @@ class InitiativesPage(PrimaryPage):
...
     parent_page_types = ['Homepage']
     template = 'wagtailpages/static/participate_page2.html'
```
Pomax, line 131: this has been left pointing to the original template until things actually work.

KalobTaulien:
I tried this, this morning:

```
    class Meta:
        verbose_name = "Participate Page"
        verbose_name_plural = "Participate Pages"
        db_table = "wagtailpages_participatepage2"
        managed = False  # With and without the managed setting toggled off 
```

Ran into the same issue. I'm thinking we'll need to run a command to delete the original content type, cascade it through the tables (clean the db of this content type), then rename the `ParticipatePage2` to `ParticipatePage`

Pomax:
let's retry this again at some later point.

Issue #4789, opened 2020-06-11
MozFest: Enhance landing page inheritance for easier author experience


kristinashu:
Same as - Enhance landing page inheritance for easier author experience #2964 - but for MozFest pages.

Child pages should inherit their parents banner image if nothing is specified for that specific page. 

Note the mozfest site has a different fallback image than the main site and if now banner image is selected that it should show the mozfest fallback image.

KalobTaulien:
@kristinashu For clarification, you'd like this to look for a banner image in this order, correct?

1. Look for a banner image on the page itself. If not found then
2. Look for a banner image on the parent page (and continue up the tree until an image is found). If not found then 
3. Fallback to the default banner image

Pomax:
https://github.com/mozilla/foundation.mozilla.org/pull/4878 exists, and is currently blocked because of review apps not actually using the same site/page system that the real mozfest site does. @cadecairos we need to look at what can be done here, because it's blocking testing of code that I'm 99.% certain of works fine =(

PR #4878, opened 2020-07-07
primary pages (mozfest + fomo) inherit their banner if not set explicitly

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4789 by applying the same logic as used in https://github.com/mozilla/foundation.mozilla.org/issues/2964: `MozfestPrimaryPage` now has the same `get_banner` function that foundation's `PrimaryPage` uses, with a template update to tap into this function.

## test pages

No banner on parent or child
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/spaces/
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/spaces/test-page/

Banner on parent, no banner on child:
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/tickets/
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/tickets/test-page/

No banner on parent, banner on child:
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/team/
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/team/test-page/

Banner on parent, different banner on child:
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/sponsors/
- https://foundation-s-mozfest-pa-2eprmx.herokuapp.com/en/sponsors/test-page/

## review app changes

In order to make the review app work properly I had to take some additional steps:
- delete the default wagtail root
- move "mozfest" pages to become child of `Root`
- delete "homepage" and all children under that
- create a new `sites` entry for the review app's domain, pointing to the mozfest page, set as default site
- *restart the dyno* (critically important)

This replicates the setup that the real mozfest website uses.


network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_banner_inheritance.py
```diff
@@ -0,0 +1,20 @@
+# A mixin that imparts banner inheritance for whatever model it gets tacked onto
+class FoundationBannerInheritanceMixin():
+
+    def get_banner(self):
+        if self.banner is None:
+            ancestors = self.get_ancestors()
+            root = ancestors[1].specific
```
mmmavis, line 7: wondering why we we `ancestors[1]` instead of `ancestors[0]`?

network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_banner_inheritance.py
```diff
@@ -0,0 +1,20 @@
...
+            # note: we need to reverse get_ancestors, because by default
+            # this gives a list that starts at the root, and then gets more
+            # specific, ending in the direct parent. We want to check in the
+            # opposite direction: start at the parent until we hit the root.
+            for page in ancestors.reverse():
+                page = page.specific
+                if page == root:
+                    return None
+                valid = isinstance(page, FoundationBannerInheritanceMixin)
```
mmmavis, line 16: Are mixins considered as "class" as well?

I checked the [doc](https://python-reference.readthedocs.io/en/latest/docs/functions/isinstance.html) of `isinstance` and it says "it returns a Boolean stating whether the object is an instance or subclass of another object."

network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_banner_inheritance.py
Pomax: ancestors[0] is a wagtail level Page literally called `Root`, rather than our own page.

network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_banner_inheritance.py
Pomax: They are. Python allows multiple inheritance, so in `class X(A,B,C)` is a class that extends three difference classes, and `isinstance(X,X)`, `isinstance(X,A)`, `isinstance(X,B)`, and `isinstance(X.C)` are all true.

mmmavis: Ah gotcha. Didn't scroll down far enough to see this line in another file: 
```class PrimaryPage(FoundationMetadataPageMixin, FoundationBannerInheritanceMixin, Page)```

Pomax:
@kristinashu I've added several test links to show off banners getting inherited, if set, or the default banner getting used if nothing is set.

@mmmavis I forgot to update the large.html hero fragment, so just set you for review again... same change as normal.html ^^

kristinashu:
Nice, it seems to be inheriting the banner as expected!

But I'm am unsure about the default banner. On the live site, I believe there is a default banner (photo of crowd with gradient overtop) that will show if staff don't add a banner. E.g. https://mozillafestival.mofostaging.net/en/speakers/
Can we keep this photo as the default? It seems like right now the default is just solid blue.

Pomax:
I've update the review app to have mozfest as a true "site root", but the way I did it is not making wagtail super happy and it seems to be generating server errors, _sometimes_... so I'm going to mark this as blocked for now and will look at things some more after the homepage updates.

Pomax:
So far it looks like the problem comes from creating a review app where the mozfest page starts life as a child of the homepage, rather than being a true root page like on staging/production. @cadecairos @patjouk is this something we could change without breaking review apps? =(

Pomax:
@mmmavis @kristinashu this PR is ready for review again, after figuring out how to make the review app use the same content hierarchy as the real mozfest site.

PR #4899, opened 2020-07-14
Minify compiled CSS file

### LINKED ISSUE #4784: Minify compiled CSS file
Minify compiled CSS file to reduce file size and load time.

We can probably do it via Webpack: https://webpack.js.org/plugins/mini-css-extract-plugin/#minimizing-for-production

mmmavis:
Closes #4784

I ended up using `postcss` to minify CSS files because 1) we are already using `postcss` 2) it's troublesome to modify webpack config to make this work. 

One surprising finding - currently our main compiled CSS file on production is ~650KB and ~400KB actually comes from inline source map that **doesn't work** 🤮 . With minification and removal of the broken inline source map code, [we are able to improve](https://github.com/mozilla/foundation.mozilla.org/pull/4899) the CSS file from

image

to 

image

😃 

I'm gonna file a separate ticket to discuss whether or not a source map for CSS is actually needed. If so, we can add it with a new PR.


package.json
```diff
@@ -27,6 +26,7 @@
     "fix:js": "npm run test:eslint:js -- --fix",
     "fix:scss": "npm run test:scss -- --fix",
     "heroku-postbuild": "npm run build",
+    "optimize:css": "postcss network-api/networkapi/frontend/_css/*.css --dir network-api/networkapi/frontend/_css",
```
mmmavis, line 29: this covers all CSS files in that dir

mmmavis: We probably want a better name for this task as our `postcss` setup doesn't only cover optimization (minification) but also autofixer. Suggestions

PR #4901, opened 2020-07-14
Add 'Cause Statement' fields to homepage model

### LINKED ISSUE #4836: Cause Statement
Fields: 
- [x] Plain text field 
- [x] Link text and link url (Confirm if internal/external urls are required; if so add validation)

mmmavis:
Closes #4836


network-api/networkapi/wagtailpages/translation.py
```diff
@@ -158,6 +158,9 @@ class HomepageTR(TranslationOptions):
         'hero_button_text',
         'hero_button_url',
         'hero_image',
+        'cause_statement',
+        'cause_statement_link_text',
+
```
Pomax, line 163:
```suggestion
```

network-api/networkapi/wagtailpages/pagemodels/base.py
```diff
@@ -468,6 +468,24 @@ class Homepage(FoundationMetadataPageMixin, Page):
...
 
+    cause_statement = models.CharField(
+        max_length=250,
+        default="",
+    )
+
+    cause_statement_link_text = models.CharField(
+        max_length=80,
+        default="",
+    )
```
Pomax, line 479: These don't get values assigned during migration, so they'll need to be `null=True` or `blank=True`, so that `inv migrate` (and `inv new-db`) don't throw an error.

mmmavis: I didn't encounter any errors with `inv new-db` or `inv migrate`, and it seems like the new fields in the migration file all have `default=''` and `null=True`. If we add `blank=True` then the field is no longer a required field. I'm wondering what I missed here?

Pomax: looks to work fine now.

Issue #4838, opened 2020-07-02
News you can use


KalobTaulien:
Fields:
- [x] 1 large feature 
- [x] 3 smaller features 

These should be page choosers and the relevant data for the design cards will be automatically pulled in. Card titles should be shortened to always fit on the same 2 lines. 

Note: We'll also need to decide how many blog author photos to display and attach an image field to blog author profiles.

Pomax:
@beccaklam is this now "News you can use"?

beccaklam:
> @beccaklam is this now "News you can use"?

Yes!

Pomax:
I've stubbed out the backend work for this over in https://github.com/mozilla/foundation.mozilla.org/pull/4903 but there's some work left to do.

Pomax:
weirdly, the PR closing did not close this issue but: this got done.

PR #4903, opened 2020-07-15
Backend work for news you can use

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4838 by setting up the backend fields and template stub for the "news you can use" section to be styled and made correct wrt the comps.


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/news_you_can_use.html
```diff
@@ -0,0 +1,41 @@
...
+                blog category
+                content snippet
+                authors
+            -->
+        </div>
+      </div>
+    {% endif %}
+    {% endfor %}
+  </div>
+</div>
```
Pomax, line 41:
```suggestion
</div>

```

network-api/networkapi/wagtailpages/pagemodels/base.py
```diff
@@ -309,6 +309,25 @@ def __str__(self):
         return self.page.title + '->' + self.highlight.title
 
 
+class HomepageNewsYouCanUse(WagtailOrderable, models.Model):
```
Pomax, line 312: It feels like this is generic enough for the homepage (which has several sections that all tap into the blog posts) that it might be worth creating an abstract superclass here, and have all the "it's another {`page`, `blog`, `panels`}" models inherit from that?

mmmavis: I would vote for having a superclass since the two classes `HomepageNewsYouCanUse` and `HomepageFeaturedBlogs` seem pretty much identical. Wanna file a follow-up?

Pomax: https://github.com/mozilla/foundation.mozilla.org/issues/4963

Pomax:
```suggestion
class HomepageNewsYouCanUse(WagtailOrderable):
```

Issue #4937, opened 2020-07-21
[Frontend] Blog Author photos


cadecairos:

mmmavis:
My work for the [Spotlight ticket](https://github.com/mozilla/foundation.mozilla.org/issues/4941) touches this. I will implement the blog author section as a template fragment and have CSS code in place so they can be reusable for other sections(e.g., the ["News you can use" section](https://github.com/mozilla/foundation.mozilla.org/issues/4940)).

Closing as work is being covered in the [Spotlight ticket](https://github.com/mozilla/foundation.mozilla.org/issues/4941).

Issue #4840, opened 2020-07-02
Blog author photos


KalobTaulien:
We need to create an easy way to get blog author photos (example use case: blog cards). 

- [ ] Explore adding a ForeignKey to the Author/Profile class. 

Note: These will be used in Blog Cards. We'll want to keep queries low or consider caching so we don't slam the database with necessary queries to the Images table.

kristinashu:
Design needs do a little more thinking around this. I've opened https://github.com/mozilla/foundation.mozilla.org/issues/4846 to be done in the July 20 sprint. Design will need need to consult engineers on this but building it out is blocked until design is done.

Pomax:
Talking through the comps, backend-wise this will require:
- updating the current Blog `author` field to a foreignKey to a new Author model with fields:
  - `name`: the author's full name
  - `image`: the image that is associated with that author
  - a migration `run` definition that automatically `get_create`s the `mozilla` author, with its associated graphic
- prepopulating the author list with the current user (if there is a name-matching `Author` instance)
  - a migration `run` definition that uplifts preexisting `author` charfields to the new `Author` model (with `get_or_create` rather than blind `create` =)
- adding a `clean` def to Blog pages so that you can't publish a page unless there is at least one Author set


Front-end wise this will require:
- implementing the new card layout, with
  - `<wbr>` between the list of icons and names, so that a line wrap splits between images/text first
  - ideally, `nowrap` on author names, so that line wrapping does not leave a name split over consecutive lines

Pomax:
Question: there is a `People` app that I don't think we're still using. Can we:
1. investigate whether that is indeed unused by now, and
2. determine whether it makes more sense to reuse, or retire

/cc @cadecairos

cadecairos:
we're exposing the model still via public API but I do not think any of our templates, pages, or snippets depend on it. 

I'm satisfied that it's safe to remove that route. 

Re-purposing is possible, but there's a bunch of existing data in there though which I doubt we need still, so starting a new model might be a cleaner approach, rather than a data migration to get it in the right format or clean it out altogether.

Pomax:
let's go for clean. I've filed https://github.com/mozilla/foundation.mozilla.org/issues/4908 to track removing the `People` app.

KalobTaulien:
@taisdesouzalessa Just to confirm, there _can_ be multiple blog authors on each blog post right? If so, how many would be the max?

taisdesouzalessa:
[...]

PR #4957, opened 2020-07-21
Feature/4840 blog author posts

KalobTaulien:
Related PRs/issues #4840 

This is the backend work and some frontend cleanup. Depends on the frontend ticket to be "complete" #4937 

Includes a data migration to convert String-based Authors to Snippets (and backwards when the migration is rolled back). 

This PR only affects this section:
image

(And is not applied to the "Featured stories" section of the Homepage) 

I've also written a small template tag called `get_authors()` which performs a little logic in order to keep the template more organized.


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
```diff
@@ -18,48 +18,60 @@ <h4 class="h4-heading">{% trans "Latest news" %}</h4>
...
-      <p>{{blog.blog.search_description}}</p>
+      <div class="body-small mb-2">
+
+{# TODO: Export this internal CSS to .scss #}
+<style>
+  .featured-blog-author-image {
+    display: inline-block;
+    width: 50px;
+    height: 50px;
+    position: relative;
```
KalobTaulien, line 50: Just a note for whoever reviews this, this isn't "proper" css, it's just stuff I whipped up to test lining up images together.

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -46,14 +49,45 @@ class BlogPageTag(TaggedItemBase):
...
+    panels = [
+        FieldPanel("name"),
+        ImageChooserPanel("image"),
+    ]
+
+    def __str__(self):
+        return self.name
+
+
+class BlogAuthorsOrderable(Orderable):
```
Pomax, line 72: let's not put `orderable` in the class name: the class definition already makes this clear.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
```diff
@@ -18,48 +18,60 @@ <h4 class="h4-heading">{% trans "Latest news" %}</h4>
...
     </div>
   </div>
   {% endif %}
   {% endfor %}
 
+
   {% if featured_blogs|length > 1 %}
   <div class="col-md-6 mb-0 pb-4">
     {% for blog in featured_blogs %}
+    {% get_authors blog as authors %}
```
Pomax, line 35: I'd like to not use a tag here if the only thing the tag does is repack the data to omit a level of indirection. Let's stick with ` {% for entry in blog.authors.all %}` and then pull out `entry.author.image` etc. - that may look uglier, but it's less magic, and less magic when it comes to Django means it's infinitely more accessible when it comes time to figure out a bug for folks less familiar with Django/wagtail in the future.

KalobTaulien: The reason I added the template tag was because we're traversing from Home Page -> Blog Page Orderable -> Blog Page -> Blog Authors. 

Here's the original work before the template tag: https://github.com/mozilla/foundation.mozilla.org/pull/4957/commits/680ed8de3dcba008056fa42b006dd6b2c3221383#diff-084370aab012384922f2ae8b4676e3d2R39-R62

Though if you feel strongly about this I can revert this work easily. Let me know what your final thoughts on this are.

Pomax: Fair, let's leave it a tag then, but let's rename it to `get_all_authors`. Small change, but makes it a little clearer that this is a custom function and a magical Django function `get_[fieldname]`.

network-api/networkapi/wagtailpages/templatetags/homepage_tags.py
```diff
@@ -29,3 +29,17 @@ def homepage_image_with_class(context, path, classname):
         'url': url,
         'classname': classname,
     }
+
+
+@register.simple_tag
+def get_authors(blog_post):
```
Pomax, line 35:
```suggestion
def get_all_authors(blog_post):
```

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/featured_blogs.html
Pomax:
```suggestion
    {% get_all_authors blog as authors %}
```

network-api/networkapi/wagtailpages/models.py
```diff
@@ -39,7 +39,8 @@
 )
 
 from .pagemodels.blog.blog import (
-    BlogPage
+    BlogPage,
+    BlogAuthor
```
Pomax, line 43: looks like this still needs to go in the `__all__` export list

Issue #4936, opened 2020-07-21
Create a LinkChooser panel that allows us to define one `url` field, but with UI for internal vs. external


Pomax:
Even if wagtail is going to be working on this, too, we should create a LinkChooser panel (and stick it in the `wagtailcustomization` app, probably?) that lets people define a model as `url = models.URLField(...)` and then  assign it a panel `panels = [ ..., LinkChooser('url'), ... ]` with UI that has a regular URL chooser as well as a page chooser for internal pages, with the "one of these needs to be real" checking before it sends the value through to the associated model field.

And this can even be _the same code_ we already thought of putting in, just wrapped by a LinkChooser so that we can rely on a single real field in our model, rather than having two fields.

/cc @KalobTaulien @cadecairos

KalobTaulien:
Hmm yeah. We could "cheat" and use a URLField for both values. However the internal_url we tend to use (with a PageChooserPanel) will use a pageId so if the slug of a page changes the link won't 404. 

I was thinking of having something like this stored in a textfield (as json)
```
{
    "external": "https://website.com", 
    "internal": 71,
    "url": ... // Checks the values and finds page 71's full url if needed
}
```
With validation to set either one or the other. And a `.url` value to maintain the same way we're already doing this. 

Though this might mean we create a custom Django field as well, much like `StreamField()`.

Pomax:
I was more thinking we define a custom model field that internally holds both a `URLField` and a foreign key to a Page, with a `@property` decorated `def url(self): ...` that lets us ask for its value. Storing JSON is something I'd like to avoid if at all possible because migrations become borderline impossible when you're dealing with having to figure out schema translations for JSON. (which is one of the reasons it would be have great if StreamFields were real models, but that's a decision that was made way too long ago to try to switch to, now).

Pomax:
It might be worth looking at https://pypi.org/project/django-composite-field/ to see if we can wrap a URLField and ForeignKey into a single composite field. We'd still have the same "it's two fields, with custom cleaning", but at least it'd be a single instance of that, which we can reuse in the various models that need it.

Pomax:
As a practical "chooser" side of things, it feels like we might be able to get away with a tailored `class CustomURLChooserPanel(MultieFieldPanel` with a hardcoded list `[FieldPanel('external'), PageChooserPanel('page')]` and then the real work is around making the frontend have something for toggling which of the two fields you need.

Pomax:
[...]

PR #4966, opened 2020-07-22
a universal link model

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/4936 by introducing a new model that wraps a URLField and a foreign key to any wagtail Page.

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/link.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/link.py
@@ -1,0 +1,76 @@
+from django.db import models
+
+from wagtail.admin.edit_handlers import FieldPanel
+from wagtail.admin.edit_handlers import PageChooserPanel
+
+
+class UniversalLink(models.model):
+    """
+    A model that can be used by other models for specifying
+    a link that is either an internal Wagtail page, or an
+    external plain link.
+
+    Used as:
+
+        field_name = models.ForeignKey(
+            'wagtailpages.UniversalLink',
+            null=True,
+            blank=True,
+            on_delete=models.SET_NULL,
+            related_name='+',
+        )
+
+    With an inline panel to surface its admin UI:
+
+        panels = [
+            InlinePanel('field_name', label='link')
+        ]
+
+    """
+
+    external_link = models.URLField(
+        blank=True
+    )
+
+    internal_link = models.ForeignKey(
+        'wagtailcore.Page',
+        blank=True,
+        null=True,
+        on_delete=models.SET_NULL,
+        related_name='internal_link',
+    )
+
+    panels = [
+        FieldPanel('external_link'),
+        PageChooserPanel('internal_link'),
+    ]
+
+    @property
+    def url(self):
+        if self.internal_link:
+            return self.internal_link.url
+        return self.external_link
+
+    def clean(self):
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/pagemodels/link.py
```diff
@@ -0,0 +1,76 @@
...
+        if not self.internal_link and not self.external_link:
+            message = "Please pick either a page or specify an external URL"
+            raise ValidationError({
+                'external_link': ValidationError(message),
+                'internal_link': ValidationError(message),
+            })
+
+    class Meta:
+        verbose_name = 'link'
+        verbose_name_plural = 'links'
```
KalobTaulien, line 76: I think we should make this an Abstract model if we're using this as a Django Mixin. 

```
class Meta:
    ....
    abstract = True
```

If we go abstract, we won't need the migration file but every other model that uses the UniversalLink will have a new migration. 

Happy to discuss further!

network-api/networkapi/wagtailpages/pagemodels/link.py
```diff
@@ -0,0 +1,76 @@
+from django.db import models
+
+from wagtail.admin.edit_handlers import FieldPanel
+from wagtail.admin.edit_handlers import PageChooserPanel
+
```
KalobTaulien, line 5: Needs `from django.core.exceptions import ValidationError`

network-api/networkapi/wagtailpages/pagemodels/link.py
```diff
@@ -0,0 +1,76 @@
...
+            on_delete=models.SET_NULL,
+            related_name='+',
+        )
+
+    With an inline panel to surface its admin UI:
+
+        panels = [
+            InlinePanel('field_name', label='link')
+        ]
+
```
Pomax, line 28:
```suggestion

    And a use in templates as:
    
        {{ page.field_name.url }}
```

network-api/networkapi/wagtailpages/pagemodels/link.py
```diff
@@ -0,0 +1,76 @@
...
+        """
+        super().clean()
+        if self.internal_link and self.external_link:
+            # Both fields are filled out
+            message = "Please ensure only one field has a value."
+            raise ValidationError({
+                'external_link': ValidationError(message),
+                'internal_link': ValidationError(message),
+            })
+        if not self.internal_link and not self.external_link:
```
KalobTaulien, line 70: This logic means one field _always_ needs to be filled out. I think this is good for most cases, but there might be a case to _not_ have either link field in use.

network-api/networkapi/wagtailpages/pagemodels/link.py
```diff
@@ -0,0 +1,79 @@
...
+            'wagtailpages.UniversalLink',
+            null=True,
+            blank=True,
+            on_delete=models.SET_NULL,
+            related_name='+',
+        )
+
+    With an inline panel to surface its admin UI:
+
+        panels = [
```
Pomax, line 25:
```suggestion
        content_panels = Page.content_panels + [
```

Pomax:
closing as just far more work than expected.

PR #4970, opened 2020-07-23
Removed todo

### LINKED ISSUE #4944: [Frontend] quote

KalobTaulien:
Closes #4944 
Related PRs/issues #4944 

This was pretty much done already by @mmmavis I believe (thank you!) 

I looked into a few different methods for passing in arguments into the `optimize_images` template tag, but nothing really stood out as "making it easier". It was either backend heavy with some math to autoscale different image sizes or frontend heavy where every include takes ~6 thumbnailing modes, which also doesnt seem very intuitive. 

I've opted to keep this the same for now and remove the todo comment. Should we see this being used in the future for, say, a StreamField, we should put this into it's own include. But for now, I'm happy where it's at. 

Happy to discuss this though!

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
@@ -135,13 +135,9 @@
     {% with quote_image=page.quote_image %}
     {% if quote_image %}
     <div class="col-md-4 col-lg-6 mb-5 mb-md-0">
-      {% comment %}
-        To-do:
-        1) Update picture_ratios.html so image sizes are passed as arguments instead of hard-corded.
-        2) Replace the following "picture" code with {% optimize_images page.quote_image %}
-      {% endcomment %}
       {% with alt_text=quote_image.alt %}
-      <picture class="cms-image {{class_names}}">
+
+      <picture class="cms-image">
           {% image quote_image width-540 as image_large %}
           {% image quote_image width-1080 as image_large_2x %}
           <source media="(min-width: 992px)" srcset="{{ image_large.url }}, {{ image_large_2x.url }} 2x"  alt="{{ alt_text }}">


network-api/networkapi/wagtailpages/templates/wagtailpages/homepage.html
```diff
@@ -90,17 +90,13 @@ <h2 class="h4-heading">{% trans "Trending projects" %}</h2>
       <a class="cta-link d-block d-md-none" href="https://www.mozillapulse.org/add">{% trans "Add your project" %}</a>
     </div>
   </div>
-  <div class="row section-quote">
+  <div class="row section-quote d-flex align-items-center">
```
mmmavis, line 93: @KalobTaulien my previous PR has passed design review. If we are making any design adjustment, let's find designers for a quick review 👀

KalobTaulien: @mmmavis Good point. This just centers the content if the image is too big and theres not enough text. 

@taisdesouzalessa thoughts on this?

taisdesouzalessa: @KalobTaulien I am not sure which component you are all talking about it here...do you have a screenshot? Tks!

KalobTaulien: @taisdesouzalessa sorry for the late reply here. 

image
It's this section. If the content gets too long the image sits at the top instead of centered with the content. I'm proposing we add vertical centering. We might not even see this as an issue if the content stays short.

KalobTaulien: Making a note that we spoke about this already. I'm resolving this. Thanks @taisdesouzalessa!

PR #4993, opened 2020-07-29
Homepage - News You Can Use

Pomax:
**(Will assign code reviewer once this passes design review)**

Eventually closes #4940 

Also note that this PR includes the factory work to make `load_fake_data` prepopulate the homepage with useful bindings.

PR #5024, opened 2020-08-05
Homepage factories

### LINKED ISSUE #4981: Audit which ia-refresh changes to the homepage still require factory code
we skipped the factory code changes for a number of changes to the homepage, leading to review apps having a homepage that requires quite a few things to be updated before you can publish what you _actually_ wanted to change.

@mmmavis @KalobTaulien this is more of an inconvenience than anything, so let's get this done but not necessarily before we finish up the front-end work =)

KalobTaulien:
Closes #4981 

Note this does not include the News You Can Use section. That's being taken care of in #4993

### DIFF
--- a/network-api/networkapi/wagtailpages/factory/__init__.py
+++ b/network-api/networkapi/wagtailpages/factory/__init__.py
@@ -1,7 +1,10 @@
 from . import (
+    areas_of_focus,
     blog,
     campaign_page,
     bannered_campaign_page,
+    homepage_areas_of_focus,
+    homepage_take_action,
     homepage_features,
     homepage,
     initiatives_page,
@@ -28,7 +31,9 @@ def generate(seed):
     participate_page_featured_highlights.generate(seed)
     styleguide.generate(seed)
     youtube_regrets_page.generate(seed)
-
+    areas_of_focus.generate(seed)
+    homepage_areas_of_focus.generate(seed)
+    homepage_take_action.generate(seed)
 
 __all__ = [
     'generate'
--- a/network-api/networkapi/wagtailpages/factory/__init__.py
+++ b/network-api/networkapi/wagtailpages/factory/__init__.py
@@ -18,12 +18,16 @@ from . import (
 
 
 def generate(seed):
+    areas_of_focus.generate(seed)
     homepage.generate(seed)
     blog.generate(seed)
+    bannered_campaign_page.generate(seed)
+    campaign_page.generate(seed)
+    # areas_of_focus.generate is required before the homepage_areas_of_focus can exist
+    homepage_areas_of_focus.generate(seed)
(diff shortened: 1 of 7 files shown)


network-api/networkapi/wagtailpages/factory/__init__.py
```diff
@@ -28,7 +31,9 @@ def generate(seed):
     participate_page_featured_highlights.generate(seed)
     styleguide.generate(seed)
     youtube_regrets_page.generate(seed)
-
+    areas_of_focus.generate(seed)
+    homepage_areas_of_focus.generate(seed)
+    homepage_take_action.generate(seed)
```
mmmavis, line 36: not a huge deal but let's order these alphabetically since it looks like that's what we have been trying to do here

network-api/networkapi/wagtailpages/factory/homepage_features.py
```diff
@@ -60,4 +61,12 @@ def generate(seed):
         for i in range(NUM_HIGHLIGHTS)
     ]
 
+    home_page.spotlight_posts = [
+        HomepageSpotlightPosts.objects.create(
+            page=home_page,
+            blog=BlogPage.objects.order_by("?").first()
+        )
+        for i in range(3)
```
mmmavis, line 69: So that this follows the existing code in this file -- let's store `3` in a variable `NUM_SPOTLIGHT_POSTS` and use that variable here

```py
NUM_BLOGS = 4
NUM_HIGHLIGHTS = 5
NUM_SPOTLIGHT_POSTS = 3 // add this line
```

network-api/networkapi/wagtailpages/models.py
```diff
@@ -3,15 +3,18 @@
 )
 
 from .pagemodels.base import (
+    AreaOfFocus,
```
mmmavis, line 6: `AreaOfFocus` and `FocusArea`.... so similar and so confusing 🙈. I'm gonna file a new ticket to rename them.

mmmavis:
Ah another request - I think it'll be useful if we can generate fake text for "Partner intro text" as well. (wondering if it should be a required field too? but this can be addressed outside of this PR)

network-api/networkapi/wagtailpages/models.py
KalobTaulien: Yes please. I originally had it called `AreaOfFocusOrderable` to explain what it is but that got renamed in another review.

KalobTaulien:
@mmmavis All updates made, thank you for these suggestions! 

> Ah another request - I think it'll be useful if we can generate fake text for "Partner intro text" as well. (wondering if it should be a required field too? but this can be addressed outside of this PR)

I added the partner_intro_text factory text, but I didn't make it a required field. I think adding a ticket for this would be best.

PR #5062, opened 2020-08-17
Implement hreflang via HTTP headers

### LINKED ISSUE #5049: Implement hreflang via HTTP headers
This enhancement is part of the PNI SEO audit (#4787). This is a low hanging fruit that should improve SEO for localized content throughout the website (not just PNI)

> The content is available in more languages. We should let crawlers know this by implementing hreflang ([via http headers](https://support.google.com/webmasters/answer/189077?hl=en)). Then other languages have a way better chance to rank in their language in search engines.

This seem like a good starting point https://github.com/mozilla/bedrock/blob/6ecb3903df3f841edde5958ee12955ff8b70d564/bedrock/base/templates/includes/canonical-url.html and [bug 1364470](https://bugzilla.mozilla.org/show_bug.cgi?id=1364470) has some relevant insights.

I’ll give it a shot and see if this is something I can fix myself.

TheoChevalier:
Closes #5049

Links localized pages in each page metadata so that the localized pages have better changes of ranking in Google.

- Google documentation: https://support.google.com/webmasters/answer/189077?hl=en

- Mozilla.org implementation: https://github.com/mozilla/bedrock/blob/master/bedrock/base/templates/includes/canonical-url.html

- Relevant discussion on why we’re excluding 3-character locale codes https://bugzilla.mozilla.org/show_bug.cgi?id=1364470

### DIFF
--- a/network-api/networkapi/context_processor.py
+++ b/network-api/networkapi/context_processor.py
@@ -19,5 +19,6 @@ def canonical_path(request):
     print(re.sub(r'^/' + lang, '', url))
     return {'CANONICAL_PATH': re.sub(r'^/' + lang, '', url)}
 
+
 def canonical_site_url(request):
     return {'CANONICAL_SITE_URL': request.scheme + '://' + request.get_host()}
--- a/network-api/networkapi/context_processor.py
+++ b/network-api/networkapi/context_processor.py
@@ -20,4 +20,4 @@ def canonical_path(request):
     return {'CANONICAL_PATH': re.sub(r'^/' + lang, '', url)}
 
 def canonical_site_url(request):
-    return {'CANONICAL_SITE_URL': settings.NETWORK_SITE_URL}
+    return {'CANONICAL_SITE_URL': request.scheme + '://' + request.get_host()}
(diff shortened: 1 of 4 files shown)


network-api/networkapi/context_processor.py
```diff
@@ -9,3 +11,14 @@ def review_app(request):
 # Used in BuyersGuide templates to check if we're using cloudinary
 def cloudinary(request):
     return {'USE_CLOUDINARY': settings.USE_CLOUDINARY}
+
+
+def canonical_path(request):
+    lang = getattr(request, 'locale', translation.get_language())
+    url = getattr(request, 'path', '/')
+    print(re.sub(r'^/' + lang, '', url))
```
Pomax, line 19: let's take this line out =)

Pomax (changes requested):
essentially R+ but let's remove the rogue print

network-api/networkapi/context_processor.py
TheoChevalier: woops! :) Thank you

Pomax (approved):

PR #5109, opened 2020-08-24
Text wrap homepage hero CTA button

### LINKED ISSUE #5105: [Homepage QA Bug] Long button gets cutoff
Originally mentioned in https://github.com/mozilla/foundation.mozilla.org/issues/5050#issuecomment-679260694

- [ ] MOBILE: reduce character counting of CTA to 26 **OR** allow button to break into multiple lines - button is getting cut in smaller screens - as Cade pointed above.
image

Note: 
It requires more discussion with the designers but this might be something we wanna consider setting in the style guide (i.e., all button text should wrap). It might require a few SCSS/DOM re-write to make button that has non-text element to align properly.

mmmavis:
Closes #5105

https://foundation-s-issue-5105-csd3hk.herokuapp.com


taisdesouzalessa:
Agree @Pomax - ideally we will always have a short CTA \0/

PR #5122, opened 2020-08-24
5101 fleshy articles

### LINKED ISSUE #5101: Publications | Flesh out article page model fields + author interface
As the publisher of a publication, I want an intuitive interface for managing the content that will be displayed on an article page so that readers can view the content on the web and that content can be easily coordinated between 'print'/pdf versions of the content. 

We've made the article page exist, but we need to give the CMS the attributes needed to fully represent an article page. 

The initial plan to do this is largely through Streamfields, mimicking other patterns in the CMS where it makes sense to have a consistent user experience. 

Concerns for MVP: 
* Where are the lines of control to be drawn over the layout of the page? Particularly around pull quotes or call out boxes
* What should we consider scope for MVP version of article page? Can we make a list of the components/blocks we want authors to be able to use. Ideally we'd keep the feature list tight and add stuff when we have justifiable reasons for doing so (i.e. it will achieve some goal of making the content more accessible)

Longer term, there may be some things to explore but aren't currently in scope:

* How does the chosen interface scale with very large/long publications?
* Many authors tend to prefer being able to copy stuff into Rich Text Fields, will a Streamfield heavy interface be a problem?

easherma:
Closes #5101 
Related PRs/issues #

## Checklist

_Remove unnecessary checks_

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?


network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -1,7 +1,27 @@
...
 from wagtail.core.models import Page
+from wagtail.core import blocks
+from wagtail.core.fields import StreamField
+from wagtail.admin.edit_handlers import FieldPanel, StreamFieldPanel
+from wagtail.contrib.table_block.blocks import TableBlock
 
 from ..mixin.foundation_metadata import FoundationMetadataPageMixin
 
+"""
+TODO:
```
KalobTaulien, line 13: Left over TODO. Could we possibly put this into a ticket for the current milestone and epic?

easherma: Sure thing. I’m hoping to explain and resolve this at standup today!

PR #5125, opened 2020-08-25
Fix 'take actions' card uneven width bug

### LINKED ISSUE #5123: [Homepage QA Bug] Take action cards have different widths
- [x] when we adjust the browser window to narrower width, this is what we see:
image

Those cards should have the same behavior that the other mobile cards in the page. 

Mockup:
https://redpen.io/mqd533ac95816f5e14

mmmavis:
Closes #5123

### DIFF
--- a/source/sass/views/home.scss
+++ b/source/sass/views/home.scss
@@ -89,10 +89,6 @@
       font-weight: bold;
     }
 
-    .card {
-      flex: 1;
-    }
-
     .join-us .btn {
       display: block;
       width: 100%;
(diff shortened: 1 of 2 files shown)


source/sass/views/home.scss
```diff
@@ -89,6 +89,10 @@
       font-weight: bold;
     }
 
+    .card {
+      flex: 1;
```
Pomax, line 93: this should also work as the bootstrap `flex-grow-1` rule, so we can try to replace the `align-items-stretch` with that

mmmavis: brilliant!

PR #5131, opened 2020-08-26
[5077] Chapter Table Template

### LINKED ISSUE #5077: Publications | Chapter table template
As per designs and as per field spec for chapter table title field.

Note: Designs show the author underneath the title of each chapter. We discussed this and it's actually going to be Chapter title with Article page titles underneath, as that's more interesting to users (we assume).

###  Design -> Dev Handoff Checklist

- [ ] Design has been finalized
- [ ] Link to original design ticket
- [ ] Link to related ticket
- [ ] Link to Redpen
- [ ] Design changes been merged into the master Sketch file
- [ ] Desktop and mobile mockups are included
- [ ] Assets are included
- [ ] Hover state for elements are designed
- [ ] Other states for elements are designed (e.g., error state for form)

KalobTaulien:
Closes #5077

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
@@ -1,5 +1,7 @@
 
 from django.db import models
+from django.db.models import QuerySet
+from typing import Union
 
 from wagtail.core.models import Page
 from wagtail.admin.edit_handlers import FieldPanel
@@ -20,12 +22,10 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
     From here the user can browse to the various sections (called chapters).
     It will have information on the publication, its authors, and metadata from it's children
 
-    TODO: this poem is beautiful, but it may not belong here
     Publications are collections of Articles
     Publications can also be broken down into Chapters, which are really just child publication pages
     Each of those Chapters may have several Articles
     An Article can only belong to one Chapter/Publication Page
-
     """
 
     subpage_types = ['ArticlePage', 'PublicationPage']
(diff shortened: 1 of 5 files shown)


network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -80,3 +84,31 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
         FieldPanel('contents_title'),
         FieldPanel('notes')
     ]
+
+    @property
+    def is_chapter_page(self) -> bool:
+        """Is this a chapter page (child-Publicationpage). Returns a bool."""
```
easherma, line 90: I think we could have something more useful here as a comment, as this is already pretty obvious with the code. 

Something like: 
"the template will render differently for the page depending on if it the page is the child of another publication page (which makes it a chapter page)"

Tangental but really hoping we come up with a more intuitive way of describing these page types..

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -80,3 +84,31 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
...
+        parent = self.get_parent().specific
+        return parent.__class__.__name__ == self.__class__.__name__
+
+    def get_chapter_number_or_none(self) -> Union[int, None]:
+        """Return the chapter page number or None."""
+        if self.is_chapter_page:
```
easherma, line 96: I think we need numbers either way, no (not just for chapter page)?

and I'm probably being dense, but if we just render these as an ordered list in html, wont they be numbered anyways? 

depending on the answers to those questions, the answer should probably be the docstring for the function :-D

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -80,3 +84,31 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
...
+            chapter_pages = list(self.get_siblings().specific())
+            return chapter_pages.index(self) + 1
+        return None
+
+    def get_child_article_pages(self) -> Union[QuerySet, list]:
```
easherma, line 101: hm, these next two functions are basically the same. I bet we can consolidate them. 

We already have the subpage_types set up to only allow ArticlePage or PublicationPage depending on the context.

Couldn't we: 
* get live children (my favorite kind of children...)
* check content type of the children
* if their content type is ArticlePage, do the thing. If their content type is Publication page, do the other thing. 

Clearly I'm not 100% sure what exactly we are trying to distinguish here for what purpose, so forgive me. 

(this made me think of an edge case...a publication page could have _some_ Article Pages and _some_ Publication pages, which could be bad? Or is fine?)

network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
```diff
@@ -47,4 +47,36 @@ <h1 class="h1-heading">
...
+            {% if not forloop.last %}
+              <div class="col-12 my-3">
+                <hr>
+              </div>
+            {% endif %}
+          </div>
+        {% endfor %}
+      </div>
+    {% else %}
+      {# Publication page #}
```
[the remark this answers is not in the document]
KalobTaulien: Publications can be nested under each other so we need to filter them out. We need to make sure 2nd level Publication Pages (and beyond) are considered "Chapter Pages", too, so we don't like both Chapters and Articles in the same index view. 

I moved the logic to the view to keep the template clean, too.

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
KalobTaulien: On Chapter pages you can have:
3.1
3.2
3.3

It's the number 3 in this case. 

For Publication Pages we can loop through each loop in the template and use `{{ forloop.counter }}` for numbering, or the numbering that comes with `<ol>`

network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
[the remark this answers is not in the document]
easherma: Sorry, this is just as much a refrence to the next couple lines: 
```
      <h1>is publication page</h1>
      <p>Show the chapters and article children in here</p>
```
I'm assuming this is a placeholder for more template work?

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
KalobTaulien: Correct on the edge case, which is why we're filtering pages with `.type(ArticlePage)` based on if the page is acting as a Chapter Page or Publication Page. This is also a reason to move nested publication pages to their own model type, such as a ChapterIndexPage.

PR #5135, opened 2020-08-26
[5076] Publication Page Authors

### LINKED ISSUE #5076: Publications | Publication page | Authors
Front and backend for authors section in the hero as per design and field spec.

###  Design -> Dev Handoff Checklist

- [ ] Design has been finalized
- [ ] Link to original design ticket
- [ ] Link to related ticket
- [ ] Link to Redpen
- [ ] Design changes been merged into the master Sketch file
- [ ] Desktop and mobile mockups are included
- [ ] Assets are included
- [ ] Hover state for elements are designed
- [ ] Other states for elements are designed (e.g., error state for form)

KalobTaulien:
Closes #5076 
Related PRs/issues #5076 

This work fetches the authors from any child or grandchild ArticlePage (which come from Orderables). Should we need to do this a different way we might want to consider closing this PR and opening a new one with similar (but different) work. 

It's the blog authors for the Publication Hero section, like below: 
image


network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
```diff
@@ -32,8 +32,29 @@ <h1 class="h1-heading">
                         {{ page.publication_date }}
                     {% endif %}
 
-                    {% if page.publication_file %}
 
+                    <div class="blog-authors d-flex flex-wrap align-items-center">
```
easherma, line 36: Is this template copy-pasted from elsewhere or is it different? I'm wondering if we have the opportunity to make a template block here.

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -80,3 +80,12 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
         FieldPanel('contents_title'),
         FieldPanel('notes')
     ]
+
+    def get_authors(self) -> set:
```
easherma, line 84: I think the intention of this is to also check child pages for authors and percolate that up to the publication page, is that correct?

In which case, cool, but I'd like to double check that that is a requirement in spec. I got the impression from talks that some reports will define authors at the 'root' level, and some at the 'article' level, and so it would be up to editors in the CMS to define where they want author stuff to show. 

If not, this is a handy feature to have, and I'd want to make sure we communicate it as available functionality. *party emoji*

network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
easherma, line 36: Is this template copy-pasted from elsewhere or is it different? I'm wondering if we have the opportunity to make a reusable template block here.

KalobTaulien: It sure is! It came from the homepage work that just got merged into master. The fragment is `blog_authors.html`. I just threw this in here for now and was thinking about refactoring this when we do the official FE work.

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
KalobTaulien: I think this is a blocker. We need clarification before proceeding with this one.

easherma: After design/dev sync today we confirmed: 

* Authors are optional and can be added to PublicationPage or ArticlePage
* Both pages will show Authors if they are selected for that page
* They will not show authors that have been selected for other pages

network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
easherma: Makes sense to me!

KalobTaulien:
@easherma Updated this PR so it uses an Orderable to explicitly select Authors per Publication page. Work is in the latest commit https://github.com/mozilla/foundation.mozilla.org/pull/5135/commits/3dbe975fec0e518ff07d8ea95f0fd81b5264d8d2 — could you quickly review once more, and if you're happy merge in?

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -72,20 +95,16 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
     )
 
     content_panels = Page.content_panels + [
-        FieldPanel('subtitle'),
-        FieldPanel('secondary_subtitle'),
-        FieldPanel('publication_date'),
-        ImageChooserPanel('hero_image'),
-        DocumentChooserPanel('publication_file'),
+        MultiFieldPanel([
```
easherma, line 98: minor quibbles:
* if Authors are optional and therefore not always entered, I'd rather have them below other required or commonly entered stuff like the Hero
* curious about the motivation to put the other hero fields into a multifield panel, im assuming thats to group things a bit more semantically. In which case wouldn't authors be apart of that since they are also in the hero?

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -72,20 +95,16 @@ class PublicationPage(FoundationMetadataPageMixin, Page):
...
+            InlinePanel("authors", label="Author", min_num=1)
```
easherma, line 99: Authors should be totally optional, `min_num=None` would be more appropriate and prevent editors/authors from getting confused about a required field if it isn't required.

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -44,6 +67,9 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
     )
 
     content_panels = Page.content_panels + [
+        MultiFieldPanel([
+            InlinePanel("authors", label="Author", min_num=1)
```
easherma, line 71: Same here, no min_num

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -23,6 +27,25 @@
...
 
+class ArticleAuthors(Orderable):
+    """This allows us to select one or more blog authors from Snippets."""
+
+    page = ParentalKey("wagtailpages.ArticlePage", related_name="authors")
+    author = models.ForeignKey(
+        BlogAuthor,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=False
```
easherma, line 38: Authors are optional

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -1,19 +1,42 @@
...
+
+class PublicationAuthors(Orderable):
+    """This allows us to select one or more blog authors from Snippets."""
+
+    page = ParentalKey("wagtailpages.PublicationPage", related_name="authors")
+    author = models.ForeignKey(
+        BlogAuthor,
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=False
```
easherma, line 26: Authors should be optional, according to specs not every publication is going to have specific authors cited

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
KalobTaulien: No reason. It's because I copy/pastas'd from some other authors work. Will move `authors` into the hero `MultiFieldPanel`

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
KalobTaulien: Yep, these are optional. It's set by the `min_num` value with `null=True`. Blank just means if an author Orderable "Add Author" button is clicked, then make sure it's filled out instead of providing an empty value.

KalobTaulien:
@easherma Can you give this a once over again? I added the additional copy field, moved the fields you suggested and removed authors being mandatory. (Just the last 4 commits)

PR #5137, opened 2020-08-27
5077 easherma patch

easherma:
Closes #
Related PRs/issues #
@KalobTaulien I mighta had too much fun with this, but I'm into it. :-D

Learned a couple neat things!

Some of the advantages of this approach as I see it: 
* less conditional checking and looping in the template. although trival, this should be more performant. and is less template stuff to manage.
* relying on django_treebeard's functionality means less code maintence on our end, as we're mainly exposing existing functionality
* relying on css and html native attributes means we have a better shot at the content being more accessible and cross-browser compatible

Disadvantages: 
* this approach is a bit less obvious and ought to be documented and understood
    * for instance, it relies on these css properties for numbering, but then tweaks them a bit. its a bit less obvious on read-through how it handles chapter pages vs publication pages, though it does work according to specs. 
* the template should be pretty portable and flexible, but it could be possible to introduce bugs if the implementation isn't fully understood and we don't currently have any test coverage to prevent this from happening

## Checklist

_Remove unnecessary checks_

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
@@ -48,42 +48,17 @@
         </div>
         {{ page.contents_title }}
 
-    {% if page.is_chapter_page %}
-      {# Chapter index page; display child article pages #}
-      <div class="container-sm background-white chapter-body">
-        {% for child in page.get_child_article_pages %}
-          <div class="row d-flex align-items-center">
-            <div class="col">
-              <div class="chapter-number">
-                <div>
-                  {{ page.get_chapter_number_or_none }}.{{ forloop.counter }}
-                </div>
-              </div>
-            </div>
-            <div class="col-md-11">
-              <a href="{{ child.url }}" class="d-block w-100 h3 chapter-article-link">
-                {{ child.title }}
-              </a>
-            </div>
-            {% if not forloop.last %}
-              <div class="col-12 my-3">
-                <hr>
-              </div>
-            {% endif %}
-          </div>
-        {% endfor %}
-      </div>
-    {% else %}
-      {# Publication page #}
+      {# page summary #}
       <div class="child-pages-summary">
       {% comment %} https://django-treebeard.readthedocs.io/en/stable/api.html#treebeard.models.Node.get_annotated_list {% endcomment %}
         {% for child, node_info in page.my_children %}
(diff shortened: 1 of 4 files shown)


source/sass/components/chapter-page.scss
```diff
@@ -18,3 +18,28 @@
 }
 
 // .chapter-article-link {}
+
```
easherma, line 21: Just kinda stuck these here for now...but we could just rename this file to publication-page and stick styles here. Also does it count as a component? I'd expect component files to be smaller elements of this page and not styles for larger elements/the page itself. We can sort this out later though.

network-api/networkapi/wagtailpages/templates/wagtailpages/publication_page.html
```diff
@@ -48,35 +48,25 @@ <h1 class="h1-heading">
...
-              </div>
+      {# page summary #}
+      <div class="child-pages-summary">
+      {% comment %} https://django-treebeard.readthedocs.io/en/stable/api.html#treebeard.models.Node.get_annotated_list {% endcomment %}
+        {% for child, node_info in page.my_children %}
+            {% comment %} 'open' means 'there are siblings so open the list tag' {% endcomment %}
+            {% if node_info.open %}
+                <ol><li data-chapter="{{page.get_chapter_number_or_none|default_if_none:''}}">
+            {% else %}
+                </li><li data-chapter="{{page.get_chapter_number_or_none|default_if_none:''}}">
```
KalobTaulien, line 59: What if there are chapter pages? This is assuming all child pages will be Article Pages, is it not? 

An example would be the design below, what is there's:
1. Article Page 
2. Chapter Page 
3. Article Page 
4. Chapter Page 
5. etc etc. 
image

easherma: It will list any type of child page. Conceptually, in that case the ArticlePage is still a chapter, even if its a single page chapter.

KalobTaulien: image

This is a Publication Page -> Publication Page -> Publication Page, it's mixed in with the Article Pages. 

I'm thinking we should create the Chapter Page at this point. Using a Publication Page as 2 different page types is getting too complex. We could create a PublicationPage and use `class ChapterPage(PublicationPage)` and add overwrites. But the newly specified content_type could make traversing easier. 

Let's have a chat about this before continuing.

Issue #5145, opened 2020-08-31
Make previous and next buttons


easherma:
* might be good to add these to the style guide
* @sabrinang how do previous and next buttons work if there isn't another sibling page to browse to? apologies if this is already noted somewhere else. 
* when @KalobTaulien and I talked about it, he mentioned maybe needing test coverage for

sabrinang:
For buttons I was thinking:
- if it is the first page, the previous button can direct to its parent
- if it is the last page, the next button can direct to its parent
- if it has no siblings, the previous button can direct to its parent (no next button)

I think this would be easier to bring the user back to see the overview of the content again than to leave the button empty

phildexter:
Where is that at also @KalobTaulien ? should this ticket still exist or is this part of the larger PR?

PR #5166, opened 2020-09-01
[5143] ArticlePage Frontend

### LINKED ISSUE #5143: Publication Article Page frontend
it exists, make it pretty. there is a lot to check here in terms of the various components as designed and where they should display on a page.

KalobTaulien:
Closes #5143 
Related PRs/issues #5143 

## Checklist

**Tests**
- [ ] Browser tested 
- [ ] Has desktop, tablet and mobile responsiveness


Zenix27 (approved):

network-api/networkapi/wagtailpages/pagemodels/article_fields.py
```diff
@@ -0,0 +1,20 @@
+"""
```
easherma, line 1: good idea to split this out!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/article_table_block.html
```diff
@@ -0,0 +1,72 @@
+{% load table_block_tags %}
+
+  </div> <!-- End .col-sm-8 -->
+  <div class="col-sm-10 offset-sm-1">
+
+  <div class="article-table-block table-responsive">
+    {% comment %}
```
easherma, line 7: great comment! lots of good template comments

source/images/right-chevron.svg
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
easherma, line 1: I'd want to hear from @cadecairos or @Pomax on this, should these live here, or maybe under images/glyps? Or are there any icon dependencies in the project that might already have chevron arrows for the links?

source/sass/components/article-page.scss
```diff
@@ -0,0 +1,316 @@
+.article-summary {
```
easherma, line 1: There is potentially alot of technical debt in this css. Some of the values being pulled, like font-family, are already defined as variables. 

The same may be true for some of these colors, and for the ones where they aren't, ideally we'd check with design and verify that they are adding new colors to the palette for the site, make new variables, and add them to the style guide.

As much as possible, we should use the Mozilla Bootstrap theme to keep things consistent. It's totally possible that there are aspects of the design spec that don't match what is defined for the site at large; for better long term maintenance it would be best for the Mozilla team to be able to reconcile all these discrepancies. 

All that being said, given our time frame it may be better to not count these issues as blocking.

easherma (commented):
see comments. wasn't going to approve since this is still marked as a draft

network-api/networkapi/templates/pages/base.html
```diff
@@ -31,7 +31,7 @@
     {% block stylesheets %}
       <link rel="stylesheet" href="{% static "_css/main.compiled.css" %}">
       <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,300,700,300i,800,900">
-      <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,400,600,700,300i">
+      <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,400,500,600,700,300i">
```
easherma, line 34: if this is something new needed for design, maybe they want a change to say one of the other font weights is fine

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -71,5 +68,54 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
             InlinePanel("authors", label="Author", min_num=0)
         ], heading="Author(s)"),
         StreamFieldPanel('body'),
+        InlinePanel("footnotes", label="Footnotes"),
         FieldPanel('sidebar_summary_title'),
     ]
+
+    @property
+    def next_page(self) -> Union[Page, None]:
```
easherma, line 76: For QA, want to make sure this behavior matches: https://github.com/mozilla/foundation.mozilla.org/issues/5145#issuecomment-683946332

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -71,5 +68,54 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
...
+        parent_sibling = self.get_parent().get_prev_sibling()
+        if parent_sibling:
+            # Check if parenet page is an ArticlePage
+            if isinstance(parent_sibling.specific, ArticlePage):
+                return parent_sibling
+            # Parent page was not an ArticlePage, return the first
+            # live child, or None
+            return parent_sibling.get_children().live().last()
+
+    def get_titles(self):
```
easherma, line 109: i'd be interested to see what affect getting these titles this way has on performance *nerd glasses emoji*

source/images/right-chevron.svg
cadecairos: looks like there's already a lot of disorganization in `source/images/`. However, I do see a `down-chevron.svg` in `glyphs` so we should probably move this there.

source/sass/components/article-page.scss
cadecairos: I'd defer to @mmmavis on situations like this. If we need to get this merged to unblock other PRs, let's make sure we open a follow-up issue to discuss potential improvements and/or work to stick to the styleguide

easherma:
@KalobTaulien when running locally, I got this error trying to view a fake article page. Maybe cause it is missing some data? Should the template be able to handle this sort of exception?
err

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
KalobTaulien: Should be very little. All stream data is kept in a single field as a dictionary, then we loop through it to find "content" streamblocks, and use BeautifulSoup to pull out the `h2` elements. There are no additional queries, the biggest thing is BS4 parsing the HTML (Or rather, the "wagtail" HTML) and that should be quite minimal as it's not ever saved as deeply nested HTML. That said, I'd be interested to see the performance on this as well. That could be a spike ticket though.

KalobTaulien:
> @KalobTaulien when running locally, I got this error trying to view a fake article page. Maybe cause it is missing some data? Should the template be able to handle this sort of exception?

@easherma I can't seem to replicate that. Does this happen still when you create a new env with `inv new-env`?

network-api/networkapi/templates/pages/base.html
KalobTaulien: Good question. I asked the design team in Slack.

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
KalobTaulien: I've re-written the functions based on the #5145 comment. I think they're small enough now and different enough that if we wanted to write tests for them it'd be easier as two functions.

network-api/networkapi/templates/pages/base.html
KalobTaulien: We're not using 500. Ill have that removed shortly.

source/js/dropdowns.js
```diff
@@ -0,0 +1,27 @@
...
+    for (let i = 0; i < dropdowns.length; i++) {
+      const element = dropdowns[i];
+      element.addEventListener('click', function(e) {
+        e.preventDefault()
+        element.classList.toggle("show")
+        const id = element.id
+        const target = document.querySelector(`[aria-labelledby="${id}"]`)
+        target.classList.toggle("d-block")
+      })
+    }
```
Pomax, line 14: Handy bit of API: `querySelectorAll` returns an array-like that does not offer `map`, but _does_ offer `forEach`:

```suggestion
    document.querySelectorAll(`.dropdown-toggle`).forEach(element =>
      element.addEventListener(`click`,  _ => {
        element.classList.toggle("show");
        const target = document.querySelector(`[aria-labelledby="${element.id}"]`);
        target.classList.toggle("d-block");
      })
    );
```

Also note that if we use proper `<button>` markup, there is no default behaviour that needs preventing.

source/js/dropdowns.js
```diff
@@ -0,0 +1,27 @@
...
+
+    // Close dropdowns when item is clicked
+    const dropdownItems = document.querySelectorAll('.dropdown-item')
+    for (let i = 0; i < dropdownItems.length; i++) {
+      const item = dropdownItems[i];
+      item.addEventListener('click', function() {
+        item.parentElement.classList.remove('d-block', 'show')
```
Pomax, line 21: Same kind of rewrite as above.

Pomax (changes requested):

Pomax:
It might be a good idea to squash-rebase before working on this some more, because 35 commits is definitely too many commits to easily diff between =)

Pomax:
I've also told heroku to rebuild the review app, so there should be a https://foundation-s-5143-artic-6u4msf.herokuapp.com up for the next few days.

network-api/networkapi/wagtailpages/factory/publication.py
```diff
@@ -97,6 +97,11 @@ class Meta:
 
 
 def generate(seed):
+    """
+    makes a batch of 3 publication pages
+    eventually I'd like to add a post_generation hooks that gives each of these pages an arbitrary number of chapters
+    and chapters an arbitrary number of articles
+    """
```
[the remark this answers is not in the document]
KalobTaulien: I think this seeding is good where it is. It does the job it's supposed to do. Ill update the docstring.

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -71,5 +68,38 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
             InlinePanel("authors", label="Author", min_num=0)
         ], heading="Author(s)"),
         StreamFieldPanel('body'),
+        InlinePanel("footnotes", label="Footnotes"),
         FieldPanel('sidebar_summary_title'),
     ]
+
+    @property
+    def next_page(self) -> Union[Page, None]:
```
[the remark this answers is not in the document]
KalobTaulien: Roger that.

network-api/networkapi/utility/templatetags/remove_richtext_wrapping.py
```diff
@@ -0,0 +1,19 @@
+from django import template
+from bs4 import BeautifulSoup
+register = template.Library()
+
+
+@register.filter
+def remove_wrapping(content=None):
+    """
+    Removes the .rich-text wrapper around richtext elements.
```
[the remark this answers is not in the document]
KalobTaulien: To remove styling from other rich-text areas on the site. This is used in the image captions on the Article Page, which is a richtext area, but we dont want all the default .rich-text styling applied to it. However, we want to keep links, bold, and italics.

PR #5167, opened 2020-09-02
Seed fix for FocusArea objects

KalobTaulien:
Closes #5139 
Related PRs/issues #5139

PR #5173, opened 2020-09-03
Run django-modeltranslation tasks after migration task

cadecairos:
... to account for data migrations that may have taken place before a model was registered for translation

Closes #5167


easherma (approved):
LGTM! Only main downside I could think of is whether or not the migration process is slowed significantly by running/checking this every time, but that dosen't feel like a blocker.

PR #5188, opened 2020-09-08
rely on python's random, rather than database

### LINKED ISSUE #5186: Blog Author random assignment breaks Percy regression tests
We're randomly assigning Blog Authors in our tests in a way that's not reproducible between builds. Let's investigate ways to make it reproducible.

Pomax:
Closes #5186

Replaces the database level randomization with python level randomization

### DIFF
--- a/network-api/networkapi/wagtailpages/factory/blog.py
+++ b/network-api/networkapi/wagtailpages/factory/blog.py
@@ -1,5 +1,5 @@
 from datetime import timezone
-from random import choice, randint
+from random import choice, randint, shuffle
 
 from django.conf import settings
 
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -54,9 +54,11 @@ def add_category(post):
 
 
 def add_authors(post):
-    authors = list(BlogAuthor.objects.order_by("?").all())
+    authors = list(BlogAuthor.objects.all())
     count = len(authors)
 
+    shuffle(authors)
```
Pomax, line 60: shuffle reorders in-place.

cadecairos (approved):
R+

PR #5212, opened 2020-09-10
Use random.choice instead of database random ordering

### LINKED ISSUE #5202: Factory code is still not deterministically random, breaking Percy
Percy is still generating new content for the same PRNG seed, making PR review far more of a hassle than it should be. This needs a fixing.

unplanned, but important enough to just get done so _everyone_ is unblocked by that noise.

Pomax:
Closes #5202

This should make Percy stable again, replacing all left-over `order_by("?")`, which is a database level randomization, with Python's `random.choice`, which respects our seed values.


mmmavis (changes requested):
Thanks for the explanation regarding `order_by("?")` vs `random.choice`.

Code looks good! Looks like there's one more place we need to update -- in `homepage_features.py`

```py
home_page.spotlight_posts = [
        HomepageSpotlightPosts.objects.create(
            page=home_page,
            blog=BlogPage.objects.order_by("?").first()
        )
        for i in range(NUM_SPOTLIGHT_POSTS)
    ]
```

Pomax:
That's already in the PR though, https://github.com/mozilla/foundation.mozilla.org/pull/5212/files#diff-608a11ec663c53b9ff8711ab42768fc8R47 ?

mmmavis (approved):
My bad. Ignore what I said earlier regarding fixing one more place. I think I had too many tabs open and probably was on a wrong branch when i checked the code lol.

cadecairos (approved):

PR #5214, opened 2020-09-10
add breakpoints and columns to styleguide

### LINKED ISSUE #5132: Add a section to the styleguide that shows off the bootstrap grid
Just to show off the base grid breakdown, in a way that can be copied for use in sketch - at the very least col12 and 12 x col1? Would we need any other sizes for usefulness or would those two cover it?

/cc @sabrinang @mmmavis

Pomax:
Closes #5132

I tried to stick to bootstrap's own data here - bootstrap doesn't have a nice "masking" red baked in, so I can't use our own `$red` without writing custom CSS for that, which didn't quite seem worth it? Happy to add it if that would be more useful though.

URL: https://foundation-s-styleguide-pwd8w3.herokuapp.com/en/styleguide/ (scroll to the "grid" section)


mmmavis (commented):
Wow cool! I like how you laid out the grid.

### From PR

image

The white space in between each column represents `gutter` from Bootstrap. I wonder if we could remove the gaps in between **rows** so it doesn't look like Bootstrap rows come with horizontal gutters?

So I tried but it looked awful and more confusing lol. Design suggestion from @sabrinang needed 😂 

image

Pomax:
Yeah, I initially had it collapsed but it just makes it a giant black blob =D

sabrinang:
I think horizontal gutter is fine as it is more confusing without it.

I was wondering if we could do something like: https://www.c-sharpcorner.com/article/bootstrap-grid-system/Images/1.png where it could help visually show different proportions of column combinations.

Pomax:
I had one of those before but I couldn't find a good way to do the ones that are missing in that image. I can just mirror their graphic though, without col-5, col-7 and col-9/col-10/col-11?

Pomax:
updated the grid to be similar to bootstrap's image, with some missing values added:
- 5 / 7
- 9 / 3
- 10 / 2
 -11 / 1

mmmavis (approved):
Perfect!

sabrinang:
This style is looking good! Can we add: 
- 6 / 6 
- 8 / 4
 
(after 5 / 7 before 9 / 3) I see that 6 and 8 col are missing too

Pomax:
ah, good catch. Added the missing 6/6 and 8/4, and use alternative layout to make it just a tad better to look at. I can make it more of a "fat triangle" if that's more useful.

spread out (current):

screenshot 325

fat triangle:

screenshot 324

sabrinang (approved):
Looks great Pomax and I like the fat triangle method, it makes it easier to scan.

PR #5259, opened 2020-09-17
Expect SafeText from RichText.__html__ in with_heading_ids proxy method

cadecairos:
We got a report that anchor links no longer worked. Upon investigation it looks like our code to insert ids into headers in RichText fields was broken:

https://github.com/mozilla/foundation.mozilla.org/blob/5a94dea934c5a13e5a85188e0100e3265b92ed72/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py#L79-L80

The above check always evaluates to `True` because `RichText.__html__()` is now passed an instance of `django.utils.safestring.SafeText`

This PR updates the check for the new expected type.

### DIFF
--- a/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
+++ b/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
@@ -2,6 +2,7 @@ import re
 from django import template
 from django.utils.text import slugify
 from django.utils.html import strip_tags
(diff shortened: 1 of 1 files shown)


cadecairos:
To test, I added an anchor to this campaign page on the review app: https://foundation-s-cadecairos-j3e1pw.herokuapp.com/en/campaigns/its-dream-contain-fall-six-learn/

Pomax (commented):
good find.

PR #5264, opened 2020-09-17
fix the RSS feed entry selection

### LINKED ISSUE #5258: RSS feed may be broken
### Describe the bug
It seems that some of the items which are spotlighted in the blog are not appearing in the RSS feed (https://foundation.mozilla.org/en/blog/rss/). At the time of writing, the following post is a spotlight that does not show up in the feed: https://foundation.mozilla.org/en/blog/2-million-award-reimagine-economics-online-journalism/

### Expected behavior
Everything in the blog should show up in the RSS feed as soon as it is published.

Pomax:
Closes #5258

Switches the RSS selector from BlogIndexPage to IndexPage, as the BlogIndexPage's `get_all_entries` filters out all featured blog items, which we definitely want for the blog page, but definitely _don't_ want for RSS purposes.

Code-reason why this was going wrong: https://github.com/mozilla/foundation.mozilla.org/blob/e5979d0f00ae9b81415a27f782b4bc236eef0bd3/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L61-L70

That L69-L70 should not run for RSS/Atom purposes

### DIFF
--- a/network-api/networkapi/wagtailpages/rss.py
+++ b/network-api/networkapi/wagtailpages/rss.py
@@ -3,7 +3,7 @@ from django.contrib.syndication.views import Feed
 from django.utils.feedgenerator import Atom1Feed
 from django.utils.text import Truncator
 
-from .models import BlogIndexPage
+from .models import IndexPage
 
 
 class RSSFeed(Feed):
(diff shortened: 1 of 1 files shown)


cadecairos (approved):

Carlos123211 (approved):

Pomax:
Reported Carlos123211 as bot, because it obviously is.

Pomax:
Hm, why is this causing a PNI page to render differently?

PR #5272, opened 2020-09-21
Update Payment pointer to MoFo Uphold account

cadecairos:
Stronghold has stopped accepting Web Monetization transactions, so we're going to use Uphold now.

### DIFF
--- a/network-api/networkapi/buyersguide/templates/bg_base.html
+++ b/network-api/networkapi/buyersguide/templates/bg_base.html
@@ -43,7 +43,7 @@
 
 
 {% block additional_head_elements %}
-  <meta name="monetization" content="$pay.stronghold.co/1a1938a9a3ec0ea4c539d779be2b7e1c9ee">
+  <meta name="monetization" content="$ilp.uphold.com/aKpM9kwnGBy4">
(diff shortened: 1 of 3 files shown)


network-api/networkapi/buyersguide/templates/bg_base.html
```diff
@@ -43,7 +43,7 @@
 
 
 {% block additional_head_elements %}
-  <meta name="monetization" content="$pay.stronghold.co/1a1938a9a3ec0ea4c539d779be2b7e1c9ee">
+  <meta name="monetization" content="$ilp.uphold.com/aKpM9kwnGBy4">
```
Pomax, line 46: can we make that identifier this an env var?

Pomax (changes requested):

network-api/networkapi/buyersguide/templates/bg_base.html
cadecairos: I suppose that'd make it easier to update in the future, but I doubt it'll change again anytime soon.

Pomax (approved):

Pomax (approved):
sweet

PR #5280, opened 2020-09-21
update PNI v4 fields

### LINKED ISSUE #5241: PNI v4 Product details backend updates
### Final Designs

criteria spreadsheet:

https://docs.google.com/spreadsheets/d/1cOOzLhkoICFIPTUmxQnrH10HZz1qq5eHF51qgj1p8QM/edit#gid=0

main field work:

- [x] remove the old fields + associated panels
- [x] add the new fields
- [x]  transform old-to-new (Software "sign in" fields)
- [x] Add a CMS checkbox for “dings or no dings”
- [x] Add “wifi required” checkbox

finesse work:

- [x] update the CMS model panel ordering so that the fields in the admin match the on-page fields ordering.
- [x] Update the factories to use the new fields

Pomax:
Fixes #5241
Manually verified `inv test` passes, as the PR does not run CI verification

See issue for the spreadsheet of fields

CMS product edit link: https://foundation-s-update-pro-wgzuq2.herokuapp.com/cms/buyersguide/generalproduct/edit/1/

As this also includes a migration that copies over the "signup" fields from `SoftwareProduct` to `Product` for any software products that were already using them, this will need to be tested against staging data, too.

- be on the base `pni-v4-2020` or `master` branches, and copy the staging data over to your local postgres docker container
- run everything and take now of what the `Zoom` software product says are its signup field values
- checkout this branch, then run `inv migrate`, then run everything
- verify that `Zoom` has the new field _names_, but with same values as the old fields


kristinashu:
Awesome!!!

### Blockers
I only see two blockers but it would be good if @caltrider could review as well.
- [ ] `Offline capable` field is currently a `check-box` but should be `Yes/No/NA/Unknown`
- [ ] Remove the helptext from `User friendly privacy policy` (note this is the "User friendly" section not the general "Privacy policy" section which does and should have the helptext field) 

### Nice to have
I think there are a few copy and order tweaks that could make things easier for content entry but will have @caltrider review it for those in more detail. Maybe we could chat tomorrow.

A key one would be to put the Minimum Security Standard in the order they appear on the site (Encryption, Strong password, Security updates, Manages vulnerabilities, Privacy policy). Particularly Privacy policy since it's in a separate section right now.

It would also be easier to use if these longer titles could be on one line rather than three. E.g.
Screen Shot 2020-09-21 at 4 26 55 PM

kristinashu:
I met with Jen and she didn't find any additional issues but would find it very helpful if the order and titles in the CMS could match the [spreadsheet](https://docs.google.com/spreadsheets/d/1cOOzLhkoICFIPTUmxQnrH10HZz1qq5eHF51qgj1p8QM/edit#gid=0). In some cases, having the shorter titles in the CMS makes the page more scanable and easier to enter the content.

kristinashu (changes requested):
Looking good just a few things mentioned above.

Pomax:
@kristinashu that is not the ordering the fields have in the spreadsheet though: does the spreadsheet need updating, or does the site need updating? 

Order in the spreadsheet:
- Does / Does not meet Minimum Security Standards
- Privacy ding
- Encryption
- Security updates
- Strong password
- Manages vulnerabilities
- Privacy policy

kristinashu:
Sorry, the order in the spreadsheet is right - `Security updates` should be above `Strong password`.

Pomax:
@kristinashu PR updated, including field with update. No idea why wagtail thinks 16.667% width makes sense, but that's been given an override.

(note, it's a new review app, so you'll need to use the new admin password)

kristinashu (approved):
Looks good!

network-api/networkapi/buyersguide/factory.py
```diff
@@ -112,44 +87,93 @@ def product_category(self, create, extracted, **kwargs):
...
 
+    uses_ai = LazyFunction(get_extended_yes_no_value)
+    ai_uses_personal_data = LazyFunction(get_extended_yes_no_value)
+    ai_is_transparent = LazyFunction(get_extended_yes_no_value)
+    ai_helptext = Faker('sentence')
+
+    phone_number = Faker('phone_number')
+    live_chat = Faker('url')
+    email = Faker('email')
+    # twitter
```
mmmavis, line 156: Should we generate fake Twitter handle?

mmmavis, line 156: let's generate fake Twitter handle as well

network-api/networkapi/wagtailcustomization/static/css/admin.css
```diff
@@ -0,0 +1,7 @@
+/*
+  This is 16% by default, which is absolutely ridiculous.
+*/
+fieldset ul li div label {
+  display: block;
+  width: auto;
+}
```
mmmavis, line 7: Can we move CSS code in this [file](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailcustomization/templates/wagtailadmin/base.html) to here? If so, I will file a new ticket.

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -206,7 +206,7 @@ def insert_panels_after(panels, after_label, additional_panels):
     )
 
     if position is not None:
-        cut = position + 2
+        cut = position + 1
```
mmmavis, line 209: Wondering why it was `position + 2` before and `position + 1` now? What are the offsets for?

mmmavis (changes requested):
Looking awesome ✨ 

I just have a few very tiny "requestions" (lol i was gonna type questions/request and my fingers accidentally created this new word).

network-api/networkapi/wagtailpages/utils.py
Pomax: it was an off-by-one bug. It's the position of where we need to insert a collection of panels inside the parents' list of panels. The idea is to insert after whatever label is indicated, so it needs to find that index, and then cut the panels list _after_ that index. For some reason, we were cutting one index too far.

network-api/networkapi/wagtailcustomization/static/css/admin.css
Pomax: yeah, that's a good idea!

Pomax: filed https://github.com/mozilla/foundation.mozilla.org/issues/5290

Pomax:
@mmmavis added tweet factories, with the percy product having an explicit twitter handle. I slipped in some zero-width non-breaking spaces, so that it _looks_ like a twitter handle, but isn't actually valid.

network-api/networkapi/buyersguide/pagemodels/products/base.py
```diff
@@ -206,12 +269,128 @@ def specific(self):
...
+        default=False,
+    )
+
+    # Company track record
+
+    track_record_choices = [
+        ('Great', 'Great'),
+        ('Average', 'Average'),
+        ('Needs Improvement', 'Needs Improvement'),
+        ('Bad', 'Bad')
```
TheoChevalier, line 346: How is this going to be localized? Do we need a template tag similar to https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/templatetags/yes_no.py#L19 ?

network-api/networkapi/wagtailpages/utils.py
mmmavis: ah got it! Thanks for the explanation!

network-api/networkapi/buyersguide/pagemodels/products/base.py
Pomax: correct.

Pomax: filed https://github.com/mozilla/foundation.mozilla.org/issues/5292 so we don't forget to do that!

Pomax:
@KalobTaulien @mmmavis I've updated this PR, with some instructions on what additional testing steps have become necessary...

KalobTaulien (approved):
Tested out the staging db, migrations work as expected. Twitter field doesnt cause me any problems on a mac/chrome.

mmmavis (commented):
image

@Pomax followed your instructions but got this after i ran `inv migrate` on this branch

Pomax:
@mmmavis  excellent catch, forgot to update the software template after removing the fields!

mmmavis (approved):
✨ ✨ ✨ ✨ ✨ ✨ ✨

Issue #5094, opened 2020-08-20
make tasks.py's "create_super_user" a task


Pomax:
It's currently not marked as `@task` so it can't be run using `inv create-super-user`, which means trying to administer a copy of the staging db is unreasonably involved.

PR #5291, opened 2020-09-23
make create-super-user an inv task

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5094


KalobTaulien (approved):
Works for me, and saves me from having to enter the docker container to create a superuser.

Pomax:
@KalobTaulien in the mean time, you can run `inv manage "shell -c \"from django.contrib.auth.models import User; User.objects.create_superuser('admin', 'admin@example.com', 'admin')\"" ` if you just want inv to do it for you

It's so convenient!

PR #5331, opened 2020-09-28
MSS section updates

mmmavis:
Closes #5237 

Code reviewer:
I fixed a bunch indentation issues so it's best to view code difference **without** whitespace changes.

----

Please only review the "Security" section. Fixes to other sections will be handled in other tickets.

image

- has ding: https://foundation-s-issue-5237-qfk8ve.herokuapp.com/en/privacynotincluded/products/percy-cypress/
- no ding: https://foundation-s-issue-5237-qfk8ve.herokuapp.com/en/privacynotincluded/products/short-of/

Issue #5237, opened 2020-09-15
Minimum Security Standard updates


jamilasnell:
### Todo
- [x] Hide MSS badge from hero banner
- [x] Re-style the criteria section (only use font styles that are already in use on production)
- [ ] Add accordion (default state is closed) - will be covered in #5257
- [x] Show privacy ding if answer is `No` - [icon](https://drive.google.com/drive/folders/1kPMOO9YoMybLN-MqziLUsmGE2EQ3hAVp?usp=sharing)
- [x] Move section down the page 

### Closed state desktop
<img width="862" alt="image" src="https://user-images.githubusercontent.com/1068783/93500801-6e3fa080-f8c9-11ea-93f2-ea15ed9e0ed1.png">

### Open state desktop
<img width="871" alt="image" src="https://user-images.githubusercontent.com/1068783/93500828-78619f00-f8c9-11ea-9c2a-300ba8e154ef.png">

### Closed state mobile
<img width="318" alt="image" src="https://user-images.githubusercontent.com/1068783/93509863-32123d00-f8d5-11ea-9aea-16f2e3176a7f.png">

### Open state mobile
<img width="251" alt="image" src="https://user-images.githubusercontent.com/1068783/93510130-9c2ae200-f8d5-11ea-87a7-e8ffcc49ead7.png">

---

###  Design -> Dev Handoff Checklist

- [x] Design has been finalized
- [x] Link to original design ticket #5152
- [X] Link to related ticket #5257
- [x] Link to [Sketch file](https://share.goabstract.com/584e7002-4d07-4ce5-aafe-dc825e076a27?sha=f0b8ca1b588425d1ef83c57a9f4126c8ff97648e)
- [x] Design changes been merged into the master Sketch file
- [x] Desktop and mobile mockups are included
- [x] Assets are included - [icon](https://drive.google.com/drive/folders/1kPMOO9YoMybLN-MqziLUsmGE2EQ3hAVp?usp=sharing)
- [x] Hover state for elements are designed - no hover state for the accordion / link: use what is on production
- [x] Other states for elements are designed (e.g., error state for form) - n/a

mmmavis:
Gonna update this ticket as this is actually blocking the [Add accordion style to MSS ticket](https://github.com/mozilla/foundation.mozilla.org/issues/5257), instead of the other way around.

mmmavis:
> Show privacy ding if answer is `No`

[...]

PR #5348, opened 2020-09-30
PNI: Small l10n fixes for product page

TheoChevalier:
So more l10n fixes on top of #5331

Related issues #5237


mmmavis (approved):
Thanks!

Should `’` always be used instead of `'` ? I started a thread on Slack regarding apostrophes uses. Let's chat there.

PR #5353, opened 2020-09-30
Exclude the job if this is the PNI feature branch

cadecairos:
This PR makes the percy job a conditional stage on this branch.


cadecairos:
I overlooked the PR branch name. will dig into the docs.

Pomax (approved):
:shipit:

PR #5376, opened 2020-10-05
Fix "copy page" js

### LINKED ISSUE #5374: Copy Page flow is not quite working
As per https://mozilla.slack.com/archives/C98GW3P7X/p1601914889018500 the copy functionality does not seem to correctly report errors (error fields are not shown after a submit)

Pomax:
Closes #5374

Updates the wagtail-modeltranslation JS to generate a "done" event, which the copy page JS can hook into so that we can show all fields with errors irrespective of whether the localization field-hiding code kicks in before our reveal-all code, or our reveal-code kicks in before the hide-all code.

I'll file a PR against wagtail-modeltranslation to get that change to their JS taken on board (though possibly with a better/different event name)


KalobTaulien (approved):
Code looks good. Tested it out, no issues which is good (I think). 

Yes let s try to get this into the modeltranslation package for less code to maintain outside the package.

Pomax:
PR filed upstream as https://github.com/infoportugal/wagtail-modeltranslation/pull/307

Issue #5372, opened 2020-10-02
Update product page template code so criteria show up correctly based on the product type


mmmavis:
Back-end ticket: https://github.com/mozilla/foundation.mozilla.org/issues/5371

All new criteria listed in this [sheet](https://docs.google.com/spreadsheets/d/1cOOzLhkoICFIPTUmxQnrH10HZz1qq5eHF51qgj1p8QM/edit#gid=0) should only be added to "general" products. "Software" product fields should stay unchanged. Once the model changes are in, we should update the template code so criteria show up correctly based on the product type.

Pomax:
merged

PR #5386, opened 2020-10-06
Update field owners

### LINKED ISSUE #5371: New criteria should only be available for "general" products
Model changes needed!

All new criteria listed in this [sheet](https://docs.google.com/spreadsheets/d/1cOOzLhkoICFIPTUmxQnrH10HZz1qq5eHF51qgj1p8QM/edit#gid=0) should only be added to "general" products. "Software" product fields should stay unchanged.

/cc @kristinashu @Pomax

Pomax:
Closes #5371
Related PRs/issues #5372

This moves all the "general only" fields from `BaseProduct` to `GeneralProduct`. as per https://docs.google.com/spreadsheets/d/1cOOzLhkoICFIPTUmxQnrH10HZz1qq5eHF51qgj1p8QM/edit#gid=0, with a custom migration to ensure that even though Django does this by removing the fields from `BaseProduct` and then building brand new fields with the same name on `GeneralProduct`, we cache the values before that destructive relocation happens, and then restore the values for all those fields once the fields have been rebuilt.

The "easiest" way to test this for devs is:

1. check out pni-v4-2020
1. run `inv new-db`
1. run `docker-compose up` and navigate to http://localhost:8000/cms/buyersguide/generalproduct/edit/1/
1. set every choice to "yes" or "true" or checked, and set every textfield you can to "this is a custom test string".
1. save that product.
1. exit docker
1. check out `update-field-owners`
1. run `inv migrate`, which should apply the fancy migration
1. run `docker-compose up`
1. load up http://localhost:8000/cms/buyersguide/generalproduct/edit/1/ again
1. look at every field you updated.

If everything went well, everything should still be yes/true/checked/custom string.

For the design/pm side of things. the important part is that the review app shows the fields that should be visible for General Products as being there in the CMS, while not showing up in the Software Product part of the CMS.


KalobTaulien (approved):
👍 LGTM, nice work with the migration caching and restoration. 

And thank you for the testing instructions!

kristinashu:
@Pomax am I doing this right? I created a page under Software:
image

But when I go to look at it it looks like a General page https://foundation-s-update-fie-sm9stq.herokuapp.com/en/privacynotincluded/products/test/.

This look right in the cms but not on the front-end.

Pomax:
@kristina this PR is only for the CMS side of things. The template changes are tracked in https://github.com/mozilla/foundation.mozilla.org/issues/5372

Pomax:
"How does it use this data?" got incorrectly moved to General from Base, updating the PR to keep that in Base since Software should also show this

mmmavis (approved):
looks great to me!

kristinashu:
I'm sorry but I'm getting a "There's nothing here, yet." when I try to View deployment. I'm on PTO tomorrow so I don't think I'll be able to review this in a timely way. Hopefully Kalob and Mavis' reviews are enough and I can double check things once it's on https://pni-v4-2020.mofostaging.net .

kristinashu (approved):
Looks good!

PR #5397, opened 2020-10-07
Homepage accessibility improvements

### LINKED ISSUE #4008: [fo.mo] Homepage Accessibility Testing
---
## Step 1️ : Automated Testing*

  
- [ ] Use `eslint-plugin-jsx-a11y` to fix a11y issues
   > Run  `npm test` and fix any issues related to the fo.mo homepage
  

- [x] Use `pa11y` to fix a11y issues
   > Run `npm start` then `pa11y http://localhost:8000/privacynotincluded/about/example-about-path` to fix any issues related to the fo.mo homepage

## Step 2️ : Web Testing*

- [x] Use [axe](https://www.deque.com/axe/axe-for-web/) browser extension to fix any remaining issues related to the fo.mo homepage
  > Available for Chrome + Firefox

## Step 3️: Manual Testing

- [x] Use [Accessibility Insights for Web](https://accessibilityinsights.io/docs/en/web/overview) browser extension _Assessment_ feature to guide you through manual testing issues related to the fo.mo homepage
  > *Currently only Available in Chrome + Edge
  > *Assessments 2-23 (_as applicable_)

- [x] Export + Upload Completed Assessments 
  > [Assessment_20201007_MozillaFoundationHom.html.zip](https://github.com/mozilla/foundation.mozilla.org/files/5341809/Assessment_20201007_MozillaFoundationHom.html.zip)

  
- [ ] Follow up ticket for further action/discussion (_if applicable_)
  > `insert follow-up ticket here` 👈 

 
---
Related Issue: #3934
_* Updates should cover the overarching key actions listed at the bottom of the [master a11y ticket](https://github.com/mozilla/foundation.mozilla.org/issues/3934)._

KalobTaulien:
Closes #4008 

In #4008 is the uploaded results from the accessibility report in the first comment. 

This PR holds _a lot_ of very small changes. 

Some things to note:
* <source> doesnt need an alt tag 
* `#primary-nav-container` got changed to `.primary-nav-container` so we didnt have two "nav" elements nested in each other (for jumping around using a screen reader)
* main content is wrapped in a `<main>`, the hero is wrapped in a `<header>`, and the footer has `contentinfo` on it. This creates 4 main sections of the site (including the `<nav>`) 
* Lots of things now have aria labels for screenreaders that dont read innertext or have different innertext 
* Headings were changed so the top of the page has an h1, and the bottom has h6s. They should be ordered from biggest to smallest down the page. But Ive added bootstrap classes to keep the styling the same. (I think there's one or two exceptions where the heading has small text and should be a smaller semantic heading)
* **Maybe the most controversial change** is the homepage hero where the brightness went from 55% to 45% in order to pass contrast tests.


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/focus_areas.html
```diff
@@ -2,7 +2,7 @@
 
 <div class="row areas-of-focus">
   <div class="col my-3 my-sm-5">
-    <h2 class="focus-heading text-center mb-4">{% trans "Our areas of focus" %}</h4>
```
mmmavis, line 5: 🙈

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
```diff
@@ -1,7 +1,7 @@
 {% load wagtailcore_tags wagtailimages_tags static %}
 
 <div id="hero">
-  <div class="banner">
+  <div class="banner bg-black">
```
mmmavis, line 4: Not a biggie but just curious how this solved a11y issue?

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/spotlight_posts.html
```diff
@@ -3,7 +3,7 @@
 <div class="row section-spotlight section-padding">
 
   <div class="col-12">
-    <h2 class="capsule-label mb-0 ml-md-4">{% trans "Spotlight" %}</h2>
+    <h5 class="capsule-label mb-0 ml-md-4">{% trans "Spotlight" %}</h5>
```
mmmavis, line 6: I checked with Tais on this - `Spotlight` is a section heading and not just for the "cover" on the left. Let's revert this back to `h2` so it can sit on the same hierarchy level as other section headings!

mmmavis:
```suggestion
    <h2 class="capsule-label mb-0 ml-md-4">{% trans "Spotlight" %}</h2>
```

mmmavis (changes requested):
Awesome work! I only checked through the code changes and didn't launch any a11y addon to do the check. Please let me know if I should be using any.

---

> * `#primary-nav-container` got changed to `.primary-nav-container` so we didnt have two "nav" elements nested in each other (for jumping around using a screen reader)

I'm wondering how changing `primary-nav-container` from an id to class helped improve a11y. Can we have it stayed as an id as it's bound to a bunch JS code (which currently is broken by the change 😢 ) and that we wanna make sure there's only at most one `primary-nav-container` on every page?

---

The brightness change on homepage hero works fine to me. I'll let designers know about this change but I'm sure they will be cool with it as well since this improves a11y!

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
KalobTaulien: There was a complaint about the background color not being set with the text color. _Somehow_ this fixed it. It changes nothing tbh, but satisfied a11y ¯\_(ツ)_/¯

KalobTaulien:
@mmmavis Oh good find about the JS. I changed those to class selectors in the latest commit. :tada: 

The reason it's a class and not an ID is because browsers were recognizing this is an aria landmark (navigation) and there's a <nav> nested inside it, so it had a nested nav landmark inside the parent one, which could be troublesome for screenreaders and assistive technologies. But for some reason (and I dont know why) but making it a class was good enough. I didnt want to rename the ID and change too much.

mmmavis:
@KalobTaulien I'm not sure if this is ready for another review? If so, please re-request review from me so this can show up on my pending review queue 😉

mmmavis (approved):
Looks good to me! Wow a11y is indeed a hard topic. Glad there are addons to help.

mmmavis (commented):
@KalobTaulien some change here seems to unintentionally break big heading size on YouTube Regrets page (as per Percy flag). (Also note that that page already has a `<main>`) Can you make sure we don't introduce UI changes before merging this PR? Thanks!

mmmavis (changes requested):
See comment above.

KalobTaulien:
> @KalobTaulien some change here seems to unintentionally break big intro text size on YouTube Regrets page (as per Percy flag). (Also note that that page already has a `<main>` so now it has two `<main>`) Can you make sure we don't introduce UI changes before merging this PR? Thanks!

@mmmavis I've added in a content wrapper block to overwrite the <main> element from bg_base.html. I'm waiting for Percy to flag more things, but in theory, this latest commit should work fine. I couldn't replicate what Percy originally found though so I'm going off of theory at this point.

mmmavis (changes requested):
Looks like there are more visual regression issues on YouTube Regrets and RegretReporter page. 🙈

KalobTaulien:
@mmmavis This visual regression looks to be fixed but Percy is showing something very different. Wondering if you can help me resolve this?

mmmavis (approved):

PR #5399, opened 2020-10-07
FieldPanel fix

### LINKED ISSUE #5367: Blog page settings bug
First published at is now overlapping the privacy button - this changed in 2.10.

KalobTaulien:
Closes #5367 

I had to rearrange the panels for the overlapping to be fixed. Not sure if this is a bug or not, I'll be exploring that shortly and if it is a bug I'll open an issue on wagtail/wagtail about it. But the PrivacyModalPanel doesn't seem to be providing enough height; that was is the issue. 

Here's what it looks like now:
image

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
@@ -1,7 +1,14 @@
 from django.db import models
 from django.conf import settings
 
-from wagtail.admin.edit_handlers import FieldPanel, InlinePanel, MultiFieldPanel, StreamFieldPanel
+from wagtail.admin.edit_handlers import (
+    FieldPanel,
+    InlinePanel,
+    MultiFieldPanel,
+    PrivacyModalPanel,
+    PublishingPanel,
+    StreamFieldPanel,
+)
 from wagtail.core import blocks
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -122,8 +129,10 @@ class BlogPage(FoundationMetadataPageMixin, Page):
         FieldPanel('tags'),
     ]
 
-    settings_panels = Page.settings_panels + [
+    settings_panels = [
+        PublishingPanel(),
         FieldPanel('first_published_at'),
+        PrivacyModalPanel(),
```
Pomax, line 135: This feels like it would impact everything, not just blogs?

KalobTaulien: Nope, just the BlogPage because it's only being set on the `BlogPage` class.

KalobTaulien:
A proper fix is coming. @zerolab opened a PR for this issue at https://github.com/wagtail/wagtail/pull/6453 but lets not wait for this to get into core wagtail with a version bump.

Pomax (approved):

PR #5339, opened 2020-09-29
PNI: Hook up CMS fields for localization

TheoChevalier:
Closes #5338

This makes various PNI fields localizable using wagtail-modeltranslation

Issue #5338, opened 2020-09-29
PNI: localize fields


TheoChevalier:
As part of the revised l10n plan, let’s hook up all CMS fields for localization using the system already in place, wagtail-modeltranslation

https://docs.google.com/document/d/1RunBlPOJ7m-KfIDwazBHL3EO6Di1MQtj9pqtTDpdZwg/edit?ts=5f7226fd#heading=h.g8i1bgv9jn4e

KalobTaulien:
Hey @TheoChevalier I requested access to this doc. I'll take a look at it tomorrow morning

TheoChevalier:
> Hey @TheoChevalier I requested access to this doc. I'll take a look at it tomorrow morning

Thanks @KalobTaulien! You should be able to access the doc now. You’re welcome to take over this PR/branch https://github.com/mozilla/foundation.mozilla.org/pull/5339

The remaining issues (I tagged you in the doc):
- Fields are duplicated for some class (e.g "name" and "name [en]"), even after running `inv l10n-sync` and `inv l10n-update`.
- Localizing the "label" field in the `ProductPrivacyPolicyLink` class doesn’t work at all (the privacy policy link panel appears empty in the admin panel)

KalobTaulien:
Hey @TheoChevalier I don't see the duplicate fields you mentioned. 

![image](https://user-images.githubusercontent.com/4743971/94831982-9ac7e200-03ca-11eb-9ff0-e33b2ec92886.png)

Though that's normal to have a `name` and `name_en` field when translating because as far as I can tell when we make a new env with `inv new-env` (or `inv new-db` it applies some raw SQL to add columns to tables.

KalobTaulien:
I've added in the label translations for `ProductPrivacyPolicyLink` 

![image](https://user-images.githubusercontent.com/4743971/94835624-2e031680-03cf-11eb-96b3-e286be210c70.png)

The fix here was adding `panels = [...]` to the Orderable model. I also removed the inherited `models.Model` class because `Orderable` already extends from `models.Model`

TheoChevalier:
Thanks a lot for fixing `ProductPrivacyPolicyLink` !

> Hey @TheoChevalier I don't see the duplicate fields you mentioned.
> 
> ![image](https://user-images.githubusercontent.com/4743971/94831982-9ac7e200-03ca-11eb-9ff0-e33b2ec92886.png)
> 
[...]

PR #5447, opened 2020-10-14
PNI Localization: second iteration

### LINKED ISSUE #5442: PNI Localize Fields (Second Iteration)
This is the second iteration of #5338. 

History: We are running into problems with #5338, namely the `inv test-python` command is not passing and it's hard to debug. This ticket is designed to re-create 5338 (and the relevant PR) in order to have a cleaner commit history so we can bisect the code easier and determine the cause of the failed test. 

#5339 is the original PR, which I'll be re-creating as well with a cleaner commit history for easier bisecting as well. 

From Theo's original comment:
> As part of the revised l10n plan, let’s hook up all CMS fields for localization using the system already in place, wagtail-modeltranslation
> https://docs.google.com/document/d/1RunBlPOJ7m-KfIDwazBHL3EO6Di1MQtj9pqtTDpdZwg/edit?ts=5f7226fd#heading=h.g8i1bgv9jn4e

There is currently a comment/sub-issue surrounding the final work where the `name` and `name_en` fields are both exposed, but only one should be shown. This may or may not be a bug, and if it is a bug it may be something we can live with as it's existence doesn't hurt the editors UX on such a small field (Category snippets, for instance). 

Both #5338 and #5339 will be closed once this ticket and relevant PR are opened.

KalobTaulien:
Closes #5442 
Related PRs/issues #5338 #5339 

## Tests failing for unknown reasons

Tests are failing because of translation migrations and we don't currently understand why. We need to get to the bottom of this.


KalobTaulien:
```
  File "/app/network-api/networkapi/buyersguide/tests.py", line 385, in test_product_view                                                                                                                                   
    self.assertEqual(response.status_code, 200, 'The product view should render once no longer a draft')                                                                                                                    
AssertionError: 404 != 200 : The product view should render once no longer a draft  
```
☝️ The latest commit prior to this comment (https://github.com/mozilla/foundation.mozilla.org/pull/5447/commits/bd2eb6896ae7038f0a924750b62271ef43c5f08b) creates this test to fail. This test previously passed before the migration file was applied. It looks a lot like the GeneralProduct has empty data in it, and I'm uncertain why. 

After this, we'll run into another issue with a similar behavior – if we can solve this issue we can likely solve the next failed test that will come up afterward.

Issue #5414, opened 2020-10-09
PNI - Verify that ProductCategories are Orderable


mmmavis:
- [x] Verify that ProductCategories are Orderable

---

**Context**

We want the categories to show up on the category nav in the order of

<img width="1129" alt="image" src="https://user-images.githubusercontent.com/2896608/95622243-f11acd80-0a27-11eb-9a86-4c1179a705b9.png">

(note that "All" will be hardcoded in template)

Pomax:
Unfortunately, as per https://docs.wagtail.io/en/stable/topics/snippets.html?highlight=order#snippets: 

>Snippets lack many of the features of pages, such as being orderable in the Wagtail admin or having a defined URL. Decide carefully if the content type you would want to build into a snippet might be more suited to a page. As per @KalobTaulien's suggestion, this is something that we would get by converting PNI from snippets-based Django app to "real" Wagtail pages app, but that is out of scope of the next 2 sprints.

However, in the mean time we can introduce an ordering so that we can at least manually fix the ordering, as per https://stackoverflow.com/questions/63472227/wagtail-how-to-order-snippets

PR #5449, opened 2020-10-14
simple sorting order

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5414 by adding naive sorting to the product category object - it's a snippet, so we can't make them Orderable

STR: fire up the admin interface, change the ordering numbers for some buyers guide product categories, then look at the pni page to see whether that ordering is reflected in the nav headings.


KalobTaulien (approved):
LGTM 🎉

Issue #5441, opened 2020-10-14
Change django.conf.urls.url() to django.urls.re_path()


Sparsh-Bansal:
The old `django.conf.urls.url()` function is replaced with new `django.urls.re_path()` function for routing .
This would help : https://docs.djangoproject.com/en/3.1/ref/urls/#url

The `django.conf.urls.url()` function is an alias for `django.urls.re_path()` in 2.2.*, and can be updated in place.

For `Django >= 3.x`,  `django.conf.urls.url()` will be removed. (https://docs.djangoproject.com/en/dev/internals/deprecation/#deprecation-removed-in-4-0)

### To Reproduce
1. Go to all the required urls.py files.
2. Change `django.conf.urls.url` to `django.urls.re_path`.
3. Change all the `url` to `re_path`

Pomax:
We will not be updating Django for probably another few months, but thank you for filing.

Sparsh-Bansal:
Sure, I will land a PR once the Django version will get updated.

Sparsh-Bansal:
Sure , I will land a PR today

PR #5452, opened 2020-10-15
Change url() to re_path()

Sparsh-Bansal:
Closes [#5441 ](https://github.com/mozilla/foundation.mozilla.org/issues/5441)

I have changed the old `django.conf.urls.url()` function with new `django.urls.re_path()` function for routing in all the required `urls.py` files.


Sparsh-Bansal:
@Pomax  Do I have to care about this `percy/foundation.mozilla.org Expected — Waiting for status to be reported` check or not?
If yes what do I need to do for this check to pass?

Pomax:
I've triggered a retest by updating the branch, I think Percy just got stuck, but it does need to either pass or fail before anything can be accepted.

Sparsh-Bansal:
@Pomax what can we do to produce some result of `percy` check ?

Pomax:
percy runs as part of the standard CI, based on the tests found in `cypress/integration/endpoint-tests,js`

Pomax (approved):

Sparsh-Bansal:
Percy test are passing [here](https://travis-ci.org/github/mozilla/foundation.mozilla.org/jobs/737204291). Then why the `percy/foundation.mozilla.org ` check is not passing.

Pomax:
Those are not the percy tests, that's the percy test _initiator_. Travis runs Cypress, which loads up pages, and creates snapshots, and then also triggers the Percy task for those same pages. Percy runs on percy.io and has a separate entry in the CI test bank (you can see it as "waiting for status to be reported"). Github and Travis have been having issues that last few days, so I fully expect this to be a matter of "percy ran, but it couldn't report the status because github's having problems".

Pomax:
Right so this keeps failing to trigger an actual Percy run, so I'm going to merge this into an local branch instead, and then see if the PR for that will cleanly pass CI. Thanks for the help!

Pomax:
refiled as https://github.com/mozilla/foundation.mozilla.org/pull/5572

Issue #5289, opened 2020-09-23
investigate replacing eslint with prettier


Pomax:
we use ESLint because it applies `prettier` rules as well as performs JSX linting, but `prettier` might at this point have full JSX support, which would allow us to cut out ESLint entirely. Especially with `prettier`'s `--write` option, it would turn "error reporting" into "and immediately fixing", which means that part of `inv test-node` wouldn't even be able to fail anymore, which would be a big win.

Pomax:
Prettier can't run on "bad" JS or SCSS, so we still need ESLint to find and flag those errors, but what we _can_ do is add "pre" tasks that run prettier on all .js, .jsx, and .scss files so that at the very least we'll never run into styling mismatches anymore.

PR #5456, opened 2020-10-15
Force prettier rules

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5289 by adding `prettier ... --write` for all js/jsx/scss tasks, so that by the time esline/stylelint run, it is impossible for them to find _style_ errors, they'll only find _actual_ errors (if there are any), because prettier can't spot those: it simply doesn't run on files with parsing errors.

I've added these tasks as part of `npm test`, so that they also run as part of `inv test` because of `inv test-node`.


package.json
```diff
@@ -35,8 +35,11 @@
     "server": "python network-api/manage.py runserver",
     "start": "docker-compose up",
     "test:procfile": "node test/test-procfile.js",
+    "test:eslint:js:pre": "prettier \"cypress/integration/*.js\" \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js --write",
     "test:eslint:js": "eslint --config ./.eslintrc.json \"cypress/integration/*.js\" \"source/js/**/*.js\" \"source/js/**/*.jsx\" webpack.config.js",
+    "test:eslint:a11y:pre": "prettier \"source/js/**/*.jsx\" --write",
     "test:eslint:a11y": "eslint --config ./.eslintrc.a11y.json \"source/js/**/*.jsx\"",
+    "test:scss:pre": "prettier \"source/sass/**/*.scss\" \"source/js/**/*.scss\" \"network-api/networkapi/{,!(frontend)/**/}*.css\" --write",
```
mmmavis, line 42: do we need to care about `.css` files?

KalobTaulien: I thought it was `pre` as in "run before something else"` but then I saw the prettier command and dont know which is which now. Could we make `:pre` something more explicit, possibly?

Pomax: @mmmavis we have a few bare .css files in overrides, we might as well format them

@KalobTaulien no, you got that right: this runs "pre" the actual `test:scss` running, blindingly applying code formatting before running code _linting_

KalobTaulien: Perfect. Thanks for adding that clarification Pomax!

mmmavis (approved):

package.json
mmmavis: Ah I forgot we have a few `.css`. Good call including them.

Issue #5204, opened 2020-09-10
Update default pages created by load_fake_data


cadecairos:
Current nav:

- Who we are (primary page)
- initiatives (initiatives page)
- news  (news page)
- What you can do (Participate page2)
- style-guide (styleguide)

Should be:

- Who we are  (primary page), already in place
- What we do (banner page), does not exist
- What you can do (Participate page2), already in place
- Blog (blog index), exists but needs to be put in the nav

Tasks:

- [x] create "what we do" fake_data page
- [x] update fake data so that "what we do" and "blog" links show_in_menu
- [x] update fake data so that old pages no longer show_in_menu

Question marks:

- [ ] ~remove initiatives page?~
  - [ ] ~remove all code for this?~
- [ ] ~remove news page?~
  - [ ] ~remove all code for this?~

Pomax:
@kristinashu are we keeping the initiatives and news pages, or should those get cleaned up (probably in a separate issue, if so)

kristinashu:
Initiatives will go away but design still has some work to do. Content of the news page will also go away but design needs to check with Kevin first (and then we're going to update the "blog" name to "news" - this was based on the IA refresh work. I believe most of this url work by design will be done in https://github.com/mozilla/foundation.mozilla.org/issues/5068.

PR #5458, opened 2020-10-16
Fake data fixes

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5204

STR: `inv new-env` (this PR secretly bumps up the node version for docker) and then run the server once that's done. The nav should show the correct items.

### DIFF
--- a/network-api/networkapi/templates/reviewapp-help.html
+++ b/network-api/networkapi/templates/reviewapp-help.html
@@ -13,15 +13,16 @@
         </div>
         <p>Test pages with placeholder content:</p>
         <ul>
-          <li><a href="/about/">About</a></li>
-          <li><a href="/style-guide/">Style guide</a></li>
-          <li><a href="/news/">News</a></li>
-          <li><a href="/initiatives/">Initiatives</a></li>
-          <li><a href="/initiatives/single-page-opportunity/">Single page opportunity</a></li>
-          <li><a href="/initiatives/multi-page-opportunity/">Multi page opportunity</a></li>
+          <li><a href="/who-we-are/">Who we are</a></li>
+          <li><a href="/what-we-do/">What we do</a></li>
+          <li><a href="/what-you-can-do/">What you can do</a></li>
+          <li><a href="/Blog/">Blog</a></li>
+          <li>-</li>
           <li><a href="/campaigns/single-page/">Single page campaign</a></li>
           <li><a href="/campaigns/multi-page/">Multi page campaign</a></li>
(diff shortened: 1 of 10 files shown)


network-api/networkapi/templates/reviewapp-help.html
```diff
@@ -13,15 +13,16 @@ <h1 class="h1-heading">Review app help page</h1>
...
-          <li><a href="/about/">About</a></li>
-          <li><a href="/style-guide/">Style guide</a></li>
-          <li><a href="/news/">News</a></li>
-          <li><a href="/initiatives/">Initiatives</a></li>
-          <li><a href="/initiatives/single-page-opportunity/">Single page opportunity</a></li>
-          <li><a href="/initiatives/multi-page-opportunity/">Multi page opportunity</a></li>
+          <li><a href="/who-we-are/">Who we are</a></li>
+          <li><a href="/what-we-do/">What we do</a></li>
+          <li><a href="/what-you-can-do/">What you can do</a></li>
+          <li><a href="/Blog/">Blog</a></li>
```
KalobTaulien, line 19: Uppercase b in `/Blog/`, is that an issue?

network-api/networkapi/wagtailpages/factory/__init__.py
```diff
@@ -19,7 +19,9 @@
 
 
 def generate(seed):
+    # the first three entries are ordered the way they are to ensure correct nav ordering
```
KalobTaulien, line 22: Glad you left this comment. 

Should we consider refactoring this at a later date so there's a _specific_ way to handle the nav ordering 9or nav creation)?

KalobTaulien (approved):
A couple of comments, but otherwise LGTM.

network-api/networkapi/templates/reviewapp-help.html
Pomax: not a true "issue" but good to keep lowercase.

Pomax:
```suggestion
          <li><a href="/blog/">Blog</a></li>
```

network-api/networkapi/wagtailpages/factory/__init__.py
Pomax: the live nav gets built by looking at the show_in_menu status of a homepage.child, but since we're literally in the step before we have any of those, and given how infrequently we change the front page, I think it's probably fine to leave it as is rather than adding code to running an explicit ordering. As long as we keep the comments that explain why things do what they do =)

PR #5464, opened 2020-10-19
Cleanup

### LINKED ISSUE #5359: ArticlePage StreamField SCSS cleanup
### File(s) or components

ArticlePage template (`article_page.html`) and it's related StreamField blocks need CSS cleanup. This work is around #5075 

### Reason for cleanup

There are some inconsistencies around things like font size and line heights that need to be checked with a designer and frontend dev. 

### Tasks:
- [x] child-pages-children-title should just use .body style but with line height override 24px instead of 27px. Need to check with the designers to see if this override is really necessary.
- [x] Check with designer about font size and line height overwrides (ie. `@include set-text-size(16px, 24px);`)
- [x] Check with designer about font-weight: 300 in `.publication-notes .rich-text p` 
- [ ] Article Summary Bar  z-index needs to be increased so it doesn't get overlapped by the footer 
- [x] Don't use BEM styling for `publication-hero-container--small`

KalobTaulien:
Closes #5359 

### Tasks:
- [x] child-pages-children-title should just use .body style but with line height override 24px instead of 27px. Need to check with the designers to see if this override is really necessary.
- [x] Check with designer about font size and line height overwrides (ie. `@include set-text-size(16px, 24px);`)
- [x] Check with designer about font-weight: 300 in `.publication-notes .rich-text p` 
- [x] Article Summary Bar  z-index needs to be increased so it doesn't get overlapped by the footer 
- [x] Don't use BEM styling for `publication-hero-container--small`

---

> - [ ] child-pages-children-title should just use .body style but with line height override 24px instead of 27px. Need to check with the designers to see if this override is really necessary.

☝️ This matches the design. And the only thing that's being overwritten is the font-size and font-family, as they should be.


mmmavis (changes requested):
> child-pages-children-title should just use .body style but with line height override 24px instead of 27px. Need to check with the designers to see if this override is really necessary.

Looks like on the PR `child-pages-children-title` is in font size 1rem/16px but we want it to be in 18px. I've asked Sabrina [on Slack](https://mozilla.slack.com/archives/C4M19QU2Z/p1603154682025500?thread_ts=1603138033.011800&cid=C4M19QU2Z) if we can just use `.body` style (without the 24px line height override). We'll know more once she gets back to us.

> Article Summary Bar z-index needs to be increased so it doesn't get overlapped by the footer

`footer` might be a typo? I'm guessing it should be `nav newsletter`?
image
STR:
1. toggle the summary dropdown
2. click on the "Newsletter" button on the top nav

KalobTaulien:
Thank you @mmmavis! I was looking in the wrong spot. 

I've adjusted the z-index's for the summary bar and the newsletter bar.

KalobTaulien:
Sabrina had said:
> The main override here is additional paragraph spacing 12px (so it is a bit futher apart) but line height itself can align to be 27px to match body that exists.

^ I removed the 16px font size in favour of 18px with the default line-height of 27px (1.5x of the font-size). I had to slightly bump down the numbering text to the left of the child page titles.

mmmavis (approved):
🎉 YAY

PR #5473, opened 2020-10-19
Enable PublicationPage as child pages

KalobTaulien:
Related Issue: #5487 

PublicationPages need to be able to exist under the following page types:

- [x] Banner Page
- [x] Campaign page
- [x] Default Page
- [x] Homepage


Pomax:
Can we file an issue? Adding new page types as children to (banneder)campaigns and opportunities may either break or require updating of the code that we use to build tertiary menus. Specifically, the code for `get_page_tree_information`, which relies on class matching.

Pomax (changes requested):
unfortunately, this needs verification of some of our menu building code for the page types that support tertiary menus.

KalobTaulien:
@Pomax Good call on that. The CampaignIndex page would have shown the PublicationPage as a child page, even though it shouldn't have. The `load more` button needed a way to pass which page to ignore so there are a couple of lines of JS that were added. 

Mind giving this one more review?

Pomax (approved):
I'm still not sure why campaigns etc. need to support this page type, if it's going to be presented completely detached from that actual set of pages with no way to get to that campaign, but function-wise this works.

Issue #5489, opened 2020-10-21
Spike: Converting PNI Products to Wagtail Pages


KalobTaulien:
Explore what it will take to convert PNI Products to Wagtail Pages. 

A few things off the top of my head we'll need (exploration first, no coding required for a spike ticket)

- [ ] ProductPage (non abstract for tree/query relationships)
  * `updates` m2m field should be converted to an Orderable
  * `product_category` is an m2m field and should be converted to an Orderable 
  * `related_products` should be an Orderable with a ForeignKey to another ProductPage (which would allow SoftwareProduct and GeneralProduct page types to be selected). **Note:** We'll want to use `.specific` in the template for page-specific pages. 
  * When a ProductPage is deleted, we'll need a post_delete wagtail hook to check and delete cloudinary images
  * `image` is a FileField and should be converted to Wagtail Image with the field being a ForeignKey to the new image 
    * The template can then use `{% image page.image fill-XXXxXXX as img %}` 
- [ ] SoftwareProductPage. **Note: this can inherit from `ProductPage` and add additional fields** 
- [ ] GeneralProductPage. **Note: this can inherit from `ProductPage` and add additional fields**
- [ ] `Product.image` is a FileField. We'll need to convert that to a Wagtail Image and use a ForeignKey to that image in the "migration script". We'll also want to make sure we manage FileFields that are supposed to be images _as_ Wagtail Images, and files that are supposed to be documents _as_ Wagtail Documents (again with ForeignKey relationships)
- [x] ~`ExtendedYesNoField` May require some fine-tuning in a Wagtail Page. I suspect that won't be the case, but let's test this out during the spike.~ **Edit: Nothing is required for this as it works as-is in a WagtailPage**
- [x] ~`CloudinaryField` does not require any updating; works fine in a Wagtail Page~
- [ ] Buyersguide URLs suggest hard coded URL values based on a single "Buyersguide" view. We will use RouteablePageMixin and `@routes` for child pages. There should be a `max_count = 1` on this "BuyersGuidePage" with the following routes. 
  * Needs a default template for buyerguide_home 
  * Needs: about/, about/why/, about/meets-minimum-security-standards/, about/methodology/, about/contact/, about/press/
  * The two big ones will be `/categories/<slug>/` and `/products/<slug>/`. We can overwrite templates and context for each routeable view as well.
  * Let's also ensure these views get added to the page sitemap with `get_sitemaps()` 
[...]

PR #5529, opened 2020-10-26
2. Add Orderables to ProductPages (converted from m2m fields) 

### LINKED ISSUE #5508: 2. Add Orderables to Product Page Types
updates, product_category and related_products need to be Orderables. These require their own models with a ParentalKey to the current page, and a ForeignKey to the same page type. These are (iirc) all going to be migrating from ManyToMany fields to Orderables so we can maintain proper ordering

Criteria:
- [ ] `updates` needs an Orderable 
- [ ] `product_category` needs an Orderable 
- [ ] `related_products` needs an Orderable

KalobTaulien:
Closes #5508 
Related PRs/issues #5489 

Criteria:
- [x] `updates` needs an Orderable 
- [x] `product_category` needs an Orderable 
- [x] `related_products` needs an Orderable 

I missed `privacy_policy_links` in the Spike spec. Added it in this PR. 
- [x] `privacy_policy_links` needs an Orderable


network-api/networkapi/buyersguide/pagemodels/product_update.py
```diff
@@ -1,7 +1,11 @@
 from django.db import models
 
+from wagtail.snippets.models import register_snippet
+from wagtail.search import index
 
-class Update(models.Model):
+
+@register_snippet
+class Update(index.Indexed, models.Model):
```
Pomax, line 8: TIL: index

Pomax: Will this also make products show up when you do regular CMS search? (worried about getting 3 pages and 300 product results when staff searches =D)

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -9,6 +15,102 @@
...
+
+    panels = [
+        SnippetChooserPanel('category')
+    ]
+
+    def __str__(self):
+        return self.category.name
+
+    class Meta:
+        verbose_name = "Product Page Category"
```
Pomax, line 48: The "Page" part is usually not super great, especially as in this case it'd be more like a `ProductPageCategoryPage`

```suggestion
        verbose_name = "Product Category"
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -36,25 +138,19 @@ class ProductPage(FoundationMetadataPageMixin, Page):
     )
     review_date = models.DateField(
         help_text='Review date of this product',
+        null=True,
```
Pomax, line 141: this should only be `null=True` inside a migration, I assume follow-up PRs give this a meaningful value and then remove it again?

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -36,25 +138,19 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
-    #     related_name='pniproduct',
-    #     blank=True
-    # )
     blurb = models.TextField(
         max_length=5000,
         help_text='Description of the product',
         blank=True
     )
-    url = models.URLField(
+    product_url = models.URLField(
```
Pomax, line 153: as this will require template updating later, let's add a comment here that this rename requires template updating. Even if we get that done a few days from now.

network-api/networkapi/buyersguide/pagemodels/product_update.py
KalobTaulien: Yep! This makes it "searchable" in the modal when using a SnippetChooserPanel 🎉

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -89,131 +185,203 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
     )
-
     privacy_policy_helptext = models.TextField(  # REPURPOSED: WILL REQUIRE A 'clear' MIGRATION
         max_length=5000,
         blank=True
     )
 
-    """
-    privacy_policy_links = one to many, defined in PrivacyPolicyLink
-    """
```
Pomax, line 183: still a good idea to keep this comment, because reverse relational property assignment is black goddamn Django magic and anyone new to django/wagtail will be confused AF about why this property seems to exist, when it's not even defined for this class.

Pomax (changes requested):
few small notes but overall still R+ on the approach

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Good idea, thanks!

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Good call, this is now an Orderable though, so I'll update the comment accordingly.

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: You're right, this should mimic what we did with the `review_date` Product model. I'll remove this and update the migration file. Good catch!

PR #5547, opened 2020-10-29
Update page title for category pages and about pages

### LINKED ISSUE #5544: PNI - update page title
Currently, all PNI pages have `Mozilla - *privacy not included` as the page title. We want them to be more descriptive (e.g., including category name etc). This should improve our analytics tracking & SEO.

This blocks the analytics implementation ticket #5427 as we need tracking events to include unique page title as part of the data sent to GA.

mmmavis:
Closes #5544 

https://foundation-s-issue-5544-w0fqlm.herokuapp.com/en/privacynotincluded/

### DIFF
--- a/network-api/networkapi/buyersguide/views.py
+++ b/network-api/networkapi/buyersguide/views.py
@@ -98,6 +98,9 @@ def category_view(request, slug):
         'category': category,
         'products': products,
         'mediaUrl': MEDIA_URL,
+        'pageTitle': pgettext(
+            'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
+            '*privacy not included') + f' - {category}',
     })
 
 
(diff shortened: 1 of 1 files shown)


network-api/networkapi/buyersguide/views.py
```diff
@@ -98,6 +98,9 @@ def category_view(request, slug):
         'category': category,
         'products': products,
         'mediaUrl': MEDIA_URL,
+        'pageTitle': pgettext(
+            'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
+            '*privacy not included') + f' - {category}',
```
Pomax, line 103: I may me misremembering, but I recall `pgettext` only working with plain strings, so we may need to separate those two elements so we can update the template to use `{% block title%}{% trans "*privacy not included" %} - {{ page.name }}{% endblock %}` (where the category name will eventually be the right localized string once we finish the PNI wagtail uplift). At least, I think we had a block for this rather than setting `<title>` directly?

@TheoChevalier can you confirm? I think we ran into something similar a few weeks ago.

Pomax (changes requested):

TheoChevalier (approved):

Pomax (approved):

Issue #5463, opened 2020-10-19
PNI - only show new products on the homepage/category pages - Nov 5


kristinashu:
We only want new products (products added from Oct 29 onward) to appear on the homepage and category pages. The old products should still be live and discoverable via a google search or a direct url link.

kristinashu:
Note on timing:
- Oct 29: Jen will start with content entry so new product page work will need to be live (her new products will be in draft for a few days)
- Nov 4: soft launch of product pages (not Homepage) - this is when this ticket should be pushed live

Pomax:
This work can move ahead, but should not land until close to release.

kristinashu:
Right, this should be pushed live on Thurs Nov 5.

PR #5560, opened 2020-11-03
Hide pre-v4 products

Pomax:
⚠️ **THIS SHOULD ONLY BE MERGED AND PUSHED WHEN WE ARE READY TO GO LIVE** :warning:

Probably the smallest PR in this entire four sprint period...

Closes https://github.com/mozilla/foundation.mozilla.org/issues/5463 by setting the product filter cutoff from the old July 1st, 2019, to October 29th, 2020.

(note that this cutoff only affects which product end up shown in our listings on the homepage/category pages)


kristinashu (approved):
This is working well!

kristinashu:
Reminder that this will need to be pushed to prod on it's own, before the main homepage updates.

Issue #5426, opened 2020-10-09
PNI - define analytics for PNI homepage


mmmavis:
Define what events should be tracked.

taisdesouzalessa:
**Started a list with things to consider:**
- [ ] "submit a product" button tap
- [ ] ding checkbox
- [ ] search box (which terms people type there? Do people use the box?)
cc: @sabrinang @natalieworth - added those based on our[ Analytics doc](https://docs.google.com/document/d/1oWwHlyreicHybrXudDocclclWfef0m1hyUsob6FyL8w/edit?usp=sharing) - anything else missing you think it is important for us to measure?

sabrinang:
Yeah those are good things to measure. Only other thing on the homepage I can think of is "category navigation" and how often they are clicked (perhaps to show which ones are more popular etc)

taisdesouzalessa:
[List of events here](https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit?usp=sharing). Tab "Buyers Guide", lines 17-20.

@mmmavis I have some questions there for you and we can meet next week in case you have any questions.
@sabrinang after some thinking I decided not to add events to the nav items because if we want to know pageviews for each category we can see it by looking at URL traffic via GA, does it make sense?

taisdesouzalessa:
I think this is done in term of defining actions. Not closing this until I have the opportunity to discuss the items with Mavis, though.

taisdesouzalessa:
Talked to Mavis and we're aligned on those. @mmmavis can I close this ticket?

mmmavis:
Yup!

Closing it now!

PR #5582, opened 2020-11-05
Add more PNI GA events

### LINKED ISSUE #5427: PNI - implement analytics for PNI homepage
Implement what's been defined in #5426

mmmavis:
Closes #5427

Implemented GA events listed in https://github.com/mozilla/foundation.mozilla.org/issues/5426#issuecomment-715572072. Note that we've been using `og:title` as the page title sent to GA (same text for all pages, `*Privacy Not Included: A Buyer's Guide for Connected Products`) and this PR changes it to using `<title>` (`*privacy not included - [category name/product name/about]`).

https://foundation-s-issue-5427-rryft3.herokuapp.com/en/privacynotincluded/

### DIFF
--- a/network-api/networkapi/buyersguide/templates/about/contact.html
+++ b/network-api/networkapi/buyersguide/templates/about/contact.html
@@ -23,6 +23,6 @@
 
     <div class="recommend-product">
       <p>{% trans "We don’t want to miss any products that might be important to you. Tell us what connected products you want reviewed in our holiday buyer’s guide and we’ll do our best to include them." %}</p>
-      <a href="https://forms.gle/UefDgjEU85Xwcf8C8" target="_blank" class="btn btn-secondary btn-recommend">{% trans "Submit a product" %}</a>
+      {% include "../fragments/submit_a_product.html" %}
     </div>
 {% endblock %}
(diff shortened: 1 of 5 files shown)


Pomax (approved):
R+ if we're fine with repeated clicks generating repeated GA events (which I think we already do anyway?)

taisdesouzalessa (changes requested):
• For the "All" tap event (row 20 of [this spreadsheet](https://docs.google.com/spreadsheets/d/1x-CP_HYaAs3oRv2cjbSFu8w5Z938E4cyHQ3hvh17c6c/edit?usp=sharing)) it seems we are capturing what people are typing in the event label and I guess we don't want that, right? It seems too far on tracking...wonder what you think, Mavis.

image

• For the focus on search, I found 2 issues: 
1. if I type searching for 3 products it triggers 4 events (I think it is because it is focused 4 times, right? It is okay to keep as is if it is too difficult to track the exact amount of times people searched, just want to understand the limitations).
PNI-ga



2. If I click on the magnifying glass multiple times it triggers the event
PNI-ga-2

mmmavis:
@taisdesouzalessa thanks for the review.

> For the "All" tap event (row 20 of this spreadsheet) it seems we are capturing what people are typing in the event label and I guess we don't want that, right? It seems too far on tracking...wonder what you think, Mavis.

image

This works as expected. In the sheet we specified that we wanna track the search query. We can change label to something else, depending on if we care more about the "not found" search query OR whether or not people use the All link.

> focus event gets fired multiple times

I'll look into this more. Is the primary goal to track whether or not people use the search box?

mmmavis:
PR updated! Instead of tracking focus in search box, we tracks only on the **first keydown in search box** in the current session.

taisdesouzalessa (approved):
It is working for me. We are now tracking if users use the search box AT LEAST one time per session. Thanks, Mavis.

Pomax (approved):

Issue #5583, opened 2020-11-05
PNI: category dropdown on product pages


sabrinang:
- [x] show `All` in its unselected, grey text state.
- [x] Reduce white space gap beneath. 

<img width="1182" alt="image" src="https://user-images.githubusercontent.com/2492510/98291506-12f06d00-1f60-11eb-9a33-ed4863de0433.png">

PR #5584, opened 2020-11-05
Mobile product nav

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5583 by making homepage/category page/product page aware of what pagetype they are, so that we can generate the nav normally for homepage/category, but for product always select "all" as default nav selection

Review URL: https://foundation-s-mobile-pro-necxpb.herokuapp.com/privacynotincluded

### DIFF
--- a/source/sass/buyers-guide/views/product.scss
+++ b/source/sass/buyers-guide/views/product.scss
@@ -1,5 +1,11 @@
 $pni-product-breakpoint-larger: $bp-md;
 
+#view-product-page {
+  #multipage-nav-mobile {
+    margin-bottom: 0;
+  }
+}
+
 .product-header {
   padding-bottom: 88px;
   position: relative;
--- a/source/sass/buyers-guide/views/product.scss
+++ b/source/sass/buyers-guide/views/product.scss
@@ -3,6 +3,10 @@ $pni-product-breakpoint-larger: $bp-md;
 #view-product-page {
   #multipage-nav-mobile {
     margin-bottom: 0;
+
+    .multipage-link.active {
+      color: $gray-60;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


mmmavis (approved):

sabrinang (commented):
Thanks for fixing the space and having a category show in the dropdown on the product page. 

On the review URL it shows as a selected state (black), will it be grey on prod?

Pomax:
@sabrinang hm, I don't know if we can easily control that: it's showing black because we're "preselecting" the category by marking it `active`. I'm not sure we can easily recolor it or get it preselected without a bigger code change =/

/cc @mmmavis

Pomax:
updated the CSS so that it's gray-60 instead of black - just need heroku to finish rebuilding the review app.

Pomax:
PR updated to show "All" in the same gray/inactive state as the other categories - heroku will need to finish rebuilding the review app again though

source/sass/buyers-guide/views/product.scss
```diff
@@ -1,5 +1,21 @@
...
+
+  .multipage-link.active {
+    color: $gray-60;
+    border-bottom: none;
+
+    &:hover {
+      color: $black;
+      border-bottom: 3px solid $black;
+    }
+  }
```
mmmavis, line 16:
```suggestion
  .multipage-link.active {
    color: $gray-60;
    border-bottom: none;
  }
```

works fine to me without the &:hover override

mmmavis (changes requested):

source/sass/buyers-guide/views/product.scss
Pomax: are you sure? because without the hover, you don't get the black, underline highlight for "all" that you get for all the other categories when you mouse over.

mmmavis: turns out this is needed

mmmavis (approved):

sabrinang (approved):
👍 unselected states for desktop/mobile, thanks!

Issue #5587, opened 2020-11-06
PNI: category pages aren't ordered by creepiness


kristinashu:
I'm pretty sure category pages aren't showing products in order of not creepy to creepy like they should. This is apparent in the order the products appear and also when the creep-o-meter doesn't always appear as not creepy to creepy.

Screen cap from "All" were Nintendo is the least creepy.

<img width="1159" alt="Screen Shot 2020-11-05 at 9 22 26 PM" src="https://user-images.githubusercontent.com/1068783/98329600-6d63ea80-1fad-11eb-8816-813036f43b35.png">

Screen cap from "Toys & Games" were Nintendo showed be first but isn't.

<img width="1182" alt="Screen Shot 2020-11-05 at 9 22 00 PM" src="https://user-images.githubusercontent.com/1068783/98329594-69d06380-1fad-11eb-8099-7885e9c218cb.png">

Pomax:
Good catch: categories never sorted on creepiness before. Thankfully this only needs a tiny change to make it do the right thing.

PR #5590, opened 2020-11-06
sort category products on creepiness

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5587 by adding creepiness sorting to the category view, similar to the homepage.


kristinashu (approved):
It's harder to tell with the fake data but looking at a couple of products it seems to be working now!

Issue #5588, opened 2020-11-06
PNI: search text doesn't reset


kristinashu:
To reproduce
1. Type in a search word
2. Click on a product
3. Click browser back button
4. Click the search icon

What happens 
- The search box is pre-filled with the search word I previously typed in and the `x` icon in the search box is missing

What should happen
- The search box is empty

Screen cap showing the text and no `x`:
<img width="1160" alt="Screen Shot 2020-11-05 at 9 20 38 PM" src="https://user-images.githubusercontent.com/1068783/98329934-317d5500-1fae-11eb-8a8a-5843da1a8ecd.png">

Pomax:
looks like we just forgot a `value=""` to ensure browsers don't use previously cached values.

PR #5591, opened 2020-11-06
force an empty value on page load

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5588


kristinashu (approved):
Good!

PR #5597, opened 2020-11-07
Various search bar styling fixes

### LINKED ISSUE #5559: PNI - adjust search bar width
Adjust search bar's width so it aligns better with mobile category nav (refer to design on the right and ignore the one on the left)

For more context see See https://github.com/mozilla/foundation.mozilla.org/pull/5551#issuecomment-720788496

mmmavis:
Closes #5559

https://foundation-s-issue-5559-7unhuh.herokuapp.com/en/privacynotincluded/

### DIFF
--- a/network-api/networkapi/buyersguide/templates/catalog.html
+++ b/network-api/networkapi/buyersguide/templates/catalog.html
@@ -30,31 +30,37 @@
 
 {% block guts %}
 <div class="project-list-section">
-  <div id="sticky-bar" class="creepiness-slider bg-white text-center">
+  <div class="container">
+    <div class="row px-3 px-sm-0">
+      <div class="col-12">
+        <div id="sticky-bar" class="creepiness-slider bg-white text-center">
 
-    <!-- temporary location for dev work -->
-    <span id="product-filter-search" tabindex="0">
-      <span class="search-icon">&nbsp;</span>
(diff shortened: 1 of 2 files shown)


sabrinang (changes requested):
The margin below the checkbox reduces when it gets into tablet, it should be same as mobile so it doesn't shift:

pni_spacing

mmmavis:
I can't reproduce the issue and also wondering why the creepy face was missing in your gif @sabrinang  🤔

sabrinang:
I think I may have scrolled down a bit, made the face stick to the bottom, and revisited the search bar area so the face isn't centre there

Pomax:
I see that margin gap in Chrome in MacOS, it feels like it's a small/medium breakpoint mismatch.

Pomax:
yeah, turns out it was setting heights using a `max-width` cascade, rather than the `min-width` overrides that all our code relies on, so I refactored that and made it use the correct height for tablet.

sabrinang (changes requested):
The mobile and tablet spacing is consistent now however on desktop search bar margins have become tall and the spacing shifts when I click on/off of the search bar:
pni_search-bar-heights

sabrinang (approved):
Thanks the margins on desktop are consistent now

Issue #5608, opened 2020-11-09
PNI: badges on related products 


kristinashu:
Todo

- [x] Remove MSS badge
- [x] Add *privacynotincluded warning badge, when applicable

Screen cap from prod
<img width="679" alt="Screen Shot 2020-11-08 at 1 46 52 PM" src="https://user-images.githubusercontent.com/1068783/98571497-6d900e80-2269-11eb-830d-0745cf27b181.png">

cc @sabrinang in the future it might be worth exploring ways to make these thumbnails look more like the homepage

PR #5609, opened 2020-11-09
Remove related mss

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5608

Removes the MSS badge from the related products "card", and removes the MSS badge graphics files and seal_of_approval fragment.

image

review URL: https://foundation-s-remove-rel-tghojm.herokuapp.com/en/privacynotincluded/products/percy-cypress/


Pomax:
This will also require a follow-up investigation to see whether we can remove all the `seal-of-approval` mixin/class code in our SCSS now too, or whether we reused that for the privacy ding, without renaming it.

Pomax:
we probably need to keep the seal-of-approval file here, and only swap out the MSS for ding..

kristinashu (changes requested):
Seems to be working as expected but can you please reduce the icon size from 30px to 25px so it is the same size as on the homepage.

Pomax:
@kristinashu updated to 25px

kristinashu (approved):
Looks good, thank you!

PR #5539, opened 2020-10-28
Updated product_page.html template to leverage Wagtail fields and Orderables

KalobTaulien:
Closes #5510 
Related PRs/issues #5489 

## Acceptance Criteria
- [x] Update templates to use `{% image ... %}` template tags, where applicable 
- [x] Update ManyToMany loops in templates to leverage Orderable model traversing

I'll add comments to the things that seem drastically different below.

PR #5622, opened 2020-11-10
4. Updated product_page.html template to leverage Wagtail fields and Orderables

### LINKED ISSUE #5510: 4. Updating template components
> Updating template components such as `{% image .... fill-XXXxXXX as img %}` and looping through Orderables.

Because FileFields should now be ForeignKeys to the WagtailImage model (or custom image model) we'll need to update the template to use the `{% image %}` template tag. 

We'll also need to loop through Orderables in the template which may look a bit different from ManyToMany loops in the template. 

Criteria:
- [ ] Update templates to use `{% image ... %}` template tags, where applicable 
- [ ] Update ManyToMany loops in templates to leverage Orderable model traversing

KalobTaulien:
Closes #5510 
Supercedes #5539 (this is a smaller/clearner PR) 
Related PRs/issues #5489 

## Acceptance Criteria
- [x] Update templates to use `{% image ... %}` template tags, where applicable 
- [x] Update ManyToMany loops in templates to leverage Orderable model traversing

I'll add comments to the things that seem drastically different below.


network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -263,16 +264,16 @@ <h2 class="h3-heading">{% trans "Company contact info" %}</h3>
...
           <div class="mb-5">
-            {% for product in product.updates %}
+            {% for product in product.product_updates.all %}
               <div class="product-update mb-4">
-                <a class="product-update-link h5-heading title" href="{{product.source}}" target="_blank">{{product.title}}</a>
-                <div class="author">{{product.author}}</div>
-                <div class="snippet">{{product.snippet}}</div>
+                <a class="product-update-link h5-heading title" href="{{product.update.source}}" target="_blank">{{product.update.title}}</a>
+                <div class="author">{{product.update.author}}</div>
+                <div class="snippet">{{product.update.snippet}}</div>
```
Pomax, line 276: this doesn't feel great: updates are not products, so we should definitely not be saying `for product in ...` here. Ideally this is just `for update in updates`, but I don't think we can "unpack" the content of the many to many relation that way, so at the very least let's turn this into `for item in  updates` and then either accessing `item.update` everywhere, or using a `with update=item.update` like how you've done `related_product`.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -118,8 +120,12 @@ class ProductPage(FoundationMetadataPageMixin, Page):
     ProductPage is the superclass that SoftwareProductPage and
     GeneralProductPage inherit from. This should not be an abstract
     model as we need it to connect the two page types together.
+
+    ProductPage is re-using the product_page.html template that PNI Products use.
```
Pomax, line 124: Let's do this a little differently: in order to keep the old app working, we want to put and templates that we need to change in their own dir, and point to those, so that even if we mess this work up (let's hope we don't of course!) the old app won't break when we push the changes, and we don't have to lock prod pushing for as long as we need to fix whatever we got wrong.

Pomax (changes requested):

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -179,7 +180,7 @@ <h3 class="pni-product-smaller-body-bold mb-1 mb-md-2">{% trans "What data does
 
 
           {% trans "Links to privacy information" as privacy_policy_links %}
-          {% include "fragments/product_privacy.html"    with  label=privacy_policy_links          value=""         links=product.privacy_policy_links                  %}
+          {% include "fragments/product_privacy.html"    with  label=privacy_policy_links          value=""         links=product.product_privacy_policy_links.all                  %}
```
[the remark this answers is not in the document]
KalobTaulien: Yeah we can because we'll be using a new template. Ill incorporate this into the next commit.

KalobTaulien:
@Pomax I made those changes, thanks for pointing those out! 

Let's also pay no attention to the typo in this branch name 👀

network-api/networkapi/buyersguide/templatetags/bg_selector_tags.py
```diff
@@ -7,7 +7,7 @@
 @register.filter
 def yes_no(value):
     """Converts boolean to yes or no string"""
-    if value is None:
+    if value is None or value == '':
```
Pomax, line 10: In that case, we can probably just use `if not value:`, since both None and `''` are falsey, unlike `No` and `Yes`.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -390,17 +393,39 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
     ]
 
+    @property
+    def product_type(self):
+        if isinstance(self, SoftwareProductPage):
+            return "software"
+        elif isinstance(self, GeneralProductPage):
+            return "general"
+        else:
+            return "standard"
```
Pomax, line 403: let's make this `unknown` because there is no such thing a "not a general or software product" right now.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -390,17 +393,39 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+                'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
+                '*privacy not included') + f' - {self.title}'
+        return context
+
     class Meta:
         verbose_name = "Product Page"
 
 
 class SoftwareProductPage(ProductPage):
+    template = 'buyersguide/product_page.html'
```
Pomax, line 421: is the `template` value not inherited?

Pomax (changes requested):

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Seemingly no, it's not. Had to add this back into the child page classes.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -390,17 +393,39 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+            return "software"
+        elif isinstance(self, GeneralProductPage):
+            return "general"
+        else:
+            return "unknown"
+
+    def get_context(self, request, *args, **kwargs):
+        context = super().get_context(request, *args, **kwargs)
+        context['product'] = self
+        context['categories'] = BuyersGuideProductCategory.objects.all()
```
Pomax, line 408: we probably want to filter for categories with `hidden=False`

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -390,17 +393,39 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+        context['mediaUrl'] = settings.CLOUDINARY_URL if settings.USE_CLOUDINARY else settings.MEDIA_URL
+        context['coralTalkServerUrl'] = settings.CORAL_TALK_SERVER_URL
+        context['pageTitle'] = pgettext(
+                'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
+                '*privacy not included') + f' - {self.title}'
```
Pomax, line 413: Just a nit, if we're using an fstring anyway:

```suggestion
        context['pageTitle'] = f'{pgettext(
          'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
          '*privacy not included'
        )} - {self.title}'
```

Pomax (changes requested):
small request are categories + string massage

Pomax (approved):

PR #5631, opened 2020-11-12
10. Added categories and routable views to the sitemap

### LINKED ISSUE #5516: 10. BuyersGuide routes need sitemap entries
> Every BuyersguidePage.route needs a sitemap entry. Every category and product needs its own sitemap entries. 

Criteria:
- [ ] Using `get_sitemap()` make sure all the routes are added. 
- [ ] Make sure all categories are added (loop through them and add them to the list that `get_sitemap()` returns. 
- [ ] Don't worry about adding products to the sitemap, they'll be added when a new page is created

KalobTaulien:
Closes #5516 
Related PRs/issues #5489 

> Every BuyersguidePage.route needs a sitemap entry. Every category and product needs its own sitemap entries. 

Criteria:
- [x] Using `get_sitemap()` make sure all the routes are added. 
- [x] Make sure all categories are added (loop through them and add them to the list that `get_sitemap()` returns. 
- [x] Don't worry about adding products to the sitemap, they'll be added when a new page is created

This can be tested by creating a new BuyersGuidePage in the Wagtail admin. Then going to http://localhost:8000/sitemap.xml and searching for /about/ and /categories/

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/products.py
@@ -537,13 +537,14 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
         Add categories and route views to the sitemap for better search indexing.
         """
         sitemap = super().get_sitemap_urls(request)
+        last_modified = self.last_published_at or self.latest_revision_created_at
         # Add all the available Buyers Guide categories to the sitemap
         categories = BuyersGuideProductCategory.objects.filter(hidden=False)
         for category in categories:
             sitemap.append(
                 {
                     "location": self.full_url + self.reverse_subpage("category-view", args=(category.slug,)),
-                    "lastmod": (self.last_published_at or self.latest_revision_created_at),
+                    "lastmod": last_modified,
                 }
             )
 
@@ -557,7 +558,7 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
             sitemap.append(
                 {
                     "location": self.full_url + self.reverse_subpage(about_page),
-                    "lastmod": (self.last_published_at or self.latest_revision_created_at),
+                    "lastmod": last_modified,
                 }
             )
         return sitemap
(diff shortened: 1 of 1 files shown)


Pomax (changes requested):
I don't see any PNI entries in http://localhost:8000/sitemap.xml ?

KalobTaulien:
@Pomax You need to create a BuyersGuidePage first. The PNI entries aren't supposed to be there because they're being converted into Pages, and every Page gets its own sitemap entry automatically – so we don't need to add the PNI products into the sitemap manually. 

Good call on the Categories though. Might as well wait.

KalobTaulien:
So change of plans. No category uplift is needed. We just needed them registered as snippets and we can use an Orderable to connect the Page to the Category. 

The category "uplift" has been happening in a couple other branches too; changing them as the need comes up - ie: changing the links to routeable links in the template. 

This is good for another review @Pomax

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -533,9 +533,35 @@ def categories_page(self, request, slug):
...
+        for category in categories:
+            sitemap.append(
+                {
+                    "location": self.full_url + self.reverse_subpage("category-view", args=(category.slug,)),
+                    "lastmod": last_modified,
+                }
+            )
+
+        # Add all the available "subpages" (routes) to the sitemap, excluding categories and products.
+        # Categories are added above. Products will be their own Wagtail pages and get their own URLs because of that.
```
Pomax, line 552: This sounds like there's no plan to uplift the categories and homepage... which would be odd. Did we not plan for those?

Pomax (changes requested):

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: There's just no need to uplift the categories. They're already an existing Django model, and if we "convert" them they'd still be plain Django model but they'd have a `@register_snippet` decorator. Unless we wanted to add more to them or do some larger migration work, like moving which app the category model lives in (which would change the database table name). Thoughts?

Pomax: If we're uplifting to a wagtail app, I don't think it makes sense to uplift the homepage, and the product pages, but not the category page: categories are effectively subclasses of the homepage, with the same template but with an overwrite in the context to make sure it uses a reduced product set. So to me that feels like "these are pages, and children of the buyersguide homepage", with product pages pointing to these new category pages.

So not "move the category model" but "make a new page type for categories, and use that, instead"?

Pomax (approved):

PR #5637, opened 2020-11-13
Temp solution - make all pages on PNI show the generic og:image

mmmavis:
temp solution until we figure out how to make category page show its own og image set in CMS

https://foundation-s-pni-og-ima-fs13yo.herokuapp.com/en/privacynotincluded/categories/pets/


kristinashu (approved):
Looks good!

Issue #5513, opened 2020-10-26
7. Create Routable Views in BuyersGuidePage


KalobTaulien:
We need `@route`s to replace the `urls.py` paths in buyersguide. 

**Note:** Do not add the view logic. Only add the routes (some include slug parameters)

Criteria: 
- [ ] Create the following routes for BuyersguidePage using a `RoutablePageMixin`:
  - [ ] about/
  - [ ] about/why/
  - [ ] about/meets-minimum-security-standards/
  - [ ] about/methodology/
  - [ ] about/contact/
  - [ ] about/press/
  - [ ] categories/`<slug>`/
  - [ ] products/`<slug>`/

PR #5651, opened 2020-11-17
Port BuyersGuide urls to BuyersGuidePage routes

### LINKED ISSUE #5514: 8. Port over the buyersguide urls to `BuyersguidePage.route`'s.
> Move the logic from each of the url views to the Page `@route`s. I'm estimating 1 point per view just in case there's some extra unforeseen issues, but this ticket is likely going to take a lot less effort than an **8**

- [x] Move logic from buyersguide `urls.py` views to the `@route`s added in the last ticket #5513
  - [x] about/
  - [x] about/why/
  - [x] about/meets-minimum-security-standards/
  - [x] about/methodology/
  - [x] about/contact/
  - [x] about/press/
  - [x] categories/`<slug>`/
  - [x] products/`<slug>`/

KalobTaulien:
Closes #5514 
Related PRs/issues #5489 

## Criteria
- [x] Move logic from buyersguide `urls.py` views to the `@route`s added in the last ticket #5513
  - [x] about/
  - [x] about/why/
  - [x] about/meets-minimum-security-standards/
  - [x] about/methodology/
  - [x] about/contact/
  - [x] about/press/
  - [x] categories/`<slug>`/
  - [x] products/`<slug>`/


network-api/networkapi/buyersguide/templates/buyersguide/bg_base.html
```diff
@@ -0,0 +1,112 @@
+{% extends "pages/base.html" %}
+
+{% load bg_nav_tags localization i18n static wagtailroutablepage_tags %}
+
+{% get_current_language as lang_code %}
+
+{% block page_title %}{% if pageTitle %}{{ pageTitle }}{% else %}Mozilla - {% blocktrans context "This can be localized. This is a reference to the “*batteries not included” mention on toys." %}*privacy not included{% endblocktrans %}{% endif %}{% endblock %}
+
+
+{# TODO: consider using a different ga_identifier? #}
```
KalobTaulien, line 10: This was copied from the original template. It's not new to this PR.

network-api/networkapi/buyersguide/templates/buyersguide/bg_base.html
```diff
@@ -0,0 +1,112 @@
+{% extends "pages/base.html" %}
+
```
KalobTaulien, line 2: This is the new buyers guide base template. Its mostly a duplicate of the original. 

Reminder: We're doing this to namespace templates a bit better, and to not change anything thats in product so we can test things thoroughly before merging anything into master.

network-api/networkapi/buyersguide/templates/buyersguide/bg_base.html
```diff
@@ -0,0 +1,112 @@
...
+        <div class="col">
+          {% if pagetype == "product" or pagetype == "about" %}
+            <a class="multipage-link active" href="{{ page.url }}">{% trans "All" %}</a>
+          {% else %}
+            <a class="multipage-link {% if not category %} active{% endif %}" href="{{ page.url }}">{% trans "All" %}</a>
+          {% endif %}
+          {% for cat in categories %}
+            {% if cat.published_product_count > 0 %}
+              {% routablepageurl page 'category-view' cat.slug as cat_url %}
+              <a class="multipage-link {% check_active_category current_category cat %}{% if cat.featured is True %} featured{% endif %}{% if category == cat.slug %} active{% endif %}" href="{{ cat_url }}" data-name="{{ cat.name }}">{{ cat.name }}</a>
```
KalobTaulien, line 96: Instead of using an `include` here I opted to _not_ copy that file and use a for loop here. The `cat.published_product_count` came with the original template.

network-api/networkapi/buyersguide/templates/buyersguide/catalog.html
```diff
@@ -1,4 +1,4 @@
-{% extends "../bg_base.html" %}
+{% extends "buyersguide/bg_base.html" %}
```
KalobTaulien, line 1: Making it more explicit since there are two of these files now.

network-api/networkapi/buyersguide/templates/buyersguide/category_page.html
```diff
@@ -0,0 +1,8 @@
+{% extends "buyersguide/catalog.html" %}
+{% load static i18n %}
+
```
KalobTaulien, line 3: Copied category page for namespacing

network-api/networkapi/buyersguide/templates/buyersguide/fragments/pni_nav_links.html
```diff
@@ -0,0 +1,4 @@
+{% load bg_nav_tags i18n wagtailroutablepage_tags %}
+
+{% routablepageurl page 'about-why-view' as about_why_url %}
+{{ pre }}<a class="{{ class }} {% bg_active_nav request.get_full_path about_why_url %}" href="{{ about_why_url }}">{% trans "About the Guide" %}</a>{{ post }}
```
KalobTaulien, line 4: Use a routable page view link as to not link to the original /privacynotincluded/ urls in the Django app

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -30,6 +35,16 @@
     image_field = ImageChooserPanel('image')
     MEDIA_URL = settings.MEDIA_URL
 
+vote_throttle_class = UserVoteRateThrottle if not settings.TESTING else TestUserVoteRateThrottle
+
+
+class ProductPageVoteBin(models.Model):
+    bin_1 = models.IntegerField(default=0, help_text='Total votes for 0%-20%')
+    bin_2 = models.IntegerField(default=0, help_text='Total votes for 21%-40%')
+    bin_3 = models.IntegerField(default=0, help_text='Total votes for 41%-60%')
```
KalobTaulien, line 44: Simplified voting "bins". See help text for details.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+        on_delete=models.SET_NULL,
+        null=True,
+        blank=True,
+        related_name='vote_bin',
+    )
+
+    @property
+    def average_vote_count(self):
+        try:
+            total = self.current_votecount / self.current_total
```
KalobTaulien, line 327: @pomax I'm wondering if these two properties should be named better. I feel like they're a bit ambiguous at the moment.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+        except ZeroDivisionError:
+            total = 0
+        return total
+
+    @property
+    def get_voting_json(self):
```
KalobTaulien, line 333: This property re-creates what the frontend is looking for, minus the `confidence` and returns a string for JavaScript to them parse.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
         )} - {self.title}'''
         return context
 
+    def serve(self, request, *args, **kwargs):
+        # In Wagtail we use the serve() method to detect POST submissions.
+        # Alternatively, this could be a routable view.
+        if request.body and request.method == "POST":
```
KalobTaulien, line 486: In Wagtail when we want to accept a POST request on a page, we tend to put it in the `serve()` method and check if the request method is POST. 

There is no `def post()` method to overwrite.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+            # If the request is POST. Parse the body.
+            data = json.loads(request.body)
+            # If the POST body has a productID and value, it's someone voting on the product
+            if data['productID'] and data["value"]:
+                # Product ID and Value can both be zero. It's impossible to get a Page with ID of zero.
+                product_id = int(data['productID'])  # ie. 68
+                value = int(data["value"])  # ie. 0 to 100
```
KalobTaulien, line 493: Originally we were checking if `value` and `product_id` were instances of ints or bools. But we can cast these to ints to force numerics from here on out.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                try:
+                    product = ProductPage.objects.get(id=product_id)
+                    # If the product exists but isn't live and the user isn't logged in..
+                    if not product.live and not request.user.is_authenticated:
+                        raise Http404("Product does not exist")
+
+                    # Save the new voting totals
+                    # TODO: Confirm with @pomax this is the intended behaviour we desire.
+                    product.current_total = product.current_total + value
+                    product.current_votecount = product.current_votecount + 1
```
KalobTaulien, line 508: @Pomax I just want to confirm this is what you were thinking for counting votes and the total votecount.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+
+                    # Add the vote to the vote bin
+                    if not product.vote_bin:
+                        # If there is no vote bin attached to this product yet, create one now.
+                        vote_bin = ProductPageVoteBin()
+                        vote_bin.save()
```
KalobTaulien, line 514: Safety fallback in case, for some reason, there isn't a OneToOne relationship with a ProductPageVoteBin.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                except ProductPage.DoesNotExist:
+                    return HttpResponse('Missing page', status=400, content_type='text/plain')
+                except ValidationError as ex:
+                    return HttpResponse(f'Payload validation failed: {ex}', status=400, content_type='text/plain')
+                except Error as ex:
+                    print(f'{ex.message} ({type(ex)})')
+                    return HttpResponse('Internal Server Error', status=500, content_type='text/plain')
+        elif not self.vote_bin:
+            # Double check a voting bin exists. It should always exist.
+            # TODO: Test the Product-to-ProductPage migration to ensure vote_bin always exists.
```
KalobTaulien, line 545: This TODO will be taken care of in a future ticket when we test porting models to Pages. I doubt we'll run into any issues here, but I wanted to give us this reminder.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+            self.vote_bin = vote_bin
+            self.save()
+
+        return super().serve(request, *args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        # When a new ProductPage is created, ensure a vote bin always exists.
+        # We can use save() or a post-save Wagtail hook.
+        if not self.vote_bin:
+            vote_bin = ProductPageVoteBin()
```
KalobTaulien, line 558: `ProductPageVoteBin`'s wont be automagically created when a page is saved. We're making sure one _always_ exists when the page is saved. This ensures future ProductPages added will have the voting bin it needs and prevents the page from 500ing and throwing an issue in the template view

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -498,39 +632,16 @@ def categories_page(self, request, slug):
...
-            'categories': BuyersGuideProductCategory.objects.all(),
-            'category': category,
-            'products': products,
-            'mediaUrl': MEDIA_URL,
-            'pageTitle': pgettext(
+        products = ProductPage.objects.filter(product_categories__category__in=[category]).live()
+        products = sorted(products, key=lambda p: p.average_vote_count, reverse=True)
+
+        context = self.get_context(request)
+        context['category'] = category.slug
```
KalobTaulien, line 639: I'm using the `category.slug` to check which category is active in the template.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -71,7 +71,7 @@ export default class CreepVote extends React.Component {
 
   sendVoteFor(payload) {
     let attribute = payload.attribute;
-    let url = `/api/buyersguide/vote/`;
+    let url = window.location.href;
```
KalobTaulien, line 74: ProductPages, when being voted on, will POST to themselves instead of posting to an explicit endpoint.

KalobTaulien (commented):
@Pomax Ive added some clarification on things in this PR to hopefully help make sense of whats going on.

network-api/networkapi/wagtailpages/migrations/0016_merge_20201116_2028.py
```diff
@@ -0,0 +1,14 @@
...
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailpages', '0015_auto_20201110_1957'),
+        ('wagtailpages', '0015_buyersguidepage'),
+    ]
+
+    operations = [
+    ]
```
Pomax, line 14: Going out on a limb and suggest manually merging 0016-0018 into a single file.

network-api/networkapi/wagtailpages/pagemodels/products.py
Pomax: This feels like a job for int_list_validator, https://stackoverflow.com/questions/1429293/storing-an-integer-array-in-a-django-database/52497141

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
         blank=True,
     )
 
+    # Un-editable voting fields. Don't add these to the content_panels.
+    current_total = models.IntegerField(default=0)
```
Pomax, line 314: Hm, what is the benefit of having an explicit `current_total` field if that value is implicit to the sum of `vote_bins`?

network-api/networkapi/wagtailpages/pagemodels/products.py
Pomax: remember that we're not so much counting "votes" as "creepiness score", so:

- `def average_vote_count` -> `def creepiness`
- `total` -> `average`
- `current_votecount` -> `current_tally`
- `current_total` -> `current_vote_count`

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+                    '1': self.vote_bin.bin_4,
+                    '2': self.vote_bin.bin_3,
+                    '3': self.vote_bin.bin_2,
+                    '0': self.vote_bin.bin_1,
+                },
+                'average': self.average_vote_count
+            },
+            'total': self.current_votecount
+        }
+        return str(json)
```
Pomax, line 350: `str` does not yield json, and dicts don't serialize to json by default, so let's use the guaranteed correct function.
```suggestion
        return json.dumps(data)
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
...
+        except ZeroDivisionError:
+            total = 0
+        return total
+
+    @property
+    def get_voting_json(self):
+        """
+        Return a dictionary as a string with the relevant data needed for the frontend:
+        """
+        json = {
```
Pomax, line 337: naming nit: this isn't json, so calling it json is misleading.
```suggestion
        data = {
```

network-api/networkapi/wagtailpages/pagemodels/products.py
Pomax: Might be handy for future readers to link to the docs page on the `serve` function in this comment too,  as it's one of the far-less-used parts of wagtail.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
         return context
 
+    def serve(self, request, *args, **kwargs):
+        # In Wagtail we use the serve() method to detect POST submissions.
+        # Alternatively, this could be a routable view.
+        if request.body and request.method == "POST":
+            # If the request is POST. Parse the body.
+            data = json.loads(request.body)
+            # If the POST body has a productID and value, it's someone voting on the product
+            if data['productID'] and data["value"]:
```
Pomax, line 490: accessing a non-existent dict key with throw a `KeyError`. Use `get` instead:
```
if data.get('productID') and data.get("value"):
```
Although as we're then also using those values inside the conditional body, let's cache them:

```suggestion
            product_id = data.get('productId', False)
            value = data.get('value', False)
            
            if product_id and value:
                try:
                    product_id = int(product_id)
                    value = int(value)
                except ValueError:
                    # generate an HttpResponseNotAllowed response and exit the function
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                # Product ID and Value can both be zero. It's impossible to get a Page with ID of zero.
+                product_id = int(data['productID'])  # ie. 68
+                value = int(data["value"])  # ie. 0 to 100
+
+                if value < 0 or value > 100:
+                    # Malicious voting attempt
```
Pomax, line 496: occam's razor suggests the actual reason here is "the devs fucked up", let's not bake user prejudice into the code =)

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                    if value <= 20:
+                        product.vote_bin.bin_1 = product.vote_bin.bin_1 + 1
+                    elif value <= 40:
+                        product.vote_bin.bin_2 = product.vote_bin.bin_2 + 1
+                    elif value <= 60:
+                        product.vote_bin.bin_3 = product.vote_bin.bin_3 + 1
+                    elif value <= 80:
+                        product.vote_bin.bin_4 = product.vote_bin.bin_4 + 1
+                    elif value <= 100:
+                        product.vote_bin.bin_5 = product.vote_bin.bin_5 + 1
```
Pomax, line 527: this becomes a heck of a lot easier with an int array. That said, if we do stick to this, let's at least use `+= 1` to reduce the typo-induced bug surface by 50%

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+
+                    # Save the product
+                    product.vote_bin.save()
+                    # Don't save this as a revision with .save_revision() as to not spam the Audit log
+                    # And don't make this live with .publish(). The Page model will have the proper
+                    # data stored on it already, and the revision history won't be spammed by votes.
+                    product.save()
+                    return HttpResponse('Vote recorded', status=201, content_type='text/plain')
```
Pomax, line 535: let's keep this a 200 (e.g. don't specify the status code) as we're not "creating new records" that can be accessed in some dedicated fashion.

Also for future reference, note that a 201 response has [requirements beyond just the number](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/201), so the above response is a bad response.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                except ProductPage.DoesNotExist:
+                    return HttpResponse('Missing page', status=400, content_type='text/plain')
```
Pomax, line 537: There is a dedicated [HttpResponseBadRequest](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpResponseBadRequest) that we should use instead.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+            data = json.loads(request.body)
+            # If the POST body has a productID and value, it's someone voting on the product
+            if data['productID'] and data["value"]:
+                # Product ID and Value can both be zero. It's impossible to get a Page with ID of zero.
+                product_id = int(data['productID'])  # ie. 68
+                value = int(data["value"])  # ie. 0 to 100
+
+                if value < 0 or value > 100:
+                    # Malicious voting attempt
+                    return HttpResponse('Cannot save vote', status=405, content_type='text/plain')
```
Pomax, line 497: There is an dedicated [HttpResponseNotAllowed](https://docs.djangoproject.com/en/2.2/ref/request-response/#django.http.HttpResponseNotAllowed) that we should use instead.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                    product.vote_bin.save()
+                    # Don't save this as a revision with .save_revision() as to not spam the Audit log
+                    # And don't make this live with .publish(). The Page model will have the proper
+                    # data stored on it already, and the revision history won't be spammed by votes.
+                    product.save()
+                    return HttpResponse('Vote recorded', status=201, content_type='text/plain')
+                except ProductPage.DoesNotExist:
+                    return HttpResponse('Missing page', status=400, content_type='text/plain')
+                except ValidationError as ex:
+                    return HttpResponse(f'Payload validation failed: {ex}', status=400, content_type='text/plain')
```
Pomax, line 539: see above note

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +480,86 @@ def get_context(self, request, *args, **kwargs):
...
+                except Error as ex:
+                    print(f'{ex.message} ({type(ex)})')
+                    return HttpResponse('Internal Server Error', status=500, content_type='text/plain')
```
Pomax, line 542: There is a dedicated [HttpResponseServerError](https://docs.djangoproject.com/en/3.1/ref/request-response/#django.http.HttpResponseServerError) that we should use instead.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -30,6 +35,16 @@
     image_field = ImageChooserPanel('image')
     MEDIA_URL = settings.MEDIA_URL
 
+vote_throttle_class = UserVoteRateThrottle if not settings.TESTING else TestUserVoteRateThrottle
+
+
+class ProductPageVoteBin(models.Model):
```
Pomax, line 41: naming problem: this is not a vote bin, this is a collection of bins, so the class has to either be `ProductPageVoteBins` or `ProductPageVotes` (I'd go with the latter, because that allows future extensions without having a then-badly-named class).

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -295,6 +310,45 @@ class ProductPage(FoundationMetadataPageMixin, Page):
         blank=True,
     )
 
+    # Un-editable voting fields. Don't add these to the content_panels.
+    current_total = models.IntegerField(default=0)
+    current_votecount = models.IntegerField(default=0)
+    vote_bin = models.OneToOneField(
```
Pomax, line 316: This is not a single bin, let's go with "votes":
```suggestion
    votes = models.OneToOneField(
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -498,39 +632,16 @@ def categories_page(self, request, slug):
...
-
-        return render(request, 'category_page.html', {
-            'pagetype': 'category',
-            'categories': BuyersGuideProductCategory.objects.all(),
-            'category': category,
-            'products': products,
-            'mediaUrl': MEDIA_URL,
-            'pageTitle': pgettext(
+        products = ProductPage.objects.filter(product_categories__category__in=[category]).live()
+        products = sorted(products, key=lambda p: p.average_vote_count, reverse=True)
```
Pomax, line 636: We may want to make this its own function that can be imported as `from ProductPage import sort_average` or the like, rather than hardcoding it in a route. Also, why are we reversing? The sorting order on the homepage and category page is the same.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -543,27 +654,7 @@ def get_context(self, request, *args, **kwargs):
...
-        #     except AttributeError:
-        #         pass
-
-        #     return 50
-        # products = cache.get_or_set(
-        #     'sorted_product_dicts',
-        #     lambda: sorted([p.to_dict() for p in Product.objects.all()], key=get_average_creepiness),
-        #     86400
-        # )
+        products = sorted(products, key=lambda p: p.average_vote_count)
```
Pomax, line 657: are we reintroducing the `cache.get_or_set` call? because 10 queries vs. 650 queries is _kind of a big deal_ =D

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
Pomax: Surely we can post to `.` instead =)

Pomax (changes requested):

network-api/networkapi/wagtailpages/migrations/0016_merge_20201116_2028.py
KalobTaulien: Done!

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: One less database query per product. Though this is only for the ProductPage and not the Home/listing page, so one more db query would be OK imo. Ill update this with the new current_total being the sum of al 5 bins, once I change the bin_1, bin_2, etc to a CharField with `int_list_validator`

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Oh I was playing with reverse to make sure lambda was sorting the way I expected. It shouldnt be reversed. Glad you caught this. 

Right, Ill move this to its own function. 👍

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Yikes yes 650 is a big deal. Im sure this will be less when we use Pages but you're still right. Ill apply caching.

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Fair! I just copied this from the last view and didnt read through it well enough. Will fix

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
KalobTaulien: Feature flag added. Good catch Pomax!

network-api/networkapi/buyersguide/templates/buyersguide/bg_base.html
```diff
@@ -0,0 +1,112 @@
...
+{% endblock %}
+
+
+{% block wagtail_metadata %}
+  <meta property="og:description" content="{% trans "Smart home gadgets, fitness trackers, toys and more, rated for their privacy & security" %}">
+  <meta property="og:image" content="{% static "_images/buyers-guide/evergreen-social.jpg" %}">
+  <meta property="og:image:type" content="image/jpg">
+  <meta property="og:image:width" content="1200">
+  <meta property="og:image:height" content="628">
+  <meta property="og:url" content="{{request.scheme}}://{{request.get_host}}{{request.path}}">
```
Pomax, line 27: we'll have to remember to replace this with the wagtail metadata block instead - if we don't have an issue open for that yet, we should add that to the epic.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -71,7 +71,7 @@ export default class CreepVote extends React.Component {
 
   sendVoteFor(payload) {
     let attribute = payload.attribute;
-    let url = `/api/buyersguide/vote/`;
+    let url = document.getElementById("product-research").dataset['ispage'] ? '' : `/api/buyersguide/vote/`;
```
Pomax, line 74:
```suggestion
    // by default, we'll be posting to the PNI Django app posting route (see its views.py)
    let url = `/api/buyersguide/vote/`;

    // If we're running for the new wagtail version of PNI, voting is a post to the product's URL instead.
    if (document.getElementById("product-research").dataset.isWagtailPage) {
        url = ".";
    }
```

Pomax (changes requested):

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -30,6 +37,33 @@
...
+
+class ProductPageVotes(models.Model):
+    vote_bin = models.CharField(default="0,0,0,0,0", max_length=50, validators=[int_list_validator])
+
+    def set_votes(self, bin_list):
+        """
+        There are 5 "bins" for votes: <20%, <40%, <60%, <80%, <100%.
+        When setting votes, ensure there are only 5 bins (max)
+        """
+        bin_list = [str(x) for x in bin_list]
```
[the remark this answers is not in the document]
KalobTaulien: We get:
```
TypeError: sequence item 0: expected str instance, int found   
```
If we're joining integers and they aren't strings.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +503,87 @@ def get_context(self, request, *args, **kwargs):
...
+                    return HttpResponseNotAllowed(f'Payload validation failed: {ex}', content_type='text/plain')
+                except Error as ex:
+                    print(f'{ex.message} ({type(ex)})')
+                    return HttpResponseServerError('Internal Server Error', content_type='text/plain')
+        elif not self.votes:
+            # Double check a voting bin exists. It should always exist.
+            # TODO: Test the Product-to-ProductPage migration to ensure `votes` always exists.
+            # If all vote bins exist, we can safely remove this elif statement.
+            votes = ProductPageVotes()
+            votes.save()
```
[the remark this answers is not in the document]
KalobTaulien: Yeah we need it, this just creates the empty instance of `ProductPageVotes` but doesn't actually save it as a row in the database. I had thought the same thing too, but I guess if there's custom logic when creating a new object like this then someone wouldnt want it to auto-save to db.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +503,87 @@ def get_context(self, request, *args, **kwargs):
...
+        return super().serve(request, *args, **kwargs)
+
+    def save(self, *args, **kwargs):
+        # When a new ProductPage is created, ensure a vote bin always exists.
+        # We can use save() or a post-save Wagtail hook.
+        save = super().save(*args, **kwargs)
+        if not self.votes:
+            votes = ProductPageVotes()
+            votes.save()
+            self.votes = votes
```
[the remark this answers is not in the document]
KalobTaulien: No not really. `serve()` is for serving the previews and published pages, `save()` is just for saving the page. I _don't think_ that `save()` gets called when a page is previewed though. Im going to run a test on this to see if we can safely remove these lines (566-573)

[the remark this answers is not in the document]
KalobTaulien: Ok tested. Lines 566-573 are needed for previewing a page that hasnt been saved yet. In theory I thought this would be the case, in reality it's the case as well. 

But if the object is created when the page is viewed regarldess, we no longer need it in the `save()` method. I would prefer to have it in both areas for data consistency though in case we need to do a vote migration in the future and there are pages that havent been served that dont have `.votes` attached to them yet.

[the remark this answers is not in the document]
KalobTaulien: Made it a method so both serve and save can get_or_create the votes.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +503,87 @@ def get_context(self, request, *args, **kwargs):
...
+        # For more on this, see the docs here:
+        # https://docs.wagtail.io/en/stable/reference/pages/model_recipes.html#overriding-the-serve-method
+        if request.body and request.method == "POST":
+            # If the request is POST. Parse the body.
+            data = json.loads(request.body)
+            # If the POST body has a productID and value, it's someone voting on the product
+            if data.get('productID') and data.get("value"):
+                # Product ID and Value can both be zero. It's impossible to get a Page with ID of zero.
+                product_id = int(data['productID'])  # ie. 68
+                value = int(data["value"])  # ie. 0 to 100
```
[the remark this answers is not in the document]
KalobTaulien: Oh yes, good catch. Fixed!

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -406,6 +477,7 @@ def is_current(self):
         cutoff = datetime(2020, 10, 29)
         return cutoff < review
 
+    @property
```
Pomax, line 480: we can probably lean-i-fy this, see https://github.com/mozilla/foundation.mozilla.org/issues/5668

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -414,6 +486,18 @@ def product_type(self):
         else:
             return "unknown"
 
+    def get_or_create_votes(self):
```
Pomax, line 489: perfect name.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +510,73 @@ def get_context(self, request, *args, **kwargs):
...
+                    return HttpResponseNotAllowed('Product ID or value is invalid')
+
+                if value < 0 or value > 100:
+                    return HttpResponseNotAllowed('Cannot save vote')
+
+                try:
+                    product = ProductPage.objects.get(id=product_id)
+                    # If the product exists but isn't live and the user isn't logged in..
+                    if not product.live and not request.user.is_authenticated:
+                        return HttpResponseNotFound("Product does not exist")
```
Pomax, line 537: just realised: we forgot the `if product is None` case, because someone could script themselves a vote bot that just votes for products from 1 to 1 million, and that's going to cause a lot of ProductPageNotFound errors O_O

Pomax (changes requested):
one missing check, but there's a few comments on the /files page that were apparently missed (thanks as ever, github)

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -426,6 +503,87 @@ def get_context(self, request, *args, **kwargs):
...
+                        votes[4] = votes[4] + 1
+                    product.votes.set_votes(votes)
+
+                    # Don't save this as a revision with .save_revision() as to not spam the Audit log
+                    # And don't make this live with .publish(). The Page model will have the proper
+                    # data stored on it already, and the revision history won't be spammed by votes.
+                    product.save()
+                    return HttpResponse('Vote recorded', content_type='text/plain')
+                except ProductPage.DoesNotExist:
+                    return HttpResponse('Missing page', status=404, content_type='text/plain')
```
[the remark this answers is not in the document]
KalobTaulien: This _just_ appeared for me. 🤔

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: Yep, I agree. I assigned that ticket to me

network-api/networkapi/wagtailpages/pagemodels/products.py
KalobTaulien: I'm going to add this in for good measure. The line above it: `product = ProductPage.objects.get(id=product_id)` should throw an exception if the product isn't found by its product_id and caught on L561.

network-api/networkapi/wagtailpages/migrations/0016_product_voting.py
```diff
@@ -0,0 +1,38 @@
+# Generated by Django 2.2.16 on 2020-11-16 20:28
```
Pomax, line 1: note that this needs to be `0017_product_voting.py` (there's already a 16)

network-api/networkapi/wagtailpages/migrations/0016_product_voting.py
```diff
@@ -0,0 +1,38 @@
...
+import django.core.validators
+import django.db.models.deletion
+
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailpages', '0015_auto_20201110_1957'),
```
Pomax, line 12:
```suggestion
        ('wagtailpages', '0016_merge_20201117_1917.py'),
```

Pomax (changes requested):
so close O_O

Pomax (approved):

Issue #2996, opened 2019-04-10
make the directory-listing-filter able to load the profile card code "from the profile html"


Pomax:
right now we have a profile card html fragment for django templating purposes, and then we have the hardcoded template in the directory-listing-filter, and it is entirely possible for those two to go out of sync. It would be nice if we could somehow copy over the fragment to the static dir, and then have the directory-listing-filter code load that and put in the relevant values, rather than using its own hardcoded html string.

Pomax:
Pomax to investigate this. Wait, that's me. I will investigate.

Pomax:
Without turning this into an API call that can respond to a request for the template HTML with  the original `profile_block.html`, updated to a form that the `directory-listing-filters` code uses, there are basically no clean ways to make this work. Instead, I'll cross document their dependence and leave it at that.

PR #5708, opened 2020-11-23
cross-document the dependence between profile_block and directory-listing-filter

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/2996


mmmavis (commented):
Code looks good but I'm wondering if this meets the goal of https://github.com/mozilla/foundation.mozilla.org/issues/2996 ? In the ticket it looks like we are hoping to get `directory-listing-filters.js` to load the template, instead of building the HTML again in JS.

Pomax:
the conclusion is basically "this is impossible, without a lot of hackery", hence this being the only change I can reasonably put in.

mmmavis (approved):
🎉

Issue #4908, opened 2020-07-15
retire People app


Pomax:
We have no idea why it's still here. Let's remove it =)

Pomax:
added into the current milestone as it's a repeat of #5504 but for the people app.

PR #5723, opened 2020-11-26
remove the people models

Pomax:
Part 1 of 2 for https://github.com/mozilla/foundation.mozilla.org/issues/4908 (part 2: https://github.com/mozilla/foundation.mozilla.org/pull/5724)

Removes the model and adds a migration to effect that.


network-api/networkapi/people/urls.py
```diff
@@ -1,11 +1,2 @@
-from django.conf.urls import url
-
-from networkapi.people.views import (
-    PeopleListView,
-    PersonView,
-)
-
 urlpatterns = [
```
KalobTaulien, line 1: Is this file needed at all anymore if the urlpatterns are empty?

KalobTaulien (approved):
Looks good to me 👍  We could probably delete `network-api/networkapi/people/urls.py` since it's basically empty now. But that's non blocking for this PR.

network-api/networkapi/people/urls.py
Pomax: it is not, but part 2 removes it too

PR #5733, opened 2020-11-30
Top level mozfest homepage in review app?

### LINKED ISSUE #5674: Review apps and dev/stage/prod do not build the mozfest site in the same way
- dev/stage/prod create mozfest as a top level site, a sibling to the foundation homepage.
- Review apps create mozfest as a child of the main site homepage.

This causes problems when we need to test backend changes that affect mozfest, as we need to know whether things work with the mozfest page being a true site root. Most importantly, moving the mozfest page to be a top level page and then updating the `sites` setting to make the mozfest page the default site landing page causes the review app to crash.

We need to fix this so that review apps build the same tree structure as local dev/stage/prod, even if that means we still need to go in and manually "undefault" the `homepage` site, then mark the `mozfest` site as default

KalobTaulien:
Closes #5674 
Review app: https://foundation-s-5674-revie-4lxipk.herokuapp.com/

**Note**: Fun little review-app caveat:
When you change the Site settings to the Mozfest Homepage, it took a few minutes before this warning went away
image
As soon as this kicked in, it seemed to work. I'm not sure if this was caused by caching or a timed event on Herokus end, but this PR officially allows Mozfest Homepage to be beside the Foundation Homepage. And we can swap the Site root pages now.

### DIFF
--- a/network-api/networkapi/mozfest/factory.py
+++ b/network-api/networkapi/mozfest/factory.py
@@ -66,10 +66,7 @@ def generate(seed):
     except MozfestHomepage.DoesNotExist:
         print('Generating a Homepage')
         # If this is a review app, make the root of the mozfest site a child of the main site
-        if is_review_app:
-            site_root = WagtailPage.objects.get(title='Homepage')
-        else:
-            site_root = WagtailPage.objects.get(id=1)
+        site_root = WagtailPage.objects.get(depth=1)
 
         home_page = MozfestHomepageFactory.create(
             parent=site_root,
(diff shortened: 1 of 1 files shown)


network-api/networkapi/mozfest/factory.py
```diff
@@ -66,10 +66,7 @@ def generate(seed):
     except MozfestHomepage.DoesNotExist:
         print('Generating a Homepage')
         # If this is a review app, make the root of the mozfest site a child of the main site
```
Pomax, line 68: we'll probably want to remove this comment too, then =)

Pomax (changes requested):

network-api/networkapi/mozfest/factory.py
KalobTaulien: Good catch. Removed!

Pomax (approved):

PR #5751, opened 2020-12-02
omggggg

Pomax:
No issue, just FINALLY a fix for the python dotenv nonsense: this line _stripped double quotes_ from env.default when copying it over to an .env file, which legacy Docker required, and modern Docker holds the exact opposite opinion on: they _must_ be used.

So this line has to go~~~


KalobTaulien (approved):

mmmavis (approved):
I can't reproduce the issue this PR is trying to fix because I'm currently running older version of Docker and am not able to update it due to some weird tech issue.

From what you described the issue and solution, this code change makes sense to me and it also works on my version of Docker! R+

mmmavis:
Just successfully updated Docker to v2.5 but I'm seeing a bunch of ```Python-dotenv could not parse statement starting at line```. 🤔

mmmavis (changes requested):
Just successfully updated Docker to v2.5 but I'm seeing a bunch of ```Python-dotenv could not parse statement starting at line```. 🤔

mmmavis (approved):
Deleted my existing `.env` file and ran `inv new-env`. `Python-dotenv could not parse statement starting at line xx` no longer appears. R+

PR #5754, opened 2020-12-03
Exclude Dear Internet from l10n

### LINKED ISSUE #5753: Exclude Dear Internet template from localization
Since we’re not translating this page, let’s exclude the template string from being extracted

TheoChevalier:
Closes #5753


Pomax:
@TheoChevalier does makemessages support a file for the exclusions, like `--conf=makemessage.yml` or the like?

TheoChevalier:
> @TheoChevalier does makemessages support a file for the exclusions, like `--conf=makemessage.yml` or the like?

It only supports glob-style patterns, we would have to add support for that in a [custom command](https://docs.djangoproject.com/en/3.1/topics/i18n/translation/#customizing-the-makemessages-command)

Pomax (approved):

Pomax:
ah, definitely more work than is worth updating this issue with =)

PR #5796, opened 2020-12-08
Custom image rendition template tag for custom image heights via stream block

### LINKED ISSUE #5785: Add short option to wide image streamblock
> Wide Image streamfield
> could we have the height be the height of the image (possibility of having more narrow wide images as it stretches out narrow images and become pixelated)?

Image is currently set to `fill-1400x410` but the height should be adjustable. 

Target area:

Criteria:
- [ ] Create a custom height field that lets the Wide Image Block's image be a custom height
- [ ] Set the default height to 410
- [ ] May require a template tag that creates a new image rendition with a custom height

KalobTaulien:
Closes #5785 

## Criteria
- [x] Create a custom height field that lets the Wide Image Block's image be a custom height
- [x] Set the default height to 410
- [x] May require a template tag that creates a new image rendition with a custom height

**Note**: Stream block changes always result in a migration even though they aren't _necessary_ because it's just a giant textfield in the database structured as JSON.


network-api/networkapi/wagtailpages/migrations/0019_auto_20201208_2004.py
```diff
@@ -0,0 +1,23 @@
...
+        ),
```
KalobTaulien, line 22: Stream blocks always result in a migration even though it's not technically required. There's a way around this if we want to get a bit hacky, otherwise, we're stuck with this.

network-api/networkapi/wagtailpages/pagemodels/customblocks/articles.py
```diff
@@ -32,6 +32,13 @@ class Meta:
 class ArticleFullWidthImageBlock(blocks.StructBlock):
 
     image = ImageChooserBlock()
+    image_height = blocks.IntegerBlock(
+        default=410,
+        help_text='A custom height for this image. The image will be 1400px wide '
+                  'by this height. Note: This may cause images to look pixelated. '
```
KalobTaulien, line 38: `image_height` is set to 410 by default, which was originally hard coded as `flll-1400x410`. We're keeping the 1400px but letting editors change the height. 

This notion enables the editors to set a height for an image they can't crop or edit on their own, and can reduce blurriness when certain image types are uploaded.

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
```diff
@@ -8,3 +8,16 @@ def optimize_images(image):
     return {
         'unoptimized_image': image
     }
+
+
+@register.simple_tag(name='custom_image_height')
+def custom_image_height(image, height=410):
```
KalobTaulien, line 14: @Pomax Not 100% certain this image tag _should_ live in the `multi_image_tag.py` file. I'm happy to move this if you think there's a better place for it.

network-api/networkapi/wagtailpages/migrations/0019_auto_20201208_2004.py
Pomax: what's one more.

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
Pomax: it might actually make more sense to make this just rename this file to `custom_image_tags.py` instead and update settings.py, if it's going to have various image related tags.

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
```diff
@@ -8,3 +8,16 @@ def optimize_images(image):
...
+    """
+    Automatically fill an image with 1400 by X pixels where X is a variable integer.
+
+    If a height is not set (False or None) default to a 410px height as per the
+    original design.
+    """
+    if not height:
+        height = 410
```
Pomax, line 22: this is redundant: if `height` is not passed, it will already be `410` because of the argument signature.
```suggestion
```

Pomax (changes requested):

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
KalobTaulien: Smart. I'll do that and replace its `{% load ... %}` values in the templates when I circle back to this PR. Thanks!

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
KalobTaulien: Yup, because I iterated on it and didnt change it haha

network-api/networkapi/wagtailpages/templatetags/multi_image_tags.py
KalobTaulien: Changed! Marking as resolved

network-api/networkapi/wagtailpages/migrations/0019_auto_20201208_2004.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Django 2.2.16 on 2020-12-08 20:04
```
Pomax, line 1: This will need to become 0022

Pomax (changes requested):
R+ but we'll have to renumber the migration and update its dependency

network-api/networkapi/wagtailpages/migrations/0022_auto_20201208_2004.py
```diff
@@ -0,0 +1,23 @@
...
+import wagtail.contrib.table_block.blocks
+import wagtail.core.blocks
+import wagtail.core.fields
+import wagtail.images.blocks
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('wagtailpages', '0018_auto_20201126_2017'),
```
Pomax, line 14:
```suggestion
        ('wagtailpages', '0021_add_missing_fields'),
```

Pomax (approved):

PR #5838, opened 2020-12-10
Use localized product urls

### LINKED ISSUE #5837: Fix PNI homepage localization urls
Going to http://localhost:8000/fr/privacynotincluded-new/ and clicking a product would go to a redirect view, and isnt accepting a locale so it 404's when using a different locale. 

Instead, let's use the product urls and leverage the localization we have, 

## Criteria
- [ ] Remove routeable loaded module from template 
- [ ] Use `{{ product.url }}` in template instead of a routable urls.

KalobTaulien:
Closes #5837

## Criteria
- [x] Remove routeable loaded moduel from template 
- [x] Use `{{ product.url }}` in template instead of a routable urls.


Pomax (approved):
nice nice

PR #5850, opened 2020-12-10
cache improvement

Pomax:
This changes the caching logic from "do all the work, then cache it" to "check the cache first, if it's `None`, do the work, otherwise don't do any work because we're already done".


benhohner (approved):

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -959,22 +960,28 @@ def product_view(self, request, slug):
 
     @route(r'^categories/(?P<slug>[\w\W]+)/', name='category-view')
     def categories_page(self, request, slug):
+        context = self.get_context(request, bypass_products=True)
```
Pomax, line 963: sneaky

benhohner (approved):
Looks good. Let's make a note to come back to this after the PNI launch

PR #5883, opened 2020-12-14
Second flip attempt

Pomax:
This moves the PNI routes to after the Wagtail routes, so that Wagtail wins, so we can flip the switch on prod


KalobTaulien (approved):
Good use of `@skip` with a TODO

PR #5902, opened 2020-12-15
Crank up the default image quality

### LINKED ISSUE #5898: Wagtail: Improve Image Quality
Images have good resolution but when uploaded they are a bit blurry. See example below: [theory of change graphic](https://foundation.mozilla.org/en/insights/trustworthy-ai-whitepaper/path-forward/introduction-theory-change/)
image

Kalob mentioned in slack:
> We could improve this too by using a `<picture>` with multiple image sources, and providing a 2x image for retina screens.
> We should also add these settings to the site:
`WAGTAILIMAGES_JPEG_QUALITY = 100`
`WAGTAILIMAGES_WEBP_QUALITY = 100`

KalobTaulien:
Closes #5898 

Default image quality for JPEG and WEBP is 85%. This PR cranks it up to 100%. We may want to keep track of how large the image assets get after deploying this just to make sure we aren't serving MASSIVE images for the sake of 100% quality. 

The issue is referring to a richtext image which is set to `width-800` by default, and afaik we can't change that. But we do have control over the image quality. 

> Ahh so this is in the content RichText stream block. I thought (for some reason) this was it's own stream block.
> So here we have very little control. It's set a width of 800 by default in the richtext area...

Docs for confirmation on this can be found here: https://docs.wagtail.io/en/stable/topics/images.html#images-embedded-in-rich-text

This PR simply sets the wagtail image quality which should help solve this problem, but not 100% fix it.


KalobTaulien:
Oh there's also this we could look into:

https://docs.wagtail.io/en/stable/advanced_topics/customisation/page_editing_interface.html#image-formats-in-the-rich-text-editor

Pomax:
I'm worried this will blow up our page size, so let's do some real testing on this before we decide whether to move forward with this. One thing that we should also look at is whether the right image encoding is being used for content - the AI ToC image, for example, should not be using JPG encoding, but PNG encoding.

KalobTaulien:
I did a little testing on localhost with (not many) real images and didn't see any benefit of using settings cranked up to 100% quality. 

I think there's a curve we should look at when measuring quality vs. file size. I feel like 85% is a good balance. And these settings only affect JPEG and WEBP images anyway. 

But when I use 100% settings for a JPEG locally, compared to a PNG I created, there's no difference between 85% and 100%. 

Top is a JPEG at 100% quality. Bottom is the PNG encoding of the same image. 
image

Filesizes when rendered in the template:
The JPEG at 100% was 124kb. 
The PNG was 173kb. 

Looking at the data, it might be best to use 100% JPEG quality as it's still a lower filesize than a it's PNG counter part (this obviously depends on the image, too). 

We could look into image optimization tools too if we want to drastically reduce filesizes while outputting the highest quality image; something like https://tinyjpg.com/ as a 3rd party service. 

As for actually using 100% instead of 85%, I personally can't the difference with this sample image on a retina display. 

Where do we go from here? I guess my recommendation would be to ditch this PR and teach the editors that richtext images should be a width of 800px and ideally hold less text with larger font. **OR** We could create them a custom stream block specifically for images that look like they fit in the richtext area - where we could have more control over it's sizing. 

CC @Pomax

Pomax:
If the PNG is larger but _actually legible_ then I don't see a benefit in cranking up JPG to 100% for no visual difference. I'd recommend we close this PR and instead focus on making sure we're using the right extensions for the type of image data we're trying to show.

/cc @sabrinang

sabrinang:
@Pomax Exported PNGs here: https://drive.google.com/open?id=1GFJPib-pBzPBedPhsKde3SAMDW7Tp_bw&authuser=sabrina%40mozillafoundation.org&usp=drive_fs

For reference, these are the original sources these are exported from:
- [google slides](https://docs.google.com/open?id=15d_ufLf4o5GuqZep4XIerBSCpVssIywSfD6oEzlnNSU&authuser=sabrina%40mozillafoundation.org&usp=drive_fs)
- [google drawing](https://docs.google.com/open?id=1uU-SRPw6_FYiA_ILQigdKfX8l4DPVEalLjtm4dA1Ggo&authuser=sabrina%40mozillafoundation.org&usp=drive_fs)

Let me know if I should remake/export in illustrator if these PNGs don't work for what we need.

Pomax:
Thanks! Running that landscape ToC graphic through tinypng, it ends up 150kb, so that's drastically smaller than the JPG version at the same size. @KalobTaulien I think the solution here is going to be education rather than code.

PR #5948, opened 2020-12-21
Remove encryption help text from MSS field

### LINKED ISSUE #5943: PNI: remove duplicate field
Currently on production the Encryption help text is being duplicated and is showing as MSS help text as well (MSS shouldn't have any helt text).

- [ ] MSS should not show help text

Screen cap from prod:
image

Pomax:
Closes #5943

Makes the field help text for the MSS field an empty string (instead of being the entirely wrong encryption help text =)


mmmavis (approved):
🎉

Issue #6000, opened 2021-01-11
Update the cypress + related tool versions


Pomax:
Cypress is a bit behind the times by now, let's bump up its version + related tools so that CI runs something more recent.

PR #6001, opened 2021-01-11
Cypress bump

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6000

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "cypress:a11y": "run-p --race cypress:docker cypress:test:a11y",
     "cypress:ci": "run-p --race server cypress:test",
     "cypress:docker": "docker-compose up",
-    "cypress:install": "npm install --no-save cypress@^3.7.0 @percy/cypress@^2.0.0 cypress-axe@^0.5.3 @testing-library/cypress@^5.0.2",
+    "cypress:install": "npm install --no-save cypress@5.6.0 @percy/cypress@v2.3.3 cypress-axe@0.12.0 @testing-library/cypress@7.0.3",
     "cypress:clean": "shx rm -rf ./cypress/screenshots",
     "cypress:test": "npm run cypress:clean && wait-on http://localhost:8000/cms && cypress run --spec ./cypress/integration/endpoint-tests.js",
     "cypress:test:a11y": "npm run cypress:clean && wait-on tcp:8000 && cypress run --spec ./cypress/integration/a11y-tests.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -87,9 +87,6 @@
     "webpack-cli": "^4.3.1",
     "whatwg-fetch": "^3.5.0"
   },
-  "engines": {
-    "node": "14.4.*"
```
Pomax, line 91: what a jerk, good riddance.

package.json
```diff
@@ -72,7 +72,7 @@
     "event-stream": "3.3.4",
     "js-cookie": "2.2.1",
     "moment": "^2.29.1",
-    "node-sass": "^4.14.1",
+    "node-sass": "^5.0.0",
```
Pomax, line 75:
```suggestion
    "node-sass": "^4.14.1",
```

package.json
mmmavis: What version of Node will it uses after removing this part from the package.json? Will it always be the latest?

Pomax: hm, good point. Whatever it's set to actually install, so for docker that's currently 14.13.1, for github actions it's "latest version of 14", but heroku prefers it's prespecified so I guess I'll set this to `>=14.4` so that v15 doesn't break things.

package.json
```diff
@@ -17,9 +20,9 @@
     "cypress:a11y": "run-p --race cypress:docker cypress:test:a11y",
     "cypress:ci": "run-p --race server cypress:test",
     "cypress:docker": "docker-compose up",
-    "cypress:install": "npm install --no-save cypress@^3.7.0 @percy/cypress@^2.0.0 cypress-axe@^0.5.3 @testing-library/cypress@^5.0.2",
+    "cypress:install": "npm install --no-save cypress@5.6.0 @percy/cypress@v2.3.3 cypress-axe@0.12.0 @testing-library/cypress@7.0.3",
     "cypress:clean": "shx rm -rf ./cypress/screenshots",
-    "cypress:test": "npm run cypress:clean && wait-on http://localhost:8000/cms && cypress run --spec ./cypress/integration/endpoint-tests.js",
+    "cypress:test": "npm run cypress:clean && wait-on http://localhost:8000/cms && cypress run --spec ./cypress/integration/endpoint-tests.js && docker-compose down",
```
Pomax, line 25:
```suggestion
    "cypress:test": "npm run cypress:clean && wait-on http://localhost:8000/cms && cypress run --spec ./cypress/integration/endpoint-tests.js",
```

PR #6007, opened 2021-01-12
Replaced 'node-sass' with 'sass' & remove old SCSS code

### LINKED ISSUE #6006: Switch from "node-sass" to "sass"
...because LibSass has been deprecated and the [migration guide](https://sass-lang.com/blog/libsass-is-deprecated#how-do-i-migrate) says

> If you’re a user of Node Sass, migrating to Dart Sass is straightforward: just replace node-sass in your package.json file with sass. Both packages expose the same JavaScript API.

mmmavis:
Closes #6006


source/js/buyers-guide/components/creep-vote/creep-vote.scss
```diff
@@ -14,23 +14,6 @@ $btn-shadow-width: 3px;
...
-      &.btn-secondary:hover {
-        cursor: auto;
-        color: inherit;
-        background: inherit;
-      }
-    }
-  }
-}
-
 .creep-vote {
```
mmmavis, line 17: Was gonna update line 21 `@extend .btn.disabled;` to `@extend .btn, .disabled;` because
Screen Shot 2021-01-11 at 5 21 48 PM
but I realized that that chunk of code is old and should be removed. I think we had it when the form had two required fields.

Pomax: that sounds fine

Pomax (approved):

PR #6009, opened 2021-01-12
Don't create a mobile menu if there are no links to use

### LINKED ISSUE #5243: Empty Menu on mobile
### Describe the bug
I created a page on the new "Donors" hub and since I don't want it to show up in menus I unchecked the item "Show in menus: Whether a link to this page will appear in automatically generated menus". However when I see this page on mobile, I see an empty menu field, like this page is part of a menu item.

### To Reproduce
1. Go to [this page](https://foundation.mozilla.org/en/giving/mozilla-guardian-giving-program/)
2. See it in "Responsive Design Mode: Mobile"

### Expected behavior
Not have an empty mobile menu (the same way we don't have an empty desktop menu)

### Screenshots
image

KalobTaulien:
Closes #5243 

**Link to sample test page**: https://foundation-s-5243-empty-ykwlbz.herokuapp.com/en/banner-page-parent/banner-page-child-1/

Simple fix here. The JavaScript originally created was looking for links in the desktop version of a Banner Campaign Page (Banner Page), and generating links for the mobile dropdown. 

If there are no links to use, return early and hide the menu.


source/js/foundation/inject-react/multipage-nav.js
```diff
@@ -12,7 +12,15 @@ export default (apps) => {
   );
 
   if (targetNode) {
-    let links = Array.from(document.querySelectorAll(`#multipage-nav a`)).map(
+    const multipageLinks = document.querySelectorAll(`#multipage-nav a`)
```
mmmavis, line 15: Good call on checking link length! 

Since we already have an `if` statement on line 14, we can combine your code with the existing line 10-14 to something like

```jsx
const targetNode = document.querySelector(
    `#multipage-nav-mobile .container .row .col-12`
  );
  const multipageLinks = document.querySelectorAll(`#multipage-nav a`);

  if (targetNode && multipageLinks.length > 0) {
```

And leave the rest of code unchanged.

mmmavis (changes requested):

source/js/foundation/inject-react/multipage-nav.js
KalobTaulien: Ahh nice, good feedback. Will implement and test shortly. Thanks Mavis!

mmmavis (approved):
Lovely!

PR #6010, opened 2021-01-12
Make ESLint cover wagtailcustomization js files

mmmavis:


Pomax (approved):
We can shorten this to `\"network-api/networkapi/wagtailcustomization/**/*.js\"` but R+ either way

Pomax (approved):

PR #6011, opened 2021-01-12
Get total product category count from ProductPage

### LINKED ISSUE #5886: PNI follow-up: make categories aware of new products
category functionality got rewritten, but the new code and templates still tap into some of the old logic. Specifically, [published_product_count()](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/pagemodels/product_category.py#L51-L53) checks for how many _original_ products are classified into this category, which determines whether or not a category shows up in the nav bar.

In order to clean this part up, we _probably_ want to switch to a new category class instead, house in `wagtailpages`, so we can start moving away from needing code from the `buyersguide` app.

We can have a look at https://stackoverflow.com/a/26472482 and see if that works with wagtail mixed in

KalobTaulien:
Closes #5886 

**Link to sample test page**: Coming soon

Get total live ProductPages from the selected category.


network-api/networkapi/buyersguide/pagemodels/product_category.py
```diff
@@ -50,8 +49,8 @@ class BuyersGuideProductCategory(models.Model):
 
     @property
     def published_product_count(self):
-        # TODO: REMOVE: LEGACY FUNCTION
-        return Product.objects.filter(product_category=self, draft=False).count()
+        from networkapi.wagtailpages.models import ProductPage
+        return ProductPage.objects.filter(product_categories__category=self).live().count()
```
Pomax, line 53: this just duplicates what we already have below, if we're doing that, let's start by making sure that `published_product_count` doesn't get called at all, replaced with `published_product_page_count` instead.

Pomax (changes requested):

network-api/networkapi/buyersguide/pagemodels/product_category.py
KalobTaulien: Solid point and I couldnt agree more. Will work on this shortly.

network-api/networkapi/buyersguide/templates/category_nav_links.html
```diff
@@ -2,7 +2,7 @@
 
 {% for cat in sorted_categories %}
 {% if cat.hidden == False %}
-{% if cat.published_product_count > 0 %}
+{% if cat.published_product_page_count > 0 %}
```
Pomax, line 5: looking at this, the better solution here is to create a copy of this file as `network-api/networkapi/buyersguide/templates/buyersguid/category_nav_links.html ` (so, with that new app namespace) and then pointing to the new file rather than the original.

KalobTaulien: Agreed. Done!

network-api/networkapi/buyersguide/templates/category_nav_links.html
```diff
@@ -1,8 +1,13 @@
 {% load bg_nav_tags %}
 
+{% comment %}
+  This is the new namespaced category_nav_links.html
+  This file is pending deletion when we delete the buyersguide app.
```
Pomax, line 5: this is the original file =)
```suggestion
  This file is pending deletion when we delete the buyersguide app.
```

network-api/networkapi/buyersguide/templates/category_nav_links.html
```diff
@@ -1,8 +1,13 @@
...
+{% endcomment %}
+
 {% for cat in sorted_categories %}
 {% if cat.hidden == False %}
-{% if cat.published_product_count > 0 %}
+{% if cat.published_product_page_count > 0 %}
```
Pomax, line 10:
```suggestion
{% if cat.published_product_count > 0 %}
```

network-api/networkapi/buyersguide/pagemodels/product_category.py
```diff
@@ -48,16 +47,11 @@ class BuyersGuideProductCategory(models.Model):
         blank=True,
     )
 
-    @property
-    def published_product_count(self):
-        # TODO: REMOVE: LEGACY FUNCTION
-        return Product.objects.filter(product_category=self, draft=False).count()
-
```
Pomax, line 55: let's put this back in for now

```suggestion
    @property
    def published_product_count(self):
        # TODO: REMOVE: LEGACY FUNCTION
        return Product.objects.filter(product_category=self, draft=False).count()

```

network-api/networkapi/buyersguide/pagemodels/product_category.py
```diff
@@ -3,7 +3,6 @@
 
 from wagtail.snippets.models import register_snippet
 
-from .products.base import Product
```
Pomax, line 6: let's put this back in for now
```suggestion
from .products.base import Product
```

network-api/networkapi/buyersguide/factory.py
```diff
@@ -44,7 +44,7 @@ def get_lowest_content_category():
     # TODO: REMOVE: LEGACY FUNCTION
     return sorted(
         [
-            (cat.published_product_count, cat)
+            (cat.published_product_page_count, cat)
```
Pomax, line 47: This can stay the old code, since `get_lowest_content_page_category` exists already
```suggestion
            (cat.published_product_count, cat)
```

Pomax (changes requested):

network-api/networkapi/buyersguide/templatetags/bg_nav_tags.py
```diff
@@ -17,7 +17,7 @@ def bg_active_nav(current, target):
 
 
 # Instantiate a list of category page links based on the current page's relation to them
-@register.inclusion_tag('category_nav_links.html', takes_context=True)
+@register.inclusion_tag('buyersguide/fragments/category_nav_links.html', takes_context=True)
```
Pomax, line 20:
```suggestion
# NOTE: this points to the new, namespaced category_nav_links. If we need to revert to the old app, change this back.
@register.inclusion_tag('buyersguide/fragments/category_nav_links.html', takes_context=True)
```

PR #6021, opened 2021-01-13
Update a deprecated stylelint rule name

mmmavis:
Fixed this warning that I just caught
```bash
Deprecation Warning: 'declaration-property-value-blacklist' has been deprecated. Instead use 'declaration-property-value-disallowed-list'. See: https://github.com/stylelint/stylelint/blob/13.7.0/lib/rules/declaration-property-value-blacklist/README.md
```


Pomax (approved):
mmm political correctness

PR #6052, opened 2021-01-19
Allow custom ToC thumbnail images

### LINKED ISSUE #5847: Consider allowing editors to upload chapter image shown on Publication's ToC
### Describe the feature request
Consider allowing editors to upload custom chapter image shown on Publication's ToC

### Additional context
#5822

/cc @sabrinang

mmmavis:
Closes #5847 

Sample test link: https://foundation-s-issue-5847-sn8tcu.herokuapp.com/en/publication-page-with-chapter-pages/
Sample edit page: https://foundation-s-issue-5847-sn8tcu.herokuapp.com/cms/pages/60/edit/


source/sass/components/publication-page.scss
```diff
@@ -50,8 +50,8 @@
 }
 
 .publication-chapter-number {
-  width: 161px;
-  height: 161px;
+  width: 160px;
```
mmmavis, line 53: 161px is a really weird number so I just changed it to 160px.

network-api/networkapi/wagtailpages/factory/publication.py
```diff
@@ -63,6 +63,9 @@ class PublicationPageFactory(PageFactory):
     hero_image = SubFactory(ImageFactory)
     publication_file = DocumentFactory()
 
+    if random() > 0.5:
```
mmmavis, line 66: This ended up adding ToC images to all pages lol... why?

network-api/networkapi/wagtailpages/factory/publication.py
```diff
@@ -63,6 +63,9 @@ class PublicationPageFactory(PageFactory):
...
+        toc_thumbnail_image = SubFactory(ImageFactory)
```
KalobTaulien, line 67: Ahh I see what you're doing here. For this we need a `@post_generation` method. 

First, import `post_generation`:
```python
from factory import post_generation
```

I also used ImageFactory from somewhere else:
```python
from networkapi.wagtailpages.factory.image_factory import ImageFactory
```
Then replace these two lines with this:
```suggestion
    @post_generation
    def toc_thumbnail_image(self, create, extracted, **kwargs):
        if random() < 0.5:
            self.toc_thumbnail_image = ImageFactory()
```

You should be able to copy and paste this to the ArticlePage factory down below as well.

network-api/networkapi/buyersguide/templates/fragments/chapter_table_of_contents.html
```diff
@@ -16,7 +17,15 @@
             <div class="row publication-row pt-3 pb-3 d-flex align-items-center">
               <div class="col-sm-1 d-md-block d-none">&nbsp;</div>
               <div class="col-auto">
-                <div class="publication-chapter-number publication-chapter-number--small"></div>
+                <div class="publication-chapter-number publication-chapter-number--small"
+                  {% with toc_image=child.specific.toc_thumbnail_image %}
+                  {% if toc_image %}
+                  {% image toc_image fill-128x128 as thumbnail %}
```
KalobTaulien, line 23: Help text suggests using a `320x320` image but the images in the templates are using `128x128`. It'll shrink down which might be what you want. Just thought I'd flag that here in case this was supposed to be `320x320`

source/sass/components/publication-page.scss
KalobTaulien: Good catch. I likely copied this CSS directly from the design.

KalobTaulien (changes requested):

network-api/networkapi/buyersguide/templates/fragments/chapter_table_of_contents.html
mmmavis: Nice flag.

Yea this is intentional. It seems like a publication page can serve two type of ToC - one with bigger thumbnail (chapter ToC) and one with smaller thumbnail (publication ToC). Since we can't distinguish them during page creation process, I set the help text using the larger thumbnail dimension so the image could render well for both cases.

image

and 

image

network-api/networkapi/wagtailpages/factory/publication.py
mmmavis: Wooohoo! Thanks! This def solves the problem I was having!

KalobTaulien (dismissed):

KalobTaulien:
Hmm looking at percy, it might not like the randomizing of toc images. Wonder if this is going to give us more percy false negatives down the road? 

Regardless, this worked beautifully for me locally. R+

sabrinang (approved):
Works for me and great to see the help text _(320×320 pixels or larger)_ included there  too! Thanks :)

mmmavis:
My last commit dismissed the approvals I got. I'm requesting another review from @Pomax.

Pomax (approved):

Pomax (approved):

PR #6056, opened 2021-01-19
Patch for tab errors

KalobTaulien:
A few things I was able to do to solve our problems.

Im using "solve" VERY loosely here. This code helps resolve the issues from the target branch, but has other code based issues worth digging into. 

For example, to solve the "Cannot find tab field for 'votes'" we need to expose `votes` in a FieldPanel, which is not an acceptable way to move forward.


Pomax:
let's create a feature branch to work against for this work, rather than master (unless it solves a problem _with_ master, of course!)

PR #6086, opened 2021-01-22
Update 'optimize:svg' task

### LINKED ISSUE #4979: Update "optimize:svg" task so svgo has the right options passed in
Currently, we have

```js
"optimize:svg": "find source/images -type f -name '*.svg' -print0 | xargs -0 -n 1 -P 6 svgo --multipass --pretty --enable=removeTitle"
```

It seems like we can optimize SVGs further by passing in appropriate [plugins](https://github.com/svg/svgo#what-it-can-do) to remove unnecessary bits like unused `id` etc.

(Related thread: https://github.com/mozilla/foundation.mozilla.org/pull/4965#discussion_r461123824)

/cc @Pomax to provide suggestion

mmmavis:
Closes #4979

Did a few tests. It seems like without passing a custom config file, SVGO will use its [default config (which includes all possible plugins)](https://github.com/svg/svgo/blob/master/.svgo.yml) to optimize SVGs. It's only when we wanna disable a certain plugin from the list, then we pass in `disable=pluginname`.

I've removed `--enable=removeTitle` from `package.json` since it's already included in the default SVGO config.


Pomax (approved):
good find

PR #6087, opened 2021-01-22
Add a new image format "full width high quality"

### LINKED ISSUE #6068: Wagtail: Blurry PNG Image
I followed steps in https://foundation.mozilla.org/en/docs/how-do-i-wagtail/photo-and-image-guidelines/ to update the JPG ToC image on whitepaper that was blurry but it is still blurry with the new PNG, see: https://foundation.mozilla.org/en/insights/trustworthy-ai-whitepaper/path-forward/introduction-theory-change/

MoFo AI Theory of Change (ToC) – Design Slide
This is the PNG and I used ImageOptim to compress size (173 kb) and you can find it in gdrive here: https://drive.google.com/open?id=1GItFxsprIs5aZpaDKmNA2KaxBTAWXITR&authuser=sabrina%40mozillafoundation.org&usp=drive_fs

mmmavis:
Closes #6068

Some background research: https://github.com/mozilla/foundation.mozilla.org/issues/6068#issuecomment-765742148

@Pomax @KalobTaulien tagging you both for review as I'm not sure if this PR can cause any side effects. I've tested it locally on different pages and scenarios and they seemed fine. Also I'm wondering if we need to add this same file to all Wagtail apps we have? If so, is there a way to just have them referenced to the same one file instead of having duplicates?

---

Add a new image format "full width high quality" that is tailored for large high def image.
(Note: I can't re-order the format list without overwriting existing code, so I'm just gonne keep the new format appear last)

- Review app test page: https://foundation-s-issue-6068-goqawn.herokuapp.com/en/publication-page-with-child-article-pages/fixed-title-article-page/
- Edit page: https://foundation-s-issue-6068-goqawn.herokuapp.com/cms/pages/51/edit/

### Steps to test (review app or locally)
1. Go to any page that allows image embed in the rich text editor (e.g., Article page).
2. In rich text editor, embed the image Sabrina included in ticket #6068 (use Goole Drive one) and choose "full width high quality" as the format
3. Preview or publish the page
4. Inspect the image. See if it's 2280px wide instead of 800px.
5. You can also test on an existing image embed and verify the rendered image quality.


network-api/networkapi/wagtailpages/image_formats.py
```diff
@@ -0,0 +1,8 @@
+from wagtail.images.formats import Format, register_image_format, unregister_image_format
+
+# We can't override existing image format,
+# so we are unregistering it and then re-registering it with new values instead.
+# Set it to 2280px because 1140px is the max width of Bootstrap's container.
+# Then we multiply that number by 2 so the image can render well on retina.
+unregister_image_format('fullwidth')
+register_image_format(Format('fullwidth', 'Full width', 'richtext-image full-width', 'width-2280'))
```
mmmavis, line 8: Or is it less risky if we add a new option instead of modifying the existing `full width` option? Something like `full width high quality`?

KalobTaulien: As far as I'm aware this should be fine. But there may be cases where we don't want an image to be 2280px wide (like if someone has a 1000px image). If the uploaded image isn't at least 2280px wide it'll stretch the image to make it wider which will blur it. 

I like the idea of adding a new high quality image format so that way there's the standard way and the new high def way. This wouldrequire a little staff education.

KalobTaulien (changes requested):
Let's go with the new `full width high quality` option you came up with so smaller images aren't stretching to meet the 2280px width requirement and becoming blurry.

network-api/networkapi/wagtailpages/image_formats.py
Pomax: do we have any control over the resize algorithm? If not, having a way to say "don't upscale" would be good.

Pomax (dismissed):
I will defer to Kalob in this case, this is a part of wagtail I've not looked at before

network-api/networkapi/wagtailpages/image_formats.py
mmmavis: @Pomax I think not putting any resizing option would render the image in its original size (e.g., no `width-*`, `fill-*` etc)

@KalobTaulien I had a similar concern when working on this, and I struggled to find a good UX solution to it. Our current image alignment/resizing options on Wagtail will resize images by default:
```py
register_image_format(Format('fullwidth', _('Full width'), 'richtext-image full-width', 'width-800'))
register_image_format(Format('left', _('Left-aligned'), 'richtext-image left', 'width-500'))
register_image_format(Format('right', _('Right-aligned'), 'richtext-image right', 'width-500))
```

How do we make it more explicitly so that users can know images will be resized automatically? Should we update the image option label so it says, for example, `Right-aligned (will automatically be resized to 500px wide)` instead of `Right-aligned`?

mmmavis:
@KalobTaulien I ended up sticking with `width-2280` for the new format. We currently don't have an option for large high def image to show up as high def large image, so I decided to make this new format tailored to it. For smaller images that are under 2280px, users will have to choose from the current existing three formats.

mmmavis:
@sabrinang I included a note in the [PR description](https://github.com/mozilla/foundation.mozilla.org/pull/6087#issue-560271881). Unfortunately, no. Well technically yes but that'll require overwriting a bunch of existing code which I think it's a bit of an overkill for this as it adds code maintenance overhead.

network-api/networkapi/wagtailpages/image_formats.py
KalobTaulien: I think this comes down to just educating staff on the CMS.

KalobTaulien:
> @KalobTaulien what do you think? Should I unregister and re-register existing image formats so they can appear in the order we want?

Personally, I don't think it's that big of a deal to have these ordered properly. But if we as a group feel its important, let's just double check what happens when you unregister and re-register with the same settings. We want to test that a "left aligned" image (for example) maintains it's left float.

KalobTaulien (approved):

KalobTaulien:
Percy be actin' strange again. @mmmavis mind taking a look at it?

mmmavis:
@sabrinang I'm going to just merge this PR. If we really wanna re-order the option list let's just file a new ticket as it requires overwriting existing settings and it could potentially break things. (should be fine but there's always a chance)

PR #6101, opened 2021-01-25
[demo] Use submenus for new BuyersGuide menu

KalobTaulien:
:warning: do not merge :warning:

Example code for wagtail menu/submenu nav item generation

### DIFF
--- a/network-api/networkapi/buyersguide/admin.py
+++ b/network-api/networkapi/buyersguide/admin.py
@@ -1,8 +1,9 @@
-from django import forms
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
 
 from wagtail.core import hooks
-from wagtail.admin.menu import MenuItem
-from wagtail.admin.forms import WagtailAdminModelForm
+from wagtail.admin.menu import Menu, MenuItem, SubmenuMenuItem
+
 
 from wagtail.contrib.modeladmin.options import (
     ModelAdmin,
@@ -11,7 +12,6 @@ from wagtail.contrib.modeladmin.options import (
 )
 
 from networkapi.wagtailpages.models import (
-    ProductPage,
     BuyersGuidePage,
     GeneralProductPage,
     SoftwareProductPage,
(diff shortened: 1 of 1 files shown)


network-api/networkapi/buyersguide/admin.py
```diff
@@ -70,8 +70,8 @@ class WagtailBuyersGuideCategoryAdmin(ModelAdmin):
     add_to_settings_menu = False
     exclude_from_explorer = False
 
-    list_display = ('featured', 'title', 'description')
-    search_fields = ('title')
+    list_display = ('featured', 'name', 'description')
+    search_fields = ('name',)
```
KalobTaulien, line 74: I noticed this was throwing an error when viewing this page. `title` -> `name`

network-api/networkapi/buyersguide/admin.py
```diff
@@ -1,8 +1,9 @@
-from django import forms
+from django.urls import reverse
+from django.utils.translation import gettext_lazy as _
```
TheoChevalier, line 2: Strings from this page, as they are only visible in the CMS, shouldn’t be localized through Pontoon, so you will want to either hardcode them, or add this page to the excluded files from localization https://github.com/mozilla/foundation.mozilla.org/blob/master/tasks.py#L23-L29

TheoChevalier (changes requested):

network-api/networkapi/buyersguide/admin.py
Pomax: I'm marking this PR as [demo]: this is not a true PR

Pomax (commented):
this should be useful when we redo the buyers guide menu - for now we're taking that menu out just to make sure staff don't get confused about where they should be updating/editing products.

network-api/networkapi/buyersguide/admin.py
TheoChevalier: Sorry for jumping the gun!

PR #6116, opened 2021-01-27
'What you can do' page - remove extra DOM node & adjust background image alignment

### LINKED ISSUE #5336: "What you can do" page: hero section
### Describe the bug
The current hero section of the What you Can do page has a duplicated image: one on top (like a banner) and one behind the copy. The screenshot below is from staging but issue is also happening on prod:
image

### To Reproduce
1. Go to '[https://foundation.mozilla.org/en/what-you-can-do/](https://foundation.mozilla.org/en/what-you-can-do/)'
2. Use the "Responsive Design Mode" > iPhone 5 (or 6/7/8)
3. See bug

### Expected behavior
When this page was built, the expected behavior was to have only the top banner.
image

However, since we evolved the homepage to have a full width banner behind the copy on mobile, it makes more sense we replicate the same behavior for this page as the screenshot below for 2 reasons: 1. consistency and 2. to push the content up in the page
image

### Additional context
@kristinashu @sabrinang @natalieworth: if you have any strong opinions/concerns in having this hero behaving the same as our homepage (as I am suggesting), please let's chat before engineers solve this bug. I'll tag it as only design for now so we can discuss it if needed. If you all agree, please give a thumbs up to this thread so I know we're all aligned. Thanks!

mmmavis:
Closes #5336

https://foundation-s-issue-5336-alkd54.herokuapp.com/en/what-you-can-do/
@taisdesouzalessa I also updated the background image position to be at "center center" instead of "top center" (what's on prod). If it doesn't look good, let me know then I'll revert the change back. "center center" is more conventional that's why I changed it to that.


Pomax (dismissed):

taisdesouzalessa (changes requested):
So good to see this image working now :)
Some small tweaks:
• On mobile (preview on iPhone 5 and 6), the button is cutting out - could we make that responsive please?
image

• The reason we used "top center" was for when we have people's faces there. We wanted the eyes to always be visible (if we used "center center" we would have the risk of showing the middle of the faces instead of the top area). Does it make sense? So my preference is to keep it "top center"

mmmavis:
@taisdesouzalessa got it. I have reverted the changes back to "top center"

For the button issue you pointed out -- button style is based on brand guide so there's not much I can do here. This was in discussion before but we've decided to keep button text unwrapped since for most cases we don't have long button copy.

taisdesouzalessa (approved):
good to go! Thanks for the note on the button, I didn't remember that conversation but now I do!

Issue #6140, opened 2021-02-01
Add Commento to the PNI product pages


Pomax:
This should be a matter of adding the following code to the product_page.html template:
```html
<script async defer src="https://cdn.commento.io/js/commento.js"></script>
<div id="commento"></div>
```

Pomax:
There may be some followup left based on notes in https://github.com/MozillaFoundation/mofo-devops/issues/805

PR #6142, opened 2021-02-01
Switch to Commento.io

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6140

This replaces all our coraltalk comment markup/code with code for using commento.io instead.


cypress/integration/a11y/global_variables.js
```diff
@@ -1,22 +1,28 @@
-const MOBILE = 'iphone-6';
-const DESKTOP = 'macbook-13';
-const viewports = [MOBILE, DESKTOP]
+const MOBILE = "iphone-6";
```
Pomax, line 1: lol `prettier`

network-api/networkapi/buyersguide/templates/product_page.html
```diff
@@ -263,7 +263,7 @@ <h2 class="h3-heading">{% trans "Company contact info" %}</h3>
     <div class="container">
       <div class="row">
         <div class="{{section_class}} mt-5 pt-3 bg-white product-section-padding-x">
-          {% if product.updates|length > 0 or coralTalkServerUrl %}
+          {% if product.updates|length > 0 %}
```
Pomax, line 266: this is the "old" product_page.html, which is no longer being used. Deleting the coral talk code still made sense, though.

network-api/networkapi/buyersguide/views.py
```diff
@@ -157,7 +157,7 @@ def product_view(request, slug):
         'categories': BuyersGuideProductCategory.objects.all(),
         'product': product_dict,
         'mediaUrl': get_media_url(settings.USE_CLOUDINARY),
-        # 'coralTalkServerUrl': settings.CORAL_TALK_SERVER_URL,
+        'use_commento': settings.USE_COMMENTO,
```
Pomax, line 160: _technically_ not needed here, since this file will get deleted when we remove the old PNI app, but it felt reasonable to change anyway just in case.

cypress/integration/a11y/global_variables.js
```diff
@@ -1,22 +1,28 @@
...
+  [".wide-screen-menu", ".nav-links"], // Desktop Nav
+  [".narrow-screen-menu-container", ".nav-links"], //Mobile Nav
+  [".donate-banner *"][(".site-footer", "#language-switcher")], // Donate Banner // Language Switcher
+  [".site-footer a.logo"], // Footer Logo
+];
 
 const EXCLUDE_PNI_CONSTANTS = EXCLUDE_CONSTANTS.concat([
-    ['#coral-talk-stream', 'iframe'] // Commenting
-  ])
+  // ["#coral-talk-stream", "iframe"], // Commenting
```
mmmavis, line 18: Do we wanna remove this part of code for good? If not, can we leave a note explaining why we decided to comment out this line instead of removing?

network-api/networkapi/buyersguide/management/commands/generate_pni_report.py
```diff
@@ -150,10 +117,6 @@ def handle(self, *args, **options):
                            'would_buy = EXCLUDED.would_buy,\n' \
                            'would_not_buy = EXCLUDED.would_not_buy'
 
-            print('Fetching Coral Comment Data')
-            comment_data = self.get_comment_counts()
-            comment_data = self.dedupe_comments(comment_data)
-
             print('Generating Upsert Query for comment data')
             sql_comments = 'INSERT INTO comment_counts\n' \
```
mmmavis, line 121: remove this and line 133 `cursor.execute(sql_comments)`?

network-api/networkapi/buyersguide/templates/buyersguide/product_page.html
```diff
@@ -328,12 +328,7 @@ <h3 class="h2-heading mb-4">{% trans "Related products" %}</h3>
...
 {% block extra_scripts %}
-  {% if coralTalkServerUrl %}
-    <script src="{{ coralTalkServerUrl }}static/embed.js" async onload="
-      Coral.Talk.render(document.getElementById('coral-talk-stream'), {
-        talk: '{{ coralTalkServerUrl }}'{% if lang_code == "en" %},
-        asset_url: '{{ product.get_parent.url }}products/{{ product.slug }}/'{% endif %}
-      });
-    "></script>
+  {% if use_commento %}
+    <script async defer src="https://cdn.commento.io/js/commento.js"></script>
```
mmmavis, line 332: Do we need both `async defer`?

network-api/networkapi/wagtailpages/templates/wagtailpages/blog_page.html
```diff
@@ -48,11 +48,7 @@ <h1 class="h1-heading">{{ page.title }}</h1>
...
 
 {% block extra_scripts %}
-  {% if coral_talk %}
-    <script src="{{ coral_talk_server_url }}static/embed.js" async onload="
-      Coral.Talk.render(document.getElementById('coral-talk-stream'), {
-        talk: '{{ coral_talk_server_url }}'
-      });
-    "></script>
+  {% if use_commento %}
+    <script async defer src="https://cdn.commento.io/js/commento.js"></script>
```
mmmavis, line 52: Do we need both `async defer`? `defer` would be ignored if both are presented I _think_?

mmmavis (changes requested):
Found some issues and I haven't tried to run it locally. Can you remind me how to run the `generate_pni_report` management task? I can't find any docs.

cypress/integration/a11y/global_variables.js
Pomax: I left this part in mostly in case we need to exclude commento, so that there's an example of the format we'll need to use.

Pomax:
```suggestion
  // ["#coral-talk-stream", "iframe"], // leaving this in for as long as is necessary to determine whether the commento element needs excluding
```

network-api/networkapi/buyersguide/templates/buyersguide/product_page.html
Pomax: yes, they do different things: `async` regulates the script load (making it a non-blocking requiest), whereas `defer` regulates loading behaviour (running the code only after the DOM has been constructed, but before DOMContentReady gets signalled)

Pomax:
you should be able to run it with `inv manage generate_pni_report`

network-api/networkapi/buyersguide/templates/buyersguide/product_page.html
mmmavis: So this essentially means the script file gets loaded asynchronously but the code doesn't run until `DOMContentReady`?

mmmavis (dismissed):
Looks good! Verified on blog page and product page with `USE_COMMENTO` set to True. Both load the script and have `<div id="commento"></div>` in DOM.

network-api/networkapi/buyersguide/templates/buyersguide/product_page.html
Pomax: until just before `DOMContentReady`, yes

Pomax:
merging to master - let's see what this does on the staging server

PR #6143, opened 2021-02-01
PNI SEO

### LINKED ISSUE #4787: SEO recommendations for /privacynotincluded/
I came across the privacynotincluded several times the last year and today I took some time to see how it does in search engines. And I found a good base of sessions coming from search but as well opportunities to grow with few small changes on the site, I collected in this doc: https://docs.google.com/document/d/1qNv74r0CcGymkLbgrHpBEEXfj4LaG7syxc_Bq1UXdlw/edit?usp=sharing

We could do some changes on titles, meta description, og:tags and the international hreflang setup to increase traffic by getting better snippets and make it easier for users finding us in search to click us.

mmmavis:
Closes #4787 

- Meta titles are set based on the [recommendation](https://github.com/mozilla/foundation.mozilla.org/issues/4787#issuecomment-652021557) but with some tweaks so the language is more aligned with our brand. 

### Meta to check
- title (meta title, og title, twitter title)
- description (meta description, og description, twitter description)
- og image and twitter image

### Testing instructions:
- [x] Go to all page types (home, category page, product page, about page) and make sure they have default meta set
- [x] Edit home and product page meta (in promote tab) on CMS, publish the page, and verify meta reflects the change

I found this handy [addon](https://detailed.com/extension/). It's def easier to use than verifying everything in view source 😂


network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -575,10 +575,10 @@ def get_context(self, request, *args, **kwargs):
...
         context['mediaUrl'] = settings.CLOUDINARY_URL if settings.USE_CLOUDINARY else settings.MEDIA_URL
         # context['coralTalkServerUrl'] = settings.CORAL_TALK_SERVER_URL
-        context['pageTitle'] = f'''{pgettext(
-          'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
-          '*privacy not included'
-        )} - {self.title}'''
+        context['pageTitle'] = f'{self.title} | ' + pgettext(
+                'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
+                'Privacy & Security Guide'
+            ) + f' | Mozilla Foundation'
```
Pomax, line 581: when you're using f strings, you don't want to also be using string concatenation:
```suggestion
        context['pageTitle'] = f'{self.title} | {pgettext(
                'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
                'Privacy & Security Guide'
            )} | Mozilla Foundation'
```

mmmavis: kk I will update. I was just following the existing code we have elsewhere.

Pomax: there's a bunch of regular strings where we do this, which can only put values "in" the string by concatenating several strings, but when using f strings you get the `{...}` syntax to do that, instead.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -626,10 +626,10 @@ def get_context(self, request, *args, **kwargs):
...
         context['mediaUrl'] = settings.CLOUDINARY_URL if settings.USE_CLOUDINARY else settings.MEDIA_URL
         context['use_commento'] = settings.USE_COMMENTO
-        context['pageTitle'] = f'''{pgettext(
-          'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
-          '*privacy not included'
-        )} - {self.title}'''
+        context['pageTitle'] = f'{self.title} | ' + pgettext(
+                "This can be localized. This is a reference to the “*batteries not included” mention on toys.",
+                "Privacy & Security Guide"
+            ) + ' | Mozilla Foundation'
```
mmmavis, line 632: Had to go with this because

image

And if I use `f'text'` and put everything in one line, flake8 will complain that the line being too long. 🤔

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1232,9 +1232,10 @@ def categories_page(self, request, slug):
 
         context['category'] = category.slug
         context['products'] = products
-        context['pageTitle'] = pgettext(
+        context['pageTitle'] = f'{category} | ' + pgettext(
                 'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
-                '*privacy not included') + f' - {category}'
+                'Privacy & Security Guide'
+            ) + f' | Mozilla Foundation'
```
mmmavis, line 1238: same issue as mentioned in https://github.com/mozilla/foundation.mozilla.org/pull/6143/files#r572438582

TheoChevalier (changes requested):
Looks like there a few things we need to update to accommodate localization. Let me know if you need more clarity/help since this is not completely straight forward

mmmavis:
@TheoChevalier thanks for the comments. Do you know when we can drop a comment (e.g., `This can be localized. This is a reference to the “*batteries not included” mention on toys.`) and when we wanna keep it? From your comments above, it seems like we are okay with having localization note in templates but not anywhere else?

Another q is that why is it called "batterires not included" instead of "privacy not included" 😄

TheoChevalier:
> @TheoChevalier thanks for the comments. Do you know when we can drop a comment (e.g., `This can be localized. This is a reference to the “*batteries not included” mention on toys.`) and when we wanna keep it? From your comments above, it seems like we are okay with having localization note in templates but not anywhere else? Is it because for those cases we didn't mention PNI in the text so that's why the comment is no longer needed?

Hey @mmmavis , generally speaking, every time we update a string that has a localization note, we should ask whether the comment is still relevant,  needs an update, or is not needed anymore. In case of doubt, I can help answering.
Here the comments are clarifying a couple things: 1/ that it’s safe to translate this brand name (because it’s often a grey area) and 2/ explain the initial reference to help them create a similar reference in their language.

So to answer your question, we only need this comment when we mention PNI, and we can have comments everywhere code permit it (both templates and the entire Python codebase)

> 
> Another q is that why is it called "batterires not included" instead of "privacy not included" 😄

See above, we’re just explaining to translators the original reference of the name

network-api/networkapi/wagtailpages/pagemodels/products.py
Pomax: python lets you break up strings over multiple lines by putting it in parens and then splitting it:
```
thing = "if this would be a long line then the following format will split it:"

longersplit = (
  "for example this is a long line"
  " and can be broken up with just"
  " a few more quote symbols"
)
```

mmmavis: Ahhhhh alright! Thanks!

network-api/networkapi/buyersguide/templates/fragments/pni_nav_links.html
```diff
@@ -1,4 +1,9 @@
-{% load bg_nav_tags i18n %}
+{% load bg_nav_tags i18n wagtailroutablepage_tags %}
 
-{% url 'about-why-view' as about_why_url %}
+{% if page %}
```
mmmavis, line 3: following https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/templates/about/about_nav.html#L4-L19

mmmavis:
@TheoChevalier @Pomax  PR updated!

@TheoChevalier thanks for the explanation. I'm still confused about how to provide a clear note to translators as I still don't get why we used `batteries` in `This is a reference to the “*batteries not included” mention on toys.` instead of just `privacy` 😬 I've tried my best putting in the note. Please suggest a better copy if mine isn't clear enough. Thanks!

mmmavis:
@Pomax tests are failing so I will ping you tomorrow for help 😬

network-api/networkapi/buyersguide/templates/fragments/pni_nav_links.html
mmmavis: @Pomax 🙄

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1142,62 +1139,62 @@ def about_page(self, request):
...
+            '*privacy not included can be localized.',
+            'Why we made *privacy not included')
         return render(request, "about/why_we_made.html", context)
 
     @route(r'^about/press/$', name='press-view')
     def about_press_page(self, request):
         context = self.get_context(request)
         context['pagetype'] = 'about'
-        context['pageTitle'] = pgettext(
+        context['pageTitle'] = gettext('Press') + ' | ' + pgettext(
```
TheoChevalier, line 1159: Just adding one more comment, to specify it’s that kind of press, not a verb/action.

```suggestion
        context['pageTitle'] = pgettext('Noun, media', 'Press') + ' | ' + pgettext(
```

TheoChevalier (commented):
Thanks @mmmavis! Your notes are looking good, just added one more to clarify "Press". And with that, it will be an r+ from me

Let me try to clarify further the batteries thing :)
“Batteries not included” was a disclosure you could find on the packaging of electrically powered toys (very common a few years ago), and it became an idiom. PNI’s name comes from this, and we’re explaining this to translators.

network-api/networkapi/wagtailpages/pagemodels/products.py
mmmavis: Thanks for making this more clear! @TheoChevalier

TheoChevalier (approved):

network-api/networkapi/buyersguide/templates/fragments/pni_nav_links.html
Pomax: That code only works for objects that extend the `RoutablePageMixin`, like the `BuyerGuidePage` (which needs to do so because it defines all the "virtual routes" for the about pages). Products aren't routable pages, so the `routablepageurl` tag won't work for them.

Pomax: Asking Kalob about this, his suggestion is to add this to our models that use the pni_nav_links:

```
def get_context(self, request, *args, **kwargs):
    context = super().get_context(request, *args, **kwargs)
    # ...if there's already context code, that goes here...
    context['about_page'] = BuyersGuidePage.objects.first()
    return context 
```

that way we can change the template from this
```
{% routablepageurl page 'about-why-view' as about_why_url %}
```
into this:
```
{% routablepageurl about_page 'about-why-view' as about_why_url %}
```

mmmavis: Updated! I've only added `context['about_page'] = BuyersGuidePage.objects.first()` to `ProductPage` and `BuyersGuidePage`. Those `about` virtual routes seem fine without having `about_page` as context.

mmmavis:
PR updated!

network-api/networkapi/wagtailpages/pagemodels/products.py
TheoChevalier: Hmm, reopening this, because it turns out ```f'{self.title} | {gettext("Privacy & security guide")} | Mozilla Foundation'``` doesn’t work: the string is not extracted by `inv makemessages`

Pomax (approved):

mmmavis:
@TheoChevalier this is ready for review 👀

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -626,10 +626,8 @@ def get_context(self, request, *args, **kwargs):
         context['categories'] = BuyersGuideProductCategory.objects.filter(hidden=False)
         context['mediaUrl'] = settings.CLOUDINARY_URL if settings.USE_CLOUDINARY else settings.MEDIA_URL
         context['use_commento'] = settings.USE_COMMENTO
-        context['pageTitle'] = f'''{pgettext(
-          'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
-          '*privacy not included'
-        )} - {self.title}'''
+        context['pageTitle'] = f'{self.title} | {gettext("Privacy & security guide")} | Mozilla Foundation'
```
TheoChevalier, line 629: This doesn’t work unfortunately, the string is not extracted when running `inv makemessages`. I assume gettext doesn’t like the `f'` syntax, or it needs to be handled differently

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1232,9 +1230,7 @@ def categories_page(self, request, slug):
 
         context['category'] = category.slug
         context['products'] = products
-        context['pageTitle'] = pgettext(
-                'This can be localized. This is a reference to the “*batteries not included” mention on toys.',
-                '*privacy not included') + f' - {category}'
+        context['pageTitle'] = f'{category} | {gettext("Privacy & security guide")} | Mozilla Foundation'
```
TheoChevalier, line 1233: Same, this string is not extracted

TheoChevalier (changes requested):
Just one small thing to fix with the `f'`, then we should be good to land this :)

network-api/networkapi/wagtailpages/pagemodels/products.py
mmmavis: Thanks for catching this. I tried running `inv makemessages` but it didn't work for me. Do I have to run another command first to it to work? I'm getting

```bash
./translation-management.sh import
Importing latest translation files from fomo-l10n repository
cp: foundation/translations/locale/: No such file or directory
```

mmmavis: note to self: follow https://github.com/mozilla/foundation.mozilla.org/blob/master/README.md#translations

TheoChevalier (approved):
r+ with the latest changes

PR #6154, opened 2021-02-02
Fix for 'header' tab mising

### LINKED ISSUE #6030: wagtail-localize: prepare MozFest site for wagtail-localize
Capture d’écran 2021-01-15 à 11 52 15

This issue happens when trying to translate the MozFest homepage with wagtail-localize ([this branch](https://github.com/mozilla/foundation.mozilla.org/tree/wagtail-2-11))

This may or may not be the same root cause as for #5953, which is still being investigated.

KalobTaulien:
Closes #6030 

Steps to reproduce:
* Create a new database with `inv new-db` 
* Add a new locale by going to Settings -> Locales 
* Edit the Mozfest Homepage 
* Translate the page into your chosen locale 
* Publish the translated page 
* View the original english page in the Wagtail admin. It should not error on you. Previews and the locale-prefixed URL should would for you as well. 

The problem and solution:
Mozfest Homepage has a "hardcoded" banner option. If the banner is hard coded, it removes the "header" (and other fields) from the `content_panels` — wagtail-localize is looking for each field to match the `content_panels`. So it was looking for "header" even though its not in a content_panel anymore. The solution is to avoid putting "header" (and other fields) into the `translatable_fields` option

### DIFF
--- a/network-api/networkapi/mozfest/models.py
+++ b/network-api/networkapi/mozfest/models.py
@@ -5,6 +5,8 @@ from wagtail.core.models import Page
 from wagtail.images.edit_handlers import ImageChooserPanel
 from wagtail.snippets.edit_handlers import SnippetChooserPanel
 
+from wagtail_localize.fields import SynchronizedField, TranslatableField
+
 from networkapi.wagtailpages.utils import (
     set_main_site_nav_information,
     get_page_tree_information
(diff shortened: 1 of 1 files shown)


Pomax (approved):
yissss

PR #6156, opened 2021-02-03
Change wagtail image settings

### LINKED ISSUE #5954: Improve the CMS image management
Our current image approach is "all images in the root collection, 20 thumbs per page" which is becoming a bit unwieldy. Let's see if we can't:

- increase the number of thumbs per page (60? 100?)
- start using collections (PNI, blog, ...?)
- try to make sure to title and tag all images so that discoverability isn't just visual but can rely on text search working?

KalobTaulien:
Closes #5954

Show 100 images per page. And show the usage. (Example image of usage below)

image

**Note**: Where it shows the number of times an image is used it doesn't include where its used in streamfield blocks. This is purely on the Field-level. The reason for this is because StreamField uses JSON behind the scenes, and it'd have to loop through every page and every streamfield to determine the number of times an image is being used (which could take a VERY long time to process)


Pomax (approved):
preeeeeetty sweet

PR #6174, opened 2021-02-06
Fix factory warning around participate page

### LINKED ISSUE #5735: Factory warning - missing Participate page
Getting this error from `participate_page_featured_highlights.generate(seed)`

image

I can't remember what the status is for Participate page. Have we sunset it? If so, we should remove `participate_page_featured_highlights` and its related code.

mmmavis:
Closes #5735 

Run `inv new-db` locally. Once that's done, search the log. The `Participate page must exist. Ensure that ...` error should not exist anymore.

Note that as a result, the fake Particapte2 page is more complete as `highlights` are being generated correctly. See Percy for differences.


Pomax (approved):
mannnnnn nice catch

PR #6188, opened 2021-02-08
So apparently comments have to go AFTER {% extends %}

### LINKED ISSUE #6152: clean up redirect template overwrite
### File(s) or components

https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailcustomization/templates/wagtailredirects/index.html

### Reason for cleanup

the heck is that top line coming from? that should be hidden~

Pomax:
Closes #6152

Live URL: https://foundation-s-cleanup-te-3qddhx.herokuapp.com/cms/redirects/ (compared to https://foundation.mozilla.org/cms/redirects/)


mmmavis (dismissed):
nice finding that we need to put comment after `extend`

Pomax:
I'm going to leave this one open for a bit so it can help determine whether a percy fix works

PR #6198, opened 2021-02-11
Add optional field 'height' to iframe_block

### LINKED ISSUE #4115: [Component] Option to set height and hide scrollbar for iFrame
**Request**
- Add option to set the height of the embedded iFrame window
- ~~Add checkbox to give option to show/hide scroll bar~~ Crossed out because we shouldn't and can't do it (see thread below)

image

**Example:** Currently the fixed height makes it hard to read/scan the content inside and the scroll bar makes it look busy and obvious it's separate from the other elements on the page (it would look more integrated and cleaner without)

image

mmmavis:
Closes #4115 

Test page: https://foundation-s-issue-4115-1y1onz.herokuapp.com/en/what-we-do/
Edit page: https://foundation-s-issue-4115-1y1onz.herokuapp.com/cms/pages/10/edit/ (play around with the new `height` field of `iFrame` block)


Pomax (approved):
lgtm

sabrinang:
@mmmavis I missed reviewing this earlier in the week, my bad! I don't see the ra links anymore, if it is up again I can review

mmmavis:
@sabrinang I've revived the review app!

sabrinang (commented):
The height field works well!

- For the URL help text, is there a way for users to know which domains are approved (white-listed) URLs? 
- For height help text, would it be clearer if it was written: “Optional pixel value for custom iFrame height” 
- probably shouldn't allow negative values (still shows as default height though)?

mmmavis:
> - For the URL help text, is there a way for users to know which domains are approved (white-listed) URLs? 

There is no good way of doing this as the white-listed URLs are stored on Heroku as a config variable. (i.e., we can't auto pull the list from Heroku so it shows up as help text on CMS)

> - For height help text, would it be clearer if it was written: “Optional pixel value for custom iFrame height” 

Sure. I will update it.

> - probably shouldn't allow negative values (still shows as default height though)? 

Interesting test case! But it should be okay to leave it as is. We make it clear that the field is for height so users are expected to enter a positive value.

sabrinang (approved):
Thanks for updating and the rest makes sense :+1:

PR #6220, opened 2021-02-16
Redirect middleware for selecting a redirect page

### LINKED ISSUE #6114: wagtail-localize: ensure redirects work correctly with wagtail-localize
There are some redirect oddities that we need to address: if I have a rule `/blogz` that goes to the `/blog` page instead, and it works for both unlocalized and localized URLs, the latter is a bit funny:

- ✔️ `/blogz` goes to `/en/blog` as expected,
- ✔️ `/en/blogz` also goes to `/en/blog` as expected
- ❌ a non-default locale code like `/fr/blogz` seems to remove the locale, apply the redirect, but forgets which locale the request was for, redirecting to `/en/blog` instead of `/fr/blog`
- ❌ even if I set my language preference to French with the language picker in our footer, `/blog/` correctly takes me to `/fr/blog`, but `/blogz` incorrectly redirects me to `/en/blog`

KalobTaulien:
Closes #6114 

To test this:
 * Setup a new locale in the /cms/ admin (ie. french)
 * Create a redirect from /test-page to a page you've chosen (ie chose the english blog page)
 * Change your browser settings to another language (ie. french)
 * Go to http://localhost:8000/test-page 
 * It _should_ take you to http://localhost:8000/fr/blog/ even though you selected the english blog page for the redirect

Also worth testing:
* Set your browser language to an unsupported language from the site (ie. If there are no Locales for spanish, set your browser to spanish) 
* Test the same redirect — because the Spanish Locale doesn't exist in the cms, it should redirect to the main selected page. 

**Note** This PR is _just_ for selecting a page when creating a new redirect. And this only works if the users language code is set in the request.

**Link to sample test page**: https://foundation-s-6114-ensur-kcjryj.herokuapp.com/english-selected-but-goes-to-es


Pomax:
Testing with:

- english locale
- german locale
- redirect: `/blogz` pointing to the english _page_ for Blog

results with browser set to english preference:

- ✔️ `/blogz` goes to `/en/blog`
- ✔️ `/en/blogz` goes to `/en/blog`
- ❌ `/de/blogz` goes to `/en/blog` (should go to `/de/blog`)
 
results with browser set to german preference:

- ✔️ `/blogz` goes to `/de/blog`
- ❌ `/de/blogz` goes to `/en/blog` (should go to `/de/blog`)
- ✔️ `/en/blogz` goes to `/en/blog`

Pomax (changes requested):
let's talk about this one, because this fixes some of the broken cases, but breaks some of the working cases.

Pomax:
Given the new wagtail-localize attempt, it seems like we won't need this anymore. Redirects appear to work properly

PR #6226, opened 2021-02-17
Change 'Become a Sponsor' footer string

benhohner:
Change footer string.
Requested by MozFest team. I thought I'd do it and save someone the time of making a ticket.
Let me know if this is not good.

live URL: https://foundation-s-switch-foo-6gy8nq.herokuapp.com/en/


Pomax (approved):
lgtm =D

PR #6230, opened 2021-02-18
Replace webpack+babel with esbuild

### LINKED ISSUE #6222: Replace webpack with esbuild
https://esbuild.github.io seems _really_ fast, and quite suited to replace webpack, based on how we use webpack. Let's speed this build up!

Pomax:
Closes #6222

This PR replaces the Webpack+Babel code with ESBuild code, which is a tool written explicitly to be the fastest, bestest js(x) bundler. And it shows: running `npm run build:js:prod` runs so fast I initially thought it was broken, but no: it generates all our code in literal milliseconds. It has a built in watch mode (=no chokidar for js(x) anymore) that is equally fast: updating a file and hitting save rebuilds all the relevant bundles (automatically, you don't say which dirs to watch, it already knows which files you're working with because it had to load all of them to build your bundles) also takes milliseconds.

This thing is amazing.

Things found:
- Sass still build as slowly as before, it would be lovely if we could figure out how to speed that up (I've made the build tasks parallel, running postcss optmize after) 
- This will not work for IE11, so we need an extra tasks that can convert our bundles to ES5 format. This probably still needs babel, but at least it doesn't need to do any jsx conversions, it'll just be a straight up "this this single ES6 js file into another single ES5 js file"

The IE11 issue, however, seems... irrelevant. I had a look at our site in IE11 and it's already pretty bad. Not "it's a hot mess" bad, but we use CSS that IE11 doesn't support, so even before any JS loads IE11 already looks "wrong". It's good enough to read our blog posts or campaigns, etc. but our JS not working in IE11 doesn't seem a problem: things don't work properly in IE11 to begin with (and presumably haven't for quite a long time now).

So after experimenting with babel to perfor for-IE11-only and for-everything ES6-to-ES5 conversion, it's just not worth it: having JS that IE can load still does not actually lead to any real improvement of site usability, so I'd say let's leave IE11 alone and just focus on improving things for the modern browser experience.


KalobTaulien (dismissed):
This runs very smoothly on my machine inside of Docker. I didn't test it outside of Docker; didn't think that was necessary at this point since we auto-run esbuild inside of Docker anyway. 

I tested a bunch of the new commands and everything seemed to work the way I expected. 

I tested changing a random .scss and random .js file, and it compiled super quickly and perfectly. Plus I didn't have to slam the hard-reload option 8+ times to see my changes in the browser, which really sold me on esbuild (along with its speed!)

This is an R+ from me, but I think @mmmavis should also take a deeper look into as this will affect her more than us. 

Is there any CI/CD changes we need to consider when moving away from webpack, or do the new commands in package.json simply replace the same commands that get executed in CI?

mmmavis (dismissed):
Wow! I'm very impressed with the JS build speed! Tested a few commands out and everything worked as expected for me.

PR #6231, opened 2021-02-18
Fix navigation & default values to None to avoid undefined var errors

### LINKED ISSUE #6163: [Publications] Previous/Next Missing Headings on Child Publication Pages
Previous/Next buttons are blank at the bottom of child publication pages (button headings on article pages are present). Currently, links to itself but should link to previous/next sections too.

Example: https://foundation.mozilla.org/en/insights/trustworthy-ai-whitepaper/path-forward/
image
Should be showing,
- Previous: Challenges with AI 
- Next: Conclusion

KalobTaulien:
Closes #6163 

Test link: https://foundation-s-6163-fix-c-4psemg.herokuapp.com/en/publication-page-with-chapter-pages/fixed-title-chapter-page/


sabrinang (dismissed):
Great 👍 I see the next/previous headings and the links work

network-api/networkapi/wagtailpages/pagemodels/publications/publication.py
```diff
@@ -127,6 +127,7 @@ def next_page(self):
         Only applies to Chapter Publication (sub-Publication Pages).
         Returns a Page object or None.
         """
+        next_page = None
```
Pomax, line 130: We probably want to set this to `self.get_parent()` instead of `None`, so that even if `self.is_chapter_page` is false, we return something with a `.url` property (otherwise `<None>url` in the template might still fail).

The following block logic might need an update though, along the lines of:
```python
next_page = self.get_parent()
if self.is_chapter_page
    sibling = self.get_siblings......
    if sibling:
        next_page = sibling
return next_page
```
```

File: network-api/networkapi/wagtailpages/pagemodels/publications/publication.py [hunk 2]
```diff
@@ -140,6 +141,7 @@ def prev_page(self):
         Only applies to Chapter Publication (sub-Publication Pages).
         Returns a Page object or None.
         """
+        prev_page = None
```
Pomax, line 144: and then the same here

Pomax (changes requested):

Pomax (approved):

PR #6247, opened 2021-02-23
Convert cloudinary images to Wagtail images in a data migration

### LINKED ISSUE #5812: Convert cloudinary images to Wagtail Images
Convert cloudinary images to Wagtail Images

May require a SPIKE ticket. May require some devops. 

Criteria:
- [x] Migration script needs to download cloudinary images from Products, create a new Wagtail Image, and attach it to the Product page
- [ ] Unit tests to test our assumptions (see comments as to why there isnt a unit test here)
- [x] ~Use s3 with cloudfront or cloudflare for cdn caching~ Using cloudflare for CDN delivery soon
- [ ] ~Remove cloudinary settings~ This will become its own ticket

KalobTaulien:
Closes #5812 

I'm opening this PR as a stand alone data migration in favour of creating a new ticket that focuses purely on removing cloudinary code. Cleanup code opened at #6248 

Testing this will be tricky. Ping me if you get stuck! 

1. In wagtailpages/pagemodels/products.py Line 513[ish] you'll see `image_field`. That needs to be replaced with this: 
```
FieldPanel('cloudinary_image'),
ImageChooserPanel('image'),
``` 
This will expose the cloudinary field and the wagtail image chooser. 

2. Upload an image to the cloudinary image field and save the PNI Product Page. The image should be stored in cloudinary now. 
3. Run the new management command with `python manage.py migrate_cloudinary`
4. Refresh the page you're editing, and the `image` field should be the same as the `cloudinary_image` field (but with a preview). And the `Images` in the admin should have the new image in the Wagtail admin.


Pomax (changes requested):
This works great, but the runtime for it is longer than I thought it'd be, so yeah... let's make this a mgmt command =P

KalobTaulien:
@Pomax I moved that to a management command. This one might take a few minutes because it has to download the images on each request (as in each page). And I simplified the testing instructions in the main comment above.

recepsirin (approved):

network-api/networkapi/management/commands/migrate_cloudinary.py
```diff
@@ -0,0 +1,50 @@
...
+from networkapi.wagtailpages.pagemodels.products import ProductPage
+
+
+class Command(BaseCommand):
+    help = 'Migrate PNI Cloudinary images to Wagtail images'
+
+    def handle(self, *args, **options):
+
+        for product in ProductPage.objects.all():
+            if product.cloudinary_image:
```
Pomax, line 21: let's add a little `print` here that tells us which product we're in (like "Processing product 23 of 317") so that running it via ssh keeps progress visible.

Pomax (changes requested):
R+ but let's add a print statement before we land this

network-api/networkapi/management/commands/migrate_cloudinary.py
KalobTaulien: Done!

Pomax (approved):

PR #6277, opened 2021-03-01
explicit target names for esbuild source

Pomax:
Fixes a server error 500 for the mozfest site (esbuild was creating index.compiled.js rather than mozfest.compiled.js)

### DIFF
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -19,13 +19,13 @@ const inProduction = mode === `production`;
 const inDir = `./source/js/`;
 const outDir = `./network-api/networkapi/frontend/_js/`;
 
-const sources = [
-  `main.js`,
-  `foundation/pages/mozfest/index.js`,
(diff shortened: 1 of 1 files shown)


esbuild.config.js
```diff
@@ -42,11 +42,9 @@ const base = {
   inject: [`esbuild.react.shim.js`],
 };
 
-sources.map((source) => {
+Object.entries(sources).map(([name, source]) => {
   const opts = Object.assign({}, base);
-  opts.entryPoints = [`${inDir}${source}`];
-  opts.outfile = `${outDir}${path
-    .basename(source)
```
Pomax, line 49: this line would take `foundation/pages/mozfest/index.js` and turn it into `index.js`, which broke the load instruction for `mozfest.compiled.js` as static asset.

KalobTaulien (approved):
Works locally for me. The _js/ dir and mozfest.compiled.js exist as well.

PR #6307, opened 2021-03-02
Replace all <figurecaption> instances with <figcaption>

### LINKED ISSUE #6305: Replace all <figurecaption> instances with <figcaption>
Replace all `<figurecaption>` instances with `<figcaption>`... because `figurecaption` does not exist as a standard HTML tag 😂 

Note: remember to check SCSS code as well

mmmavis:
Closes #6305


Pomax (approved):
how did we get so manyyyyy O_O

Issue #6313, opened 2021-03-03
Remove old BuyersGuideProductCategory model


KalobTaulien:
We need to remove the old BuyersGuideProductCategory model. There are some intricacies with this, such as imports, factories and tests that rely on the old model. 

We'll likely just need to update most of the imports to import from wagtailpages.BuyersGuideProductCategory instead of the old model. 

Dependency on #6012

Acceptance criteria:
- [ ] Remove old BuyersGuideProductCategory code.
- [ ] Swap out old BuyersGuideProductCategory code for factories and tests for the new wagtailpages.BuyersGuideProductCategory code 
- [ ] Make sure any context or @properties that are relying on the old code continue to work with the new model
- [ ] Ensure `inv new-db` continues to work (it'll fail if migrations are broken somewhere) 
- [ ] Ensure `inc new-inv` continues to build new environments
- [ ] Ensure `inc test-python` still passes on all expected tests.

PR #6317, opened 2021-03-03
Migrations to create new PNI category model, and migrate old data to new model.

### LINKED ISSUE #6012: Move `BuyersGuideProductCategory` model to `wagtailpages` app
Spin off from #5886

> we probably want to switch to a new category class instead, house in wagtailpages, so we can start moving away from needing code from the buyersguide app.
> 
> We can have a look at https://stackoverflow.com/a/26472482 and see if that works with wagtail mixed in

**Note**: My estimate of `4`is quite conservative. It could very well be an `8`. It's not a very straightforward task. There's also [this realpython](https://realpython.com/move-django-model/) article that could help shed some light on moving an entire model with ForeignKeys attached to it.

KalobTaulien:
Closes #6012 
Related PRs/issues #6313 (6313 is a follow up ticket)

All we did was:
1. Copy the old model to a new app (so we can remove the old model later) 
2. Copy the data from one model to the other (allowing pk's to change as needed) 
3. Copy the `category` field to `category_new` and fill it with the new model data
4. Remove the old `category` field
5. Rename `category_new` to `category` so the site picks up where it left off, and everything works as expected.
6. De-register the old BuyersGuideProductCategory snippet so editors continue to work with the new model instead of the old model

Testing steps: 
- [x] Run `inv new-db`. Migrations in wagtailpages numbered 0043 and 0044 should be executed without complaints. 
- [x] Edit a PNI product in the admin
- [x] Under the Product Categories are orderables. It should look like nothing has changed but if you click "Edit" on one of the snippets, it should bring you to a URL with `/wagtailpages/` in it. ie. /cms/snippets`/wagtailpages/buyersguideproductcategory/`1/
- [x] Pull down a copy of the staging database, note the category(ies) on a PNI product. Run the new migrations against localhost. The categories should look the same but the edit URL will change to reflect that the snippet is now in the `wagtailpages` app
- [x] Check the pni detail view to make sure the template still works as expected.


Pomax:
Something's not quite right with the migrations, because testing this on staging data shows categories changing:

image

I'm pretty sure Google Nest Hub Max was not a valentine's day product =D

Pomax:
Nav menu definitely also breaks now:

image

Pomax (changes requested):
this doesn't seem to be working correctly yet.

network-api/networkapi/wagtailpages/migrations/0044_buyersguideproductcategory_2.py
```diff
@@ -0,0 +1,42 @@
...
+    for product_page_category in ProductPageCategory.objects.all():
+        if hasattr(product_page_category, 'category'):
+            category = product_page_category.category
+            # Find the "new" model instance of the old category. Make sure
+            # there is only one returned, not a full QuerySet
+            new_category = BuyersGuideProductCategory.objects.filter(name=category.name).first()
+            # Assign the new temporary "category_new" field. This will be
+            # renamed later
+            product_page_category.category_new = new_category
+            product_page_category.save()
```
Pomax, line 19: It feels like this is not doing what it should be thanks to M2M...

KalobTaulien:
@Pomax This is ready for re-review. I pulled down staging and ran the migrations together and it works now. I ran it 3 times in a row because there were inconsistencies in the results I was getting, but it's working as expected now. 

I'd appreciate if you could also give this a run using the staging database please

Pomax:
Much better, but it looks like some of the category settings are not getting carried over. Compare the ordering, and entries, 

Foundation site:

image

This branch:

image

Pomax (changes requested):
so close O_o

KalobTaulien:
@Pomax This is working for me now. I pulled the staging database, ran the migrations, and got the same categories for both, all in the right order. Could you give this another try?

Localhost:
image

Foundation staging site:
image

Pomax (approved):
hurray!

PR #6323, opened 2021-03-05
Fix `migrate_cloudinary` management command: Now uses .specific()

### LINKED ISSUE #6322: Fix cloudinary migration command
Previously when running `inv manage migrate_cloudinary` it was resetting the un-specific model data because it was using a parent class. 

We need to adjust this so it uses the specific classes and doesn't reset the data on each page.

KalobTaulien:
Closes #6322

Acceptance criteria:
- [ ] When models are saved, the cloudinary image should be copied to the wagtail image field. 
- [ ] When images are copied, the data for SoftwareProductPage and GeneralProductPages should be left alone (unchanged) and _only_ the `image` field should be updated. 
- [ ] Confirm the data in the PNI product pages don't revert back to default data 

Testing steps:
- [ ] Enable cloudinary in your local settings 
- [ ] Pull down staging database
- [ ] Confirm you have staging data. Take note of which fields are set to what value (or for testing, set all boolean fields to checked, all choicefields to "yes", all text fields to "DEV TEST" this way you can see if anything was changed easily) 
- [ ] Run `inv manage migrate_cloudinary` 
- [ ] Double check the Wagtail images exist but all the other data is left intact.


network-api/networkapi/tests.py
```diff
@@ -465,3 +467,136 @@ def test_software_page_export_fields(self):
...
+            software_product_page.image.filename,
+            f'{software_product_page.cloudinary_image.public_id}.{software_product_page.cloudinary_image.format}',
+        )
+        self.assertNotEqual(
+            general_product_page.image.filename,
+            f'{general_product_page.cloudinary_image.public_id}.{general_product_page.cloudinary_image.format}',
+        )
+
+        # Run the management command which will convert Cloudinary images to Wagtail images.
+        call_command("migrate_cloudinary")
```
Pomax, line 586: before we run this, let's convert `software_product_page` and `general_product_page` to dicts so we can use those as references later

network-api/networkapi/tests.py
```diff
@@ -465,3 +467,136 @@ def test_software_page_export_fields(self):
...
+        self.assertTrue(hasattr(general_product_page, 'image'))
+        # Assert General Product Page "specific" fields have not changed
+        self.assertEqual(general_product_page.offline_capable, 'Yes')
+        self.assertEqual(
+            general_product_page.offline_use_description,
+            'Although it is unclear how offline capabilities work'
+        )
+        # Assert Software Product Page "specific" fields have not changed
+        self.assertEqual(software_product_page.recording_alert, 'Yes')
+        self.assertEqual(software_product_page.handles_recordings_how, 'Sample software recording description')
```
Pomax, line 602: And then here, we can iterate through the previous created dicts, where we verify every field with an skip for the image fields.

Pomax (changes requested):

network-api/networkapi/tests.py
```diff
@@ -465,3 +467,136 @@ def test_software_page_export_fields(self):
         self.assertIn("Host controls", export_fields)
         self.assertIn("Easy to learn and use", export_fields)
         self.assertIn("Easy to learn and use help text", export_fields)
+
+
+class TestCloudinaryMigration(TestCase):
+
+    def setUp(self):
+        pni_homepage = BuyersGuidePageFactory.create(
```
Pomax, line 475: let's `from django.conf import settings` and then make setUp return if `USE_CLOUDINARY` isn't `True`.

network-api/networkapi/tests.py
```diff
@@ -465,3 +467,136 @@ def test_software_page_export_fields(self):
...
+            product_url='http://example.com/general-percy',
+            price=999,
+            worst_case='Visual regression fails',
+            # software product fields
+            handles_recordings_how='Sample software recording description',
+            recording_alert='Yes',
+        )
+
+    def test_fields_remain_untouched_and_wagtail_image_is_changed(self):
+        # Test fields before the command is run
```
Pomax, line 556: let's `from django.conf import settings` and then make this test log a string along the lines of `not running test_fields_remain_untouched_and_wagtail_image_is_changed: USE_CLOUDINARY is not set` and then exit the function if `settings.USE_CLOUDINARY` isn't `True`.

Pomax (changes requested):
Let's do some dict magic:
```
from django.forms.models import model_to_dict

    ....

    def test...(self):
        ...
        software_page = Page.objects.get(pk=self.software_product_page.id)
        software_product_page = software_page.specific
        software_product_dict = {
            k: v
            for k, v in model_to_dict(software_product_page).items()
            if k not in model_to_dict(software_page)
        }

        # etc.
```

This should give dicts with only the specific-page relevant fields, so we can blind-compare them

network-api/networkapi/tests.py
```diff
@@ -465,3 +470,164 @@ def test_software_page_export_fields(self):
...
+        self.assertTrue(hasattr(new_general_product_page, 'image'))
+        # Assert General Product Page "specific" fields have not changed
+        self.assertEqual(new_general_product_page.offline_capable, 'Yes')
+        self.assertEqual(
+            new_general_product_page.offline_use_description,
+            'Although it is unclear how offline capabilities work'
+        )
+        # Assert Software Product Page "specific" fields have not changed
+        self.assertEqual(new_software_product_page.recording_alert, 'Yes')
+        self.assertEqual(new_software_product_page.handles_recordings_how, 'Sample software recording description')
```
Pomax, line 615: Do we still need these lines, given the dict compare?

KalobTaulien: Nope. Will remove!

Pomax (approved):
R+ as long as we make sure the migration numbering is correct wrt the master branch because obviously =D

PR #6342, opened 2021-03-09
Add support for Scout APM

tomusher:
Adds Scout APM package and enables it as a Django app, pulling settings from environment variables.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -8,6 +8,8 @@ airtable-python-wrapper==0.13.0
     # via wagtail-airtable
 anyascii==0.1.7
     # via wagtail
+asgiref==3.3.1
+    # via scout-apm
 beautifulsoup4==4.8.2
     # via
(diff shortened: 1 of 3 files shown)


requirements.txt
```diff
@@ -182,11 +197,12 @@ tqdm==4.15.0
     # via wagtail-inventory
 unidecode==1.1.1
     # via wagtail
-urllib3==1.25.9
+urllib3[secure]==1.25.9
```
mofodevops, line 200:
```
urllib3[secure]==1.25.9
```
Probably nothing, but will this change the functionality of other dependent packages?

benhohner, line 200:
```
urllib3[secure]==1.25.9
```
Probably nothing, but will this change the functionality of other dependent packages?

benhohner (approved):
R+ if comment above doesn't apply and API key added, and tests pass

PR #6345, opened 2021-03-09
Advanced table block (use with care)

### LINKED ISSUE #6250: Advanced Publication Table Block
We need a more advanced table block that can hold information such as:

- [x] Links 
- [x] Lists 
- [x] Centering text in a cell 
- [x] Bold 
- [x] Italics 

Other features it will need are:

- [ ] Merging cells left and right 
- [ ] Merge cells top and bottom
- [x] Heading styling (first row and left column)
- [ ] ~Accessibility title~ (same as below)
- [x] Optional table caption (richtext)

KalobTaulien:
Closes #6250 

**Link to sample test page**: https://foundation-s-6250-advan-e20z7x.herokuapp.com/en/publication-page-with-child-article-pages/fixed-title-article-page/


sabrinang (commented):
So awesome to see the links, lists, centering, styling variations available in a table!

**Feedback for advance table block**
- can we reduce the cell padding on the sides so it is similar to the simple table and we can fit more text in the cell?
- when bold is applied, it isn't as bold as the bold seen in the simple table
image


- Column width: Does anything change when it changes to “1”? It looks the same as “0” default so that might be confusing with the help text:
_“Enter the number of extra cell columns you want to merge together. Merging a cell column will expand a cell to the right.”_
- when cells are merged, can the alignment still be left aligned with the initial cell?
image

- is it possible to align the checkbox in wagtail with the other fields?
image

network-api/networkapi/wagtailcustomization/static/css/admin.css
```diff
@@ -53,8 +53,3 @@ fieldset ul li div label {
 #wagtail .wrapper input[type="checkbox"] {
   vertical-align: middle;
 }
-
-#wagtail .wrapper input[type="checkbox"]:before {
```
mmmavis, line 57: @Pomax do you remember what this was for? I found the PR but couldn't recall the details. Was this to fix the user group checkbox misalignment? If so, checkboxes look fine without these rules now so we should be able to safely remove this part of SCSS.

image

Pomax: that's what we had them for, yes. If they're no longer needed, let's remove them \o/

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
```diff
@@ -0,0 +1,51 @@
...
+      {% else %}
+        {# For every other row that's not a header... #}
+        <tr>
+          {% for cell in row.value %}
+            {# If the first cell is suposed to be blue; when "column" is checked #}
+            {% if self.column %}
+              <td
+                {% if cell.value.column_width %}colspan="{{ cell.value.column_width }}"{% endif %}
+                class="{% if forloop.first and self.column and not cell.value.column_width %}blue{% endif %} {% if cell.value.centered_text %}text-center{% endif %}"
+                width="20%"
```
mmmavis, line 33: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td

`width` on `<td>` is obsolete. Let's use CSS instead.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
```diff
@@ -0,0 +1,51 @@
...
+        <thead>
+          {% for cell in row.value %}
+            <th
+              scope="col"
+              {% if cell.value.column_width %}colspan="{{ cell.value.column_width }}"{% endif %}
+              class="blue {% if cell.value.centered_text %}text-center{% endif %}"
+            >
+              {{ cell.value.content|richtext }}
+            </th>
+          {% endfor %}
```
mmmavis, line 22: missing `<tr>`

```suggestion
          <tr>
          {% for cell in row.value %}
            <th
              scope="col"
              {% if cell.value.column_width %}colspan="{{ cell.value.column_width }}"{% endif %}
              class="blue {% if cell.value.centered_text %}text-center{% endif %}"
            >
              {{ cell.value.content|richtext }}
            </th>
          {% endfor %}
          </tr>
```

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
```diff
@@ -0,0 +1,51 @@
...
+
+<div class="article-table-block table-responsive">
+  <table class="table">
+    {% if self.caption %}
+      <caption>{{ self.caption }}</caption>
+    {% endif %}
+
+    {# Loop through every row #}
+    {% for row in self.table %}
+      {# If the "header" is checked, the first row will be a <thead> with a blue background #}
```
mmmavis, line 11: In case we decide to change the highlight colour in the future, could we update the comment from "blue" background to "highlighted" background?

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
```diff
@@ -0,0 +1,51 @@
...
+            >
+              {{ cell.value.content|richtext }}
+            </th>
+          {% endfor %}
+        </thead>
+      {% else %}
+        {# For every other row that's not a header... #}
+        <tr>
+          {% for cell in row.value %}
+            {# If the first cell is suposed to be blue; when "column" is checked #}
```
mmmavis, line 28: Typo in "suposed" :) 

Could we change "blue" to be "highlighted" for the same reason above? Thanks.

source/sass/components/article-page.scss
```diff
@@ -322,6 +357,11 @@
     @include set-text-size(16px, 24px);
     font-family: $font-family-sans-serif;
     font-weight: 600;
+
+    .rich-text * {
+      color: $white !important;
```
mmmavis, line 362: `!important` 🙈 

I'm gonna file a separate ticket as we should really avoid using `!important` (I know there are a few occurrences in the codebase already 😬 ).

source/sass/components/article-page.scss
```diff
@@ -312,6 +312,41 @@
...
+      p {
+        max-width: initial;
+      }
+
+      b {
+        font-weight: bold;
+      }
+
+      a {
+        color: $dark-blue;
```
mmmavis, line 331: @sabrinang what style should links on article page content be? Do they appear with underline? How about hover state style?

source/sass/components/article-page.scss
```diff
@@ -312,6 +312,41 @@
...
+
+      ol,
+      ul,
+      p {
+        margin-bottom: 12px;
+      }
+
+      ol > li::marker {
+        font-weight: normal;
+        font-size: 16px;
```
mmmavis, line 342: I see that on line 234-236 we have

```scss
ol > li::marker {
    font-weight: bold;
  }
```
which sets ANY ordered list's marker on Article page to have bold weight. Can we just change that value from `bold` to `normal`? Or do we specifically want markers in **tables** on Article page to appear differently (that is, to have `normal` weight instead of `bold`)?

mmmavis (changes requested):

source/sass/components/article-page.scss
sabrinang: @mmmavis I'm good with same anchor link styles we have in style guide already _(so underline on hover)_

KalobTaulien: I believe these are the same in both tables already.

source/sass/components/article-page.scss
KalobTaulien: Not sure how this got in here to be honest. I'm going to remove it.

source/sass/components/article-page.scss
KalobTaulien: Yeah :/ this gets into some nasty specificity. So either an !important declaration _or_ it's a crazy long selector (and maybe multiples). I'm happy to go with whichever is easier for us to maintain. I opted for !important to cut down the amount of code we ship in our css file ¯\_(ツ)_/¯

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
KalobTaulien: Good suggestion. Absolutely, yes, let's do this.

danielfmiranda (approved):

mmmavis (approved):
Thanks for making the changes 🎉

Issue #6344, opened 2021-03-09
Fix the code around tags, to make the site responsive again


Pomax:
We should find our taggit hotspots, and fix them:

![image](https://user-images.githubusercontent.com/177243/110510149-c22c8480-80b7-11eb-9bad-b9820fc0a38b.png)

- [x] find the places we're performing tag logic: foundation meta mixin
- [x] figure out how we can optimize that code: https://github.com/mozilla/foundation.mozilla.org/pull/6346
- [x] add Tags to our settings menu: see above PR

PR #6346, opened 2021-03-09
make the default social share image a one-time lookup

Pomax:
Related PRs/issues #6344 

- adds `Tags` to the wagtail settings menu, as we did not surface those anywhere before
- changes the mixin code from a flood of "perform a tag lookup in every constructor" + "perform an image lookup in every page's metadata lookup", to a single "cache the social share image, once, ever"
- adds an image with tag "social share image" to the db during the `load_fake_data` run, so that pages that rely on the social share image for their cards (e.g. Blog cards) aren't suddenly without content


network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_metadata.py
```diff
@@ -33,23 +37,40 @@ def get_meta_description(self):
...
+        # get the tag with name "social share image"
+        default_share_tag_name = 'social share image'
+        tag, create = Tag.objects.get_or_create(name=default_share_tag_name)
+        default_social_share_tag = tag
+
+        # then find an image in the CMS that uses that tag:
+        default_social_share_image = None
+        default_social_share_images = Image.objects.filter(tags=default_social_share_tag)
+        if default_social_share_images.count() > 0:
+            default_social_share_image = default_social_share_images.first()
```
KalobTaulien, line 54:
```suggestion
        find_default_social_share_image = Image.objects.filter(tags=default_social_share_tag).first()
        if find_default_social_share_image:
            default_social_share_image = find_default_social_share_image
```
☝️ This way the .filter() will stop searching through the DB when it finds the first instance

network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_metadata.py
```diff
@@ -33,23 +37,40 @@ def get_meta_description(self):
...
+
     def get_meta_image(self):
+        # If we have a local social share image, use that
         if self.search_image:
             return self.search_image
 
+        # If note, walk up our ancestor chain and use the first social
```
KalobTaulien, line 61:
```suggestion
        # If not, walk up our ancestor chain and use the first social
```

KalobTaulien (approved):
A couple of things to consider changing, but not worth me blocking you.

network-api/networkapi/wagtailpages/pagemodels/mixin/foundation_metadata.py
Pomax: I rewrote it _from_ that because it was in a try/except. What happens when you call `first()` on an empty queryset?

benhohner:
@Pomax looks like the blog images aren't displaying in Percy, is this normal?

Pomax:
@ben nope that was a genuine error, the old code relied on magic behaviour if the `social share image` tag did not exist, where it would select "the first image in the database" instead of "the first image in the database with the social share tag" if the tag didn't exist, leading to the mixin never returning `None` despite the code clearly being written with the intention to return `None`.  I've updated our `load_fake_data` command to explicitly create an image with that tag now, so that should now yield the same views as before, but because it should, rather than by accident =)

KalobTaulien (approved):

PR #6349, opened 2021-03-10
Make PNI test data use fake product images

### LINKED ISSUE #5905: PNI follow-up: make PNI test data product images use our "fake product" images
Follow-up to https://github.com/mozilla/foundation.mozilla.org/issues/5874: it would be a nice if we could use the old product images, which means:

- [ ] figure out how to get the "product images" in our static asset dir into the CMS as wagtail images
- [ ] switch from `ImageFactory()` to picking a random (using `choice`) wagtail `Image` as target

KalobTaulien:
Closes #5905 

### Acceptance criteria:
- [x] figure out how to get the "product images" in our static asset dir into the CMS as wagtail images
- [x] switch from ImageFactory() to picking a random (using choice) wagtail Image as target

This will now create proper Wagtail Image objects from the images in our `products/` directory when loading new data and throw the images in a `pni products` Collection.

It will create 5 Wagtail Images, and then when creating PNI product pages it'll randomly select one of the 5 images.

There's also a new somewhat large utility function to create wagtail images on the fly. It's documented but doesn't have unit tests yet. I'll create a backlog ticket to write unit tests for that function.


network-api/networkapi/management/commands/load_fake_data.py
```diff
@@ -68,6 +69,15 @@ def handle(self, *args, **options):
         social_share_tag, created = Tag.objects.get_or_create(name='social share image')
         images[0].tags.add(social_share_tag)
 
+        # Create one PNI product for every image we have in our media folder
+        product_images = ['babymonitor.jpg', 'drone.jpg', 'nest.jpg',
+                          'teddy.jpg', 'echo.jpg']
+        for image in product_images:
+            create_wagtail_image(
+                f'/app/network-api/media/images/placeholders/products/{image}',
```
KalobTaulien, line 77: `/app/...` comes from Docker since that's the standard way to setup a docker project and we follow that rule, thus the hard coded image path.

KalobTaulien:
CI failed. Will look at this on Monday!

network-api/networkapi/buyersguide/factory.py
```diff
@@ -253,7 +253,11 @@ class Meta:
 
     @post_generation
     def set_image(self, create, extracted, **kwargs):
-        self.image = ImageFactory()
+        """Creates a new Wagtail Image for each PNI Product."""
+
+        pni_images = Image.objects.filter(collection__name='pni products')
```
Pomax, line 258: yesss

network-api/networkapi/management/commands/load_fake_data.py
Pomax: We can use some `os.path` magic in combination with `__file__` here, similar to what we did in the cloudinary migration script

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -1,10 +1,18 @@
+import ntpath
```
Pomax, line 1: Is there an issue to take this out of the other code path(s)?

Pomax (changes requested):

Pomax:
There's also some CI failures that will need to be addressed.

network-api/networkapi/wagtailpages/utils.py
KalobTaulien: We use this to extract the base filename from an image source if an explicit name is not provided: `image_name = ntpath.basename(img_src)`. This just makes sure the Wagtail Image has a "title" that matches the filename. The title can be anything but if we don't know the context of the image being created, it's probably safest to fallback to the filename

network-api/networkapi/management/commands/load_fake_data.py
KalobTaulien: Done, good suggestion! We needed this anyway because CI doesnt run seem to have the same paths as our local docker setup (ie. /app/ isnt accessible, or something something that made CI fail)

Pomax (approved):

Issue #5715, opened 2020-11-24
Unify BlogAuthor and ArticleAuthor into an Author


Pomax:
If BlogAuthor and ArticleAuthor are the same thing, just duplicated code, we should unify those into a PageAuthor or something and then use that everywhere instead of duplicating functionality.

Pomax:
downgrading the estimate to a 2, but I might bump it down to a 1 depending on how long migration work takes.

Pomax:
downgrading to 1, Django was able to identify that this was a simple rename without needing migration surgery

PR #6353, opened 2021-03-11
rename BlogAuthor to ContentAuthor

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/5715 by renaming the `BlogAuthor` model to `ContentAuthor` instead, and moving it into its own file. Django was able to turn this into the smallest migration possible, being a single model rename instruction.

The `BlogAuthors` and `ArticleAuthors` have been left as is, but updated so that they both use `ContentAuthor` as their actual author model.

I've assigned this to multiple reviewers just to make sure enough people gave it a testing to make sure we're confident this works fine. I would recommend running first creating a db on `master`, taking note of the authors on some blog/publication/article pages, then switching branches, running `inv migrate` and verifying those authors are still there. In the CMS, the snippet menu should now have `ContentAuthor` with the same list as `BlogAuthor` on master.


network-api/networkapi/wagtailpages/models.py
```diff
@@ -99,7 +102,7 @@
     BanneredCampaignPage,
     BlogIndexPage,
     BlogPage,
-    BlogAuthor,
+    ContentAuthor,
```
mmmavis, line 105: move this down so the list is in alphabetical order?

network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -90,10 +90,10 @@ class Meta:
     title_text = Faker('sentence', nb_words=3, variable_nb_words=False)
 
 
-class BlogAuthorFactory(DjangoModelFactory):
+class ContentAuthorFactory(DjangoModelFactory):
```
mmmavis, line 93: Maybe as a followup? 

Since `ContentAuthor` is being used on both fake Blog Pages and fake Article pages, can we move this to its own file so it's not so embedded in blog page generation?

mmmavis (changes requested):
I followed the testing instructions and everything turned out as expected.

Left two small code change suggestions.

network-api/networkapi/wagtailpages/factory/blog.py
Pomax: good call, I filed https://github.com/mozilla/foundation.mozilla.org/issues/6359

mmmavis: Awesome!

mmmavis (approved):
Lovely!

KalobTaulien (changes requested):
Works well. I tested this with the blog page, article page, and the snippets area. 👌 Just need to bump the migration file to `0048` it seems

KalobTaulien (approved):

PR #6377, opened 2021-03-15
Management command to generate pni products

### LINKED ISSUE #6373: Create a management command to pre-create PNI image renditions and PNI admin renditions.
When turning off cloudinary, we've run into a problem where 287 PNI images are trying to be resized all at once, and we hit out 30s timeout on Heroku. 

We need a management command that will:
- [ ] Look for existing PNI products and programatically generate wagtail image renditions. This should include:
  - [ ] /privacynotincluded/ image, and 
  - [ ] All of the images `pni images` collection in the Wagtail admin to images load fast in the backend as well.

KalobTaulien:
Closes #6373 

To run this:
`inv manage create_pni_image_renditions` 

It will print a status during creation. 

To test this:
Go into your shell, and run:
```
from networkapi.wagtailpages.pagemodels.products import ProductPage

page = ProductPage.objects.last().specific() 
page.image.renditions.all() # Should return a queryset of image rendition objects.
```

### DIFF
--- a/network-api/networkapi/management/commands/create_pni_image_renditions.py
+++ b/network-api/networkapi/management/commands/create_pni_image_renditions.py
@@ -1,0 +1,30 @@
+from django.core.management.base import BaseCommand
+
+from networkapi.wagtailpages.pagemodels.products import ProductPage
+
+
+class Command(BaseCommand):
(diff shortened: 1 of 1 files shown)


Pomax (approved):
let's get this ball rolling. Some more.

PR #6382, opened 2021-03-16
PNI - Add 'excluded_categories' picker

### LINKED ISSUE #6025: [PNI] add a category-picker to the BuyersGuidePage (homepage) that is used to exclude products from the listing
Have categories that purely organizational categories (e.g. not for showing the category nav) and can be used to exclude products from showing in `catalog.html` product sets.

Use case: have a "valentine's day" category that shows in the nav as well as a "sex toys" category that doesn't show in the nav, assign all the sex toys to both categories, and then be able to tell the homepage to exclude listing any product that has "sex toys" as (one of) its categories.

- [x] allow category creation in the CMS
- [x] allow categories to specify whether they show up in the nav menu
- [x] add an `excluded_categories` field and `snippetChooserPanel("excluded_categories")` to the BuyersGuidePage model, that can be populated with categories that should be excluded from the homepage. (e.g adding "sex toys" would remove all products in that category, even if those products are also in other categories, from the list on the PNI homepage, but not from PNI category pages)
- [x] update the queryset that gets used for `catalog.html` so that products that are in any of the categories in `excluded_categories` get filtered out before the template gets rendered

mmmavis:
Closes #6025


network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1343,10 +1371,11 @@ def get_context(self, request, *args, **kwargs):
                 self.cutoff_date,
                 authenticated,
                 key,
-                ProductPage.objects.all()
+                ProductPage.objects.exclude(product_categories__in=self.excluded_categories.all().values_list('id'))
```
mmmavis, line 1374: @Pomax this PR is still WIP but can I get a hand for this line? I don't know how to properly exclude ProductPages that fall under any categories we have in `excluded_categories`. This line does nothing and I also tried it without `.values_list('id')` but got 
```
ValueError at /en/privacynotincluded/

Cannot use QuerySet for "ExcludedCategories": Use a QuerySet for "ProductPageCategory".
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1343,10 +1371,11 @@ def get_context(self, request, *args, **kwargs):
...
             )
 
-        context['categories'] = BuyersGuideProductCategory.objects.filter(hidden=False)
+        context['categories'] = BuyersGuideProductCategory.objects.exclude(
+            excludedcategories__in=self.excluded_categories.all())
```
Pomax, line 1378: We still want to show these categories in the nav, we just don't want to show products from those categories in the main listing.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1149,6 +1149,23 @@ class Meta:
         verbose_name = "General Product Page"
 
 
+class ExcludedCategories(Orderable):
+    """This allows us to select one or more blog authors from Snippets."""
+
+    page = ParentalKey("wagtailpages.BuyersGuidePage", related_name="excluded_categories")
+    excluded_category = models.ForeignKey(
```
Pomax, line 1156: let's make this `category`

network-api/networkapi/wagtailpages/pagemodels/products.py
Pomax: Note that this is now a "through" relation, so instead of `product_categories`, you need to test using `product_categories__category`, and the same goes for the exclude categories:

```
        ...
        products = cache.get(key)

        exclude_ids = [excats.excluded_category.id for excats in self.excluded_categories.all()]

        if not kwargs.get('bypass_products', False) and products is None:
            products = get_product_subset(
                self.cutoff_date,
                authenticated,
                key,
                ProductPage.objects.exclude(product_categories__category__id__in=exclude_ids)
            )

        ...
```

Pomax: although if you rename that field from `excluded_category` to just `category`, that will change the above code to:
```
exclude_ids = [excats.category.id for excats in self.excluded_categories.all()]
```

network-api/networkapi/buyersguide/views.py
```diff
@@ -15,6 +15,7 @@
 from rest_framework.permissions import AllowAny, IsAuthenticated
 from rest_framework.response import Response
 
+from networkapi.wagtailpages.pagemodels.products import BuyersGuidePage
```
mmmavis, line 18: not sure how to test my changes in this file

network-api/networkapi/buyersguide/templates/buyersguide/catalog.html
```diff
@@ -139,6 +139,12 @@ <h1 class="h1-heading text-center mt-0">{{ page.header }}</h1>
               <div class="product-company body-small">{{product.company}}</div>
               <div class="product-name body">{{product.title}}</div>
             </a>
+            {% comment %}
+            FOR PR TESTING PURPOSES. THE NEXT THREE LINES OF CODE BELOW ARE TO BE REMOVED.
+            {% endcomment %}
+            {% for cat in product.product_categories.all %}
+            {{ cat }} <br>
+            {% endfor %}
```
mmmavis, line 147: TO BE REMOVED

network-api/networkapi/wagtailpages/pagemodels/products.py
mmmavis: Ahhh I see the differences now. Thanks for the explanation.

network-api/networkapi/buyersguide/templates/buyersguide/catalog.html
Pomax: it would be hilarious if we accidentally left this in =D

network-api/networkapi/buyersguide/views.py
Pomax: We don't use this file anymore, it's part of the old PNI app, so we only need changes to the wagtailpages code

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1192,13 +1209,24 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
     def get_banner(self):
         return self.hero_image
 
+    def get_excluded_categories(self):
+        return [
+            {"name": category.category.name} for category in self.excluded_categories.all()
```
Pomax, line 1214: just to avoid confusion, let's make this use the same name as later in the code:
```suggestion
            {"name": excats.category.name} for excats in self.excluded_categories.all()
```

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1285,13 +1313,15 @@ def categories_page(self, request, slug):
...
         products = cache.get(key)
+        exclude_cat_ids = [excats.category.id for excats in self.excluded_categories.all()]
 
         if products is None:
             products = get_product_subset(
                 self.cutoff_date,
                 authenticated,
                 key,
                 ProductPage.objects.filter(product_categories__category__in=[category])
+                                   .exclude(product_categories__category__id__in=exclude_cat_ids)
```
Pomax, line 1324: pretty sure flake8 doesn't like this much indent =D

Pomax (changes requested):
some notes, and it looks like CI has some complaints, too. But it's pretty close!

network-api/networkapi/buyersguide/views.py
mmmavis: Alright! Will revert these changes!

Pomax (changes requested):
Ah, noticed one last thing: because we're caching products, we want to make sure that when the homepage gets saved, we clear the cache, and changes to the "exclude categories" means we shouldn't be serving cache with those products in them anymore.

This should do the trick:

```
@hooks.register('after_publish_page')
@hooks.register('after_unpublish_page')
@hooks.register('after_delete_page')
def invalidate_cache(request, page):
    """
    When a product is created, or updated, or deleted, invalidate the product cache.
    """
    if isinstance(page, ProductPage) or isinstance(page, BuyersGuidePage):
        cache.clear()
```

Which adds the `or isinstance(page, BuyersGuidePage)` bit to make sure that we invalidate not just on product changes, but also on homepage changes.

mmmavis:
@Pomax does this existing code do the trick? 😃 https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/wagtail_hooks.py#L106-L114

Pomax:
haha, it does, Kalob ended up touching the same hooking code =D

Pomax (approved):

mmmavis:
🎉

PR #6384, opened 2021-03-16
updated image size to reduce number renditions

### LINKED ISSUE #6378: Reduce the number of PNI images in the /privacynotincluded/ template
Inside of `network-api/networkapi/buyersguide/templates/buyersguide/catalog.html` is a ` <picture class="product-thumbnail">` element. 

Inside of that picture element is a bunch of `<source>` elements. 

One of them is a 360x360 image rendition. The other is a 370x370 image rendition. We need these to be exactly double (or half, if you will) other image so we have a high res and low res image in each <source> 

I think we should turn:
```django
{% image product.image fill-185x185 as img_1x %}
{% image product.image fill-370x370 as img_2x %}
```

Into 
```
{% image product.image fill-180x180 as img_1x %}
{% image product.image fill-360x360 as img_2x %}
```

This way we can leverage having 2 360x360 images and cut the number of renditions down by one per image (or roughly reduce the image rendition computation cost of the page by about 290) 

Message @KalobTaulien for clarification on this ticket, if needed.

danielfmiranda:
Closes #6378 

## Checklist

**Changes in Models:**


mmmavis (changes requested):
We aren't updating packages in this PR so there shouldn't be changes introduced in package-lock.json. Can we remove it from the commit, please?

Pomax (changes requested):
Note that with https://github.com/mozilla/foundation.mozilla.org/issues/6395 thrown into the mix, we don't just need to update one set's dimensions, we should take out the other three source sets as well.

network-api/networkapi/buyersguide/templates/buyersguide/catalog.html
```diff
@@ -114,10 +114,10 @@ <h1 class="h1-heading text-center mt-0">{{ page.header }}</h1>
               <source
                 media="(max-width: 414px)"
```
Pomax, line 115: note that as we're reducing the source to a universal "same x1 and x2 at all sizes", we don't need this `media` attribute anymore, and the other `<source>` collections still need to be removed

network-api/networkapi/buyersguide/templates/buyersguide/catalog.html
```diff
@@ -114,10 +114,10 @@ <h1 class="h1-heading text-center mt-0">{{ page.header }}</h1>
...
                 {% if USE_CLOUDINARY %}
-                  srcset="{% cloudinary_url product.cloudinary_image quality=50 fetch_format="auto" crop="fit" width=185 %} 1x, {% cloudinary_url product.cloudinary_image quality=50 fetch_format="auto" crop="fit" width=370 %} 2x"
+                  srcset="{% cloudinary_url product.cloudinary_image quality=50 fetch_format="auto" crop="fit" width=180 %} 1x, {% cloudinary_url product.cloudinary_image quality=50 fetch_format="auto" crop="fit" width=360 %} 2x"
                 {% else %}
-                  {% image product.image fill-185x185 as img_1x %}
-                  {% image product.image fill-370x370 as img_2x %}
+                  {% image product.image fill-180x180 as img_1x %}
+                  {% image product.image fill-360x360 as img_2x %}
```
Pomax, line 120: we want 360 for the x1 size, and 720 for the x2 size

danielfmiranda:
HI Everyone,

I am closing this PR as I have mistakenly added package-lock.json to the list of changes.
 
So instead of keeping this one, which will show a history of adding, deleting, and then re-adding the package-lock.json file, 
I am going to close this PR and create a new one, which will then just show the changes made to the desired file which is "**catalog.html**"

mmmavis:
Thanks for the update @danielfmiranda .

Don't worry about "polluting" the commit history next time as the commits are squashed when we hit the PR merge button. 😉  https://github.com/mozilla/foundation.mozilla.org/commits/master

PR #6391, opened 2021-03-17
Localized PNI categories and products

### LINKED ISSUE #6064: wagtail-localize: Test localized category reassignment (product/blog)
If we localize PNI categories, as well as blog categories, what happens to all the "other" locales? E.g. update a product category in the English locale, does that product still live in the "old" category for the French page instance?

(and the same for blog posts, but that's less pressing as we rarely, if ever, change categories for those)

KalobTaulien:
Closes #6064 

Steps for user testing:
1. Create a new env. 
2. Create a new locale (Settings -> Locales -> Create a new one; ie. French) 
3. Create a new Buyers Guide Product Category snippet (I made a brand new one that was easier to track). And then create a translated version of that same category (ie. One English and One French version). 
4. Edit an existing English PNI product and change the category to your new English category you made in step 3. 
5. Translate the page and edit the French version - change the category to your French category you made in step 3. 
6. Go to http://localhost:8000/en/privacynotincluded/ to see your English product. Your English category should show up too. 
7. Go to http://localhost:8000/fr/privacynotincluded/ to see your French product. The French category should show up too. 
8. Click the category at the top of the page to filter your products. Make sure only your English product shows up in the English category, and only the French product shows up in the French category.

☝️ That's a lot of steps. Ping me on slack if anyone needs assistance. 

Requesting Theo for user testing review and Pomax for code review.

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/products.py
@@ -1252,6 +1252,13 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
         SynchronizedField('slug'),
     ]
 
+    def get_language_code(self, request):
+        """Accepts a request. Returns a language code (string) if there is one. Falls back to English."""
+        default_language_code = 'en'
+        if hasattr(request, 'LANGUAGE_CODE'):
+            default_language_code = request.LANGUAGE_CODE
+        return default_language_code
+
     @route(r'^about/$', name='how-to-use-view')
     def about_page(self, request):
         context = self.get_context(request)
@@ -1326,6 +1333,7 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
     def categories_page(self, request, slug):
         context = self.get_context(request, bypass_products=True)
         slug = slugify(slug)
+        language_code = self.get_language_code(request)
 
         # If getting by slug fails, also try to get it by name.
         try:
@@ -1342,7 +1350,10 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
                 self.cutoff_date,
                 authenticated,
                 key,
-                ProductPage.objects.filter(product_categories__category__in=[category])
+                ProductPage.objects.filter(
+                    product_categories__category__in=[category],
+                    locale=Locale.objects.get(language_code=language_code)
+                )
             )
 
         context['category'] = category.slug
@@ -1386,9 +1397,15 @@ class BuyersGuidePage(RoutablePageMixin, FoundationMetadataPageMixin, Page):
 
     def get_context(self, request, *args, **kwargs):
         context = super().get_context(request, *args, **kwargs)
+        language_code = self.get_language_code(request)
 
         authenticated = request.user.is_authenticated
-        key = 'home_product_dicts_authed' if authenticated else 'home_product_dicts_live'
+        key = (
+            f"home_product_dicts_authed_{language_code}"
+            if authenticated
+            else f"home_product_dicts_live_{language_code}"
+        )
+
         products = cache.get(key)
 
         if not kwargs.get('bypass_products', False) and products is None:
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1252,6 +1252,13 @@ def get_banner(self):
         SynchronizedField('slug'),
     ]
 
+    def get_language_code(self, request):
+        """Accepts a request. Returns a language code (string) if there is one. Falls back to English."""
+        default_language_code = 'en'
```
Pomax, line 1257: grab this from the settings?  https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/settings.py#L413

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1386,20 +1397,30 @@ def get_sitemap_urls(self, request):
...
+        language_code = self.get_language_code(request)
 
         authenticated = request.user.is_authenticated
-        key = 'home_product_dicts_authed' if authenticated else 'home_product_dicts_live'
+        key = (
+            f"home_product_dicts_authed_{language_code}"
+            if authenticated
+            else f"home_product_dicts_live_{language_code}"
+        )
+
```
Pomax, line 1408: feels like we can keep the original, then follow up with `key = f'{key}_{language_code}'`

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1342,7 +1350,10 @@ def categories_page(self, request, slug):
                 self.cutoff_date,
                 authenticated,
                 key,
-                ProductPage.objects.filter(product_categories__category__in=[category])
+                ProductPage.objects.filter(
+                    product_categories__category__in=[category],
+                    locale=Locale.objects.get(language_code=language_code)
+                )
```
Pomax, line 1356: we're doing this a few times, it might be worth investigation the possibility of `ProductPage.objects.locale(language_code).filter(...)`, or at the very least create a prefilter function along the lines of:

```python
def get_localized(ModelType, request):
    language_code = ... or settings.LANGUAGE_CODE
    return ModelType.objects.filter(locale=Locale.objects.get(language_code=language_code))
```

so that this code becomes:
```python
products = get_product_subset(
    ...,
    get_localized(ProductPage, request).filter(product_categories__category__in=[category])
)
```

Pomax (dismissed):

KalobTaulien:
I've run into a bit of a hiccup here. My instructions above are _wrong_. We need Alias Pages (ie. French pages) to be able to select their own snippets. Otherwise what we're doing is editing an English page and changing it's category - which may not reflect the exact behaviour we want out of wagtail-localize. 

Ideally, wagtail-localize would provide a way for us to select Snippets in the translation-view (Alias Page). Currently, I can't seem to enable that behaviour. 

I'm moving this ticket to blocked until I can get some more information around this.

TheoChevalier:
> I've run into a bit of a hiccup here. My instructions above are _wrong_. We need Alias Pages (ie. French pages) to be able to select their own snippets. Otherwise what we're doing is editing an English page and changing it's category - which may not reflect the exact behaviour we want out of wagtail-localize.

Ah, I was going to ask about this!

> 
> Ideally, wagtail-localize would provide a way for us to select Snippets in the translation-view (Alias Page). Currently, I can't seem to enable that behaviour.

> 
> I'm moving this ticket to blocked until I can get some more information around this.
> 
> Edit: On second thought, there's an easier way around this.
> 
> We can select multiple categories (ie. English Entertainment and French Entertainment) and then filter products based on their categories. Then auto-aliasing can kick in and only one page needs to manage the category content. And that would unblock this ticket.

Can’t we use the same mechanism that Karl mentioned to get the snippets synchronized, like any other field? I think he suggested we could just mark the snippet as synchronized? Or is that what you mean with auto-aliasing?

My general feedback on this is that don’t really want something like this to be manual, as it would be way too hard to keep in sync for all locales, doesn’t really scale,  and is hard to track. I.e. if a category is added to an existing product in English, you’d have to :
1. be notified about that change
2. go through all the translations to make the same change
3. QA all the translations to make sure you picked the right category for all of them/didn’t forget one

KalobTaulien:
Yeah so what I think we're going to end up doing @TheoChevalier is managing all the categories through the English page. So there could be Entertainment (English), Entertainment (French), Entertainment (Spanish) etc. At least that would be a good first step to finishing this ticket, and then we can do back and get more clever with this so they only need to select one category. But I feel like that's a tad out of scope for this ticket. Definitely open to feedback though!

su0641 (approved):

TheoChevalier:
Closing in favor of #7124

PR #6393, opened 2021-03-17
Simple caching

### LINKED ISSUE #6392: Further PNI caching
We need to explore the reason the PNI homepage is taking so long to load it timesout for users. 

—

On further inspection, it looks like since we moved away from cloudinary we're running into very slow loading pages. It's likely the PNI homepage is somewhat DDoS'ing itself with database queries due to 9 image renditions per image. 290 image x 9 renditions = 2610 unique database queries. On localhost I'm getting at least 440 queries _with_ the ORM query being cached in the Django view.

We already have some caching in the backend being applied which cuts my local queries down from 660..ish to 440..ish, but we aren't caching the product rendition queries — a quick solution is to cache that part of the template. 

When I applied caching on the template side of things, I got the queries down from 440 to 26 and load times (with a profiler installed) from 188s to 11.2s.

KalobTaulien:
Closes #6392 

* Removed .specific() as it's not needed. Though it wasn't doing any harm, it just wasn't needed. 
* Added template layer caching to help alleviate the `{% image ... %}` template tag issue where each function call results in a database query.


Pomax (approved):
let's see how far this gets us!

Pomax (changes requested):
rescinded: categories no longer work with this fix

KalobTaulien:
@Pomax Should be resolved now. We're using a `template_cache_key` and the `request.LANGUAGE_CODE` to make each cache fragment more unique.

Pomax:
holding off because it sounds like there might be a better solution tomorrow.

KalobTaulien:
PNI Home page products are now cached for anonymous users, and _not_ cached for logged in users. 

When a PNI Product Page is modified, the cache is cleared. 

We're clearing all cache on the site, however, since looping through every available language code times every category creates a very big matrix of cache keys to delete. 

This work takes localhost queries down from 240 to 26 queries when cached. And with debug toolbar profiling enabled, takes the page load time down from 48 seconds to 8 seconds when using the cached version of products.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1297,6 +1296,7 @@ def categories_page(self, request, slug):
         context['category'] = category.slug
         context['products'] = products
         context['pageTitle'] = f'{category} | ' + gettext("Privacy & security guide") + ' | Mozilla Foundation'
+        context['template_cache_key'] = category.slug
```
Pomax, line 1299: this will break later when we switch to "literally different pages for different locales", let's work the locale into the key even if with wagtail-modeltranslation that isn't a problem yet.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -1350,6 +1350,7 @@ def get_context(self, request, *args, **kwargs):
         context['products'] = products
         context['web_monetization_pointer'] = settings.WEB_MONETIZATION_POINTER
         context['about_page'] = BuyersGuidePage.objects.first()
+        context['template_cache_key'] = 'pni_home'
```
Pomax, line 1353: missing locale

network-api/networkapi/wagtailpages/wagtail_hooks.py
```diff
@@ -100,3 +101,15 @@ def before_delete_page(request, page):
...
+@hooks.register('after_edit_page')
+@hooks.register('after_delete_page')
+@hooks.register('after_publish_page')
+@hooks.register('after_unpublish_page')
+def manage_pni_cache(request, page):
+    if isinstance(page, ProductPage):
+        # Clear all of our Django-based cache.
+        # This is easier than looping through every Category x Language Code available
+        # To specifically clear PNI-based cache.
+        cache.clear()
```
Pomax, line 115: can we add/invalidate pages in a cache set? Because that would save a lot of time =D

Pomax (changes requested):

network-api/networkapi/wagtailpages/wagtail_hooks.py
Pomax: Also, question: does this invalidate the code we already have in place in the `products.py` file (the `invalidate_cache` function)?

KalobTaulien: Oh look at this, this exists already! We should the invalidate_cache function in products.py and use the code in wagtail_hooks.py — only because hooks are _supposed_ to live in wagtail_hooks.py.

KalobTaulien: Thats what I ended up doing in a new commit.

KalobTaulien: Oh also I forgot to mention before closing this. I don't think there's a global way across all caching backends to get a cache by its name and wildcard its arguments.

network-api/networkapi/wagtailpages/wagtail_hooks.py
```diff
@@ -100,3 +101,14 @@ def before_delete_page(request, page):
     if isinstance(page, ProductPage) and page.votes:
         # Delete the vote from ProductPages
         page.votes.delete()
+
+
+@hooks.register('after_delete_page')
+@hooks.register('after_publish_page')
+@hooks.register('after_unpublish_page')
+def manage_pni_cache(request, page):
+    if isinstance(page, ProductPage):
```
Pomax, line 110: We'll want to also invalidate the cache if the BuyersGuidePage gets updated because that'll invalidate which products to show (e.g. the cutoff date changes, etc)

Pomax (changes requested):

Pomax (approved):

PR #6398, opened 2021-03-17
updated CSP values in env.default to remove single quotation marks and added spaces around the actual data

### LINKED ISSUE #6352: remove quotes from CSP values in env.default
We had to put quotes around our env CSP values in the past in order not to get a million dotenv errors in Docker, but it looks like they've changed how that works yet again, and the combination of double and single quotes now breaks env var reading, causing the site to not load correctly because instead of `'self'` the browser now gets `"'self'`, which is an invalid CSP target.

- [x] update `env.default` so that all CSP values use `CSP_THING_SRC=" ... "`, with spaces around the actual data

danielfmiranda:
Closes #6352 
Related PRs/issues #

## Checklist


Pomax (changes requested):
The quotes are `self` are not optional: they _have_ to be there as per the CSP spec, so all those `self` etc where you removed the single quotes need to become `'self'` etc.  again =)

danielfmiranda:
Hi @Pomax, just re-added the single quotes around "self" and "none". Thanks!

Pomax (approved):
:+1:

Issue #6395, opened 2021-03-17
Consider reducing the thumbnail images we serve on the PNI imagepage 


Pomax:
## Summary

The PNI homepage uses 8 different sized thumbnails for product cards, and with 136 products that means 1088 database queries to get all our images, and that is about 1000 too much (normally, hyperbole, but in this case, pretty much literally 1000...) causing the page to time out because wagtail has to resolve each of those to real files before it send the page HTML to the user.

We're looking at turning this into something manageable, and the most immediate win would be to only use a single x1 and a single x2 size for PNI thumbnails, set to something that works for all sizes (probably: 360x360 for x1, 720x720 for x2).

@kristinashu would this be something we could do, or would there be objections for reasons we didn't think of?

<details>
  <summary><strong>Backlog From slack</strong></summary>

[pomax] @kristinashu @KalobTaulien soooo we have an interesting problem relating to image sizes

[pomax] basically we have so many different source sizes for the homepage/category page view that wagtail is hitting its performance limit just trying to service all the [source] instructions =)

[pomax] would it be possible to reduce the number of resolutions we declare in https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/buyersguide/templates/buyersguide/catalog.html#L82-L135 ?

[pomax] (the problem being that wagtail apparently sees an {% image ... %} instruction, queries the database, then sees a new {% image ... }% instruction, queries the database again, eight times for each product, 136 products on the homepage = that's a lot of database work before we can even render the page for the user =S) (edited) 

[ben] Is it possible to move this logic into the view so we can group the db queries and pass them into the template via context?

[pomax] there's no view function like in plain Django, so not as easily as you would hope, but kalob did mention an interesting idea around a new template tag that can batch a whole bunch of different sizes as a single operation.

[pomax] even so, reading through this &lt;picture&gt; set, I'm a little puzzles, since all these sizes are very similar

[pomax] but also the smaller the viewport, the large an image we serve? unless I'm reading this wrong. It looks like on desktop, it's 260x260 / 520x520 but on mobile it's 360x360 / 720x720

[Kalob Taulien] Its because the bootstrap grid kicks in and makes each image wider at a certain point, and then smaller again when the breakpoints kick in. That threw me off at first too

[pomax] just a single intermediate size to cover all sizes feels like it might solve a lot of our problems here, too.

[...]

Issue #6248, opened 2021-02-23
Remove Cloudinary


KalobTaulien:
We need to remove Cloudinary after #6247 lands. 

There are settings to remove, dependencies to remove, functions to remove, model fields to remove, and template code to remove — all in favour of Wagtail Images. 

Most of this should be pretty quick but we may want to spend some extra time making sure the templates still render as we expect, and that there isn't additional data in the context processors being added to every request. 

Example: There are things like this in `products.py` 

```python
if settings.USE_CLOUDINARY:
    image_field = FieldPanel('cloudinary_image')
    MEDIA_URL = settings.CLOUDINARY_URL
else:
    image_field = ImageChooserPanel('image')
    MEDIA_URL = settings.MEDIA_URL
```

We can remove all of that minus the ImageChooserPanel, which we'll place directly into the `content_panels` of the `ProductPage` class. Currently that swaps out ImageChooserPanel() for a FileField() to upload to cloudinary.

KalobTaulien:
I'm going to wait until migrations are squashed to finish this ticket to keep migrations clean without having to do some fancy migration surgery.

PR #6401, opened 2021-03-18
Dynamically serve PNI images

### LINKED ISSUE #6400: Dynamically serve PNI images using Wagtail ServeView
PNI images have too many renditions to load and look for on each page request causing timeouts. 

Implement this feature: https://docs.wagtail.io/en/stable/advanced_topics/images/image_serve_view.html 

The idea is to _not_ look for all images in a single HTTP request, but to instead spread them out so each image becomes it's own HTTP request to generate a rendition on the fly in a non-blocking manner.

KalobTaulien:
Closes #6400
Related PRs/issues #6378 #6248 #6395 

Replaces `{% image .. %}` with `{% image_url %}` so when an http request is made to the image URL, it will dynamically create an image rendition and serve it. Caching is included, so we don't need to worry about creating multiples of the same renditions over and over again. 

**Note**: I did _not_ adjust the number of thumbnails being generated. That will be taken care of in other upcoming tickets.


network-api/networkapi/urls.py
```diff
@@ -48,6 +48,11 @@
         RedirectView.as_view(url='/docs/how-do-i-wagtail/'),
         name='how-do-i-wagtail'
     ),
+    re_path(
+        r'^images/([^/]*)/(\d*)/([^/]*)/[^/]*$',
+        ServeView.as_view(action='redirect'),
+        name='wagtailimages_serve',
+    ),
```
Pomax, line 55: Small nit, but let's make this its own `image_url_tag_urls.py` file with some comments that explain that this lets us use an `image_url` tag in our templates, and what it does, in the `wagtailcustomizations` dir, and then include that patthern it exports here? Just to make it much more obvious to future devs (including ourselves) that this is a customization

Pomax (changes requested):

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
+"""
+Dynamic Wagtail Images.
+https://docs.wagtail.io/en/stable/advanced_topics/images/image_serve_view.html
+
+Typically with Wagtail we use use an {% image img.object fill-XXXxYYY %} template tag
```
Pomax, line 5:
```suggestion
Typically with Wagtail we use an {% image img.object fill-XXXxYYY %} template tag
```

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
...
+to create an image rendition. These renditions are blocking processes, which is typically
```
Pomax, line 6:
```suggestion
to create an image rendition. Creatingh renditions is a blocking process, which is typically
```

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
...
+fine for a standard page with a normal amount of images. Every rendition costs one
+database query, and if the rendition doesn't exist, it needs to also create the image.
+On a page with 200 images, that's 200 extra database queries, plus rendition-creation
+time. If you have 200 images, with 10 renditions each for mobile support, that's 2000
+database queries plus rendition-creation time, which is unacceptably slow.
+
+Instead of performing all those queries up front, along with performing rendition
+creation upfront, we can use dynamic image URLs to generate an image URL and when
+that URL is requests, Wagtail will find the original image, create the rendition, and
```
Pomax, line 15:
```suggestion
that URL is requested, Wagtail will find the original image, create the rendition, and
```

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
...
+serve the image. This off loads the heavy lifting in the upfront HTTP request and defers
```
Pomax, line 16:
```suggestion
serve the image. This offloads the heavy lifting in the intiial HTTP request and defers
```

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
...
+the rendition-creation to a different HTTP request. To take advantage of this, we use
+the custom URL in this file, along with a slightly different template tag and syntax.
```
Pomax, line 18:
```suggestion
the custom URL declared in this file, along with a slightly different template tag and syntax.
```

Pomax (changes requested):
mostly phrasing/typos =)

KalobTaulien:
I'm not sure why exactly, but tests are failing. 

They were originally failing because there were no images attached to a ProductPage in the tests, but then I added that and got another similar error. The test class is  `networkapi.buyersguide.tests.TestBuyersGuidePage`

```
  File "/app/dockerpythonvenv/lib/python3.7/site-packages/wagtail/images/views/serve.py", line 20, in generate_image_url                                                                                                    
    signature = generate_signature(image.id, filter_spec, key)                                                                                                                                                              
AttributeError: 'str' object has no attribute 'id' 
```

This is suggesting the `product.image` doesn't have an actual image attached to it, but I had thought I solved that already. The problem comes up in the about routable pages. The test I was specifically working with was `def about_url_test(..)` and `def test_buyersguide_about_how_to_use_view(..)` (and commented out the rest of them for simplicity)

My current thinking is that `BuyersGuidePage.get_context(...)` isn't providing the right images to the products, or just that some products are missing images. I noticed get_context() is called twice, one with a ProductPage that **has** an image, and a second time where the same ProductPage **doesn't** have an image. 

That is where I'm at right now.

network-api/networkapi/image_url_tag_urls.py
```diff
@@ -0,0 +1,34 @@
+"""
```
Pomax, line 1: let's put this in `networkapi/wagtailcustomization`?

KalobTaulien:
Soooooo ran into a funny thing where a bunch of our buyersguide pages fail to load through tests, but load perfectly when user testing. Pomax and I both spent a chunk of time trying to resolve this issue so we can use dynamic iamge URLs but we couldnt find a solution in a timely manner. 

Instead we're going to cut out the renditions we don't need so we can at least flip the switch to fully remove cloudinary and unblock those tickets. Then we'll come back to this a bit later if we determine we still need dynamically generated image URLs.

Pomax (approved):

PR #6405, opened 2021-03-18
6390 download button update

### LINKED ISSUE #6390: Download button on Publication Page is off-brand
### Describe the bug
Download button on Publication Page is off-brand. I didn't check Article Page but the button there might be off-brand as well.

### To Reproduce
1. Go to https://foundation.mozilla.org/en/insights/trustworthy-ai-whitepaper/
2. Inspect the "Download PDF" button (default state and hover state)
3. Button is in a different style than what we have in the [style guide](https://foundation.mozilla.org/en/style-guide/)

### Expected behavior
Button should follow style guide. Need to verify this with designers but I'm guessing the button should be in dark theme `.btn-primary` styling. e.g., we need to make sure on hover, the download icon is black instead of white.

### Screenshots
Default state
image
Hover state
image

danielfmiranda:
Closes #6390 

**Link to sample test page**: https://foundation-s-6390-downl-qafc85.herokuapp.com/en/publication-page-with-child-article-pages/

**Changes in Models:**

### DIFF
--- a/source/sass/components/publication-page.scss
+++ b/source/sass/components/publication-page.scss
@@ -46,18 +46,22 @@
   }
 }
 
-.download-button::before {
-  content: "";
-  display: inline-block;
-  width: 18px;
-  height: 18px;
-  margin-right: 10px;
-  background-image: url(../_images/glyphs/download-file.svg);
-  transition: $button-icon-transition-speed;
-}
+.download-button {
+  &::before {
+    content: "";
+    display: inline-block;
+    width: 18px;
+    height: 18px;
+    margin-right: 10px;
+    background-image: url(../_images/glyphs/download-file.svg);
+    transition: $button-icon-transition-speed;
+  }
 
-.download-button:hover:before {
-  filter: brightness(0);
+  @include hover-focus-active {
+    &::before {
+      filter: brightness(0);
+    }
+  }
 }
 
 @media (max-width: $bp-md) {
--- a/source/sass/components/publication-page.scss
+++ b/source/sass/components/publication-page.scss
@@ -53,7 +53,7 @@
   height: 18px;
   margin-right: 10px;
   background-image: url(../_images/glyphs/download-file.svg);
-  transition: filter 0.2s ease-in-out, -webkit-filter 0.2s ease-in-out;
+  transition: filter 0.2s ease-in-out;
 }
 
 .download-button:hover:before {
--- a/source/sass/components/publication-page.scss
+++ b/source/sass/components/publication-page.scss
@@ -53,7 +53,7 @@
   height: 18px;
   margin-right: 10px;
   background-image: url(../_images/glyphs/download-file.svg);
-  transition: filter 0.2s ease-in-out;
+  transition: $button-icon-transition-speed;
 }
 
 .download-button:hover:before {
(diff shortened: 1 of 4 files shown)


source/sass/components/publication-page.scss
```diff
@@ -53,6 +53,11 @@
   height: 18px;
   margin-right: 10px;
   background-image: url(../_images/glyphs/download-file.svg);
+  transition: filter 0.2s ease-in-out, -webkit-filter 0.2s ease-in-out;
```
danielfmiranda, line 56: I added this line as without it, the svg icon and the text of the button seemed to be out of sync when changing color. feel free to ping me if you would like me to show you what I mean :)

mmmavis: Good call. This is how we implemented [`.btn-secondary` the external link version](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/buttons.scss#L131-L142) as well.

Two things here:

- Nice that you included vendor prefix! Good news is you don't have to worry about it on this repo! We have `autoprefix` in place so vendor prefixes will auto get added after compilation! 
- Since `.download-button::before` share quite a bit of rules as what we have for the [`.btn-secondary` the external link version](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/buttons.scss#L131-L142), let's create a new mixin to refactor things a bit. Have a look at the existing examples in [mixins/_buttons.scss](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/mixins/_buttons.scss) and give it a try. You can always ping me if you have questions. 🤓

mmmavis (changes requested):
Let's create a new mixin!

sabrinang (approved):
:+1: Button icon looks good!

mmmavis:
Sorry for the confusion. In my previous comment I didn't realize the external link button has icon on the right side and while download button has the icon on the left. Let's just turn the transition rule into a variable and not worry about adding a new mixin.

danielfmiranda:
Hi Mavis, no problem! I have created the variable, and updated any buttons that use an SVG icon to use the new variable like discussed

source/sass/components/publication-page.scss
```diff
@@ -53,6 +53,11 @@
   height: 18px;
   margin-right: 10px;
   background-image: url(../_images/glyphs/download-file.svg);
+  transition: $button-icon-transition-speed;
```
mmmavis, line 56: For code readability, let's nest these two blocks together & use the `hover-focus-active` mixin as we'd like the transition to kick in for focus/active state as well

```scss
.download-button {
  &::before {
    ...
    ...
    ...
  }

  @include hover-focus-active {
    &::before {
      filter: brightness(0);
    }
  }
}
```

mmmavis (changes requested):
PR's looking great! Just one more change then this should be ready to go!

source/sass/components/publication-page.scss
danielfmiranda: Hi Mavis,

Great suggestion! I have added it to the latest commit

mmmavis (approved):
Lovely 🌮 🍰 🍺

PR #6406, opened 2021-03-18
Reorder some SCSS imports

### LINKED ISSUE #6404: Re-order some imports in main.scss and bg-main.scss
Some imports like [`resets.scss`, `type.scss`, `button.scss`](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/main.scss#L42-L44) ... etc should probably be imported first in main.scss (e.g., move up in the file) as they contain the base style rules for [components](https://github.com/mozilla/foundation.mozilla.org/blob/master/source/sass/main.scss#L15-L38) to follow. By doing so, we can potentially clean up unnecessary overrides in the component scss files. 

Setting the estimate to `1` for now as I'm being optimistic that this won't introduce any regression issues. 🍀

mmmavis:
Closes #6404

I don't know why Percy is flagging an issue on article page. I compared the review app page and the page called but I couldn't find any SCSS difference in the hero section. I'm guessing this is safe to merge... as the difference Percy flagged seems to be a tiny offset?

This is where Percy flags the visual regression issue
- https://foundation-s-issue-6404-5fvwnt.herokuapp.com/en/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/

Local page
- http://test.example.com:8000/en/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/

### DIFF
--- a/source/sass/buyers-guide/bg-main.scss
+++ b/source/sass/buyers-guide/bg-main.scss
@@ -12,9 +12,12 @@ html {
 
 // Site-wide
 
+@import "../resets";
 @import "../glyphs";
 @import "../mixins";
 @import "./mixins"; // Buyer's Guide specific mixins
+@import "../type";
+@import "../buttons";
 
 // PNI specific
 
@@ -48,9 +51,6 @@ html {
 
 // Misc
 
-@import "../resets";
-@import "../type";
-@import "../buttons";
 @import "../global";
 @import "../utilities";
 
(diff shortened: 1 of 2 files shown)


Pomax:
yeah, this seems a small enough to offset difference to not really bother trying to figure out...

Pomax:
yeees, forgot to hit "approve", but: approved =D

PR #6407, opened 2021-03-18
removed all instances of the internet health check link and related files

### LINKED ISSUE #4103: Remove IH snippet
A lot of our CMS templates automatically add this to the bottom of the page. We should take a look at this again, make sure that this is still the best approach and that the copy is still relevant.
image

Notes:
- The goal for this was to reduce our bounce rate. 
- Copy is pretty out of date and maybe we can do something that will work on everywhere.
- I'm also curious about how many people actually click on this.
- Wonder if showing related content via tagging will be a more effective way to get people to stay and explore the site.

And making a note about Pomax's comment in slack:

> does it make sense to have the "this is part of a broader movement" on artifact pages? e.g. https://foundation.mozilla.org/en/artifacts/x-ray-goggles/

danielfmiranda:
Closes #4103 

**Link to sample test page**:
Campaign Page [here](https://foundation-s-4103-remov-tle1s4.herokuapp.com/en/campaigns/answer-loss-itself/)
Initiative Page [here](https://foundation-s-4103-remov-tle1s4.herokuapp.com/en/initiatives/single-page-opportunity/)

## Checklist

**Changes in Models:**


Pomax (changes requested):
There's a comment in homepage.js that we'll want to update, too.  It feels like it's no longer even remotely related to what it does =)

danielfmiranda:
Hi Pomax, thanks for the catch! 

I have updated the comment to make it a little more descriptive of what the file houses.

Pomax (approved):
nice!

PR #6413, opened 2021-03-19
removed all other source sets, besides 360x360 and 720x720

### LINKED ISSUE #6378: Reduce the number of PNI images in the /privacynotincluded/ template

danielfmiranda:
Closes #6378 #6413 

**Link to sample test page**: [here](https://foundation-s-4103-remov-tle1s4.herokuapp.com/en/privacynotincluded/)

## Checklist

**Changes in Models:**


KalobTaulien (approved):
This works as expected! R+

PR #6425, opened 2021-03-24
Fix for stray divs in article page with wide image

### LINKED ISSUE #6386: Investigate mismatched HTML tag issues on article pages
### Describe the bug
There are HTML validation errors around mismatched HTML tags on our sample article pages. I'm guessing the cause might be some block template are missing opening/closing tags.

### To Reproduce
1. Go to `/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/` locally or on review app.
2. View source code
3. Scroll through and look for highlighted mismatched HTML tag errors

### Expected behavior
No mismatched HTML tags

### Screenshots

image

image

danielfmiranda:
Closes #6386 

**Link to sample test page**: [here](https://foundation-s-6386-misma-l6tkti.herokuapp.com/en/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/)

## Checklist

**Changes in Models:**


danielfmiranda:
Hi Mavis, 

I believe I found the cause for the 3 stray divs, if you check the source code now everything looks like it has its own closing div :)

Tagging Kalob as I was hoping he can double check in case there was a reason for the 3 divs! 

Thanks all

KalobTaulien:
There is a reason for the missing opening div elements. They get injected from a streamfield template when the page is built. It allows us to break out of the bootstrap grid, create something "wider" in the template, and re-enter the bootstrap grid so everything stays lined up nicely. 

Take a look at [this template](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/article_full_width_image_block.html) for instance

Pomax:
@KalobTaulien remember to hit that "changes requested" =D

KalobTaulien:
This needs some more discussion about what the goal is here. If there are extra tags being exposed we need to figure out where they are being generated.

mmmavis:
The main goal is to make sure the final page HTML code is valid (see #6386).

mmmavis:
Hi @danielfmiranda , is this PR ready for review?

danielfmiranda:
Hi @mmmavis, not yet, going to take this issue back to the in progress lane to double check a couple of things. Thank you!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/article_full_width_image_block.html
```diff
@@ -24,9 +24,10 @@
     </figure>
   </div>
 
-{% if not is_last %}
-  {% comment %} Restart the Article Block bootstrap grid {% endcomment %}
-  <div class="container">
-    <div class="row">
-      <div class="offset-sm-1 col-sm-10 col-12 article-blocks">
-{% endif %}
+  {% comment %} Restart the Article Block bootstrap grid. If it is the last item, hide the empty container to keep from pushing footer down a couple px{% endcomment %}
```
danielfmiranda, line 27: Hi @KalobTaulien @mmmavis,

I believe that the 3 mismatched divs were coming from **article_page.html**, as there are 3 closing divs on lines 45,46,47, that are expecting to close the three container divs on lines 16,17,18.

After finding that the error was only reproduced when there was a wide image block at the end of the article page, I then found that the **wide_image_block** file starts by closing the 3 container divs from **article_page.html** on lines 4,5,6, but then does **not** reopen them if the image block is the last item in the for loop.

To solve the issue, I removed the if statement so the image block can open the three container divs **article_page.html** is expecting to close, which removes the error. 

However, this did create a empty container object that pushed the contents below down just a bit.

To alleviate this, I then added a if statement in the class, which checks if the image block is the last item in the for loop. If so, it adds the "hidden" class to remove the extra container div to avoid anything else being pushed down in the page, which I confirmed with percy. 

If you have any questions feel free to let me know and I can explain it through a zoom call :) Thank you!

KalobTaulien (approved):

mmmavis (changes requested):
Tested a few scenarios and they all looked good. Nice work!

One thing that can be improved -- instead of adding a new utility class `hidden`, let's just use Bootstrap's `d-none`.

danielfmiranda:
Hi @mmmavis thank you very much and thanks for the catch! I have it now using the d-none class and removed the custom class from the CSS files.

mmmavis (approved):

PR #6426, opened 2021-03-24
Make PrivacyPolicyLinks translatable

### LINKED ISSUE #6257: Privacy Policy Links need localization
Currently, the Privacy Policy Links (PNI) aren't being translated and we don't know why. Let's explore why they aren't and how we can get them to work with wagtail-localize, and apply it. 

Acceptance criteria:
- [x] Make privacy policy links (Orderable) translatable using wagtail-localize 

Testing:
- [ ] Make sure the PrivacyPolicyLinks orderable is translatable
- [ ] When translating a page, test to see if the translations work on the newly translated page

KalobTaulien:
Closes #6257

When translating a page, we can now update the privacy policy label and the URL. I think it's good we translate the URL. For example: many products have privacy policies in different languages - so a Spanish product might have a Spanish privacy policy we can link to. 

image

This needed the migrations (in this order). We needed to bootstrap the model, then convert it to a translatable model. This requires a few migrations [as per the documentation](https://docs.wagtail.io/en/stable/advanced_topics/i18n.html#making-snippets-with-existing-data-translatable)


TheoChevalier (changes requested):
I’ve been thinking a while about this, and while I totally agree those URLs should be customizable, I don’t think translators should be the ones doing it. I mean, there are pros and cons for doing it, but ultimately I think we should treat them like other URLs, and override them as needed from the admin.

That said, we can always revisit this later since it’s easy to switch the field from `synchronized` to `translatable`, but for now would you mind leaving it as synchronized?

Thanks!

Pomax:
closing this for now, given Theo's comment.

Issue #5887, opened 2020-12-15
PNI follow-up: remove the buyersguide app


Pomax:
- [ ] make sure all app files have already been disabled/deleted
- [ ] create a management command to
  - [ ] deal with the table rename for the product category
  - [ ] remove the buyersguide tables from the db using pure SQL
- [ ] update wagtail:
  - [ ] remove the buyersguide nav section

Pomax:
marking as blocked until we have all the things we still rely on in the old app, moved over to the `wagtailpages` app

KalobTaulien:
Dev note: Let's also ensure we remove the old templates that aren't being used. 2x check the views manually since there are duplicate templates that are now namespaced - we want to make sure we remove the _correct_ templates.

PR #6428, opened 2021-03-24
Delete old pni models, related code and add removal migrations 

### LINKED ISSUE #5521: PNI follow-up: delete PNI product models and related code, plus migrations
> Delete PNI products (in their own ticket) and cleanup the old code 

Criteria:
- [ ] Remove PNI Product models 
  - [ ] base
  - [ ] general
  - [ ] software 
- [ ] Create migrations to remove the models from the database 
- [ ] Remove any related code (ie. factories) surrounding the old product models.

KalobTaulien:
Closes #5521 

Deleted the models, most of the related code, and added removal migrations. I didnt merge the migrations because we have a squash coming up — but as a good practice I can merge them if decided we should do that. 

**Notes**: 
* This removes the ProductPrivacyLink snippet which has data in production, but should not be used at all anymore. 
* There's also a tonne of additional code we can remove in the buyersguide app but that should be taken care of in #5887.
* Since the app is not being used anymore, I replaced instances of `Product.` with `ProductPage.` just so Django doesn't complain. Ultimately that will all be removed in #5887 as well.


Pomax (approved):
as far as I can tell this is good to go. Tests pass, manual verification of the homepage, category, and some products show this still works, so... let's get it in there.

PR #6431, opened 2021-03-24
added default URL to video URL field

### LINKED ISSUE #5364: Video StreamField block needs a factory URL
## Steps to find problem

1. Create a new env or db with `inv new-env` or `inv new-db` 
2. Log into the admin
3. Edit the homepage
4. Publish Homepage
5. Notice the validation error from the missing URL on the `video` block

### The fix requires
The fix requires a URL field for the `video` block. 

### Considerations
Might want to see what platform the URL is looking for, or if its a generic URL, so we don't end up putting a filler/lorem URL when the URL should be a linked to YouTube, for example.

danielfmiranda:
Closes #5364 

**Link to sample test page**: [here](https://foundation-s-5364-strea-1tem1t.herokuapp.com/en/blog/initial-test-blog-post-with-fixed-title/)

**Changes in Models:**


network-api/networkapi/utility/faker/streamfield_provider.py
```diff
@@ -159,7 +159,7 @@ def generate_video_field():
         # Fake embed url will lead to timeout error for Percy.
         # Solution: either we provide a valid embed url or use an empty string.
         # See details: https://github.com/mozilla/foundation.mozilla.org/issues/3833#issuecomment-562240677
-        'url': '',
+        'url': 'https://www.youtube.com/embed/83fk3RT8318',
```
Pomax, line 162: can we at least have a URL that is owned by us? I'd like to make sure we're not driving traffic to youtube just because as testing,

Pomax (changes requested):

network-api/networkapi/utility/faker/streamfield_provider.py
danielfmiranda: Hi @Pomax! 

Since using the example.org URL would prevent the video component from rendering for testing, I took a look at the mozilla youtube channel and found [this](https://www.youtube.com/watch?v=COU5T-Wafa4) video on "how to choose strong passwords" from ~10 years ago. I was wondering if this would be a good candidate for the default URL? 

Tagging @KalobTaulien in case he has any input as well :) 

Thanks all!

mmmavis:
@danielfmiranda just a reminder to move PRs in review to Zenhub's `Review/QA` pipeline 😉

network-api/networkapi/utility/faker/streamfield_provider.py
mmmavis: @Pomax the example YouTube video included here is under Firefox's YouTube account: https://www.youtube.com/watch?v=83fk3RT8318 so we should be good. Or we can find a Mozilla video if that's more preferred. As Daniel mentioned above, we need a real embed url for the video embed to show up properly so unfortunately example.com URLs won't work for this use case.

Pomax (approved):

PR #6443, opened 2021-03-26
Update translation management script with new buyersguide path

### LINKED ISSUE #6442: Migrate buyersguide app translations
#6435 moved the buyersguide app, but translation files are still set up for the old path.

I’m moving the files here https://github.com/mozilla-l10n/fomo-l10n/pull/40, we will just need to update the management script here https://github.com/mozilla/foundation.mozilla.org/blob/master/translation-management.sh

TheoChevalier:
Closes #6442

If you want to test, you need to setup the translation repo first: https://github.com/mozilla/foundation.mozilla.org#translations
Then run `inv makemessages`


translation-management.sh
```diff
@@ -37,16 +37,16 @@ case $command in
   "import")
     echogreen "Importing latest translation files from fomo-l10n repository"
     cp -r "${L10N_REPO}foundation/translations/locale/" "network-api/locale/"
-    cp -r "${L10N_REPO}foundation/translations/networkapi/buyersguide/locale/" "network-api/networkapi/buyersguide/locale/"
-    cp -r "${L10N_REPO}foundation/translations/networkapi/buyersguide/templates/about/locale/" "network-api/networkapi/buyersguide/templates/about/locale/"
+    cp -r "${L10N_REPO}foundation/translations/networkapi/wagtailpages/templates/buyersguide/locale/" "network-api/networkapi/wagtailpages/templates/buyersguide/locale/"
+    cp -r "${L10N_REPO}foundation/translations/networkapi/wagtailpages/templates/about/locale/" "network-api/networkapi/wagtailpages/templates/about/locale/"
```
Pomax, line 41: hm, this might be a "bug", in that the BG about page templates should be in the BG dir, not the top template dir... We can land this first, then fix this, then land a new PR for that fix, or we can fix it first and then land this PR with the updated location, which would you prefer @TheoChevalier?

Pomax (changes requested):

translation-management.sh
TheoChevalier: Let’s probably land this now so that we’re able to land string changes again, then update the path again. Let’s coordinate before actually moving the about page templates though, I’ll need to move their counterparts in the fomo-l10n repo, and I’d like to check it’s not breaking things

Pomax: I'll file a new issue for that work, then.

Pomax: https://github.com/mozilla/foundation.mozilla.org/issues/6477

Issue #6381, opened 2021-03-16
squash wagtailpages migrations


Pomax:
In order to get some semblance of sanity back, it's time to squash our migrations:

part 1:

- [x] create the django squash migration for `wagtailpages`
  - [x] preserve all as-part-of-migrations data generation (but not conversion) functions
- [x] apply that migration to production

part 2:

- [x] remove all old migrations
- [x] apply that removal to production

part 3:

- [ ] remove the squashed migration and generate a fresh single step db bootstrap
  - [ ] preserve all as-part-of-migrations data generation (but not conversion) functions
- [ ] apply that removal to production

Pomax:
part three will be covered in its own issue here: https://github.com/mozilla/foundation.mozilla.org/issues/6491

PR #6470, opened 2021-03-30
initial squashings

Pomax:
Part one of #6381 

- [ ] `inv migrate` on top of a prod/stage copy should not actually "do" anything.
- [ ] `inv test` should still work

This will need to land and get rolled out all the way to prod before part 2 can happen (which is removing the old migration files and removing the `initial=True` and `replaces=[...]` from the squash migration)


KalobTaulien (approved):
Looks good to me! Everything seems to work as far as I can tell. Just need to update the branch before merging

PR #6471, opened 2021-03-30
6465 wider iframes on blog and publication pages

### LINKED ISSUE #6465: Wider iframe and wider images on blog and publications
We need to create slightly wider image format (col-12, currently uses col-8 on lg) to enable our content creators to increase engagement with their graphics, as they're currently quite small.

- [ ] wide iframe on blog
- [ ] wide image on blog
- [ ] wide iframe on publications
- [x] wide image on publications

KalobTaulien:
Closes #6465 

**Link to sample test page**: 
* Article Page with wide Iframe and Wide..ish image: https://foundation-s-6465-wider-zy8rl8.herokuapp.com/en/publication-page-with-child-article-pages/fixed-title-article-page/
* Blog page iwth wide iframe and wide image: https://foundation-s-6465-wider-zy8rl8.herokuapp.com/en/blog/initial-test-blog-post-with-fixed-title/

## acceptance criteria 
- [x] wide iframe on blog
- [x] wide image on blog
- [x] wide iframe on publications
- [x] wide image on publications

**Note**: This makes the iframe template component very technical. More technical than a stream block template _should_ be, but also keeps the code a bit more dry — so instead of having two virtually-identical iframe templates, we have just one. 

For this work, we need to break out of existing grids and use a wider column setting. That's why you'll see new `</div>`'s in the templates.


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/iframe_block.html
```diff
@@ -1,14 +1,44 @@
...
-</div>
+{% if page_type == "blog" %}
+  {# Break out of the blog_page.html containment grid #}
+      </div> <!-- End .blog-body -->
+    </div> <!-- End .cms -->
+  </div> <!-- End .col-lg-8 -->
+{% elif page_type == "article" %}
+  {# Break out of the article_pag.html containment grid #}
+  </div> <!-- end .offset-sm-1.col-sm-10.col-12.article-blocks -->
+{% endif %}
```
benhohner, line 19: I think we should find a better way to do this, long term. How do we make it so we can have the same block across multiple templates without conditionals? Maybe some refactoring around where the bootstrap `.row` or `.container` classes are defined?

benhohner (approved):

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/iframe_block.html
mmmavis: This was my concern too when we decided to allow wider blocks (that break container) on certain page types. I can't think of other ways to build this without restructuring all our page template code. 😬

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/annotated_image_block.html
```diff
@@ -1,13 +1,30 @@
...
-  </figcaption>
-  {% endif %}
-</figure>
+
+{% if value.wide_image %}
+    </div> <!-- End .blog-body -->
+  </div> <!-- End .cms -->
+</div> <!-- End .col-lg-8 -->
+
+<div class="col-lg-12">
```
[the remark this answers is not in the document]
mmmavis: Let's get a designer to review this PR! Gonna tag Kristina on this.

mmmavis:
@kristinashu we would like to get a designer to review this PR. Could you help assign one for us? :D

kristinashu:
Thank you Mavis for including design in this review! It's exciting to see these wider widths. Sabrina would be best but since she's still on PTO I'll take a stab at it. Here are a few layout issues that I see. Note I'm only looking at the image component and not the iFrame since I can't see iFrames in PR but my assumption is that it has the same layout as the image component..

## Different wide widths for blog and publications
My assumption is that "wide width" would be the same for both the blog template and the publications template. The blog version of wide width being the ideal version. Is it possible to make the publication's wide the same as the blog?

Screen cap from PR
dif-widths


## Different spacing for blog and publications
Spacing seems inconsistent between the image component on blog and publications. Spacing includes: above the image, between the image and the caption, and below the caption. These should be as consistence as possible between each other as well as the other established components (e.g. Video embed). 

Screen cap from PR
spacing


## Check box doesn't change anything on publications template 
Selecting or not selecting the check box for wide image on the publications template doesn't seem to change anything. 

Screen cap from PR
box-checked

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/annotated_image_block.html
```diff
@@ -1,13 +1,30 @@
...
+      {{ value.caption }}
+      {% if value.captionURL %}</a>{% endif %}
+    </figcaption>
+    {% endif %}
+  </figure>
+
+{% if value.wide_image %}
+</div>
+
+<div class="py-4 py-md-5 col-lg-8 offset-lg-2">
```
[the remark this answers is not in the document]
KalobTaulien: Yes there is. Because its a stream block it could be used anywhere in the future. The ideal scenario would be using 2 annotated image blocks for both publications and blog (one each), and keep that pattern in the future for future use cases. It creates less DRY code, but keeps it flexible because the blog vs publication templates were setup differently. And a future page where we need to break out of the normal grid would likely be different yet again.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/iframe_block.html
KalobTaulien: @mmmavis Same here, I cant think of another way to do this other than to introduction a new stream block. Like a "blog_iframe" and a "publication_iframe". But becuse we're already using iframes, the migration would be tricky to write which we want to avoid if possible. So then we fall into the situation where there's an "iframe", "blog_iframe" and "publication_iframe" (where the normal "iframe" is on both paths, so there's 2 iframe blocks on each page, which gets confusing for editors). 

We could maybe rename the normal iframe to "normal iframe" and create a new one called "wide iframe" and split "wide iframes" into 2 files, one for publications and one for blog.

Both options are messy I think. Im happy to switch techniques in this PR if that's what we think is going to be most maintainable moving forward. Thoughts @mmmavis @benhohner ?

KalobTaulien:
Note: Ben, Mavis, Pomax and myself have a meeting today to discuss a better implementation of streamfields that's the same across blog pages and article pages. 

It'll likely become a large ticket, or even an epic.

KalobTaulien:
This needs discussion as to whether or not we want to merge this in as-is, or wait for a proper streamfield cleanup that's being addressed in #6517. I'm going to move this to *Needs Discussion*

benhohner:
I recommend we merge it because the suggested refactor is large and we're not sure when we can get to it, so at least people can get value right now!

KalobTaulien:
@benhohner Sounds good. Ill move this back up to to Up Next and look at implementing some of the design feedback before merging this in

KalobTaulien:
I've made some adjustments based on your feedback @kristinashu - and thank you for the in-depth review! 

I'll get the review app setup again with some new links and demos with iframes soon. 

One thing to note is the spacing between content and iframes/wide images on blog and publications aren't identical. We're running into issues making this work without getting into insane code debt. But when we refactor all these blocks this will be fixed because each block won't have custom "break out of grid" logic in it, so it'll be more consistent after that work. I think our best bet is to land this as close as we can get it, without it being perfect, and then taking a serious look at the work involved around the refactor. 

For those just jumping in: the stream block refactor would turn blog posts and article page stream blocks into their own sections without having to consider logic for breaking out of the grid, or determining if it's the last block, or having to possibly deal with previous block templates for consistent spacing. The spike issue is referenced here #6517

KalobTaulien:
@kristinashu Here are the new demo pages. 

* Article Page with wide Iframe and Wide..ish image: https://foundation-s-6465-wider-zy8rl8.herokuapp.com/en/publication-page-with-child-article-pages/fixed-title-article-page/
* Blog page with wide iframe and wide image: https://foundation-s-6465-wider-zy8rl8.herokuapp.com/en/blog/initial-test-blog-post-with-fixed-title/

**Note** the spacing is getting _really_ funky now because of the complex way we have these blocks setup. We're super aware of that and when we refactor these blocks to be their own individual rows instead of breaking in and out of a parent container, that will be fixed. I think we should prioritize that work next sprint so when this lands, it doesn't look too "off" for too long.

kristinashu (approved):
Looks good! Love the photo shoot!

PR #6494, opened 2021-04-06
Some tests for the create_wagtail_image function

### LINKED ISSUE #6350: `create_wagtail_image` function needs unit tests.
To ensure that `create_wagtail_image` always works the way we're expecting it to work in the future, we should write some unit tests. 

Namely, we need to test:
- [ ] Fetching an image from a URL and saving it as a Wagtail Image (but have this disabled so CI doesn't take longer and download an image on every run, but can be tested locally)
- [ ] Fetching an image from a local resource (local file) and saving it as a Wagtail Image 
- [ ] Test automatic creation of a new collection
- [ ] Test image generation without a collect (should be in the "Root" Collection)

KalobTaulien:
Closes #6350 

Some basic tests to increase confidence that this function works.


network-api/networkapi/tests.py
```diff
@@ -613,3 +616,52 @@ def test_fields_remain_untouched_and_wagtail_image_is_changed(self):
...
+        """A generic test to ensure the image is created properly."""
+        new_image = create_wagtail_image(
+            self.image_path,
+            image_name='fake teddy.jpg',
+            collection_name='pni products'
+        )
+        # Image was created
+        self.assertTrue(bool(new_image))
+        # Image has a collection and is in the proper collection
+        self.assertTrue(new_image.collection_id)
```
Pomax, line 636: I assume `..._id` is a number, not a bool?

network-api/networkapi/tests.py
```diff
@@ -613,3 +616,52 @@ def test_fields_remain_untouched_and_wagtail_image_is_changed(self):
...
+        self.assertEqual(new_image.title, 'teddy.jpg')
+        self.assertEqual(new_image.collection.name, 'Root')
+
+    def test_new_collection(self):
+        new_image = create_wagtail_image(
+            self.image_path,
+            image_name='fake teddy.jpg',
+            collection_name='brand new collection'
+        )
+        self.assertEqual(new_image.collection.name, 'brand new collection')
```
Pomax, line 652: Small nit, but this makes it impossible for typos to break this test in the future:
```suggestion
        collection_name = 'brand new collection'
        new_image = create_wagtail_image(
            self.image_path,
            image_name='fake teddy.jpg',
            collection_name=collection_name
        )
        self.assertEqual(new_image.collection.name, collection_name)
```

network-api/networkapi/tests.py
```diff
@@ -613,3 +616,52 @@ def test_fields_remain_untouched_and_wagtail_image_is_changed(self):
...
+
+    def test_existing_collection(self):
+        NEW_COLLECTION_NAME = 'first collection'
```
Pomax, line 655: this doesn't seem to be a const, probably better as `collection_name`?

Pomax (changes requested):

network-api/networkapi/tests.py
KalobTaulien: Correct, it's a number.

Pomax: Do we want `self.assertIsNotNone(new_image.collection_id)` here?

Pomax (approved):
R+ either with or without the assert update.

PR #6495, opened 2021-04-06
rebuilding the wagtailpages migration

### LINKED ISSUE #6491: migration squashing part 3: generate a fresh migration
Followup to #6381 

We've complete step 1, which replaced the 1-100-squash and additional migrations 2 through 55 into a single file, and we've completed step 2, which removed all the original migrations leaving only the newly squashed single migration file, leaving step 3:

- [ ] regenerate fresh migrations, so we no longer have a migration file saddled with "createmodel" -> "addfield" -> "alterfield" -> "removefield" -> "removemodel" code that ends up running, but having a net result of "nothing"

Note: this may require either some surgery or squashing of PNI, as wagtailpages 0001 relies on PNI migration 0008, while PNI migration 0009 relies on wagtailpages 0001.

Pomax:
Closes #6491
Related PRs/issues #6381 

This replaces the original squash migration with a "from scratch" migration to cover what our code now looks like. Any data migration functions were removed, but the "create default" code for blogs, PNI categories, and areas of focus (as listed on the homepage) were kept.

This also updates a PNI migration, because it had a dependency that... did nothing? But did get in the way of creating this new migration file.

### Testing

- Obvious `inv test-python` should pass
- `inv new-db` should work just fine.
- grab a copy of staging, run migrations, there should be no errors.
  - PNI should still work =D


network-api/networkapi/buyersguide/factory.py
```diff
@@ -43,17 +43,6 @@ def get_extended_yes_no_value():
     return get_random_option(['Yes', 'No', 'NA', 'CD'])
 
 
-def get_lowest_content_category():
```
Pomax, line 46: removed mostly because of test interference.

network-api/networkapi/buyersguide/factory.py
```diff
@@ -75,151 +64,6 @@ class Meta:
     snippet = Faker('sentence')
 
 
-"""
```
Pomax, line 78: this might as well go, if I'm touching this file anyway. It does nothing, and keeping it sure as heck is dirty.

network-api/networkapi/buyersguide/migrations/0009_remove_product_category.py
```diff
@@ -7,7 +7,6 @@ class Migration(migrations.Migration):
 
     dependencies = [
         ('buyersguide', '0008_auto_20210126_1943'),
-        ('wagtailpages', '0001_squashed_0055_auto_20210325_1757'),
```
Pomax, line 10: Removed, because it does nothing, but got in the way of generating a clean new migration for the wagtailpages app.

KalobTaulien (approved):
Everything works as intended, nice job!

PR #6498, opened 2021-04-06
don't restrict wagtail URLs to slugs-only

### LINKED ISSUE #6464: Add support for spaces in PNI categories back in
We took it out, so that resolution on slug is the only way to get to categories, but we're seeing request for the old format, e.g.:

```
Mar 29 00:47:24 foundation-mozilla-org heroku/router at=info method=GET path="/en/privacynotincluded/categories/Toys%20&%20Games/" host=foundation.mozilla.org request_id=f6e92e2c-4c76-42b7-8e0c-f4c1ea866a25 fwd="54.70.40.11,172.69.23.51" dyno=web.2 connect=1ms service=80ms status=500 bytes=1811 protocol=https
```

Alternatively, we could have that at least return the PNI homepage instead of a 500 server error?

Pomax:
Closes #6464

This makes URLs like http://localhost:8000/en/privacynotincluded/categories/health%20&%20exercise/ work again, by making sure that the wagtail URL matcher doesn't only look for slug patterns, but "anything".


KalobTaulien:
A couple tests need some adjustments it seems. Once that's done feel free to merge. I gave an R+ as of now but happy to re-review the test work.

KalobTaulien (approved):

PR #6507, opened 2021-04-07
Remove cloudinary

### LINKED ISSUE #6248: Remove Cloudinary

KalobTaulien:
Closes #6248 

Removes cloudinary code but keeps the cloudinary dependency as a package, which is needed for a couple of migrations. Everything else was removed. 

**Notes**:
We'll want to remove cloudinary environment variables from staging when testing, and then remove them from production when deploying to prod.


Pomax (approved):
so. very. nice.

PR #6523, opened 2021-04-08
Convert highlight images

### LINKED ISSUE #6482: make `highlight` app use wagtail for its images
- [x] rewrite `networkapi/highlights` to use wagtail images instead of S3 file uploads
- [x] create a migration that turns all remote highlight images into wagtail images.
- [ ] change the highlights factory code so that it uses the same random wagtail image code that blogs etc. use

This will unblock https://github.com/mozilla/foundation.mozilla.org/issues/6102

I suspect this will be similar to our PNI image uplift, although nowhere near as likely to go completely wrong =)

KalobTaulien:
Part 1 of 2. 
Partially addresses #6482 
Second PR: #6526

Steps to test:
1. Copy staging database 
2. Set your S3 environment vars (I made sure mine were working by seeing images load on /privacynotincluded/ and checking the img src's)
3. Apply the migration in this PR 
4. Run `inv manage convert_highlight_images`

We need to run `convert_highlight_images` in production before merging in step 2. Step 2 removes the management command and finishes the ticket with a secondary migration. I took the management command route instead of relying on a data migration that could take a while to run. 

Let's roll this out _before_ applying step 2.


Pomax (approved):
this looks like it's working quite well.

PR #6526, opened 2021-04-08
Convert highlight images part 2

### LINKED ISSUE #6482: make `highlight` app use wagtail for its images

KalobTaulien:
Part 2 of 2.
Closes #6482 
Related PRs/issues #6523

* Renames the new image field to the new image field. 
* Removes the management command
* Updates the /api/highlights/ endpoint to continue using an image URL rather than an image ID


Pomax (approved):
as far as I can tell this all works as intended.

KalobTaulien:
Fixed a bunch of stuff. But percy complained about one particular page. @Pomax I'd love if you could take a look and tell me if it's me that broke something or if you know if something else got merged into master and is causing the visual regression to happen.

Pomax:
@KalobTaulien Right, I have NO IDEA what Percy's doing. Locally this all works fine. Even running cypress locally shows things work just fine. So...  I'm going to say let's ignore Percy because this is just bs.

Pomax:
If I use `inv new-db` directly on this branch, then while I cannot reproduce this error through localhost testing, I **CAN** if I run `npm run cypress:test:manual` with a new test file:

STR
- create a file called `what-you-can-do.js` in the cypress integration dir, with the following content:

```js
const FIXED_DATE = new Date(2019, 1, 1).getTime();

describe(`Visual regression testing for foundation.mozilla.org`, () => {
  beforeEach(function () {
    cy.clock(FIXED_DATE);
  });

  it(`What can you do page`, function () {
    cy.visit(`/en/what-you-can-do/`);
    cy.window().its(`main-js:react:finished`).should(`equal`, true);
    cy.wait(500);
    cy.screenshot();
  });
});
```
- run `npm run cypress:install` (if not installed) followed by `npm run cypress:test:manual`.
- In the Cypress application, run the "what you can do" test
- observe the page it renders. If this happens too fast, it's screenshooting it to the cypress/screenshots dir, so you can reivew it there.


I get the following page render result:

Visual regression testing for foundation mozilla org -- What can you do page\

So that's in line with what Percy's seeing.

Pomax:
Using a long-timout in the cypress test, the DOM for these images is:

image

but looking at that image, the asset dimensions are 100x58. Does the `fill-350x197` force-upscale, or does it leave that image at 58px height? If so, that explains the problem.

Pomax:
- run `docker-compose up`
- load http://localhost:8000/en/what-you-can-do in Chrome: works fine.
- load http://localhost:8000/en/what-you-can-do in Firefox: works fine
- load http://localhost:8000/en/what-you-can-do in Cypress: broken.

However, if  I explicitly tell Cypress to use my Chrome or my Firefox, it works.

So I'm starting to wonder if Percy is just maybe using ancient browsers?

Pomax:
Let's try this again, with cypress updated to 6.9

Pomax (changes requested):
Let's redo this PR, because it's just too weird. Let's see if we can split this PR up into a way that the image url stays `highlight.image.url` and see if _that_ works.

Pomax:
I think this is the straw to break the camel's back: trying to figure what's even going wrong on Percy's side is a sysyphean process and it's costing us more to try to figure this out than it does to just roll visual diffing that can run as a github action.

https://github.com/mozilla/foundation.mozilla.org/issues/6643
https://github.com/MozillaFoundation/ci-image-diff

Pomax:
So things are finally behaving, although it looks like the blog author factory needs a fix - even with `RANDOM_SEED` clamped, blogs seem to be generating with different authors

Pomax (approved):
finally

PR #6535, opened 2021-04-12
Add Cloudflare cache invalidation support

tomusher:
Adds support for Cloudflare cache invalidation so we can get caching enabled


KalobTaulien (approved):
LGTM. Thanks for adding this Tom!

KalobTaulien (changes requested):
Looks like tests aren't passing. Havent looked into why yet. But we should dig into that.

tomusher:
I don't think there's anything in this change that should be causing these tests to fail - any idea what might be going on?

Pomax:
Testing the failing tests "with real data on the running server" does not seem to run into any trouble (vote posting through a product page is HTTP 200 for example), what are these two tests doing...

KalobTaulien (approved):

Pomax (approved):

PR #6546, opened 2021-04-14
Add caching to RSS/Atom feeds

### LINKED ISSUE #6374: Add caching for the atom/rss feeds to drive down their response times
Similar to #6371 but for rss/atom feeds:

image

Pomax:
Closes #6374 

This adds caching to the RSS/Atom `items()` function so that we only hit the database once a day for the current list of blog posts to include in our feed. It's self-invalidating, with the value defaulting to one day but overridable using a `FEED_CACHE_TIMEOUT` environment variable.

I tested this by setting that var to `10`, with a `print('caching')` inside the `if feed_set is None: ...` block so I could see it building or hitting the cache by reloading the RSS page on http://localhost:8000/en/blog/rss/

### DIFF
--- a/network-api/networkapi/wagtailpages/rss.py
+++ b/network-api/networkapi/wagtailpages/rss.py
@@ -1,4 +1,5 @@
 from django.conf import settings
+from django.core.cache import cache
 from django.contrib.syndication.views import Feed
 from django.utils.feedgenerator import Atom1Feed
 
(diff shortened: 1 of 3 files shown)


network-api/networkapi/wagtailpages/rss.py
```diff
@@ -16,18 +17,27 @@ class RSSFeed(Feed):
...
 
-        # If that doesn't yield the blog page, pull using the universal title
-        if index is None:
-            index = IndexPage.objects.get(title__iexact='Blog')
+        if feed_set is None:
+            # If we don't have an active cache, we build one: pull the index page
+            # specifically using the English page title, as an IndexPage rather than
+            # as a BlogIndexPage, to make sure we're not filtering out all the
+            # "featured" posts (which we need to do for site content purposes).
+            index = IndexPage.objects.get(title_en__iexact='Blog')
```
KalobTaulien, line 28: With .get() if it doesn't find the index page it'll throw a `IndexPage.DoesNotExist` error. We'll want to `try/catch` this just in case the page titled "Blog" has its title changed in the future.

network-api/networkapi/wagtailpages/rss.py
```diff
@@ -16,18 +17,27 @@ class RSSFeed(Feed):
...
 
-        blog_pages = index.get_all_entries().order_by('-first_published_at')
+            # If that doesn't yield the blog page, pull using the universal title
+            if index is None:
+                index = IndexPage.objects.get(title__iexact='Blog')
```
KalobTaulien, line 32: Same same here. We want to avoid `IndexPage.DoesNotExist` exceptions.

KalobTaulien (changes requested):
Looks mostly good! Just need to catch a couple possible exceptions. Caching looks nice and easy to control!

KalobTaulien (approved):

PR #6552, opened 2021-04-15
Added option for subscribing to newsletter through basket-client instead of SQS

### LINKED ISSUE #6544: Switch Petition newsletter signup to Basket API instead of SQS
Let's use the Python Basket Client if possible

https://pypi.org/project/basket-client/
https://basket-client.readthedocs.io/en/latest/

danielfmiranda:
Closes #6544

**Link to home page (to test newsletter sign up)**: https://foundation-s-petition-b-phogx2.herokuapp.com/en/ 
**Link to petition page**: https://foundation-s-petition-b-phogx2.herokuapp.com/en/campaigns/story-provide-least-training-new-difficult/

## Checklist
**Changes in Models:**


dev-requirements.txt
```diff
@@ -8,6 +8,11 @@ certifi==2020.4.5.1
     # via
     #   -c requirements.txt
     #   requests
+    #   urllib3
```
danielfmiranda, line 11: Updated when I added basket-client to the requirements.

.github/workflows/continous-integration.yml
```diff
@@ -45,6 +45,7 @@ jobs:
         options: --health-cmd pg_isready --health-interval 10s --health-timeout 5s --health-retries 5
     env:
       ALLOWED_HOSTS: localhost,mozfest.localhost,default-site.com,secondary-site.com
+      BASKET_URL: https://basket-dev.allizom.org/
```
danielfmiranda, line 48: Added these so they can be used during testing wagtail CI.

requirements.txt
```diff
@@ -10,6 +10,8 @@ anyascii==0.1.7
     # via wagtail
 asgiref==3.3.1
     # via scout-apm
+basket-client==1.0.0
```
danielfmiranda, line 13: updated when I added basket-client

network-api/networkapi/settings.py
```diff
@@ -114,6 +119,12 @@
 # Build paths inside the project like this: os.path.join(BASE_DIR, ...)
 BASE_DIR = root()
 
+# Basket client configuration
+BASKET_URL = env('BASKET_URL')
```
danielfmiranda, line 123: The basket URL is set to use the staging environment of basket in env.default.

We dont need the basket API key at the moment, but I left it here as Kalob mentioned it might be a good idea in case we need it in the future. 

The subscribe method in env.default is defaulting to BASKET

network-api/networkapi/campaign/views.py
```diff
@@ -194,6 +203,13 @@ def petition_submission(request, petition):
         }
     })
 
+    if settings.MOFO_NEWSLETTER_SUBSCRIBE_METHOD == 'BASKET':
```
danielfmiranda, line 206: If I understood our conversation with Pmac correctly, we just want to subscribe them to the newlsetter through basket, and then send their petition information to SQS which is what this block of code doing.

@benhohner was wondering if you could please confirm that this is the desired flow :)

Pomax:
Is this still WIP? If so, you probably don't want to set reviewers quite yet. If not, you probably want to remove "WIP" from the title =D

danielfmiranda:
Thank you for the catch Pomax! 😆 Removed WIP from the title

env.default
```diff
@@ -28,6 +28,15 @@ X_FRAME_OPTIONS=DENY
 XSS_PROTECTION=True
 XROBOTSTAG_ENABLED=False
 
+#basket
```
benhohner, line 31:
```suggestion
# Basket config:
```

env.default
```diff
@@ -28,6 +28,15 @@ X_FRAME_OPTIONS=DENY
...
+
+# See the Basket client docs https://basket-client.readthedocs.io/en/latest/install.html
+BASKET_URL=https://basket-dev.allizom.org/
```
benhohner, line 34: So on Prod we'll have to set this environment variable to the basket prod URL?

env.default
```diff
@@ -28,6 +28,15 @@ X_FRAME_OPTIONS=DENY
...
+BASKET_API_KEY=
+MOFO_NEWSLETTER_SUBSCRIBE_METHOD=BASKET
```
benhohner, line 36: Can you move this to a new section that says something like `# Newsletter config:` as it's not really related to Basket specifically?

network-api/networkapi/campaign/views.py
benhohner: Correct, this is the desired flow!

benhohner, line 206:
```suggestion
    if settings.MOFO_NEWSLETTER_SUBSCRIBE_METHOD == 'BASKET' \
        and request.data['newsletterSignup'] is True:
```
I believe this is the correct way to combine the two if statements into one statement while following the [PEP8 79 character line limit](https://www.python.org/dev/peps/pep-0008/#maximum-line-length). I believe linter should automatically do the right thing.

env.default
```diff
@@ -28,6 +28,15 @@ X_FRAME_OPTIONS=DENY
 XSS_PROTECTION=True
 XROBOTSTAG_ENABLED=False
 
+#basket
+
+# See the Basket client docs https://basket-client.readthedocs.io/en/latest/install.html
+BASKET_URL=https://basket-dev.allizom.org/
+BASKET_API_KEY=
```
benhohner, line 35:
```suggestion
```
I know we agreed to add this originally, but I don't think we'll ever need this in donate, so it's probably good to remove.

network-api/networkapi/settings.py
```diff
@@ -114,6 +119,12 @@
...
+BASKET_API_KEY = env('BASKET_API_KEY')
```
benhohner, line 124:
```suggestion
```
I know we agreed to add this originally, but I don't think we'll ever need this in donate, so it's probably good to remove.

benhohner (changes requested):
Looks good. Couple minor adjustments. Great work Daniel!

network-api/networkapi/campaign/views.py
```diff
@@ -194,6 +203,13 @@ def petition_submission(request, petition):
...
+        if request.data['newsletterSignup'] is True:
```
benhohner, line 207:
```suggestion
```

benhohner (changes requested):
I missed one thing, not sure how to add to the above review. Sorry, hopefully this isn't too confusing.

network-api/networkapi/campaign/views.py
danielfmiranda: Hi @benhohner, just double checking, this line is here because when a user is submitting a petition they are given the choice to additionally sign up to the newsletter. 

So the 2nd if statement is there to say, if the user submits a petition, and we switched over to basket, and they clicked the additional "sign me up for the newsletter button", it will then subscribe them using basket and then send their petition info to SQS. 

If they sign up for the petition, but do not sign up for the newsletter, it will just send their petition sign up to SQS :)

Would you still like me to remove this line?

env.default
danielfmiranda: Thank you Ben!! 🙌

network-api/networkapi/campaign/views.py
```diff
@@ -214,6 +230,7 @@ def send_to_sqs(sqs, queue_url, message, type='petition'):
         }, 201)
 
     try:
+
```
benhohner, line 233:
```suggestion
```

benhohner (approved):
R+ Looks good!

PR #6563, opened 2021-04-20
Moved tests

### LINKED ISSUE #6440: Move buyersguide app tests.py to wagtailpages app tests file
Currently the buyersguide app has a tests.py file. We want to eventually delete this app entirely, so we need to move the buyersguide/tests.py file to wagtailpages/tests.py. 

Imports will need to be updated. But other than that, this should be a straight forward ticket.

KalobTaulien:
Closes #6440

Just needed to move tests file out of the buyersguide app. Literally copied and pasted it into the wagtailpages app and that's it.


danielfmiranda (approved):
Hi Kalob! Looks good to me, approving!

PR #6568, opened 2021-04-22
Remove CSRF from petition and signup forms. 

### LINKED ISSUE #6566: Make CSRF tokens unneeded for petition signups
> This works on production, but on staging it doesn't

We need to make sure a petition signup form doesn't require a csrf token when submitting a petition. 

Example url: https://foundation.mozilla.org/en/campaigns/hulu-its-time-ad-disclosure/

KalobTaulien:
Closes #6566 

We apparently cant use csrf exempt views with drf's api_view. So I opted for regular Django views that return a JsonResponse instead of a drf.Response. 

Notes:

- you will need to add a `campaign_id` value when manually testing petition/signup CTA, and make sure your `.env` is up to date with a `BASKET_API_ROOT_URL`  set to https://basket-dev.allizom.org/
- you will also need to set the `newsletter` field for your signup snippet to `mozilla-foundation`

### DIFF
--- a/network-api/networkapi/campaign/views.py
+++ b/network-api/networkapi/campaign/views.py
@@ -1,4 +1,3 @@
-from rest_framework.response import Response
 from rest_framework import status
 from django.core.exceptions import ObjectDoesNotExist
 from django.views.decorators.csrf import csrf_exempt
@@ -58,9 +57,10 @@ logger = logging.getLogger(__name__)
 def signup_submission_view(request, pk):
     # We need to re-write the data that's coming in from the XMLHttpRequest.
     # XMLHttpRequest's send data through the request.body, not request.POST despite it being a POST method
+    # request.POST is supported for unit tests
     new_body = request.body.decode("utf-8")
     request.data = json.loads(new_body)
-    # TODO: Test this view
+
     try:
         signup = Signup.objects.get(id=pk)
     except ObjectDoesNotExist:
@@ -78,13 +78,16 @@ def signup_submission_view(request, pk):
 def petition_submission_view(request, pk):
     # We need to re-write the data that's coming in from the XMLHttpRequest.
     # XMLHttpRequest's send data through the request.body, not request.POST despite it being a POST method
-    new_body = request.body.decode("utf-8")
-    request.data = json.loads(new_body)
-    # TODO: Test this with working sqs env vars from staging or RA apps
+    # request.POST is supported for unit tests
+    if request.POST:
+        request.data = request.POST
+    else:
+        new_body = request.body.decode("utf-8")
+        request.data = json.loads(new_body)
     try:
         petition = Petition.objects.get(id=pk)
     except ObjectDoesNotExist:
-        return Response(
+        return JsonResponse(
             {'error': 'Invalid petition id'},
             status=status.HTTP_400_BAD_REQUEST,
         )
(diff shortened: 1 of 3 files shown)


network-api/networkapi/campaign/views.py
```diff
@@ -52,10 +52,15 @@ def send_message(self, QueueUrl, MessageBody):
...
 
-@api_view(['POST'])
-@parser_classes((JSONParser,))
-@permission_classes((permissions.AllowAny,))
+@csrf_exempt
+@require_http_methods(['POST'])
 def signup_submission_view(request, pk):
+    # We need to re-write the data that's coming in from the XMLHttpRequest.
+    # XMLHttpRequest's send data through the request.body, not request.POST despite it being a POST method
+    # request.POST is supported for unit tests
```
Pomax, line 60: `XMLHttpRequest` was replaced by the Fetch API a long time ago (in web years), let's call this a "network request" instead.

network-api/networkapi/campaign/views.py
```diff
@@ -52,10 +52,15 @@ def send_message(self, QueueUrl, MessageBody):
...
+    new_body = request.body.decode("utf-8")
+    request.data = json.loads(new_body)
```
Pomax, line 62: Note that `json.loads` can throw a `ValueError`, so we'll want to try/except this.

network-api/networkapi/campaign/views.py
```diff
@@ -68,14 +73,21 @@ def signup_submission_view(request, pk):
...
 
-@api_view(['POST'])
-@parser_classes((JSONParser,))
-@permission_classes((permissions.AllowAny,))
+@csrf_exempt
+@require_http_methods(['POST'])
 def petition_submission_view(request, pk):
+    # We need to re-write the data that's coming in from the XMLHttpRequest.
+    # XMLHttpRequest's send data through the request.body, not request.POST despite it being a POST method
+    # request.POST is supported for unit tests
```
Pomax, line 81: same comment note.

network-api/networkapi/campaign/views.py
```diff
@@ -68,14 +73,21 @@ def signup_submission_view(request, pk):
...
+    if request.POST:
+        request.data = request.POST
+    else:
+        new_body = request.body.decode("utf-8")
+        request.data = json.loads(new_body)
```
Pomax, line 86: same `except` note

benhohner (approved):
R+

network-api/networkapi/campaign/views.py
```diff
@@ -145,9 +157,8 @@ def signup_submission(request, signup):
 # handle Salesforce petition data
 def petition_submission(request, petition):
     cid = petition.campaign_id
-
     if cid is None or cid == '':
-        return Response(
+        return JsonResponse(
             {'error': 'Server is missing campaign for petition'},
```
Pomax, line 162: let's make this `campaign id`

Pomax (changes requested):
let's make sure we're safe.

network-api/networkapi/campaign/views.py
KalobTaulien: It's _actually_ using an XmlHttpRequest in the JavaScript though. But I can update, Im not picky in that.

network-api/networkapi/campaign/views.py
KalobTaulien: Good catch!

danielfmiranda (approved):
Hi Kalob,

Thanks again for taking the time to handle this, as I know this was a interesting problem to work on. 

After testing, looks like everything is working to me! So I am approving, thanks!

PR #6574, opened 2021-04-23
quote unsafe-inline targets

Pomax:
No related issue: this quotes the `unsafe-inline` value, which should have quotes, and has quotes on staging, and production, but we forgot to add for our `env.default` file.


danielfmiranda (approved):
Hi Pomax, I reviewed the code that you updated, as well as the review app and the changes in percy. 

Everything looks good to me! I am approving 👍

PR #6584, opened 2021-04-26
Fixed erroring tests

KalobTaulien:
We just needed Dummy cache instead of wagtail's base layer of caching. 

I also used .get(depth=1) to select the first page in treebeard instead of getting the first home page. It's just a tad more robust this way.


tomusher:
Great! Should we use this cache on the entire test suite?

PR #6601, opened 2021-04-27
updated petition signing data to have everything under the data label 

### LINKED ISSUE #6600: Update Petition Data sent to SQS format
### Describe the feature request
Shawn Collins requested that instead of having the petition submission data in its original format, where the users information can be found in the form object:
image

He would like everything nested under the "data" object like so: 
image

danielfmiranda:
Closes #6600 
**Link to sample test page**:

Example of data now being sent, note that everything is under the "Data" object instead of nested one level deeper in a "form" object. 
image

## Checklist

**Changes in Models:**


network-api/networkapi/campaign/views.py
```diff
@@ -172,13 +172,16 @@ def petition_submission(request, petition):
         )
 
     data = {
+        "json": True,
```
danielfmiranda, line 175: Since "json" will always = True and "event_type" will always = "crm_petition_data",
I moved these two lines into the "data" object, that way when we send the message to SQS it will be all one flattened object like requested.

network-api/networkapi/campaign/views.py
```diff
@@ -215,11 +218,7 @@ def petition_submission(request, petition):
     message = json.dumps({
         'app': settings.HEROKU_APP_NAME,
         'timestamp': datetime.now().isoformat(),
-        'data': {
```
danielfmiranda, line 218: Please see comment above

benhohner: Looks good, but can you please confirm with Shawn that the app and timestamp fields are fine being outside of the data object?

benhohner (approved):

network-api/networkapi/campaign/views.py
Pomax: except now you've changed the object that gets sent. Note that the original code sends a payload fo the form:
```
{
  app: settings.HEROKU_APP_NAME,
  timestamp: datetime.now().isoformat(),
  data: {
    json: true,
    form: {...lots of key/value pairs...},
    event_type: 'crm_petition_data'
  }
}
```
That is, it's an object with three keys, that inform Basket what format the `form` property has (it will be a JSON serialized object), and which event type this is.

With your change, we would now be sending:
```
{
  app: settings.HEROKU_APP_NAME,
  timestamp: datetime.now().isoformat(),
  data: {
    json: true,
    /* form is no longer defined */
    event_type: 'crm_petition_data',
   ...lots of key/value pairs...
  }
}
```

Pomax (changes requested):

Pomax (approved):

danielfmiranda:
@benhohner I reached out to Shawn via the same email thread that he requested the data changes in. Going to wait for his confirmation before merging in this branch!

benhohner:
> @benhohner I reached out to Shawn via the same email thread that he requested the data changes in. Going to wait for his confirmation before merging in this branch!

Thanks Daniel. Can you also please make sure that the original data format is being sent until we change the feature flag? We need to keep sending the original format to basket on prod until the prod switchover on May 25th. Not sure if this is done already or not.

danielfmiranda:
Hi @benhohner, 

Just added the feature flag functionality and the app is ready for re-review. Please let me know if you have any questions! Thank you

benhohner (approved):
Looks great! R+

Issue #6463, opened 2021-03-30
Improve the blog category/tag filter code to prevent n+1 warnings


Pomax:
The "code improvement" part of https://github.com/mozilla/foundation.mozilla.org/issues/6371

Rewrite the category/tag selection code for blogs so that it's more efficient and leads to fewer n+1 errors:

![screenshot 715](https://user-images.githubusercontent.com/177243/111192590-775ab300-8576-11eb-9c71-721649924ee7.png)

Pomax:
reopened due to https://github.com/mozilla/foundation.mozilla.org/pull/6606

PR #6602, opened 2021-04-27
Add caching for index pages (with invalidation on child del/pub/unpub)

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6463

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -55,6 +55,7 @@ env = environ.Env(
     HEROKU_BRANCH=(str, ''),
     HEROKU_PR_NUMBER=(str, ''),
     HEROKU_RELEASE_VERSION=(str, None),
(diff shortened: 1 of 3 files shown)


network-api/networkapi/settings.py
```diff
@@ -55,6 +55,7 @@
     HEROKU_BRANCH=(str, ''),
     HEROKU_PR_NUMBER=(str, ''),
     HEROKU_RELEASE_VERSION=(str, None),
+    INDEX_PAGE_CACHE_TIMEOUT=(int, 60*60*24),
```
Pomax, line 58: With cache invalidation added, this might actually be too generous, but it's an env var, so we can always play with making the timeout longer via heroku

KalobTaulien (approved):
This looks great to me, thanks for adding in the hooks to invalidate the cache!

Issue #6604, opened 2021-04-27
PNI search has broken


Pomax:
Go to https://foundation.mozilla.org/en/privacynotincluded/, try to search

I'm seeing queryselector errors around a `btn-recommend`, which might be preventing the actual search code from having loaded.

Estimating as 1 based on the fact that we don't _actually_ know what's wrong yet.

PR #6605, opened 2021-04-27
fix PNI search filter

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6604

Fixes the PNI search filter by making sure that we don't just set `d-none` to hide elements but also remove the `d-flex` class (otherwise, `d-flex` wins, and `... d-flex ... d-none ...` will stay visible)


KalobTaulien (approved):
Glad this was something easy and not something crazy. R+. Feel free to merge when CI passes.

PR #6608, opened 2021-04-28
cache.delete

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6463 this time with the proper cache delete call, and without treating `.specific` as a function

testing steps:
- fire up this branch
- add `print("clearing cache!")` at the same depth as cache.delete
- edit a blog page, then publish it.


KalobTaulien (approved):
LGTM. STR worked as expected. No 500s on my end.

PR #6611, opened 2021-04-29
Fix PNI UI l10n

### LINKED ISSUE #6610: PNI UI translations are broken
Similarly to #6512, UI translations are missing and we need to add the `buyersguide` template directory to the `LOCALE_PATH` var.

This is happening because the `locale` folder is no longer at the root of a package listed in `INSTALLED_APPS`, so it needs to be explicitly added to `LOCALE_PATH`.

TheoChevalier:
Closes #6610

Prod homepage in French:
Capture d’écran 2021-04-29 à 11 34 02

French homepage on https://foundation-s-fix-pni-ui-x9lzo0.herokuapp.com/fr/privacynotincluded/:

Capture d’écran 2021-04-29 à 12 24 05


Pomax (approved):
nice catch!

PR #6632, opened 2021-05-04
Localize Newsletter strings in Take action section

### LINKED ISSUE #6631: Localize newsletter form in Take Action section
The newsletter sign up form in the Take Action section of the homepage has a header and description coming from a template, so we can easily localize this.

TheoChevalier:
Closes #6631

Realized this one was not JSX content, so we should totally translate it using our regular process


Pomax (approved):
oh man, good catch O_o

PR #6636, opened 2021-05-04
Removed basket related tests

### LINKED ISSUE #6635: Basket test get rate limited: do we still need basket testing
Now that basket is handled by its own python library, do we still need tests? As long as our data shape is correct, we should now be able to trust that the code we call works, becuase the library has its own tests?

This is slightly pressing as basket's rate limiting causes CI to fail, which means our PRs become blocked.

/cc @KalobTaulien @benhohner @danielfmiranda

KalobTaulien:
Closes #6635 

I deleted the entire test file because each of the tests will inevitably hit basket at the end of each request, and we were getting rate limited on it. Because basket is reasonably well covered for tests, we don't need these tests anymore.


Pomax (approved):
pretty much the most direct solution

PR #6684, opened 2021-05-12
updated task name and all references, also added alias

### LINKED ISSUE #6633: turn `create-super-user` into `createsuperuser`
### File(s) or components

tasks.py

### Reason for cleanup

The Django command is `python manage.py createsuperuser` so our tasks.py file using `create-super-user` is a little odd, and we should probably rename it to `createsuperuser`, with a `create-super-user` alias so that command still works.

danielfmiranda:
Closes #6633

Updated the task's name and references from "create-super-user" to "createsuperuser".

Also added an alias to call the command with "create-super-user" like requested.

## Checklist

**Changes in Models:**

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -68,8 +68,8 @@ def l10n_block_inventory(ctx):
     manage(ctx, "block_inventory")
 
 
-@task
-def create_super_user(ctx):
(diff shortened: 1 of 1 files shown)


Pomax (approved):
hurray!

PR #6689, opened 2021-05-12
Use dynamic image urls to decrease time to first byte

### LINKED ISSUE #6626: Reduce image and rendition queries on the wagtailpages/homepage.html template.
Scout is reporting lots of wasted time looking for images and image renditions on the homepage. 

A few things to look into, in a prioritized list: 
- [ ] [Dynamic images](https://docs.wagtail.io/en/stable/advanced_topics/images/image_serve_view.html) - we do this one the PNI Homepage already 
- [ ] Reducing the number of image renditions we actually need. Remember, each rendition = 1 query OR if the image rendition size doesn't exist yet then it's going to process the image (crop it). We want to reduce this time as it makes time to first byte take longer. 
- [ ] If all else fails, lets look into template caching entire sections of the home page. 
  - [ ] We'll need to invalidate the cache when a related page is changed (using a wagtail hook most likely) and invalidate the cache when the homepage is saved.

:point_up: Explore each option one by one, see above. We may not need all options.

KalobTaulien:
Closes #6626 

### Things to try:
- [x] [Dynamic images](https://docs.wagtail.io/en/stable/advanced_topics/images/image_serve_view.html) - we do this one the PNI Homepage already 
- [ ] Reducing the number of image renditions we actually need. Remember, each rendition = 1 query OR if the image rendition size doesn't exist yet then it's going to process the image (crop it). We want to reduce this time as it makes time to first byte take longer. 
- [ ] If all else fails, lets look into template caching entire sections of the home page. 
  - [ ] We'll need to invalidate the cache when a related page is changed (using a wagtail hook most likely) and invalidate the cache when the homepage is saved.

I've implemented Dynamic image URLs on the homepage in several spots (not all due to some other issues that came up). But this should, hopefully, reduce page load times drastically and defer the image rendition and process to a new HTTP request. 

If this passes CI we should implement this in `master` and see how it goes. If the homepage is still slow, we'll reduce image <source>'s being generated (which won't speed things up too much) and add some caching as well. If we take this route, it'll go in its own ticket. 

🤞 Hope Percy doesn't hate this!

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html
@@ -7,8 +7,7 @@
   {% for author in authors %}
     <div class="author-image d-flex align-items-center justify-content-center {% if author.image is None %}no-avatar{% endif %}" aria-label="{{ author.name }}">
     {% if author.image %}
-      {% image author.image fill-100x100 as img %}
-      <img src="{{ img.url }}" alt="{{ author.name }}" title="{{ author.name }}" />
+      <img src="{% image_url author.image "fill-100x100" %}" alt="{{ author.name }}" title="{{ author.name }}" />
     {% endif %}
     </div>
   {% endfor %}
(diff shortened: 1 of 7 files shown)


Pomax:
I almost don't believe it, but: 
image

Pomax (approved):
:shipit:

PR #6718, opened 2021-05-18
updated the test url list to all urls

### LINKED ISSUE #6717: [ci testing] update URL list to all the URLs
We're currently testing only a handful of URLs, time to switch over to all the URLs in our urls in our integration endpoints file and see if things are still performant.

Pomax:
Closes #6717


Pomax:
So this needs an update to `ci-image-diff`, where the code runs but puts all the actual network request function calls into a `request_queue` with a small size, and then once the full request queue's been built, we `join()` that so that requests occur at a reasonable rate.

Pomax:
followup: check if the github action can do a review approval/change request, so that PRs are block unless a "request changes" gets dismissed.

Pomax:
This seems to be running unacceptable slow with a concurrency queue of 10

Pomax:
updated ci-image-diff to use async batching instead of trying to make its idiotic semaphore system work. Rerunning...

Issue #6702, opened 2021-05-17
Step 1. Upgrade Django ato Django 3, and Wagtail to 2.12


KalobTaulien:
## Acceptance criteria

**Note** This will need to be 2 PRs. 

- [ ] Upgrade Wagtail to 2.12.* and upgrade Django to 3.0.*. Keep django-modeltranslation and wagtail-modeltranslation. We'll remove these later when upgrading to Wagtail 2.13 and Django 3.2.

KalobTaulien:
I split this into a second PR to house the Collection fix on its own. It can be found at #6720 and is blocking this ticket.

PR #6721, opened 2021-05-19
Collection fix via data migration

### LINKED ISSUE #6720: Fix multiple root Collections
When multiple Collections have a `depth=1` then none show up in the settings. 

And there's a wagtail 2.12 migration that looks for the root collection using `.get(depth=1)` but it returns more than one in our code base. We need to make sure there's only one. 

Luckily, it looks like the _other_ Collections that aren't `root` (`id=1`) are factory generated Collections. 

## Acceptance criteria
- [ ] Create a data migration to delete unused Collections with a depth of 1 (that aren't root, we _need_ the root collection for django-treebeard happiness)

KalobTaulien:
Closes #6720 
Related PRs/issues #6702 

**Note**: This migration _needs_ to be run in production before #6702 can be applied. #6702 will fail if the watail migrations are run before the migration in this PR. 

Steps to test: 
1. Checkout this branch
2. Pull down the staging database
3. Run `inv manage clean_collections` 

All migrations should run. And if you go to http://localhost:8000/cms/collections/ you should see our collections show up the way they're supposed to. 

Let's also double check that the Collections that end up being deleted in the 0008 migration are the in fact factory collections, and not actually being used. As far as I was able to tell today, this is wiping out unused collections from staging's database.


network-api/networkapi/wagtailpages/migrations/0008_collection_fix.py
```diff
@@ -0,0 +1,18 @@
+# Generated by Django 3.0.14 on 2021-05-17 22:53
+from django.db import migrations
+
+
+def delete_factory_collections(apps, schema_editor):
+    Collection = apps.get_model('wagtailcore', 'Collection')
+    Collection.objects.filter(depth=1).exclude(id=1).delete()
```
Pomax, line 7: I'd feel more comforatble if this grabbed every collection with `id` > 1, moved any images in side of it into the id=1 collection (if any), and then deleted each collection once empty, instead.

Pomax (changes requested):

network-api/networkapi/management/commands/clean_collections.py
```diff
@@ -0,0 +1,33 @@
...
+
+class Command(BaseCommand):
+    help = 'Remove unused Collections with a depth of 1, and move respective images to the root collection'
+
+    def handle(self, *args, **options):
+        root = Collection.objects.get(name='Root')
+
+        # Let's confirm root is ID #1, as it should be just to be safe.
+        if root.id != 1:
+            print("Incorrect root Collection ID. Stopping this script to prevent any mistakes")
```
Pomax, line 15: nice

Pomax (approved):

Issue #4290, opened 2020-03-04
Language switcher doesn’t work in some cases


TheoChevalier:
There’s a specific case in which the language switcher stops working. It works a first time, but then when you land on a different locale again (e.g. you clicked on a hardcoded /en link), you can’t switch back.

Here are the STR:

- Set French as your preferred language for Web content (it would work with any locale we support)
- Open `foundation.mozilla.org/en` or any page in English
- Use the switcher to pick French
- You get redirected to /fr as expected.
- Now open `foundation.mozilla.org/en` again (or any page in English)
- Try to switch to French

→ You are not redirected

mmmavis:
I'm not familiar with the i10n setup so it'll take me some time to figure this out. Since this is a unique case and that I don't have the capacity to work on it this week, I'm going to put this in the next milestone. @Pomax do you know what might be causing this and how complicated it will be to fix it?

TheoChevalier:
Fixed by #6765

PR #6765, opened 2021-05-27
Fix the language picker

### LINKED ISSUE #6710: Step 8. Make the footer language picker work.
This may or may not be a problem, I'm not entirely sure. 

I'm going to assume we're going to run into things we need to work on for this task. This task is largely unknown and could possibly _not_ be a problem, or may be a hard one to solve.

## Acceptance criteria

- [x] Change the language picker in the footer of the website to properly set the language of the site

Pomax:
Closes #6710 by making sure that the `next` form field points to the same page, but with the locale stripped off, so that Django can correctly further-redirect  from unlocalized to "localized, but for the new prefered locale".

This may also address https://github.com/mozilla/foundation.mozilla.org/issues/4290

Testing:

- load up https://foundation-s-wagtail-lo-6oyjbi.herokuapp.com/
- switch to "nederlands" with the footer picker
- notice the page text is now different (there is an nl localised homepage in the cms).
- now switch to "english" with the footer picker
- notice the page is back to `/en/`

### DIFF
--- a/network-api/networkapi/templates/fragments/language_switcher.html
+++ b/network-api/networkapi/templates/fragments/language_switcher.html
@@ -4,7 +4,6 @@
 {% get_local_language_names as languages %}
 
 <form action="{% url 'set_language' %}" method="post">
-  {% csrf_token %}
   <input name="next" type="hidden" value="{% get_unlocalized_url page current_language %}" />
   <div class="d-flex flex-column flex-md-row align-items-md-center justify-content-center mb-md-0">
     <label class="h5-heading mb-0 d-flex align-items-center globe-glyph medium" for="language-switcher">{% trans "Language" %}</label>
--- a/network-api/networkapi/templates/fragments/language_switcher.html
+++ b/network-api/networkapi/templates/fragments/language_switcher.html
@@ -1,15 +1,17 @@
-{% load i18n localization %}
+{% load i18n localization wagtailcore_tags %}
+
+{% get_current_language as current_language %}
+{% get_local_language_names as languages %}
 
 <form action="{% url 'set_language' %}" method="post">
-  <input name="next" type="hidden" value="{{ request.path }}">
+  {% csrf_token %}
+  <input name="next" type="hidden" value="{% get_unlocalized_url page current_language %}" />
   <div class="d-flex flex-column flex-md-row align-items-md-center justify-content-center mb-md-0">
     <label class="h5-heading mb-0 d-flex align-items-center globe-glyph medium" for="language-switcher">{% trans "Language" %}</label>
     <select name="language" id="language-switcher" class="mt-3 mt-md-0 ml-md-3 w-100 form-control" onchange="this.form.submit()">
-      {% get_current_language as LANGUAGE_CODE %}
-      {% get_local_language_names as LANGUAGES %}
-      {% for CODE, NAME in LANGUAGES %}
-        <option value="{{ CODE }}"{% if CODE == LANGUAGE_CODE %} selected{% endif %}>
-          {{ NAME | capfirst }}
+      {% for language_code, language_name in languages %}
+        <option value="{{ language_code }}"{% if language_code == current_language %} selected{% endif %}>
+          {{ language_name | capfirst }}
         </option>
       {% endfor %}
     </select>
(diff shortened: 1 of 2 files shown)


Pomax:
A potential bug: 
- I added Dutch as locale
- I clicked "translate this page" for the homepage
- this page has an EN field `hero_headline` with content `Plant bad attention painting clearly. Hold off allow maybe significant stop.`, which is too long, but the `load_fake_data` process doesn't care.
- I exported the strings as .po file and added a dutch translation:
```
msgctxt "hero_headline"
msgid "Plant bad attention painting clearly. Hold off allow maybe significant stop."
msgstr "Plant slecht attentieschilderij duidelijk. Wacht sta toe misschien stop."
```
- uploading this .po file and publishing the page leads to an error, claiming the field content is too long. 
- I made the string drastically shorter, but it still complains:

{5C23F111-2458-4457-A107-68C7D05C77A4}

- I figured I'd blank the string again, so that it doesn't _have_ a translation:

```
msgctxt "hero_headline"
msgid "Plant bad attention painting clearly. Hold off allow maybe significant stop."
msgstr ""
```
- uploading this .po file and publishing _does not remove the translation_

So it feels like there might be several bugs here.

TheoChevalier:
> uploading this .po file and publishing does not remove the translation and preserves the last translation.

I suspect uploaded .po files are treated differently and are _merged_ only, while actually deleting a translation from Pontoon will work as intended.
Here you should be able to just click the Edit button, remove the translation and hit save + publish


> I went to the original page and updated the field to have a shorter string, and published that.

You need to sync the page after publishing to propagate the change (this allows for copy review before content gets sent for translation)

TheoChevalier (approved):
Did further testing (and added other locales for the homepage, too), and I can confirm this fixes #4290 🎉

network-api/networkapi/templates/fragments/language_switcher.html
```diff
@@ -1,15 +1,17 @@
-{% load i18n localization %}
+{% load i18n localization wagtailcore_tags %}
+
+{% get_current_language as current_language %}
+{% get_local_language_names as languages %}
 
 <form action="{% url 'set_language' %}" method="post">
-  <input name="next" type="hidden" value="{{ request.path }}">
+  {% csrf_token %}
```
KalobTaulien, line 7: We'll want to figure out how to do this without CSRF to continue using CloudFront in production without running into caching issues.

KalobTaulien (changes requested):
Code looks good 👍  Can confirm the intended code does the job. 

Just need to figure out the CSRF thing with CloudFlare. Wonder if we can wrap the `set_language` view in a `csrf_exemp` decorator? 

**Edit**: I made another comment (below) about removing or using different DRF settings for manual page translations.

KalobTaulien:
Things I found during testing. 

When manually translating through the UI we need to remove this from our settings (or change it to allow writing)
image

```python
REST_FRAMEWORK = {
    'DEFAULT_PERMISSION_CLASSES': [
        'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly',
    ]
}
```
☝️ With that setting on (which it currently is) the terminal says:
> AssertionError: Cannot apply DjangoModelPermissionsOrAnonReadOnly on a view that does not set `.queryset` or have a `.get_queryset()` method.

This is something I missed in the initial setup. But we can add it at any point in time (and since this might be our last small PR for wagtail-localize it might be good to add it here).

network-api/networkapi/templates/fragments/language_switcher.html
```diff
@@ -1,15 +1,17 @@
...
+  <input name="next" type="hidden" value="{% get_unlocalized_url page current_language %}" />
```
Pomax, line 8:
```suggestion
  <input name="next" type="hidden" value="{% get_unlocalized_url page current_language %}" />
```

Pomax: urls.py already has csrf exemption:
```
    # set up set language redirect view
    path('i18n/setlang/', csrf_exempt(set_language), name='set_language'),
```

KalobTaulien (approved):

PR #6773, opened 2021-05-31
Article page now able to use at any point in the wagtail tree

### LINKED ISSUE #6730: [Publications] Single Article Page
Currently article pages exist only as child pages to a publication page that is created. There have been requests to use only an article page (without a parent publication page)

**Note: added:** Currently Article Pages _must_ live within a Publication Page. We need to enable them to live anywhere in the wagtail tree. 

## Criteria:
- [x] Need to unrestrict Article Pages 
- [x] May need to give pages the permission to have Article pages as a child page (`subpage_types = ['...', 'articlepage']`) 
- [x] Will need to "ignore" the parent page and next/prev pages and the buttons in the template associated with them since they wont have any real siblings and the parent page could be _any_ page type. 
  - [x] We want to ensure that Article Pages under Publication Pages still function as expected.

danielfmiranda:
Closes #6730

Article pages are now available to use anywhere in the wagtail tree, to see examples, please visit the following links below.

Please note that the article pages under publications, marked in bold, still show the breadcrumbs and bottom navigation as before. 

**Links to sample test pages**:
Link to Test Article Page Under Homepage: [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/test-article-page-under-homepage/)
Test Article under Blog: [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/blog/test-article-page-under-blog/)
Test Article Under Campaigns: [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/campaigns/test-article-page-under-campaigns/)
Test Article under Initiatives: [link]( https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/initiatives/test-article-under-initiatives/)
Test article under news: [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/news/test-article-page-under-news/)
**Article page under a publication page with child articles:** [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/publication-page-with-child-article-pages/fixed-title-article-page/)
**Article page under a publication page with chapter pages:** [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/
)
Article page under privacy not included: [link](https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/privacynotincluded/test-article-page-under-privacy-not-included/)

## Checklist

**Changes in Models:**

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
@@ -40,13 +40,11 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
 
     """
 
-    Article belong to PublicationPages
-    An Article can only belong to one Chapter/Publication Page
+    Articles can belong to any page in the Wagtail Tree.
     An ArticlePage can have no children
-
-    ? If these only belong to PublicationPages, should be extra explicit and call it PublicationArticlePage?
+    If not a child of a Publication Page, page nav at bottom of page
+    and breadcrumbs will not render.
     """
-    parent_page_types = ['PublicationPage']
     subpage_types = []
     body = StreamField(article_fields)
 
@@ -112,6 +110,14 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
         StreamFieldPanel('body'),
         InlinePanel("footnotes", label="Footnotes"),
     ]
+
+    @property
+    def is_publication_article(self):
+        parent = self.get_parent().specific
+        if "Publication" in parent.title:
+            return True
+        else:
+            return False
 
     @property
     def next_page(self):
--- a/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
@@ -9,8 +9,7 @@ from wagtail.images.edit_handlers import ImageChooserPanel
 from wagtail.snippets.edit_handlers import SnippetChooserPanel
 
 from networkapi.wagtailpages.models import ContentAuthor, PublicationPage
-from networkapi.wagtailpages.utils import get_plaintext_titles
-from networkapi.wagtailpages.utils import set_main_site_nav_information
+from networkapi.wagtailpages.utils import get_plaintext_titles, set_main_site_nav_information, TitleWidget
 
 from ..mixin.foundation_metadata import FoundationMetadataPageMixin
 from ..article_fields import article_fields
@@ -92,7 +91,12 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
         help_text="Show social share buttons on the side"
     )
 
-    content_panels = Page.content_panels + [
+    content_panels = [
+        FieldPanel(
+            "title",
+            classname="full title",
+            widget=TitleWidget(attrs={"class": "max-length-warning", "data-max-length": 60})
+        ),
         MultiFieldPanel([
             InlinePanel("authors", label="Author", min_num=0)
         ], heading="Author(s)"),
--- a/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/publications/article.py
@@ -113,12 +113,9 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
 
     @property
     def is_publication_article(self):
-        parents = self.get_ancestors()
-        parent_publication_pages = parents.filter(title__icontains="Publication")
-        if parent_publication_pages:
-            return True
-        else:
-            return False
+        parent = self.get_parent().specific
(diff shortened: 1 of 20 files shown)


danielfmiranda:
Hi @sabrinang! 

I have a quick question about this ticket, in the description the ask was to have article pages available for use anywhere in the wagtail tree, and at the moment with this pull request it is set to do so. 

However, I just wanted to double check that we wanted users to be able to make articles under the privacy not included section? 

Thanks in advance for your response, and for checking out my PR!

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -113,6 +111,15 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
         InlinePanel("footnotes", label="Footnotes"),
     ]
 
+    @property
+    def is_publication_article(self):
```
danielfmiranda, line 115: Hi @KalobTaulien! 

Thanks in advance for checking out my PR, I was wondering if I can get some feedback on this function which I am using to tell the template that it is rendering an article page underneath a publication page? 

Right now I have it checking if any of the ancestors have the word "Publication" in them, however, I was concerned as I was not sure how likely we might possibly rename publication pages to something else in the future. 

Thanks again!

sabrinang:
Awesome, I checked in the CMS that `article page` option does appear in add child page in all of the test links you mentioned.

**Feedback**
- I'm thinking it might be nice to include for /docs (where we do more internal documentation) if we ever need one-pagers the option is available but probably unlikely to be used

- I’m not sure what is going on with /news for the future in general but this might not be needed there but won't hurt to include

- Regarding privacynotincluded: It feels separate overall as the existing page types right now for it are only "General Product Page" and "Software Product Page". However, I am not in the loop if they are planning to publish long form content in their roadmap that needs to specifically live under /privacynotincluded

@kristinashu Do you have any more context or thoughts regarding this? Just want to make sure I don't miss anything as I review

kristinashu:
Nothing really to add! +1 to adding it to /docs and not having it on PNI.

sabrinang (changes requested):
@danielfmiranda 
Please add the article page option to `/docs` and you can exclude from `/privacynotincluded`

Everything else looks great!

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
```diff
@@ -113,6 +111,15 @@ class ArticlePage(FoundationMetadataPageMixin, Page):
...
+        parents = self.get_ancestors()
+        parent_publication_pages = parents.filter(title__icontains="Publication")
+        if parent_publication_pages:
+            return True
+        else:
+            return False
```
KalobTaulien, line 121: Yeah this is too fragile because eventually I'm sure a publication won't have "Publication" in the title. 

I think we just need to check if the parent page is an instance of a PublicationPage. Some pseudo-ish code:

```python
parent = self.get_parent() 
if isinstance(parent, PublicationPage):
    return True 
return False 
```

I don't think we need to check _all_ ancestors, just the parent page because Article Pages were initially supposed to only live under Publications, so the prev/next and breadcrumbs are already built for that based on the Parent always being an Article page.

KalobTaulien (changes requested):
This is a great first attempt, well done @danielfmiranda ! 

You're right that filtering ancestor pages by its title could be better. I added some pseudo code (I didn't check if it works, heads up) to hopefully help make some more sense out of it.

network-api/networkapi/wagtailpages/pagemodels/publications/article.py
danielfmiranda: Thanks for the feedback Kalob! I appreciate it.

I have updated it to instead return whether or not the parents .__class__ is a "PublicationPage" model, similar to publication's 
[is_publication_page()](https://github.com/mozilla/foundation.mozilla.org/blob/76da7f8f5a06fa153a5b50bb0d21085c664bf494/network-api/networkapi/wagtailpages/pagemodels/publications/publication.py#L121) method.

danielfmiranda:
Hi @sabrinang, done and done! You can now add article pages under /docs/ and I have removed it from PNI. thanks!

KalobTaulien (approved):

danielfmiranda:
Hi @sabrinang! 

The /docs/ directory in the foundation site is using a mini-name-space site object called /docs/ which the review sites are not creating by default, here is a recreation of what a article would look like under one that I made called /docs/ for you to give a look! https://foundation-s-6730-singl-8euw3i.herokuapp.com/en/docs/test-article-page-under-docs/ 

Thanks

sabrinang (approved):
Great thanks for sharing the link to the /docs, looks good!

PR #6778, opened 2021-05-31
allow authenticated users to hit the wagtail localize endpoints

### LINKED ISSUE #6776: Step 8.5. Ensure we're using the correct DRF permissions
When manually translating through the UI we need to remove this from our settings (or change it to allow writing)
image

```python
REST_FRAMEWORK = {
    'DEFAULT_PERMISSION_CLASSES': [
        'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly',
    ]
}
```
☝️ With that setting on (which it currently is) the terminal says:
> AssertionError: Cannot apply DjangoModelPermissionsOrAnonReadOnly on a view that does not set `.queryset` or have a `.get_queryset()` method.

This is something I missed in the initial setup. But we can add it at any point in time (and since this might be our last small PR for wagtail-localize it might be good to add it here).

Pomax:
Closes #6776 by opening all DRF endpoints as long as people are authenticated. We can tighten this further if we need to, but any locale changes folks would be tempted to make get overwritten by Pontoon, so leaving this a matter of policy, rather than a matter of code, feels fine for now.

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -477,7 +477,7 @@ if env("FRONTEND_CACHE_CLOUDFLARE_BEARER_TOKEN"):
 # Rest Framework Settings
 REST_FRAMEWORK = {
     'DEFAULT_PERMISSION_CLASSES': [
-        'rest_framework.permissions.IsAuthenticated',
+        'rest_framework.permissions.IsAdminUser',
     ]
 }
 
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -477,7 +477,7 @@ if env("FRONTEND_CACHE_CLOUDFLARE_BEARER_TOKEN"):
 # Rest Framework Settings
 REST_FRAMEWORK = {
     'DEFAULT_PERMISSION_CLASSES': [
-        'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly',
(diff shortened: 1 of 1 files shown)


TheoChevalier:
> any locale changes folks would be tempted to make get overwritten by Pontoon,

Unfortunately that’s not how it works, changes made from the admin interface will replace translations on Pontoon…
Can we set things up like on donate and make anything wagtail-localize only accessible to myself and admins?

Note that the UI part of Wagtail core might already be only visible to admins, so this could be enough?

KalobTaulien (approved):

Pomax:
@TheoChevalier Donate is set up just different enough that we can't use its method, but there is a https://www.django-rest-framework.org/api-guide/permissions/#isadminuser that I can switch this to instead, good call.

network-api/networkapi/settings.py
```diff
@@ -477,7 +477,7 @@
 # Rest Framework Settings
 REST_FRAMEWORK = {
     'DEFAULT_PERMISSION_CLASSES': [
-        'rest_framework.permissions.DjangoModelPermissionsOrAnonReadOnly',
+        'rest_framework.permissions.IsAuthenticated',
```
Pomax, line 480:
```suggestion
        'rest_framework.permissions.IsAdminUser',
```

TheoChevalier (approved):

PR #6779, opened 2021-05-31
Soft character limit

### LINKED ISSUE #6515: Implement a Character limit Widget
Goal: users need some visual feedback to not create long titles. We also want to balance that with existing title lengths, and allow people to go over if absolutely necessary. 

Solution: "Create a JavaScript snippet that counts down the remaining characters when entering a title. Turns red when its over 60 characters to provide a visual cue that the title is getting too long." (similar to Twitter)

Tasks: Create a custom Widget and inject JS with a Media object, and a classname on the Widget.

KalobTaulien:
Closes #6515 

**Link to sample test page**: https://foundation-s-6515-chara-xrvxty.herokuapp.com/cms/pages/55/edit/

☝️ if you edit the page title just under it will be a character counter. It should also work with modeltranslation fields to prove that the character counter is scoped to its closest input field.

### DIFF
--- a/network-api/networkapi/wagtailcustomization/static/wagtailadmin/js/max-length-field.js
+++ b/network-api/networkapi/wagtailcustomization/static/wagtailadmin/js/max-length-field.js
@@ -1,0 +1,35 @@
+/**
+ * Even listeners to let editors know when they reached the recommended maximum length of a title.
+ * This is paired with max-length-fields.css and wagtailpages/wagtail_hooks.py
+ */
+
+document.addEventListener('DOMContentLoaded', () => {
+  const maxLengthFields = document.querySelectorAll(".max-length-warning");
+  console.log(maxLengthFields)
+
+  if(maxLengthFields.length) {
+    maxLengthFields.forEach((node) => {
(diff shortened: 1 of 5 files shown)


network-api/networkapi/wagtailcustomization/static/wagtailadmin/js/max-length-field.js
```diff
@@ -0,0 +1,34 @@
+/**
+ * Even listeners to let editors know when they reached the recommended maximum length of a title.
```
danielfmiranda, line 2:
```suggestion
 * Event listeners to let editors know when they reached the recommended maximum length of a title.
```

Found a small typo, other than that everything looks great!

danielfmiranda (approved):
Hi Kalob, 

Thanks for tagging me in this PR! 

I found it super useful to see how we can make changes in the wagtail admin, and will  keep note of this PR for future reference when its my turn to make some wagtail admin changes 😄 

I did find one small typo which I marked in the comments, however everything else looks great to me, so I am approving 👍 that way you wont have to wait for me to re-review over a typo

PR #6777, opened 2021-05-31
Snippet translations + translatable fields

KalobTaulien:
Closes #6708
Related PRs/issues #6703 

## Acceptance criteria 

> Note: Don't forget to reference [the document from Step 5 #6707](https://docs.google.com/spreadsheets/d/1AU7pxboZoBiVcbTpZO-VUMAhxZwx5mdEH_FBLfo2D-w/edit#gid=0)

- [x] Create migrations for all snippets. [This document](https://docs.wagtail.io/en/stable/advanced_topics/i18n.html#making-snippets-with-existing-data-translatable) will be _very_ helpful. We need to bootstrap migrations for each model. 
  - [x] Make sure all snippet models are translated. 
  - [x] Make sure all snippet fields have `translatable_fields = [ .. ]` set on them 
  - [x] Make sure tests still pass. May need to do some adjusting for new language headers in the `request` object, particularly around redirects. 
- [ ] Get Theo to review this (specifically, Theo) to make sure all the fields are the right translation type.

PR #6855, opened 2021-06-10
Campaign page translation fix

### LINKED ISSUE #6851: [Extra] Step 6: Wagtail-localize runs into an issue with the campaign cta's.
The campaign CTA's seem to throw an error when we try to translate a page that uses those snippets. 

We need to explore what the problem is and come up with a working solution to allow them to be translated properly. 

See this comment for more details https://github.com/mozilla/foundation.mozilla.org/issues/6777#issuecomment-852173696

KalobTaulien:
Closes #6851 
Related comment: https://github.com/mozilla/foundation.mozilla.org/pull/6777#issuecomment-852173696

This should work now. I needed to put the translation mixin on the actual class, not the parent class. That means reverting a migration and re-applying the standard 3 migrations for translating existing snippet content. 

I kept these all as separate migration files as to not conflict with the localization/staging instance.


network-api/networkapi/wagtailpages/migrations/0016_add_localization_to_subclasses.py
```diff
@@ -0,0 +1,35 @@
...
+    dependencies = [
+        ('wagtailcore', '0062_comment_models_and_pagesubscription'),
+        ('wagtailpages', '0015_remove_localization_from_cta_model'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='petition',
+            name='locale',
+            field=models.ForeignKey(editable=False, null=True, on_delete=django.db.models.deletion.PROTECT, related_name='+', to='wagtailcore.locale'),
```
Pomax, line 18: do we need to catch the error this will throw if someone tries to delete it while a campaign makes uses of it?

KalobTaulien: Great question, I have no idea. @TheoChevalier thoughts on this?

Pomax: turns out: it does nothing. I tried deleting an in-use petition, and it just...worked O_o

TheoChevalier: I’d say it’s good to have, but not blocking

TheoChevalier (approved):
I’ve been able to check the issue is fixed!

network-api/networkapi/wagtailpages/migrations/0016_add_localization_to_subclasses.py
Pomax: alright, let's get this in.

PR #6866, opened 2021-06-15
Localization synching and creating management commands

### LINKED ISSUE #6869: Localization: Add synching and locale creation management commands
We need a couple of management commands to dynamically create:
1. New locales for each of our locales in our settings 
2. Sync page trees from English to new locales 

## Acceptance criteria
- [ ] Need a command to create new locales in our admin settings. Must be idempotent to run multiple times.
- [ ] Need a command to synchronize English pages with other locales (ie. English -> German). Must be idempotent to run multiple times.

KalobTaulien:
Closes #6869

Two new management commands. 

### 1. create_locales
`python manage.py create_locales` will loop through all our language settings and create new locales if they don't exist. This can be run multiple times. 

### 2. sync_locale
`python manage.py sync_locale` will list all your available language codes and ask you to type one in. It will skip English because we shouldn't sync English with English (and the English Locale should exist for us already) 

It will then verify the language_code you entered is in our settings. 

It'll get_or_create the source locale and the target locale objects. 

Then it'll create a LocalSynchonization object using those locals, or find the one that already exists. 

Lastly, it'll run `.sync_trees()` to sync a locale against English (it always uses English as the source locale). 

This should be safe to run multiple times too. 

### Testing steps 
1. Checkout this branch 
2. Run `python manage.py create_locales` 
3. Run it again and notice it didn't create a second set of Locales (making this safe to run multiple times in production) 
4. Run `python manage.py sync_locale`. You'll need to specify a specific language from the list it will give you. 
5. Run the exact same command a second time. Notice there is _not_ a second set of pages, making this safe to run multiple times in production. Your Wagtail page explorer should show English pages, and also your new locale pages.

### DIFF
--- a/network-api/networkapi/management/commands/create_locales.py
+++ b/network-api/networkapi/management/commands/create_locales.py
@@ -1,0 +1,13 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from wagtail.core.models import Locale
+
+
+class Command(BaseCommand):
+    help = 'Look for and create locales if they do not exist. This can be run multiple times if needed.'
+
+    def handle(self, *args, **options):
+        for language_code, name in settings.WAGTAIL_CONTENT_LANGUAGES:
+            locale, created = Locale.objects.get_or_create(language_code=language_code)
+            if created:
+                print(f"Create new locale: {name}")
--- a/network-api/networkapi/management/commands/sync_locale.py
+++ b/network-api/networkapi/management/commands/sync_locale.py
@@ -1,0 +1,38 @@
+from django.conf import settings
+from django.core.management.base import BaseCommand
+from wagtail.core.models import Locale
+from wagtail_localize.models import LocaleSynchronization
+
+
+class Command(BaseCommand):
+    help = 'Sync pages with original English pages'
+
+    def handle(self, *args, **options):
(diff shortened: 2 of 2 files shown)


TheoChevalier (changes requested):
I like the idea, and it seem to work pretty well, I’ve just found an issue with hyphenated locale codes, they are wrongfully flagged as invalid

Capture d’écran 2021-06-16 à 18 03 38

I’m also wondering whether we could (later/in a follow up) run these commands automatically when deploying the app? We will always want to match the locale list in the settings anyway, so if it fixes the front end timeout at the same time, why not?
One concern would be the additional time it’s taking to deploy the app, but hopefully it wouldn’t do anything until we actually add a new locale to the settings file?

KalobTaulien:
Oh good catch @TheoChevalier ! I'll test out the Frissian and Portoguese locales. 

And yes, that would be a great idea to throw these commands into our deployment process. I actually don't know how to do that, but I suspect it's no different than the other commands we have running. The sync command might need to take an argument though, but that's easy to adjust in the future if we go this route. 

I'll work on the immediate fixes right away. Thanks again!

TheoChevalier (approved):
Looking good, thanks!

PR #6872, opened 2021-06-16
Make tests pass

### LINKED ISSUE #6709: Step 7. Passing tests
I remember tests not passing for various reasons when we tried to use wagtail-localize the first time. In some cases, tests may need to be re-written or deleted. There will be a lot of exploration in this ticket named around language headers and testing in different browser languages. 

## Acceptance criteria 

- [ ] Tests ticket: 
  - [ ] We might end up at a point where we need to work on _just_ tests for a while. I recall some of these being rather tricky. First up: Make sure cypress tests for redirects work.
  - [ ] Make sure redirect tests continue to redirect to the appropriate language

KalobTaulien:
Closes #6709 

Not much to do here except flake8 some files and change the default sitemap URL. 

### Steps to run:
1. `inv test-python`. We should see a few skipped tests, but no errors. The skipped tests are expected. 
2. `inv test-node`. Everything should look good here. 
3. `inv test` to test node and python together - never hurts to re-run a test suite.

### DIFF
--- a/network-api/networkapi/campaign/views.py
+++ b/network-api/networkapi/campaign/views.py
@@ -12,6 +12,7 @@ import logging
 import json
 from networkapi.wagtailpages.models import Petition, Signup
 
+
 def process_lang_code(lang):
     # Salesforce expects "pt" instead of "pt-BR".
     # See https://github.com/mozilla/foundation.mozilla.org/issues/5993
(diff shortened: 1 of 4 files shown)


danielfmiranda (approved):
Hi Kalob! 

Just finished following the instructions that you wrote and can confirm that everything is passing, approved!

PR #6875, opened 2021-06-17
Marked fields are sync or translatable

### LINKED ISSUE #6867: Mark snippet fields translatable/synchronized
We will need to backport changes from https://github.com/mozilla/foundation.mozilla.org/commit/8b44930d796f8bbf4c92b4d8daef6b0c829a97d4 following rules on https://docs.google.com/spreadsheets/d/1AU7pxboZoBiVcbTpZO-VUMAhxZwx5mdEH_FBLfo2D-w/edit#gid=134886357

KalobTaulien:
Closes #6867 

Added snippet field translation support from [this document](https://docs.google.com/spreadsheets/d/1AU7pxboZoBiVcbTpZO-VUMAhxZwx5mdEH_FBLfo2D-w/edit#gid=134886357) 

One thing to note is the CTA model is not translatable. But it's child classes _are_, so the fields from the CTA model get passed into the translatable_fields for Signup() and Petition()

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/products.py
@@ -108,6 +108,12 @@ class BuyersGuideProductCategory(TranslatableMixin, models.Model):
         blank=True,
     )
 
+    translatable_fields = [
+        TranslatableField('name'),
+        TranslatableField('description'),
+        SynchronizedField('slug'),
+    ]
+
     @property
     def published_product_page_count(self):
         return ProductPage.objects.filter(product_categories__category=self).live().count()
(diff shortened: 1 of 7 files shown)


network-api/networkapi/wagtailpages/translation.py
```diff
@@ -1,3 +1,5 @@
+# TODO: REmove this and other translation.py files.
+# They aren't needed but keep coming back when we merge master into our localization branch
```
KalobTaulien, line 2: Added this because it keeps coming back. It's not doing any harm being here, so I'm leaving this for "final cleanup" when localization gets merged into the master branch.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -108,6 +108,12 @@ class BuyersGuideProductCategory(TranslatableMixin, models.Model):
         blank=True,
     )
 
+    translatable_fields = [
+        TranslatableField('name'),
+        TranslatableField('description'),
+        SynchronizedField('slug'),
```
KalobTaulien, line 114: Fields that aren't supposed to be translated at all are ignored.

TheoChevalier (changes requested):
Just a few missing fields in the petition snippets, but otherwise I think it’s good!

PR #6876, opened 2021-06-17
Allow banner pages to be further localized

### LINKED ISSUE #6868: Banner page localization
Banner pages, e.g. [What We Do in German here](https://foundation-localize.herokuapp.com/cms/pages/2363/edit/)  is missing most content (Intro & body fields, primarily)

KalobTaulien:
Closes #6868 

Because this page inherits from the PrimaryPage, we need to re-set the fields on the child model. There's a cleaner way to do this, but for now this works. 

The "cta" (Its a Petition() ForeignKey) and the "signup" (It's a ForeignKey to a Signup()) may not translate properly on this branch, but should translate fine on #6867 (PR #6875)


TheoChevalier (approved):
lookin’ good! 🚀

PR #6896, opened 2021-06-19
6554 - Refactor blog authors template code so publication pages can use it

### LINKED ISSUE #6554: Refactor blog authors template code so publication pages can use it
The majority of the code in this part of [publication_hero.html](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/publication_hero.html#L42-L73) is almost identical to [blog_authors.html](https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html).

Let's 

1. first check with the designers and see if the styling of both types should match
2. determine how much code we can extract from `blog_authors.html` (and SCSS) so publication pages can use it
3. refactor code and rename "blog_authors/blog-author" in code and file name to something more generic

fessehaye:
Closes #6554

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/author-images.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/author-images.html
@@ -1,0 +1,11 @@
+{% load i18n wagtailimages_tags %}
+
+<div class="author-images d-flex flex-wrap">
+    {% for author in authors %}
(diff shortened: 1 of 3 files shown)


KalobTaulien (approved):
Looks good! 🚀 

It'd be awesome if we could add caching to the author-images.html template fragment at some point in the future. Every author image is a database query, and even something as simple as template fragment caching could be beneficial.

PR #6938, opened 2021-06-25
6599 - Move Wagtail streamblock-specific SCSS

### LINKED ISSUE #6599: Move Wagtail streamblock-specific SCSS to ./scss/wagtail/blocks
### File(s) or components

all SCSS files that contain streamblock-specific code

### Reason for cleanup

We should start moving Wagtail streamblock specific SCSS to `scss/wagtail/blocks` instead of leaving them scattered around in various files and places

fessehaye:
Closes #6599 

I skipped the pulse listing since it was a react component.

Let me know if I skipped a streamblock

Should be no visual changes


source/sass/main.scss
```diff
@@ -62,7 +65,6 @@
 @import "./views/youtube-regrets-reporter";
 @import "./views/dear-internet";
 @import "./views/publication";
-@import "./views/article";
```
Pomax, line 65: this is not itself a streamfield block, but a page type, so let's keep that in `views`, and only move the `// StreamBlocks` part to its own .scss file in the `blocks` dir.

source/sass/main.scss
```diff
@@ -38,6 +38,9 @@
 @import "./wagtail/blocks/dear-internet-letter";
 @import "./wagtail/blocks/feature-quote";
 @import "./wagtail/blocks/image-text-mini";
+@import "./wagtail/blocks/paragraph";
```
Pomax, line 41: let's rename this one to `rich-text`, as the block gives folks a WYSIWYG editor that lets them insert far more than just paragraphs, but the only custom styling we needed was around the way paragraphs got spaced.

Pomax (dismissed):

source/sass/main.scss
```diff
@@ -39,6 +39,8 @@
 @import "./wagtail/blocks/feature-quote";
 @import "./wagtail/blocks/image-text-mini";
 @import "./wagtail/blocks/link-button.scss";
+@import "./wagtail/blocks/rich-text";
```
Pomax, line 42: it looks like this still needs the `blocks/article-blocks` here

Pomax (changes requested):

Pomax (approved):
comfirmed that everything still looks the same, in it goes!

PR #6939, opened 2021-06-25
6348 - Update the blog "featured post" cards to have a uniform height

### LINKED ISSUE #6348: Update the blog "featured post" cards to have a uniform height
we're currently only restricting featured blog images to a width 540px but with an unspecified height, which can lead to a blog page that looks like this:

image

Let's:

- [ ] determine the appropriate max-height for featured post images 
- [ ] update https://github.com/mozilla/foundation.mozilla.org/blob/master/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_index_feature.html#L12 to include this new value

fessehaye:
Closes #6348

Small PR. Just setting a fixed height so the image are in a 16:9 ratio


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_index_feature.html
```diff
@@ -9,7 +9,7 @@
 {% endcomment %}
 
 <div>
-  {% image blog.specific.get_meta_image width-540 %}
+  {% image blog.specific.get_meta_image width-540 height-303 %}
```
KalobTaulien, line 12:
```suggestion
  {% image blog.specific.get_meta_image fill-540x303 %}
```

KalobTaulien: A cool thing Wagtail has is "focal" points. On uploaded images you can select a part of the image that's important, and when we crop using `fill` it'll try to crop closer to the focal point instead of randomly cropping at an edge or dead center.

KalobTaulien (dismissed):

KalobTaulien (approved):

PR #6951, opened 2021-06-29
6909 - Standardize our Prettier RC config for jinja

### LINKED ISSUE #6909: Standardize our Prettier RC config for jinja
## Acceptance Criteria:
- [ ] When saving a page, .HTML files need to not break in django templates. 

This also includes investigate if there's a Prettier plugin for jinja or base django templating

fessehaye:
Closes #6909 

So there is no support for jinja templates for prettier 😞, but there were a few things that I included to make working with prettier easier. I also tried using nunjunks formatters since they are really similar but it also didn't work haha.

*IMPORTANT* : You do need to have prettier installed for vscode to use it. Installing the extension is not enough. It should not be a problem if you did `inv new-env` but if it still does not work you can just `npm ci or npm install`!

1. There is a `.prettierignore` added to the root that will omit prettier on save on all html files!
2. there is an extensions.json file that includes recommended vscode Extensions to install for your workspace. You can view the recommended extensions by clicking on the tab and typing in `@recommended` in the search bar!

If you want to add other extensions to the repo you think are useful: https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions

I added one that allows you to toggle the formatter pretty easily just in case:

image

3. There is a .prettierrc added so we are all formatting files the same way regardless of what are local defaults are. The current config is what all of our current JS files follow at the moment. You can also test it with `npm run precommit`

Make sure you have `"editor.formatOnSave": true` on your VScode settings!

Future nice things:
* Prettier also affecting SASS instead of the stylelint stuff we have
* ~~removing the gitignore on `.vscode/settings.json` so we can add it to the repo~~ Done!


.prettierrc
```diff
@@ -0,0 +1,7 @@
+{
+    "printWidth": 80,
+    "singleQuote": false,
+    "trailingComma": "es5",
+    "useTabs": false,
+    "embeddedLanguageFormatting": "off"
+}
```
danielfmiranda, line 7:
```suggestion
}

```

.prettierignore
```diff
@@ -0,0 +1 @@
+*.html
```
danielfmiranda, line 1:
```suggestion
*.html

```

danielfmiranda (changes requested):
Hi Simon! 

Thanks so much for the great and easy-to-follow instructions in this PR, it made following the instructions to test a breeze! Everything looks good on my end, just suggested two new lines in a couple files. 

Other than that everything looks, and works great!

danielfmiranda (approved):
Hi Simon! 

Thanks for making those requested changes 👍, after giving it a second look, everything makes sense to me! 

I think that having the prettier settings standardized would be helpful in making sure that everything will follow an easy-to-read format. 

I also think that having prettier ignore HTML files will be helpful so they dont break the jinja templates without the user having to go into their settings! 

Approving

PR #6957, opened 2021-06-30
Moved factory and updated related code

### LINKED ISSUE #6436: Move buyersguide factories to wagtailpages app
There are factory generators in `networkapi/buyersguide/factory.py`. We need to move these over to the `wagtailpages/` app so we can eventually remove the buyersguide app entirely. 

Related PR: #5887

KalobTaulien:
Closes #6436

Moved factory.py into buyersguide.py. Updated a few paths as well. 

Local testing:
1. `git checkout 6436-move-bg-factories`
2. `inv new-db`
3. Check localhost:8000/en/privacynotincluded/ 
4. Run `inv test-python`


Pomax (approved):
yep, this works a treat!

PR #6959, opened 2021-06-30
Basic YouTube 2021 regrets page

### LINKED ISSUE #6891: Create a wagtail page for the new YT Regrets template (backend)
Need a backend dev to create a new URL with an empty template for frontend devs to work with to inject each of the sections into the template. 

This will help keep the page cleaner and more maintainable if each section is in an include file and simply included into the main template page. 

## Acceptance criteria
- [ ] URL pattern for the YouTube Regrets page ~(may conflict with the existing YouTube regrets URL and may require a Wagtail page redirect)~ https://foundation.mozilla.org/campaigns/regrets-reporter/findings
- [ ] Create an empty template for the frontend developers to start working with 
- [ ] Ensure this URL is added to our sitemap.xml 

## Unknowns: 
* ~Is this replacing the current youTube regrets page?~ No. ~If so, we can strip out the old fields from the page and re-use the original template. If we did that, the sitemap would auto-generate this URL for us and we wouldn't need to create a URL path.~

KalobTaulien:
Closes #6891 

We just need a _very_ simple Wagtail page to hold just about all of the frontend work for the new YouTube Regrets work. I labeled it YouTubeRegrets 2021 because we're keeping the old content still. 

All the actual work for this year's YT Regrets is going to be in the frontend templates and translated with `{% translate %}` rather than going through wagtail-localize (not landed yet) or wagtail-modeltranslation (about to be removed). But we still needed wagtail-modeltranslation support for this PR.

### Primary testing criteria:
- [x] Can create a page at this URL: localhost:8000/en/campaigns/regrets-reporter/findings/


KalobTaulien:
Forgot to mention: 
I also figured we could toss this in `master` since it's not complete, there's a `max_count` on this new page type of 1, and it'll prevent frontend conflicts between the teams.

network-api/networkapi/wagtailpages/pagemodels/youtube.py
```diff
@@ -88,3 +88,20 @@ def get_context(self, request):
         return set_main_site_nav_information(self, context, 'Homepage')
 
     template = 'wagtailpages/pages/youtube_regrets_reporter_page.html'
+
+
+class YoutubeRegrets2021Page(FoundationMetadataPageMixin, Page):
+
+    template = 'wagtailpages/pages/youtube_regrets_reporter_page_2021.html'
```
Pomax, line 95: `youtube regrets reporter` is a browser extension, so let's make sure this is `youtube_regrets_page_2021.html`

Pomax (changes requested):
almost R+, but let's rename that template.

network-api/networkapi/wagtailpages/pagemodels/youtube.py
KalobTaulien: Ah, good catch. I wasn't aware of that. Template path and file name updated!

Pomax (approved):

PR #6962, opened 2021-06-30
keep repo as editor-agnostic as possible

fessehaye:
Removing some VSCode exclusive files from the repo as well as the devContainer files since no one uses them anymore.


Pomax (approved):
nice!

PR #7067, opened 2021-07-12
update all homepage fragments to pull from localized content

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7034
Related PRs/issues https://github.com/mozilla/foundation.mozilla.org/issues/7066

This updates the template fragments used by the homepage, making sure to use the `.localized` property for any page that is a foreignkey/M2M/etc to another page in the CMS.

These updates are based on looking at both the `Homepage` class in `base.py` and identifying fields that are ForeignKey, or are properties defined through other models using a `related_name`, as well as by looking at each template file and looking for the use of `.url`, and updating everything to use the localized page for fields where it feels like that made sense.

PR #7068, opened 2021-07-12
Fix all the localized properties

### LINKED ISSUE #7066: Update templates to use `.localized` for page-chooser based content
Content that is linked via page chooser (or ForeignKey / M2M to wagtailcore.Page or subclass of Page) needs to reference its content through the `.localized` property, to make sure that locales generate the correct localized link (if the linked page has an alias or full translation).

- [ ] homepage (see https://github.com/mozilla/foundation.mozilla.org/issues/7034)
- [ ] ...more...?

Step 1: audit the templates/models to build out the above list.
Step 2: fix all of them.

Pomax:
Builds on https://github.com/mozilla/foundation.mozilla.org/pull/7067, (first few commits are from that PR)

Closes #7066 by updating all the other template entries that point to pages that should be explicitly referencing the localized versions.... I _think_

It's possible I missed some, but outside of QA I'm not sure how we'd find them, and it's probably a number that is okay if we find them "as time goes on"?

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
+++ b/network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
@@ -68,13 +68,13 @@
       {# User is not logged in. Return cached results. 24 hour caching applied. #}
       {% cache 86400 pni_home_page template_cache_key_fragment %}
         {% for product in products %}
-          {% include "../fragments/buyersguide_item.html" with product=product %}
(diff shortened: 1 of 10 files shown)


KalobTaulien (approved):
Code looks good. Nice use of the {% with %} tag as well, that's great to see it being used!

TheoChevalier (approved):

Issue #5736, opened 2020-12-01
Set up a parallel mozfest pipeline


Pomax:
https://github.com/mozilla/foundation.mozilla.org/pull/5733 fixes the fact that review apps deploy a _different_ page hierarchy than local dev/staging/proc, which makes testing mozfest changes basically impossible. However, the way it does that is to restore the old functionality where we have to manually switch the `Sites` default site over from homepage to mozfest. That's not ideal, so a probably better solution is to set up a separate review pipeline for mozfest... the complications:

- the code is one repo, so Percy is probably going to be problematic.
- we only want things to rebuild in the respective pipelines if code that _affects_ those pipelines changes.

Alternatively, if we can figure out a way to do "subdomain hosting" on heroku, so things work identically as on localhost (where the distinction is `localhost:8000` vs `mozfest.localhost:8000`), that would solve a lot of problems...

/cc @tomusher to see if there's ideas he might have

tomusher:
Couple of options I can think of here:

1. Given some way to identify whether a given branch is a mozfest branch (maybe if the branch contains the string 'mozfest'), run a `postdeploy` script which changes the default site automatically.
2. Run a `postdeploy` script on all deployments that adds appropriate domains based on the branch name to the Heroku app, and uses the DNS provider API to create those records. e.g. pushing to the `fix-locale` branch would:
   1. Create the `fix-locale-abc123` review app
   2. Add the domains `fix-locale-abc123.review.mozillafoundation.org` and `mozfest.fix-locale-abc123.review.mozillafoundation.org` to the review app
   3. Call the Route 53 API to add the Heroku-generated CNAMEs
   4. Update the Wagtail sites to use these generated names
3. Add middleware to detect the hostname and override the default site. e.g. if you're accessing a review app with the branch `mozfest-locale-change`, the term `mozfest` in the hostname would force the current site to be the MozFest site.

The second option is much more complicated but may be more future-proof when there is a need for additional subdomains.

Pomax:
@tomusher @KalobTaulien I'm with Kalob in that option 2 sounds like the best option, but it also feels like setting up #3 as "immediate fix" while we figure out how to make #2 happen would be a worthwhile investment.

Pomax:
@benhohner @tomusher with Mozfest rapidly approaching (it's in ~4 weeks from the time of this comment) it sounds like we probably want to set this up the "quick and dirty" way for now, and then do things the proper way after Mozfest has concluded?

[...]

PR #7073, opened 2021-07-13
Add script to generate DNS records for different Sites in review apps

tomusher:
As part of #5736, this adds a script which adds DNS records for the foundation and mozfest site to AWS on the creation of review apps, then tears them down when the review app is removed.


Pomax (commented):
let's land this and see how it holds up!

Pomax (approved):

Pomax (changes requested):
I spoke too soon: flake8 is unhappy with the formatting, so we'll need to fix that before we can land this.

benhohner:
@tomusher @Pomax what do we need to do to finish this up?

Issue #6992, opened 2021-07-06
add a `related_posts` field with pagechooser


Pomax:
- [x] update the BlogPost page with an explicit related_posts field
- [x] add a PageChooser for related posts

(This might make the most sense using a "related_posts" mixin, but that might complicate things too much in terms of finding the appropriate model type.)

Issue #6756, opened 2021-05-26
update our templates to tap into the new `related_posts` field


Pomax:
update the blog (and campaign?) template to use the new `related_posts`field instead of the original `get_entries_by_tag` functionality

Issue #6755, opened 2021-05-26
Write a management command for presetting `related_posts`


Pomax:
Write a management command that, for each blog page with an empty `related_posts` list, performs the tag filtering, finds the top 3 results, and binds them as `related_posts` relations.

Issue #6754, opened 2021-05-26
Write a save rule for blog posts


Pomax:
Move the current tag filter code to its own function that can be used in a Blog page save hook, so that the tag code runs only if no related content is set, in order to fill the `related_posts` field,

PR #7076, opened 2021-07-13
Related posts for blog

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6992
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6754
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6756
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6755

(Each of those issues turned out to really be more one or two checkboxes in a multi-task issue, hence why a relatively small PR closes all of them)

Dev Testing:
- clone a copy of the staging db
- apply migrations
- verify that blog posts have zero related posts in their new admin view all the way at the bottom
- run `inv manage ensure_related_posts`
- verify that (many) blog posts now have three related posts (determined using the same code we were already using for dynamically determining related posts)
Note that some blog posts will have zero related posts: any blog post without tags set will (just like right now) have no related posts.

Design testing:
- log into the CMS using https://foundation-s-related-po-azdlow.herokuapp.com/cms and have a look at the new `related posts` fields at the bottom of individual blog pages, such as on https://foundation-s-related-po-azdlow.herokuapp.com/cms/pages/13/edit/. The way it works is that if you save a blog page, it will see if there are any related posts set, and if that number is any number less than 3 it will do a related post search based on the posts's tags in the "promote" tab, and automatically fill up the list of related posts to 3 entries. For example: if you delete all three related posts on the "initial test blog post" and then publish, you should see the list of related posts repopulated when you go to edit the page again. Similarly, creating a new blog post and giving it some tags without specifying any related posts explicitly should populate the list of related posts automatically on publish. Also, if you _do_ set any related posts explicitly, but fewer than 3, the automated list population process should leave the ones that were already there alone, and only pad out the list of related posts to 3.


kristinashu:
Thanks Pomax! My memory on this is rusty but isn't the purpose of this work to remove all the tags? The backup feature you described seems to rely on tags but is that just temporary?

Otherwise, I followed your description and everything seems good but I found one issue:

I edited the post so it only had 2 related posts. When previewing the post it only showed the 2 related ones.

Screen Shot 2021-07-14 at 8 03 17 AM

But then after publishing it, I guess it pulled from a tag but it is showing a duplicate.

Screen Shot 2021-07-14 at 8 03 04 AM

There shouldn't be any duplicates. And ideally, Preview and Publish would show the same results.

kristinashu (commented):
See above

Pomax:
@kristinashu before we can remove tags we first need to add in the related posts "based on tags" so that all the old posts get their related fields set correctly. The next step will then be removing tags and making related posts staff picks instead of "automatically picking them based on tags still".

kristinashu:
Oh I see, sounds good!

network-api/networkapi/wagtailpages/templates/wagtailpages/blog_page.html
```diff
@@ -20,7 +20,6 @@
 {% block body_id %}blog{% endblock %}
 
 {% block subcontent %}
```
Pomax, line 22:
```suggestion
{% block subcontent %}

```

network-api/networkapi/management/commands/ensure_related_posts.py
```diff
@@ -0,0 +1,25 @@
...
+        total_posts = all_posts.count()
+
+        print('Total number of posts to update:', total_posts)
+
+        for i, post in enumerate(all_posts):
+            print(
+                f"Processing post {i+1} out of {total_posts}, "
+                "{post.related_posts.all().count()} preexisting related posts"
+            )
+            post.save_revision().publish()
```
KalobTaulien, line 24: This will publish draft pages as well. 

We may want to check if a page is live first. iirc it's something like this:
```python
if post.live:
    post.save_revision().publish()
else:
    post.save_revision()
```

KalobTaulien (changes requested):

network-api/networkapi/management/commands/ensure_related_posts.py
Pomax: oh sh- good catch!

Pomax:
I've updated the code to:

1. make sure no duplicates are automatically added,
2. have the preview show what the published page will look like, and
3. make sure draft pages don't suddenly become published pages (thanks again for catching that one, Kalob!)

network-api/networkapi/management/commands/ensure_related_posts.py
```diff
@@ -0,0 +1,30 @@
...
+                f"Processing post {i+1} out of {total_posts}, "
+                "{post.related_posts.all().count()} preexisting related posts"
+            )
+
+            post.save_revision()
+
+            print(f'-  post {i+1} updated to {post.related_posts.all().count()} related posts')
+
+            if post.live:
+                post.publish()
```
KalobTaulien, line 30:
```suggestion
                post.save_revision().publish()
```
Published pages always fallback to a saved revision.

network-api/networkapi/management/commands/ensure_related_posts.py
```diff
@@ -0,0 +1,30 @@
...
+
+        for i, post in enumerate(all_posts):
+            print(
+                f"Processing post {i+1} out of {total_posts}, "
+                "{post.related_posts.all().count()} preexisting related posts"
+            )
+
+            post.save_revision()
+
+            print(f'-  post {i+1} updated to {post.related_posts.all().count()} related posts')
```
KalobTaulien, line 27: This f-string is acting strange. Terminal shows me:
```
Processing post 23 out of 537, {post.related_posts.all().count()} preexisting related posts
                                       ^ Not formatted 
```

KalobTaulien (changes requested):
I think we're close!

network-api/networkapi/management/commands/ensure_related_posts.py
Pomax: wow okay that one's weird.

Pomax:
```suggestion
            total_related_posts= post.related_posts.all().count()
            print(f'-  post {i+1} updated to {total_related_posts} related posts')
```

KalobTaulien (approved):

PR #7106, opened 2021-07-20
Hardcode How Do I Wagtail menu entry

### LINKED ISSUE #7105: Hardcode string in wagtail_hooks.py
#6420 moved a string from a file excluded from localization into a file that’s not.

Since it’s just one string and we don’t want it translated, I suggest we stop translating the string in the first place by hardcoding it, and then remove the file exclusion. It should make it easier to maintain in the future and removes _some_ code :)

TheoChevalier:
Closes #7105

Hey @fessehaye, for context, that string was excluded from localization by the line in task.py, but I think it’s better/more maintainable if we just don’t wrap it for localization in the first place.

And for even more context on why we’re not translating that string — because I see how it can be confusing — it’s because we’re only translating content that is visible by end-users, here’s it’s only visible by admins/CMS users.

Hope this helps!


fessehaye (approved):
Thanks for the help! Everything looks good!

Issue #6518, opened 2021-04-07
utility.middleware.py breaks without TARGET_DOMAINS set


Pomax:
The following code relies on TARGET_DOMAINS existing, which is not true when you're creating a new heroku app or pipeline
```
from django.http.response import HttpResponseRedirectBase
from django.conf import settings

hostnames = settings.TARGET_DOMAINS
referrer_value = 'same-origin'

if settings.REFERRER_HEADER_VALUE:
    referrer_value = settings.REFERRER_HEADER_VALUE

   ...
```

PR #7149, opened 2021-07-28
catch a missing TARGET_DOMAINS variable

Pomax:
This mostly fixes an issue we ran into when we set up the parallel localization pipeline, where things weren't working because TARGET_DOMAINS was missin.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/6518


network-api/networkapi/utility/middleware.py
```diff
@@ -2,6 +2,10 @@
 from django.conf import settings
 
 hostnames = settings.TARGET_DOMAINS
+
+if len(hostnames) == 0:
```
Pomax, line 6: we're checking `len()` here because settings.py loades in the target domains as `TARGET_DOMAINS=(list, []),` defaulting to an empty list if no env. var of the same name is found.

danielfmiranda (approved):
Hi Pomax, 

Thanks for tagging me in this ticket, as well as explaining what the code is doing to solve the problem. It made checking the PR easy to follow and understand. 

Approved!

Issue #6757, opened 2021-05-26
Create a new "blog entries" streamfield component


Pomax:
this should be a manual picker for 1-or-more blog entires, using a PageChooser, to replace the (few) instances where we're currently dynamically querying for blog posts by tag.

Pomax:
Let's reopen this one until the handful of pages that use the old component have been updated

Pomax:
I've update all instances that used tag filtering in the "Recent blog entires" and replaced them with the "blog set" block instead.

Pomax:
(some pages still use the recent blog entries component, but for category filtering, which is fine)

PR #7168, opened 2021-08-03
blog set component to replace recent-entries-by-tag

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6757 by introducing a replacement component that lets staff explicitly pick which blog pages to show as a series of cards, rather than trying to filter all blog posts based on tag similarity.

Truest STR:
- set up a clone of the prod db, then
- edit the `Election integrity and platform accountability` page: remove everything except the `The U.S. Elections 2020` component.
- hit "preview", then
- add a new "blog set" component under the original component it and add all the blog pages that the page preview shows.
- reload the preview: you should now see identical content twice, once from the "tag filtering super slow" component, and once from the "immediately resolving" page-chooser based component.


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/blog_set_block.html
```diff
@@ -0,0 +1 @@
+{% extends "./recent_blog_entries.html" %}
```
Pomax, line 1: straight up pass-through: the context variables are all the same, so it will (at least, it should =D) render identically.

KalobTaulien (approved):

PR #7175, opened 2021-08-05
6987 Customize Tailwind Spacing Units

fessehaye:
Redoing spacing units based on Figma design system, should be reviewed before people try to use tailwind for other tickets not related to the Tailwind Epic!

Also rebuilt the dev-environment for reloading HTML/tailwind changes.

On a separate terminal from the docker one, run `npm run tailwind` and trying changing the html files and their classes.
Changes should appear on `localhost:3000`

Old Default Spacing:
https://tailwindcss.com/docs/customizing-spacing

New Spacing: 
```
    spacing: {
      0: "0",
      px: "1px",
      1: "0.25rem",
      2: "0.5rem",
      3: "0.75rem",
      4: "1rem",
      5: "1.5rem",
      6: "2rem",
      7: "3rem",
    },
```

### DIFF
--- a/network-api/networkapi/context_processor.py
+++ b/network-api/networkapi/context_processor.py
@@ -16,3 +16,8 @@ def canonical_path(request):
 
 def canonical_site_url(request):
     return {'CANONICAL_SITE_URL': request.scheme + '://' + request.get_host()}
+
+
+def env_debug(request):
+    debug_flag = settings.DEBUG
+    return{"debug_flag": debug_flag}
--- a/network-api/networkapi/templates/pages/base.html
+++ b/network-api/networkapi/templates/pages/base.html
@@ -31,7 +31,9 @@
     {% block commento_meta %}{% endblock %}
 
     {% block stylesheets %}
+      {% if debug_flag %}<link rel="stylesheet" href="{% static "_css/tailwind.compiled.css" %}">{% endif %}
       <link rel="stylesheet" href="{% static "_css/main.compiled.css" %}">
+      
       <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Nunito+Sans:400,300,700,300i,800,900,400i">
       <link rel="stylesheet" href="https://fonts.googleapis.com/css?family=Zilla+Slab:300,400,600,700,300i,400i,600i">
     {% endblock %}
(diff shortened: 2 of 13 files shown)


KalobTaulien (changes requested):
As far as I can tell the code looks good. But percy has flagged some margin/padding changes. Let's get a designer to review this as well.

sabrinang (dismissed):
For the figma design file, I've added the sizing/spacers on the [grid & layout page](https://www.figma.com/file/JT9Dvdyc96AqQ3HJlnA9Qq/?node-id=312%3A354) and in the plug-in: 
image

@fessehaye I was able to add 0, px, 2.5 as figma components however I can't add "2.5" as a design token in [the plugin](https://docs.tokens.studio/tokens/creating-tokens) as the . separates the name and doesn't store the token in the plugin. Any tips on how you are naming fraction values and I don't know if you had the same issue in the code?

network-api/networkapi/context_processor.py
```diff
@@ -16,3 +16,7 @@ def canonical_path(request):
 
 def canonical_site_url(request):
     return {'CANONICAL_SITE_URL': request.scheme + '://' + request.get_host()}
+
+def env_debug(request):
+    debug_flag = settings.DEBUG
+    return{"debug_flag":debug_flag}
```
KalobTaulien, line 22:
```suggestion
    return {"debug": settings.DEBUG}
```

network-api/networkapi/templates/pages/base.html
```diff
@@ -31,7 +31,9 @@
     {% block commento_meta %}{% endblock %}
 
     {% block stylesheets %}
+      {% if debug_flag %}<link rel="stylesheet" href="{% static "_css/tailwind.compiled.css" %}">{% endif %}
```
KalobTaulien, line 34:
```suggestion
      {% if debug %}<link rel="stylesheet" href="{% static "_css/tailwind.compiled.css" %}">{% endif %}
```

KalobTaulien (changes requested):
This is looking good. Let's use a standardized "debug" term for our context processors so that it matches the backend. Everything else looks good. Ping me for a re-review when it's ready!

sabrinang (approved):
Great thanks for updating them to whole numbers for the naming Simon!

PR #7161, opened 2021-08-02
removed anchor link for sub-publication pages

danielfmiranda:
Closes #7136 

**Link to sample test page**: https://foundation-s-7136-publi-8sfeai.herokuapp.com/en/publication-page-with-chapter-pages/

**Steps to test**:
1.) Visit https://foundation-s-7136-publi-8sfeai.herokuapp.com/en/publication-page-with-chapter-pages/, and scroll down to view the child publication pages.
2.) Hover over the title of the child publication pages, note that they are no longer clickable as requested.
3) Hover over the 1st chapter of the publication page table of contents, it should be clickable and link correctly.
4.) If everything is working correctly, testing is done! If there are any additional requests, please let me know and I will implement them ASAP. Thank you! 

## Checklist

**Changes in Models:**
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations)

PR #7197, opened 2021-08-09
updated publication template links

danielfmiranda:
Closes bug caused by #7161 where links to publication pages without any child pages would not render a link. 

What this fix does is check if the page has any child pages, if it does not, render a link to the single page. 
If the publication page has child pages, then only render links to the child pages like requested. 

**Link to sample test page**:
Publication with child article pages: https://foundation-s-fix-public-6gm86f.herokuapp.com/en/publication-page-with-child-article-pages/
Publication with chapter pages: https://foundation-s-fix-public-6gm86f.herokuapp.com/en/publication-page-with-chapter-pages/

## Checklist

**Changes in Models:**


Pomax (approved):
this looks good to me!

PR #7210, opened 2021-08-10
fix the article-image-block template

Pomax:
Hotfix for the tiktok campaign

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/article_image_block.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/article_image_block.html
@@ -10,10 +10,11 @@ no-gutters
 {% endblock %}
 
 {% block main_block_class %}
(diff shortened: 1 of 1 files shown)


danielfmiranda (approved):
thanks for the catch @pomax this makes sense to me as well, as it appears as if the articles without an image render correctly, while the “research-findings” article page which does use an image, returns a 500 error. 

Approving!

danielfmiranda:
Here are some test pages that I have created:

The first is a publication page, underneath the campaign section in the CMS, since this is where the tik tok pages are housed in production: https://foundation-s-fix-articl-cge4pv.herokuapp.com/en/campaigns/test-publication-page-under-campaigns/

The second page is an article page, which we have added a image block to, in order to test if this was the problem since this is the page that is returning the 500 error in production: https://foundation-s-fix-articl-cge4pv.herokuapp.com/en/campaigns/test-publication-page-under-campaigns/test-article-page-under-a-publication-under-a-campaign-page/

PR #7215, opened 2021-08-11
Moved consts inside of functions to prevent runtime/migration conflicts

KalobTaulien:
This introduces some not-so-DRY code, but it's minimal and thought this would be OK considering our timeframe today. 

Kept the consts as local consts in each function too keep this PR as small as possible.

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/mixin/snippets.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/mixin/snippets.py
@@ -1,8 +1,6 @@
 from django.conf import settings
 from wagtail.core.models import Locale
 
-DEFAULT_LOCALE = Locale.objects.get(language_code=settings.LANGUAGE_CODE)
-
 
 class LocalizedSnippet():
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/pagemodels/mixin/snippets.py
```diff
@@ -1,17 +1,16 @@
...
-
 
 class LocalizedSnippet():
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     @property
     def original(self):
+        DEFAULT_LOCALE = Locale.objects.get(language_code=settings.LANGUAGE_CODE)
+
```
Pomax, line 12:
```suggestion
   DEFAULT_LOCALE = None

    def __init__(self, *args, **kwargs):
        if self.DEFAULT_LOCALE is None:
            self.DEFAULT_LOCALE = Locale.objects.get(language_code=settings.LANGUAGE_CODE)
        super().__init__(*args, **kwargs)

    @property
    def original(self):
```

KalobTaulien: Added to the latest commit

Pomax (approved):
Filed https://github.com/mozilla/foundation.mozilla.org/issues/7217 for drying this back out, but for now: in it goes~!

PR #7231, opened 2021-08-17
WIP: added new rich_text_block field for paragraph and content blocks

### LINKED ISSUE #6920: Create "paragraph"/"content" streamfield blocks.
On the following pages:

We loop through all the different streamfield blocks, but have an if statement specifically for 'paragraph' or 'content' blocks, since we do not have a template file for them. 

```
{% for block in page.body %}
{% if block.block_type == 'paragraph'%}
<div class="container {{block.block_type}}-block">
  <div class="row justify-content-center">
    <div class="streamfield-content">
      {% include_block block with parent_page=page page_type="blog"%}
    </div>
  </div>
</div>
{% else %}
{% include_block block with parent_page=page page_type="blog" %}
{% endif %}
{% endfor %}
```

I think that a better solution would be to create a new template file for paragraph and content blocks, that way instead of having a "if" statement in the files above, we can just loop through the streamfields like so: 

```
{% for block in page.body %}
     {% include_block block with parent_page=page page_type="blog" %}
{% endfor %}
```

danielfmiranda:
Closes #6920 

**Link to sample test page**:

## Checklist

**Changes in Models:**
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations)

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/rich_text_block.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/rich_text_block.html
@@ -1,0 +1,11 @@
+{% extends "./base_streamfield_block.html" %}
+
+{% comment %}
+This block is used for rich-text streamfields, such as the 'paragraph'
+and 'content' blocks found on Article, Blog, Campaign, Oppurtunity, and Primary pages. 
+{% endcomment %}
+
+
+{% block block_content %}
+    {{ value }}
+{% endblock %}
--- a/network-api/networkapi/wagtailpages/pagemodels/base_fields.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/base_fields.py
@@ -14,7 +14,8 @@ base_fields = [
             'h2', 'h3', 'h4', 'h5',
             'ol', 'ul',
             'link', 'hr',
-        ]
+        ],
+        template='wagtailpages/blocks/rich_text_block.html',
     )),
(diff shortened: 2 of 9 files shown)


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/rich_text_block.html
```diff
@@ -0,0 +1,11 @@
+{% extends "./base_streamfield_block.html" %}
+
+{% comment %}
+This block is used for rich-text streamfields, such as the 'paragraph'
+and 'content' blocks found on Article, Blog, Campaign, Oppurtunity, and Primary pages. 
+{% endcomment %}
+
+
+{% block block_content %}
+    {{ value }}
```
KalobTaulien, line 10: This value will be richtext, right? We'll need to add `{{ value|richtext }}` instead. 

Under the hood, Wagtail stores links to pages, images and documents as non-valid HTML. Such as (and Im paraphrasing here) `<a type="link" page-id="42">Text in here</a>` and `|richtext` internalizes that and looks up the page id from the markup it's given. 

Edited: `{% load wagtailcore_tags %}`

KalobTaulien (changes requested):
One small thing, but everything else looks nice and tidy. Thanks for doing all this cleanup Daniel!

PR #7234, opened 2021-08-18
make sure we select a single blog index page by specifying locale

Pomax:
```
<marcwalsh> Hello, Solana and I are both getting a 500 error when visiting https://www.mozillafestival.org/en/newsletter/ however the homepage loads without issue.
```

This was caused by selecting the Blog index without a locale, so it was getting eight results instead of one result, making `.get` crash out. 

STR: grab a copy of prod or staging, make sure to set the site settings back to `localhost` on port `8000` for foundation and `mozfest.localhost` on port 8000 for mozfest (probably restart to make those settings stick since they're network bindings, which don't like updating on the fly) and then try the http://mozfest.localhost:8000/en/newsletter/ url. It should now work, even if you change locale codes in that URL or switch languages in the lang picker in the page footer.

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/customblocks/recent_blog_entries.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/customblocks/recent_blog_entries.py
@@ -1,8 +1,12 @@
 from django.apps import apps
+from django.conf import settings
+from django.template.defaultfilters import slugify
+
 from wagtail.core import blocks
+from wagtail.core.models import Locale
 
 from ..blog.blog_category import BlogPageCategory
-from django.template.defaultfilters import slugify
+from networkapi.wagtailpages.utils import get_locale_from_request
 
 
 class RecentBlogEntries(blocks.StructBlock):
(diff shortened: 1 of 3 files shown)


network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -45,16 +46,6 @@
 ]
 
 
-def get_language_code_from_request(request):
```
Pomax, line 48: I was going to move this one into utils, only to (re)discover we already had a far better implementation for the same thing there.

network-api/networkapi/wagtailpages/pagemodels/customblocks/recent_blog_entries.py
```diff
@@ -54,8 +60,8 @@ def get_context(self, value, parent_context=None):
         if tag and not category:
             tag = slugify(tag)
             query = tag
-            blogpage.extract_tag_information(tag)
-            entries = blogpage.get_entries(context)
+            blog_page.extract_tag_information(tag)
```
Pomax, line 63: Naming change just to bring the name in line with similar code in BlogPage

KalobTaulien (approved):

Issue #7233, opened 2021-08-17
Wagtail Localize - testing result aggratation issue


Pomax:
leave comments to this issue with things you found, and then we can turn those into issues when we're done testing.

Pomax:
https://foundation.mofostaging.net/fr/blog/will-big-tech-stop-election-disinformation-in-germany/  still shows the related blog post links as `/en/`, so that's probably a missing template update for blog cards

Pomax:
> [tchevalier] K, just localized the campaign page: https://foundation.mofostaging.net/de/campaigns/election-integrity-and-platform-accountability/
> Looks like the blog post links are /en/ :confused:
> Picking a blog page now
>@pomax translating https://foundation.mofostaging.net/fr/blog/youtube-recommendations-tips-tricks-limit/
> [pomax] I'm pretty sure I know why: blog posts point to actual pages, and the block code does not "correct" the locale

Issue #6965, opened 2021-07-01
Related content section is pulling blog posts from all locales


TheoChevalier:
Check the Related Content section at the bottom of https://foundation-localize.herokuapp.com/en/blog/lets-unfck-internet/
It’s now pulling content from all locales instead of the current one.

<img width="1209" alt="Capture d’écran 2021-07-01 à 17 12 35" src="https://user-images.githubusercontent.com/1294206/124148022-27e9b580-da87-11eb-93c5-81db79c243a2.png">

Pomax:
This should no longer be the case when we switch to "real" related content (i.e. staff will pick related blog posts using a page chooser, see #https://github.com/mozilla/foundation.mozilla.org/issues/6753

benhohner:
Do we need a temporary fix until we get #6753 completed so we can launch? Theo mentioned there's an issue with page choosers at the moment?

TheoChevalier:
> Do we need a temporary fix until we get #6753 completed so we can launch? Theo mentioned there's an issue with page choosers at the moment?

Here’s the ticket about the page choosers https://github.com/mozilla/foundation.mozilla.org/issues/7034

PR #7236, opened 2021-08-18
fix the relocalized-url tag, and update some more templates to use it

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6965 while also fixing up the `relocalized_url` template tag necessary for resolving that issue.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/7233 because of the widened scope.

The STR will require grabbing a copy of the most recent production data, so that you have real localized content to test on.

- when on `/de/` the http://localhost:8000/blog index page should link to `/de/` links for all cards rather than `/en/`
- same for the http://localhost:8000/campaigns index page.
- the PNI homepage should have categories and products that all link to the correct locale (`en` on EN, `de` on DE, etc)
- the PNI "all" category and logo should point to the homepage on the current locale
- "related products" for PNI products should have URLs for the current locale
- the homepage should also point to `/de/` things when on http://localhost:8000/de/ etc.

Also note that the blog card has had the code for "showing tags" removed, since we haven't exposed that in years and we're already working on taging tags out (for UI purposes).

### DIFF
--- a/network-api/networkapi/wagtailpages/templatetags/localization.py
+++ b/network-api/networkapi/wagtailpages/templatetags/localization.py
@@ -5,6 +5,7 @@ from django import template
 from django.conf import settings
 from django.utils.translation import get_language_info
 from wagtail.contrib.routable_page.templatetags.wagtailroutablepage_tags import routablepageurl
+from networkapi.wagtailpages.utils import get_language_from_request
 
 register = template.Library()
 
@@ -54,8 +55,10 @@ def get_unlocalized_url(page, locale):
 
 
 # Force-relocalize a URL
-@register.simple_tag()
-def relocalized_url(url, locale_code):
+@register.simple_tag(takes_context=True)
+def relocalized_url(context, url):
+    request = context['request']
+    locale_code = get_language_from_request(request);
     if locale_code == DEFAULT_LOCALE_CODE:
         return url
     return url.replace(f'/{DEFAULT_LOCALE_CODE}/', f'/{locale_code}/')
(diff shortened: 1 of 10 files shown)


TheoChevalier (changes requested):
Working well overall, just spotted a couple areas we will need to fix, that don’t seem hard to fix (?)
Also tested the Blog set component, blog titles have the localized URL, but categories don’t. 

Capture d’écran 2021-08-19 à 17 08 22

Capture d’écran 2021-08-19 à 17 53 25


Capture d’écran 2021-08-19 à 17 57 27

TheoChevalier (approved):
Actually, the category URL was fixed in the other PR, and I just noticed the faker title issue happens on another PR, so it’s definitely not caused by this code

Pomax:
I don't know why, but this PR causes a redirect test that passes in `master` to fail. I'll have to dig into why.

Pomax:
So https://github.com/mozilla/foundation.mozilla.org/pull/7236/files#diff-2a95e5e43636ba7a64323d87d452a5a27c79dad85ed2daab647f6c0f0105826cR59-R61 is breaking redirects, somehow, and I've love to dig into the whole "why".

Pomax:
Right, found the problem here: `privacynotincluded/products/blah` would fail because the products.py code was directly referencing the `relocalized_url` template tag, instead of calling the "non-template-tag-code" that should have backed `relocalized_url`, so e31c7a1bfce8bfb1c25cff03b9e6d8adf206cff1 fixes that.

network-api/networkapi/wagtailpages/templatetags/localization.py
```diff
@@ -53,19 +54,26 @@ def get_unlocalized_url(page, locale):
...
 
+# Force-relocalize a URL
+@register.simple_tag(takes_context=True)
+def relocalized_url(context, url, locale_code=None):
+    if locale_code is None:
+        request = context['request']
+        locale_code = get_language_from_request(request, True)
+    return relocalize_url(url, locale_code)
+
+
```
Pomax, line 71:
```suggestion
# Force-relocalize a URL
@register.simple_tag(takes_context=True)
def relocalized_url(context, url):
    request = context['request']
    locale_code = get_language_from_request(request, True)
    return relocalize_url(url, locale_code)


```

Pomax:
@TheoChevalier @KalobTaulien one last pass: tests pass now, and I'm pretty sure we're good to go now, but just in case I had a typo in that last commit =D

TheoChevalier (approved):
Thanks for digging that! Looking good

Issue #7122, opened 2021-07-22
[wagtail-localize] Localize blog categories


TheoChevalier:
Similar to #7120 but for blog categories:

- [x] Category names should appear translated on blog pages and /blog
- [x] Create a slug field that’s synchronized with translations or slugify the English name for all locales.

I’ve done a test with the Advocacy category in French, it’s showing up here https://foundation-localize.herokuapp.com/fr/blog/category/militantisme/ instead of `blog/category/advocacy`

<img width="1264" alt="Capture d’écran 2021-07-22 à 15 55 02" src="https://user-images.githubusercontent.com/1294206/126651301-3e3e5af3-b984-4b44-89c1-c2a9cdec6f73.png">

PR #7237, opened 2021-08-18
Localize categories

Pomax:
Part 2 of https://github.com/mozilla/foundation.mozilla.org/issues/7122

Builds on #7445 

This fix introduces a new templating tag `localized_version` which can be used with anything that supports the `.get_translation(locale)` function, with template updates that use this tag to make sure there's a handle to the URL-correct localized category, for showing users category names in their own language.

## Testing

I tested this by setting up a new environment, then
1. adding a new locale
1. translating the blog categories
1. translating the "initial blog post" as well
1. loading localhost:8000/blog and verifying it moves me to my browser's locale URL
1. loading localhost:8000/en/blog and verifying all links we correct
1. loading localhost:8000/.../blog for the locale I picked, and verifying all links we correct for that locale
    1. noting that the featured posts do not show in the "rest of the posts" set
1. clicking through to the category that has the "initial post" in it
    1. seeing the initial post listed, with localized title
    1. changing the locale in the URL back to `en` and verifying the same content but in English now shows


TheoChevalier (changes requested):
A couple templates are broken due to the tag not being loaded, but otherwise it works well!

Now realizing my comment on the other PR about this link not being localized probably belongs to this one instead.
Capture d’écran 2021-08-19 à 17 57 27

TheoChevalier (changes requested):
It’s better if I *actually* submit my review comment… (using `localizedroutablepageurl` to localize the links in cards)

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html
```diff
@@ -17,7 +17,8 @@
         {% else %}not published yet{% endif %}
       {% with category=page.specific.category.first %}
       {% if category and blog_index %}
-        | <a class="category" href="{{blog_index.url}}category/{{category.slug}}">{{ category }}</a>
+        {% localized_version category as localized_category %}
+      | <a class="category" href="{{blog_index.url}}category/{{category.slug}}">{{ localized_category }}</a>
```
TheoChevalier, line 21:
```suggestion
      | <a class="category" href="{{blog_index.localized.url}}category/{{category.slug}}">{{ localized_category }}</a>
```

TheoChevalier (commented):
Just found one `.localized` missing, but otherwise looks good!

Pomax:
suggestion merged =D

TheoChevalier (approved):
Thanks! Looks like everything is working :)

Pomax:
marking as blocked until 7236 is merged in.

KalobTaulien (approved):

TheoChevalier (approved):
R+ with just a suggestion to merge

TheoChevalier (changes requested):
Actually, did additional testing and it seems there’s been a regression along the way 😟

With a localize Advocacy snippet, I can’t get http://localhost:8000/fr/blog/category/advocacy/ to show a localized name + intro on the page, instead it’s available at http://localhost:8000/fr/blog/category/militantisme/ which shouldn’t even be a valid URL.

I’m pretty sure a previous version of this PR had the correct behavior?

Pomax:
This should be all the things now: the `index` and `blog_index` have been updated so that their links and content are pulled from the current locale, whatever that may be, and some of the templates have been updated because they had a hardcoded `/blog` and making it explicitly localized felt future-proofy.

TheoChevalier (changes requested):
Getting cloooooooose 🤞

Unfortunately it’s crashing when opening a category that is not translated:
Capture d’écran 2021-09-23 à 14 28 56

Also noticed translated slugs are still valid URLs http://localhost:8000/fr/blog/category/militantisme/, but this can be fixed in a follow up

Pomax:
This should finally do it...? 🤞

TheoChevalier (approved):
🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀🚀

PR #7269, opened 2021-08-27
Use faker directly when creating review app password

tomusher:


danielfmiranda (approved):
Hi Tom, 

Thanks so much for adding this fix! 
Taking a look at the review app channel, I can see that the password is displaying correctly now for this PR. Approving!

network-api/networkapi/management/commands/review_app_admin.py
```diff
@@ -19,15 +19,9 @@ def handle(self, *args, **options):
...
-            password = Faker(
-                'password',
-                length=16,
-                special_chars=True,
-                digits=True,
-                upper_case=True,
-                lower_case=True
-            )
-            User.objects.create_superuser('admin', 'admin@example.com', str(password))
+            fake = Faker()
```
Pomax, line 22: just to match the rest of the files where we're directly using faker, let's move this instantiation up to just after line 9 and call it `faker` rather than `fake`.

Pomax (changes requested):

Pomax:
@tomusher it looks like this PR got lost in traffic: could I ask you to update it?

Pomax:
given that reviewe apps work, I'm going to consider this PR obsolete.

Issue #7290, opened 2021-08-30
factory code ends up showing computer-nonsense


Pomax:
From https://github.com/mozilla/foundation.mozilla.org/pull/7236:

<img width="830" alt="Capture d’écran 2021-08-19 à 17 08 22" src="https://user-images.githubusercontent.com/1294206/130100826-f3bb4fc0-b01e-420d-a508-cc521575a49f.png">

Pomax:
This looks to be a `Faker` issue, where we're not constructing the correct faker object and then calling its functions.

PR #7303, opened 2021-09-01
make sure to use the right faker

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7290

Switches the wrong Faker (from `factory`) for the right Faker (from `faker`) in code that isn't related to Django model factories.

The review app's homepage should now show normal text in the hero banner "black box", as well as the "take action" section.


TheoChevalier (approved):
Looks good!

PR #7304, opened 2021-09-01
RSS Reader Fix

### LINKED ISSUE #7261: RSS feed returns 500 Server Error
1. Go to https://foundation.mozilla.org/en/blog/rss/

Screen Shot 2021-08-26 at 3 43 55 PM

danielfmiranda:
Closes #7261 

**Link to sample test page**: https://foundation-s-7261-rss-e-iaa1ux.herokuapp.com/en/blog/rss/

**Steps to test**:

1. Visit the link above 
2. Add it to your rss reader of choice
3. The page should render correctly with no 500 error and should appear in the RSS reader correctly as well.
**Note:** Due to an ongoing redirect bug on review apps, it appears that all links are redirecting to a mozfest URL instead of the review apps correct URL. 
Due to this, my RSS reader seemed to be rendering the pages correctly:
Screen Shot 2021-09-01 at 2 23 10 PM

But will link to a bad URL if clicked on.

## Checklist

**Changes in Models:**
- [x]  [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations)


network-api/networkapi/wagtailpages/rss.py
```diff
@@ -16,7 +18,16 @@ class RSSFeed(Feed):
     feed_url = 'https://foundation.mozilla.org/blog/rss/'
     description = 'The Mozilla Foundation Blog'
 
+    def __call__(self, request, *args, **kwargs):
+        # get locale from request header, or fall back to the default EN locale.
+        request_locale = get_locale_from_request(request)
```
danielfmiranda, line 23: get_locale_from_request will either return the users locale, or will default to the english one. 

So we are setting the locale here, and then grabbing it in the "items" function, with no need to set default there.

Pomax (approved):
beautiful

PR #7316, opened 2021-09-03
Updated weights for quote blocks and callouts

### LINKED ISSUE #7287: [QuoteBlock/Callout Block] - Make font one weight heavier
There is a request to make the text in both the quote block and call out block one weight heavier.

For example if it is a normal weight, make it bold. However if we do not support anything bolder than 'bold' we should investigate the difference in page load time before adding a new font weight to the site

- [ ] ping Ben when complete so he can let Katie (stakeholders) know.

danielfmiranda:
Closes #7287

**Link to sample test page**:
Link to updated style guide: https://foundation-s-7287-quote-9aj1mz.herokuapp.com/en/style-guide/

**Steps to test**:

1. Visit [the updated style guide](https://foundation-s-7287-quote-9aj1mz.herokuapp.com/en/style-guide/) and scroll to the bottom to take a look at the Callout and Quoteblock component.
2. If both appear OK with the new font weights, testing is done! 

## Checklist

**Changes in Models:**
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/master/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations)


danielfmiranda:
Hi @sabrinang I had a quick question regarding this ticket:

Here is the quote block with two different font widths, 500 and Bold
500
quote-500-font-weight
Bold
quote-bold-font-weight

I was wondering which font weight was desired, as the 500 font width would require us to import a new font. Which is not difficult to do, I would just like to double check with @Pomax that it would be OK for our load times.

The bold font weight is already imported, so that would only require us just making a one-line change in the code, but I am not sure if it was too bold for the requirements of this ticket.

Thank you!

sabrinang:
hey @danielfmiranda I don't have complete context of the requested change but from the comparison my preference would be 500 if it doesn't impact load times severely as the bold weight feels too visually heavy compared to our heading styles.

sabrinang (approved):
The weight updates for both components look fine to me 👍

fessehaye (approved):

kristinashu:
@danielfmiranda sorry but I think there was a misunderstanding in the original ticket. The request wasn't to make all the text bolder but to have the ability to make **certain words** in the text bolder in order to highlight them.

Issue #6493, opened 2021-04-06
Remove unused management commands


KalobTaulien:
There are at least 2 identified management commands we need to remove because they are no longer used. 

### Acceptance criteria
- [ ] remove_milestones.py 
- [ ] image_migration.py

Any other management commands we no longer need, please open a new ticket for so we can keep track of the work without confusing tickets.

PR #7332, opened 2021-09-07
remove some dead management commands

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6493

As far as I can tell all these commands were one-offs, and are no longer required. No code should refere to them, no docs should mention them, and no scheduled tasks on Heroku should be invoking them.


danielfmiranda (approved):
Hi Pomax! Thanks for tagging me in this PR, after giving everything a look through, everything looks good to me! Approving 👍

PR #7338, opened 2021-09-08
[WIP] Switch YT 2021 dependencies from npm to CDN

### LINKED ISSUE #7308: Pull youtube-regrets JS libraries from CDN, rather than bundling them in
We currently bundle a bunch of pure frontend libraries into our site's app bundle, even though they're only used on single pages, which means that we're serving a lot of data that most folks will never need, as well as forcing a bundle redownload when these library's versions get bumped, even if nothing else changed. Let's take these out of our bundle, and instead load them from CDN using normal script tags. 

These include:

- chart.js
- gsap.js
- count-up.js
- swiper.js

But I might even still be missing some.

Pomax:
Closes #7308


Pomax:
Note: cypress seems to fail on tailwind, not sure why.

Issue #7347, opened 2021-09-09
mystery "data-block-key" attribute on Wagtail streamfield content


Pomax:
It looks like all CMS streamfield content now renders with a mysterious `data-block-key` attribute, and we should look into why it's there, and whether we can remove it because that's pretty unnecessary bloat.

- [x] as immediate remediation, let's update our rich text HTML rewriter to make sure it can cope with these attributes being here.
- [x] as long-term solution, let's find out why these attributes are there at all.

Pomax:
Note that this is on a deadline due to the fact that this affects a campaign going live on the 16th.

Pomax:
the immediate problem was solved, and the "real" solution hopefully surfaces as part of https://github.com/wagtail/wagtail/issues/7341

PR #7348, opened 2021-09-09
incorrect back reference

Pomax:
Addresses part one of #7347 by fixing the heading-linking code we have in place. This was using the wrong back-reference `\1` on in the regex `<h((\d)[^>]*)>(.+?)<\/h\1>`, which matches the first group: `((\d)[^>]*)`. As long as there are no attributes, this will be equivalent to matching the heading number, but once there's an attribute like in `<h2 data-block-key="...">` it will match "everything between the `h` and `>`" so it starts looking for the closing tag `</h2 data-block-key="...">` and obviously that is going to fail. Instead it should use the second group as backreference, which is only the `(\d)` part.

### DIFF
--- a/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
+++ b/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
@@ -25,7 +25,7 @@ __original__html__ = RichText.__html__
 # 1. "everything after the h" for the opening tag
 # 2. the heading number
 # 3. the tag's text content
-heading_re = r"<h((\d)[^>]*)>(.+?)</h\2>"
+heading_re = r"<h((\d)[^>]*)>(.+?)<\/h\2>"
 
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
```diff
@@ -25,7 +25,7 @@
 # 1. "everything after the h" for the opening tag
 # 2. the heading number
 # 3. the tag's text content
-heading_re = r"<h((\d)[^>]*)>(.+?)</h\1>"
+heading_re = r"<h((\d)[^>]*)>(.+?)</h\2>"
```
Pomax, line 28:
```suggestion
heading_re = r"<h((\d)[^>]*)>(.+?)<\/h\2>"
```

KalobTaulien (approved):

PR #7366, opened 2021-09-10
updated migrations for help text

danielfmiranda:
This is a PR which includes migrations reflecting updates to iframe blocks help text from whitelist to allowlist.


Pomax (approved):
let's roll out :shipit:

PR #7370, opened 2021-09-13
6987 Style Guide Design System updates

### LINKED ISSUE #6987: 4 - Improve the style guide
- [x] include spacing tokens in styleguide
- [x] include icons in styleguide
- [x] include streamblock molecules in styleguide
- [x] add molecules that aren't in the style guide
- [x] reorganize into atomic design [design input]
- [x] possibly move to storybook [design input]

fessehaye:
Closes #6987 

* adding spacing demonstration to style-guide
* adding streamfields to style-guide
* added icons to style-guide
* added local links to the style-guide for each header


danielfmiranda (approved):
Hi Simon! Just took a look at the files changed/percy comparison and everything looks good to me, thanks for taking the time to do this and adding all the components to the style guide!

sabrinang (changes requested):
- spacing section looks good
- anchor links work well 
- Icon section: looks like some of these assets aren’t icons but images and we should curate these as some are dated, stretched etc. Are these currently all the icons that are live on the site?
- `image`, `image text mini` & `video` usually fills the container width but not here. Video especially looks very small.
- You have an extra `image` section below quote and above card_grid (remove)
- `profile_listing` looks differently styled than what we have as the “profile listing” “profile by id” or “profile directory” components (where it has a line on top with smaller images)
- Missing `blog set`, `airtable`, & `typeform` components

sabrinang (dismissed):
- Icons: there are some icons still stretched/squished can you display at regular ratio (if not, remove)
- image text mini usually fills the container width 
image

sabrinang (approved):
Great thanks for fixing the widths and icon ratios. I think the icon section can use some attention with the sizing etc. but that would be good for a separate ticket

PR #7397, opened 2021-09-16
Replace USE_THOUSAND_SEPARATOR setting with intcomma filter

### LINKED ISSUE #7355: Turn off `USE_THOUSAND_SEPARATOR` so that templating of values without filters don't magically autofilter
We're currently running with Django's `USE_THOUSAND_SEPARATOR` set to True, which automatically converts any and all numbers used in templating to a string representation with separators in it, which means you can't rely on numbers actually being numbers. Let's turn this off (which is the default) and instead make sure that any number that needs to be localised _explicitly_ does so through the use of the [intcomma](https://docs.djangoproject.com/en/3.2/ref/contrib/humanize/#intcomma) filter instead.

- [ ] when we do, also roll back https://github.com/mozilla/foundation.mozilla.org/pull/7354

TheoChevalier:
Closes #7355

### DIFF
--- a/network-api/networkapi/settings.py
+++ b/network-api/networkapi/settings.py
@@ -200,6 +200,7 @@ INSTALLED_APPS = list(filter(None, [
     'django.contrib.staticfiles',
     'django.contrib.redirects',
     'django.contrib.sitemaps',
+    'django.contrib.humanize',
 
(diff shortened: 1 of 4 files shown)


Pomax (approved):
nice!

PR #7405, opened 2021-09-17
7393 - Update content width

### LINKED ISSUE #7393: [PNI] Product Page - update content width
- [ ] update the content width to match col-8 (@sabrinang to confirm that's the true new width?)

Current layout:

image

Relevant part of the product page comp:

image

Comps on figma: https://www.figma.com/proto/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?page-id=548%3A72128&node-id=611%3A77813&viewport=283%2C48%2C0.08&scaling=min-zoom&starting-point-node-id=611%3A77813

fessehaye:
Closes #7393

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/buyersguide/product_page.html
+++ b/network-api/networkapi/wagtailpages/templates/buyersguide/product_page.html
@@ -25,7 +25,7 @@
 <div class="text-center product-header bg-product-image{% if product.draft %} draft-product{% endif %}">
   <div class="container">
     <div class="row">
-      <div class="col-12 col-lg-10 offset-lg-1 product-header-content {% if product.privacy_ding %}show-privacy-ding{% endif %}">
+      <div class="tw-w-full large:tw-w-8/12 tw-mx-auto tw-relative product-header-content {% if product.privacy_ding %}show-privacy-ding{% endif %}">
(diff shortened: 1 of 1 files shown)


sabrinang (commented):
@fessehaye Which specific product page URL in the deployment should review? When I click on a product it is Bad Request (400) for me.

danielfmiranda (approved):
Hey Simon! Looks good to me, approving!

sabrinang (dismissed):
The text contents/updates/related products should align with the 8 col grid so it seems should be a bit wider:
screencap-grid
 
(I used a bootstrap grid overlay browser extension to check)

In the figma mockup, you can also go to View > Show Layout Grids for alignment reference

sabrinang (changes requested):
- the cut out sections can align to the edge of column 2 and 11 for the top section and dotted banner below
- the content below the dotted banner will be the same width as the content above to align with edge of column 3 and 10 as you have updated it
- highlighted in pink lines for your reference
image

image

sabrinang (changes requested):
Cut outs are good now but the text after the dotted banner should also align _(similarly to text above to the edge of col 3 and 10)_:
image

Issue #7398, opened 2021-09-16
PNI: Search in Snippets --> Buyers Guide Product Updates Returns Server Error (500)


caltrider:
### Describe the bug
When I try to search for a news article I've added to Snippets --> Buyers Guide Product Updates it returns a Server Error (500)

### To Reproduce
1. Go to https://foundation.mozilla.org/cms/snippets/wagtailpages/update/?locale=en
2. Click on search
3. Search for anything, like Facebook or Slack
4. See error --> Gives me a "Server Error (500)"

### Expected behavior
I expect it to return all the articles we've uploaded that match the search request

### Desktop (please complete the following information):
 - OS: [e.g. iOS] Mac iOS
 - Browser [e.g. chrome, safari] Firefox
 - Version [e.g. 22] 91.0.2

### Additional context
It makes it hard to find and edit news articles

PR #7414, opened 2021-09-20
add the missing FilterField for localized snippet filtering

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7398

Without this fix, the server will throw the following error when DEBUG is set to True:

image

And thankfully the error just tells us what the solution is =)

(although I don't know why it just doesn't "do that" without us having to write code, if it already knows both what's wrong, and what the fix is...)


danielfmiranda (approved):
Hi Pomax! Looks good to me, I am approving 👍

Issue #6688, opened 2021-05-12
Focus areas need factory images


KalobTaulien:
The "Focus Areas" on the Foundation homepage need factory images. On localhost they don't show anything. We could use standard blue boxes for these as the image itself in review apps isn't super important. 

For frontend reference, the code is in focus_area.html

KalobTaulien:
This is a bit trickier because I believe this is an Orderable, with a reference to another model. And if I also remember correctly, the initial models are populated by a migration. 

So we'll need to grab the objects _after_ the objects are created, but _before_ the factory code is finished running so we can add blank images as factory code. 

Alternatively, we could just grab random images from the the Wagtail Image object and use whatever has been generated by image factories already.

PR #7418, opened 2021-09-20
update areas of focus and assign them images during load_fake_data

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/6688

- updated Areas of Focus labels/text in the migration file (as one of the olders migration files, this will only affect load_fake_data/review apps, which is what we want)
- updated the homepage factory to assign areas of focus an image during load_fake_data


danielfmiranda (approved):
Hi Pomax,

Just tested this out and can confirm that images are now being added as well as the new text. 

Approving!

Issue #7392, opened 2021-09-14
[PNI] Product page - remove SoftwareProduct from the CMS


Pomax:
13-27 September sprint:

- [x] Keep the model around, keep the data, but remove any mention of the SoftwareProduct in the CMS nav/permissible child pages, etc.

Later:

- [ ] clean up the model + data later, once folks have gone over all the old softare products and deemed them removable (if not removed by staff themselves already).

Pomax:
part 1 is done, moving issue to backlog so it can be planned into a future sprint, with updated estimate.

tbrlpld:
There is 14 `SoftwareProductPage` items left on product, but they are all in draft state and not live.

```
>>> from networkapi.wagtailpages.pagemodels.products import SoftwareProductPage
>>> SoftwareProductPage.objects.count()
14
>>> SoftwareProductPage.objects.live().count()
0
```

PR #7419, opened 2021-09-20
hide the softwareproductpage child page type

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7392 by removing the `SoftwareProductPage` as valid child for the `BuyersGuidePage`. Or rather, commenting it off. This only affects how the admin behaves, and does not affect preexisting sofware products's on the website itself.

Log into the admin interface, navigate to the `privacy not included` page, and add a new child page. It should no longer offer the choice between general or software, and immediately open an edit interface for a general product page.


danielfmiranda (approved):
Hi Pomax,

Tested and can confirm that creating a child page underneath the Privacy Not Included page now defaults to the GeneralProduct Page.

Approving!

Issue #7349, opened 2021-09-09
Fix the auto-heading-link code for wagtail streamfield RichText headings 


Pomax:
Rich text headings with apostrophes lead to fragments ids with `x27` in them: 

```html
<h2 data-block-key="3n47r"><a class="fragment-id" id="someonex27s-fancy-header"></a><br>Someone's Fancy Header</h2>
```

This should be id `someones-fancy-header` instead. We'll need to update the `add_id_attribute` function in `wagtailcustom_tags.py to make sure we slugify the heading text before we set it as `id` attribute.

Marked as "deadline" as there is a campaign that launches on September 16 that relies on this functionality not being broken.

PR #7420, opened 2021-09-20
make sure to unescape HTML entities before slugifying

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7349 by applying `html.unescape` to the text content before stripping tags and slugifying.

STR:
- load the admin
- load the test blog post and change the first heading in the body to `it's working`
- publish and view the page
- inspect the heading in dev tools: it should now contain an `<a id="its-working">` rather than the `<a id="itx27s-working">` we would be generating without this fix

### DIFF
--- a/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
+++ b/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py
@@ -1,4 +1,5 @@
 import re
+import html
 from django import template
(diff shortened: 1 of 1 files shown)


danielfmiranda (approved):
Hi Pomax! 

Awesome fix, I just tested it out following the instructions and can confirm that everything is working as it should! 

Here is a screenshot:
Screen Shot 2021-09-20 at 6 00 37 PM

Approving 👍

Issue #7440, opened 2021-09-21
PNI voting seems to be broken for (some) products?


Pomax:
Looking at the database, there are a bunch of products with zero-vote data:
```
wagtail=# select distinct(title) from (
    select
        page.id, page.title as title, product.page_ptr_id, votes.id
    from
        wagtailcore_page as page,
        wagtailpages_productpage as product,
        wagtailpages_productpagevotes as votes
    where
        votes.vote_bins='0,0,0,0,0'
        and product.votes_id = votes.id
        and page.id = product.page_ptr_id
) as subquery order by title;

                      title
--------------------------------------------------
 BlueJeans
 Discord
 Doxy.me
 Facebook Messenger (and Facebook Messenger Kids)
 FaceTime
 Google Duo/Google Hangouts/Google Meet
 GoToMeeting
 Houseparty
 Jitsi Meet
 Marco Polo
 Ring Indoor Cam
 Signal
 Skype
 Slack
 Teams
 Telegram
 Threema
 Viber
 Webex
 WeChat
 WhatsApp
 Zoom
(22 rows)
```

And we need to figure out why voting does not work for these products, specifically.

kristinashu:
This seems to be product wide. I any testing this while logged into the CMS.

**Dating app: https://foundation.mozilla.org/en/privacynotincluded/hornet/**
Before voting: 151
<img width="881" alt="image" src="https://user-images.githubusercontent.com/1068783/134221812-94aef6a7-536b-4dbd-9d35-3e2fc005be72.png">
During vote: 152
<img width="872" alt="image" src="https://user-images.githubusercontent.com/1068783/134221875-edb5ec71-8ddf-4cd1-b46d-eb46e223f938.png">
After voting: 151
<img width="867" alt="image" src="https://user-images.githubusercontent.com/1068783/134221916-086d908c-f15e-4aa2-8461-8c86dd254eb3.png">

**Nintendo Switch: https://foundation.mozilla.org/en/privacynotincluded/nintendo-switch/**
Same thing, vote count is not increasing.

Pomax:
Looking at the database, it does not seem that new votes are getting saved, with the vote bins showing `28,24,57,22,20`, which sums to 151 rather than some higher number based on additional test votes.

Pomax:
running vote testing against our `load_fake_data` db shows no votes getting saved either, so it looks like there's a way to locally reproduce and figure out what's going on.

Pomax:
[...]

PR #7442, opened 2021-09-21
add CSRF token to the voting process

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7440 by ensuring that there is a Django csrf token available in the page source for our JS to tap into and send back to the server during the POST request.

STR for verifying it's broken on prod:
1. visit https://foundation.mozilla.org/en/privacynotincluded/nintendo-switch/
1. open dev tools, and open the network tab
1. clear your network tab's log so far
1. cast a creepiness vote 
1. look at the POST request made: it should be `403`, and if you click it and look at the request headers, it'll show `x-csrftoken: undefined` at the bottom of the header list
1. For double confirmation, reload the page, view-source, and confirm there is no string that contains `csrf` anywhere in the page source.

STR for verifying it's fixed in this patch:
1. open any PNI product in the review app
1. open dev tools, and open the network tab
1. clear your network tab's log so far
1. cast a creepiness vote 
1. look at the POST request made: it should be `200`, and if you click it and look at the request headers, it'll show `x-csrftoken: (some long string)` at the bottom of the header list
1. For double confirmation, reload the page, view-source, and confirm there is a hidden input element with name `csrfmiddlewaretoken` in the page source.

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/buyersguide/product_page.html
+++ b/network-api/networkapi/wagtailpages/templates/buyersguide/product_page.html
@@ -82,6 +82,7 @@
             <div class="creep-vote-target mb-5 mt-3 mt-md-4 p-5" data-product-name="{{product.title}}" data-product-type="{{product.product_type}}">
               <input type="hidden" name="productID" value="{{ product.id }}">
               <input type="hidden" name="votes" value='{{ product.get_voting_json | safe }}'>
+              {% csrf_token %}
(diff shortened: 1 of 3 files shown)


danielfmiranda (approved):
Hi Pomax! Reviewed,tested, and approved. I followed the instructions you have provided and can confirm that the fix works on the review app

PR #7461, opened 2021-09-23
Localize blog category title/share metadata

### LINKED ISSUE #7460: Add custom title to blog categories
Blog categories currently don’t have a dedicated title to be displayed on the category page and as a SEO title.
Instead we’re concatenating the category name with " Blog" and applying `titlecase()`.

From a localization standpoint, that’s quite wrong, so let’s get a dedicated field where staff can pick the title they want then translate this.

TheoChevalier:
Closes #7460

While that code is fresh in your mind, I’m improving the localization even further:

- Localizing share metadata: seo title + description + image
- Adding proper field for the page title instead of the concatenation with "Blog"


network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -114,17 +103,33 @@ def filter_entries_for_category(self, entries, context):
...
+        # then explicitly set all the metadata that can be localized, making
         # sure to use the localized category for those fields:
         locale = get_locale_from_request(context['request'])
         try:
             localized_category = category.get_translation(locale)
         except ObjectDoesNotExist:
             localized_category = category
 
-        context['index_title'] = titlecase(f'{localized_category.name} {self.title}')
+        context['index_title'] = localized_category.title
```
Pomax, line 114: Let's make sure to fall back to `titlecase(f'{localized_category.name} {self.title}')` if `localized_category.title` is empty

Pomax (changes requested):
Let's make sure there's always a page title =D

Pomax (approved):

PR #7455, opened 2021-09-23
7294 - PNI category name untranslated in page title

fessehaye:
Closes #7294

PR #7478, opened 2021-09-24
Localize PNI category name in page title

### LINKED ISSUE #7294: PNI category name untranslated in page title
On PNI category pages, the category name in the page title is left untranslated 

Capture d’écran 2021-08-31 à 11 56 59

TheoChevalier:
Closes #7294
Follow up from #7455

Turns out my comment in #7455 ended up in `Pending` state 🙃

We can’t pass strings dynamically to `gettext()`, here we just want to use the translation served by wagtail-localize


fessehaye (approved):
Thanks for the help!

Issue #7481, opened 2021-09-24
Load more results button on Blog doesn't work


benhohner:
### Describe the bug
When you click more results button on blog, more results don't load.

### To Reproduce
1. Go to https://foundation.mozilla.org/en/blog/
2. Click Load more results
3. Nothing happens

### Expected behavior
More blog posts should load

PR #7482, opened 2021-09-24
make sure locale fallback for get_entries gets assigned to the correct variable

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7481

To test, you'll need a copy of the stage db, as `load_fake_data` does not have enough blog posts to reproduce this behaviour.

STR:
- load the `/blog` page
- click the "load more" button
- more entries should load.

Compared to the current behaviour on https://foundation.mozilla.org/blog where clicking the button causes a server error 500 on the Fetch url  (see dev tools "network tab" when clicking the button to confirm that error).

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/index.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/index.py
@@ -119,7 +119,7 @@ class IndexPage(FoundationMetadataPageMixin, RoutablePageMixin, Page):
         DEFAULT_LOCALE = Locale.objects.get(language_code=DEFAULT_LANGUAGE_CODE)
 
         if 'request' in context:
-            locale = get_locale_from_request(context.get('request'))
+            locale = get_locale_from_request(context['request'])
         else:
             locale = DEFAULT_LOCALE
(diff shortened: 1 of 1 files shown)


network-api/networkapi/wagtailpages/pagemodels/index.py
```diff
@@ -117,9 +117,14 @@ def get_entries(self, context=dict()):
         """
         DEFAULT_LANGUAGE_CODE = settings.LANGUAGE_CODE
         DEFAULT_LOCALE = Locale.objects.get(language_code=DEFAULT_LANGUAGE_CODE)
-        locale = get_locale_from_request(context.get('request', DEFAULT_LOCALE))
+
+        if 'request' in context:
+            locale = get_locale_from_request(context.get('request'))
```
Pomax, line 122:
```suggestion
            locale = get_locale_from_request(context['request'])
```

danielfmiranda (approved):
Hi Pomax! Just tested this and it seems to be working well. Approved!

PR #7491, opened 2021-09-24
new function in products.py to get the data, software pages not working

danielfmiranda:
Hi Simon! 

Here is the PR with the get_tabs_data function in order to populate the tabs with the security information. 

I am not sure if this is the best method to get the data, but if this helps you populate the tabs and get the front end working, I am thinking we can run this by Pomax to see if there is a better method to get the data sent to the react component :D 

Please note that some AI fields are still missing, I must have missed them in my PR to add them and will have a pr that adds the rest of them soon. 

These fields are: 

- help text for is the ai untrustworthy.
- help text for ai transparency
- can the user control the ai, and the help tex pertaining to that. 

If you find any other fields missing, please let me know!


danielfmiranda:
Closed as this is not needed anymore, we have found a better way to pass the data through

PR #7518, opened 2021-09-28
Synchronize inlinePanel fields with translations

### LINKED ISSUE #7517: InlinePanel fields not kept in sync on translated pages
Turns out InlinePanel fields are not `synchronized` by default, so we need to explicitly add them to the `override_translation_fields` var.

TheoChevalier:
Closes #7517

Tested locally, fields are properly synchronized


Pomax (approved):
just that little bit closer to sanity.

Issue #7448, opened 2021-09-22
PNI ordering of news updates doesn't work in the CMS


caltrider:
When I try to reorder the news updates with up/down arrows in the CMS they don't seem to save and reorder. They move around but once I save, they seem to revert back to the order they were before. 

### To Reproduce
1. Go to any PNI product page in the CMS. Go down to the product updates section near the bottom.
2. Click on the up or down arrows to reorder a news updates
3. Scroll down save draft. 
4. Now go down and see if they updates are reordered in the order you'd hoped. They keep jumping around or not reordering for me. 

### Expected behavior
The updates reorder into the order I placed them and stay that way once saved and published.

Pomax:
This appears to be a two parter:
1. we forgot to add the `Orderable.Meta` to the product update meta class, which is kind of necessary (either no meta class, or make sure the meta class extends `Orderable.Meta`). This fixes the "saving" part of the problem
2. even with the `Orderable.Meta` superclass, and an explicit `ordering = ['sort_order']`, the actual ordering in terms of templating does not seem to change, which means that even if the ordering is correct in the CMS, that order doesn't make it onto the actual page users look at.

Pomax:
nope, I am an idiot, of course the live page won't update if you only save a draft.

PR #7530, opened 2021-09-29
update the ProductUpdate meta class to include ordering

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7448 by updating the `ProductUpdate` meta class, which was missing the Orderable.Meta superclass, and was missing the `sort_order` ordering field.

STR for current bad behaviour:
- run the pni-q3-2021 branch and look at any PNI product page in the CMS
- scroll down to the "updates" seection, take note of the ordering
- reorder the updates, and "save draft"
- note that the ordering that you changed it to did not get saved

STR for fix:
- repeat all steps on this branch, after `inv migrate` to apply the model change to the DB
- in the last step, note that the ordering _did_ change 
- publishing the page will then also has this new ordering when looking at the live page


danielfmiranda (approved):
Hi Pomax!

Just reviewed and tested, can confirm that the items are now ordering correctly.

Approved!

PR #7539, opened 2021-10-01
7377 mozilla researched

danielfmiranda:
Closes #7377 

**Link to sample test page**:

## Checklist

**Changes in Models:**
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

PR #7553, opened 2021-10-04
New localization approach for Mozilla researched

TheoChevalier:
Hey @danielfmiranda, the localization approach in #7539 needed some changes to:
- Support plural forms on "hour"
- Allow localizers to move the link/the number of hours around in the sentence

Also now that all the context in included in one single string, I’ve removed the localization note.

However, this approach creates some styling issues when you move things around, e.g. in French with
`Mozilla a effectué <span %(span_attrs)s>%(time_researched)s heures</span> de <a %(link_attrs)s>recherches</a>`

in particular spaces added with CSS
Capture d’écran 2021-10-04 à 12 24 14

Could you have a look please?

Feel free to merge this PR into your branch, or use this branch as a follow up.


danielfmiranda (commented):
Thanks for adding this in @TheoChevalier! I will merge it into my current branch and then I will request your review for the PR

Issue #7502, opened 2021-09-27
PNI creepy ratings on non-English pages is in the wrong order


Pomax:
How to Reproduce:

This was flagged on Twitter https://twitter.com/dtdan03/status/1441837328323031044 and was confirmed by staff
- Go to https://foundation.mozilla.org/fr/privacynotincluded/categories/video-call-apps/
- Notice apps like https://foundation.mozilla.org/fr/privacynotincluded/threema/ that have rated least creepy, are appearing towards the bottom of the page (it should be at the top of the page like it is on English version)

Pomax:
Looking at the DOM, the code shows:

![image](https://user-images.githubusercontent.com/177243/135351984-00f63f44-274b-4fe7-ad52-43c6066fb376.png)

However, looking at Viber (the first item), it has 224 votes, with a spread of 22%, 21%, 36%, 10%, and 11% which means it cannot possibly have an average creepiness of 0,18834080717488788

However, it _does_ have an average creepiness near 19% so it looks like the numbers aren't getting corrrectly turned into percentages.

Pomax:
Looking at the stats, it appears that voting is storing values "for locales" rather than "for the main product". Solving this one will require

- [ ] updating the POST handling to make sure that given a product ID, we find the _original_ product to update the votes for
- [ ] update the product model to find the original product on which to check the creepiness values
- [ ] write migrations that merge votes from all locales back to the original product.

Pomax:
Current code in `products.py`:

```python
    @property
    def total_vote_count(self):
        return sum(self.get_or_create_votes())

    @property
    def creepiness(self):
        try:
            average = self.creepiness_value / self.total_vote_count
        except ZeroDivisionError:
            average = 50
        return average
```

This needs those `self` replaced with `product`, based on finding the "real" product given a localized product.

Similarly, the POST code currently reads:

```python
        if request.body and request.method == "POST":
            data = json.loads(request.body)
            if data.get("value"):
                ...
                try:
                    product = ProductPage.objects.get(pk=self.id)
                    ...
```

where this product, too, should be resolved to the original English product, not the localized product.

Pomax:
[...]

PR #7555, opened 2021-10-04
Fix POST handling of localized PNI products

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7502

This gives PNI products an `.original_product` accessor for finding the English version of a product (based on product slug), with a refactor of the code that fetches the `DEFAULT_LOCALE` and `DEFAULT_LOCALE_ID`, which are required to make sure that we can find the original English version of products. We didn't have "one place to get them" so there were a bunch of files that all reinvented the wheel, all of those now use the same function to get default locale information.

This also update some of the `@property` functions for `ProductPage` so that total votes and creepiness are derived from the original product, and _not_ a localized alias/published version of that product, both of which are used in the average-creepiness calculation for ranking items on the homepage/category pages.

Finally, it makes sure to always template the creepiness value into the product page in an `unlocalized` fashion, so that the JS that hooks into it doesn't get confused by numbers that use commas instead of period for decimal points.

Bad case STR:
- be on the main branch,
- run an `inv copy-stage-db` to grab a copy of the staging database,
- run `inv migrate` just to make sure the db is at the tip of our migration chain
- run `docker-compose up` and load up http://localhost:8000/en/privacynotincluded/categories/video-call-apps/ as well as http://localhost:8000/fr/privacynotincluded/categories/video-call-apps/ in two tabs

This will show a different ordering, where on the English page, you should see Threema, Signal, FaceTime, and Jitsi Meet as top four, whereas on the French page, there are very different elements at the top.

- DOM inspect the pages, and you should see a list of `<figure...>` elements that have `data-creepiness` that starts at ~13 and ends at ~82 for the English page
- on the French page, you'll see these values start at `0,10660980810234541` (or similar), with comma instead of period, and end at `98,6` (or similar), again with comma.

Fixed case STR:
- `docker-compose down`
- check out this branch
- `docker-compose up`
- reload the two category pages
- both should show identical ordering
- inspecting the DOM should show identical `data-creepiness` values

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/products.py
@@ -643,7 +643,7 @@ class ProductPage(AirtableMixin, FoundationMetadataPageMixin, Page):
 
     @property
     def original_product(self):
-        return self.alias_of or get_original_by_slug(ProductPage, self.slug)
+        return get_original_by_slug(ProductPage, self.slug)
 
     def get_or_create_votes(self):
         """
@@ -659,12 +659,16 @@ class ProductPage(AirtableMixin, FoundationMetadataPageMixin, Page):
 
     @property
     def total_vote_count(self):
-        return sum(self.get_or_create_votes())
+        # Voting only happens on the original product, not "self"
+        product = self.original_product
+        return sum(product.get_or_create_votes())
 
     @property
     def creepiness(self):
+        # Creepiness is tied to the votes on the original product, not "self"
+        product = self.original_product
         try:
-            average = self.creepiness_value / self.total_vote_count
+            average = product.creepiness_value / product.total_vote_count
         except ZeroDivisionError:
             average = 50
         return average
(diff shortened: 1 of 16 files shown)


network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -636,14 +641,34 @@ def get_status_for_airtable(self):
             return "Live"
         return "Draft"
 
+    @property
```
Pomax, line 644: this just got moved up, so that it's next to the other vote-related function/properties

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -636,14 +641,34 @@ def get_status_for_airtable(self):
...
+    def original_product(self):
+        return get_original_by_slug(ProductPage, self.slug)
+
+    def get_or_create_votes(self):
```
Pomax, line 648: this just got moved up, so that it's next to the other vote-related function/properties

network-api/networkapi/wagtailpages/tests.py
```diff
@@ -375,33 +375,35 @@ def test_get_or_create_votes(self):
 class WagtailBuyersGuideVoteTest(APITestCase, BuyersGuideTestMixin):
 
     def test_successful_vote(self):
+        product_page = self.product_page
```
Pomax, line 378: This is basically a noop refactor, I thought I would be able to add tests in this PR, but it turns out that's much more work, so I filed #7559 to tackle that monster.

TheoChevalier (approved):
Did some testing with the staging DB, couldn’t spot any issue

PR #7567, opened 2021-10-05
improve the copy script to automatically find the right container names

Pomax:
No associated issue, this makes the copy script look for the container names based on `docker ps`, so it no longer has any hardcoded container names. It also now force-drops the db, so that even if the database is in use (even if it's not, but there's some .lock file somewhere in the postgres docker image) things keep working.


danielfmiranda:
Worked great for me @Pomax!

danielfmiranda (approved):
Works great for me @Pomax!

Issue #7284, opened 2021-08-30
Create new snippet for call campaigns integrating Callpower


benhohner:
We need to integrated the Callpower call activism tool into Fo'Mo. It allows you to enter your phone number and receive a call that connects you with one of more of your local representatives. The purpose of this is for the new head of the FCC election. We'd like to reach out to people to get them to express support for a progressive ex-mozillian candidate.

- [ ] Create a new campaign snippet type for call campaigns
- [ ] Reference code from previous implementation (Pomax to find)
- [ ] Add a checkbox for signing up for the email newsletter.

benhohner:
@benhohner to add docs from call with Josh

benhohner:
Notes from meeting with Josh:

- Can submit directly to callpower with a form post.
- Add a js hook on the submit action, to send to both CRM and callpower
- Response in json says Call queued
- Often want to show someone what to say when they're talking. Can either store in callpower or on site. Use that to show a modal or something.
- "We're calling you now. Here's what to say."
- Phone will call with delay, can set delay
- https://github.com/OpenSourceActivismTech/call-power/blob/master/INTEGRATION_JS.md#calling-endpoints-directly, easiest way, if you have devs, don't need to use
- Testing?
  - Make a test campaign, run as many calls through it as you like
- Volume
  - Set up on couple thousand calls per month.
  - If we end up running hundreds of thousands of calls per month, let Josh know, he can bump up server resources.
- Gotchas
  - Their rep. If you call one number, you can overwhelm the target.
  - There is a rate limit, so people don't call numbers more than a couple times an hour

Pomax:
Looking at https://github.com/OpenSourceActivismTech/call-power this appears to be a "deployable by us" server, is that the actual intention here?

benhohner:
We're just looking to call their RESTful API to submit a call action to it
when someone fills in their phone number. We definitely don't need to host
anything!

Pomax:
@ben I cannot find any documentation on that RESTful API or which endpoint(s) to use, could you (or someone else) either drop that information into this issue, or communicate that to me through non-public channels?

Pomax:
While it's easy to implement this as extension to CTA, having existing campaigns allow for "any extension of a CTA" is proving fairly complicated in wagtail (as in, not sure if we can do this) so I'm asking for help on the wagtail slack, but if this isn't possible we may need to rethink how we want to do this.

benhohner:
[...]

PR #7577, opened 2021-10-06
Add a Callpower CTA

Pomax:
Will close https://github.com/mozilla/foundation.mozilla.org/issues/7284 when finalized.

Demo URL: https://foundation-s-genericise-ftncey.mofostaging.net/en/campaigns/initial-test-bannered-campaign-with-fixed-title/

edit page: https://foundation-s-genericise-ftncey.herokuapp.com/cms/pages/76/edit/
snippet edit page: https://foundation-s-genericise-ftncey.herokuapp.com/cms/snippets/wagtailpages/callpower/edit/27/

## Testing

there are several "secret" buttons for toggling the various states, because we don't have a test campaign set up atm that allows actual calls to go through, so:

1. Confirm that phone number and zipcode have validation: phone number should only allow a 10 digit US phone number, and zipcode should only allow a five digit US zipcode. You should get errors if that's not the case and you press "make the call".
2. the "show thank you" button should trigger what happens when calling works and the callpower API accepts the make-a-call button click.
3. the "bad fields" button should trigger the "you filled in a phone number we can't call" flow, keeping users on the form but showing them an error text. Once the user edits the phone number or zipcode, this text should disappear again.
4. the "rate limited" button should trigger what happens when the user's placed too many calls per hour already
5. the "show error" button should trigger what happens when things go horrible wrong and we have no idea why.

## Testing against staging/production

To make sure the migrations don't mess with preexisting data, please also test this with the staging database:
1. `inv copy-stage-db`
2. fire up `docker-compose up` and head over to http://localhost:8000/campaigns
3. in a separate tab, open https://foundation.mozilla.org/campaigns
4. open a bunch of campaigns on both, and verify that anything with a petition looks the same between localhost and staging. If they don't, that's a problem =)

## Notes

- the testing buttons must be removed before merging

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta/callpower.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/tags/cta/callpower.html
@@ -41,11 +41,6 @@
     <h5>{% trans "Thank you for all your calls!" %}</h5>
     <p>{% trans "Unfortunately, it looks like you’ve reached the hourly limit. Please try again in an hour." %}</p>
   </div>
-
-  <button class="btn btn-primary test-toggle-ty" style="width:100%; opacity: 0.1; margin-top:4em;">Show thank you</button>
-  <button class="btn btn-primary test-toggle-bad" style="width:100%; opacity: 0.1; margin-top:0.25em;">Bad fields</button>
-  <button class="btn btn-primary test-toggle-limit" style="width:100%; opacity: 0.1; margin-top:0.25em;">Rate limited</button>
-  <button class="btn btn-primary test-toggle-err" style="width:100%; opacity: 0.1; margin-top:0.25em;">Show error</button>
 </div>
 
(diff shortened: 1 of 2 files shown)


Pomax:
@TheoChevalier added you just to make sure I didn't miss an obvious place where things should be translated but I forgot to add the appropriate code for that.

kristinashu:
Yes, that looks good Pomax!

kristinashu:
The text looks good. But please also make the form field outline red when there are errors in them so it aligns with the petition form.

- [ ] Update red form field outline when there is an error 

Screen cap from PR
image

Screen cap from petition form on prod
image

Pomax:
@richbrennan reflagging you for review, I updated the migration file to use the pk rather than title/header combination, since pk worked perfeclty fine, but I also made it much easier for you to do local verifcation against staging - I've added some additional instructions to the first comment.

Pomax:
@fessehaye reflagged you just to make sure the now final JS makes sense =)

PR #7637, opened 2021-10-18
[7571] On the BanneredCampaignPage, only show root page title if the …

### LINKED ISSUE #7571: child pages of "Giving" don't show their own title
STR:
- Go to https://foundation.mozilla.org/cms/pages/1437/edit/
- See the title and header text in the CMS
- See the title on the live page https://foundation.mozilla.org/en/giving/happy-holidays-mozilla/ that says "Giving"
- It should say "Happy Holidays from Mozilla"

Jesse W is trying to create a new page under /giving and is having this same issue. Page is launching next week.

richbrennan:
…menu is shown too

Closes #7571

This PR contains a change to the Banner Page template to only show the parent page title if the menu is shown:

- Page with menu, parent title shown: https://foundation-s-fix-7571-g-pex1xa.mofostaging.net/en/campaigns/multi-page/concern-letter-nor/
- Page without menu, pages own title shown: https://foundation-s-fix-7571-g-pex1xa.mofostaging.net/en/giving/happy-holidays-from-mozilla/
    - Edit: https://foundation-s-fix-7571-g-pex1xa.herokuapp.com/cms/pages/165/edit/

The menu is shown when one or more child pages of the parent have 'Promote' > 'Show in menus' ticked.

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/campaigns.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/campaigns.py
@@ -17,10 +17,7 @@ from modelcluster.contrib.taggit import ClusterTaggableManager
 from .modular import MiniSiteNameSpace
 from .primary import PrimaryPage
 from .mixin.foundation_metadata import FoundationMetadataPageMixin
-from ..utils import (
-    get_page_tree_information,
-    get_content_related_by_tag
-)
+from ..utils import get_content_related_by_tag
 
 
 class CTA(models.Model):


Pomax:
Good find!

Pomax:
Hm, one thing that percy flags: `https://foundation-s-fix-7571-g-pex1xa.mofostaging.net/en/campaigns/initial-test-bannered-campaign-with-fixed-title/` no longer uses its title, but now uses its header instead. 

@kristina for a banner page with both title and header, what should happen in this case?

kristinashu:
Yikes, thanks for flagging that @Pomax! It should show the Title at the top of the page. TBH I don't really understand the differences in the Title and the Header.

Pomax:
@kristina I just did a test against `main` and it does the same thing: giving is a banner page, and so has 1 level of children that it will show up in the nav. Children of children do not show up, so giving→holiday→sub won't show.

Pomax:
@richbrennan @kristinashu I'm okay with removing the header field if we're not using it, but then let's make sure we also test this against our current real data in the staging db. @richbrennan you should be able to run `inv copy-stage-db` locally for testing.

kristinashu:
Thank you for the list Rich. So this update won't change anything on /campaigns pages? The original issue isn't happening on those pages so they don't need any update. I'm just checking because they are our most used pages.
E.g.
https://foundation.mofostaging.net/en/campaigns/election-integrity-and-platform-accountability/
https://foundation.mofostaging.net/en/campaigns/tell-amazon-connected-devices-must-have-privacy-policies/

Pomax:
Having said that, https://foundation.mofostaging.net/en/giving no longer seems to show _any_ content, despite the CMS clearly showing that there is (see https://foundation.mofostaging.net/cms/pages/1126 and https://foundation.mofostaging.net/cms/pages/1126/edit), so that's definitely a problem. @richbrennan it might make sense to put this PR on pause for a bit to see if we first write out the exact behaviour we want to have happening here (with @kristinashu's input/sign-off) before we then update the code to implement that behaviour.

richbrennan:
**Page/template investigation**

This investigation covers what is displayed in the H1 element on applicable pages as per the original issue. I haven't looked at the menus that use the header field as that's a different component, do we need to investigate these too, if so would a separate issue be less likely to cause confusion?

The title/heading logic that displays the H1 is contained in the `primary_heroguts.html` template. This is used by the following pages:

- Banner Page: `bannered_campaign_page.html`
- Homepage: `homepage.html`
- Primary Page: `primary_page.html`
- MozFest Homepage: `mozfest_homepage.html`
- MozFest Primary page: `mozfest-base.html`

The H1 is populated by these variables in the following priority order:
1. root title
2. page header
3. page title

root is a custom variable (an instance of a page) who's value is set depending on the page's parent type, see individual descriptions below.

**Individual page logic**

Homepage
The title/heading logic of the `primary_heroguts.html` template is not used on the homepage.

Banner Page
If an ancestor page (i.e. a page above this one in the page tree) is also a Banner page, the root variable is set to be the highest ancestor Banner page up in the page tree. This ancestor's title will be shown in the H1: https://foundation.mofostaging.net/en/what-we-fund/fellowships/senior-fellows/
If there are no other Banner pages as ancestors then the root is set to the page itself, so the pages' own title is shown in the H1.
I don't believe that header will ever be shown in the H1 on a Banner page, because its always overridden by the root variable (ancestor page or the actual page).

Primary Page
The Primary page uses the same logic as the Banner Page for determining the root variable and what to output in the H1.
Example Primary page with a Primary page ancestor,  so root is set to the ancestor whos title is displayed in the H1: https://foundation.mofostaging.net/en/internet-health/trustworthy-artificial-intelligence/
Example Primary page without a Primary page ancestor, so root is set to itself: https://foundation.mofostaging.net/en/press-center/

MozFest Homepage
The title/heading logic of the `primary_heroguts.html` template is not used on the MozFest homepage.

MozFest Primary Page
The MozFest Primary page uses the same logic as the Banner Page for determining the root variable and what to output in the H1.
Example MozFest Primary page with a MozFest Primary page ancestor,  so root is set to the ancestor whos title is displayed in the H1: https://mozillafestival.mofostaging.net/en/on-demand/virtual-dialogue-and-debates/big-tech-and-germanys-election/
Example MozFest Primary page without a MozFest Primary page ancestor, so root is set to itself: https://mozillafestival.mofostaging.net/en/technologists/ai-irl-hackathon-building-trustworthy-ai/


**Questions/issues**

> Thank you for the list Rich. So this update won't change anything on /campaigns pages? The original issue isn't happening on those pages so they don't need any update. I'm just checking because they are our most used pages. E.g. https://foundation.mofostaging.net/en/campaigns/election-integrity-and-platform-accountability/ https://foundation.mofostaging.net/en/campaigns/tell-amazon-connected-devices-must-have-privacy-policies/

The first link is a Banner page with a Campaign Index page ancestor (parent), so root will be the page itself and its own title will show in H1.
The second link is a Campaign page which is not affected.

With regard to the last three comments related to staging, none of my changes are on staging and these seem to related to the menu rather than the H1, so maybe we discuss and spin off into a separate issue?
https://github.com/mozilla/foundation.mozilla.org/pull/7637#issuecomment-947888039
https://github.com/mozilla/foundation.mozilla.org/pull/7637#issuecomment-947893781
https://github.com/mozilla/foundation.mozilla.org/pull/7637#issuecomment-947897649

Issue #7617, opened 2021-10-14
[PNI] - Subcategories - create backend subcategory functionality


Pomax:
- [ ] extend PNI category snippets with a "parent" field so that new categories can be created that are "children of" some other category.
- [ ] add some convenience functions: `get_parent` and `get_children` to make sure we can easily tap into those relations in our templates

Issue #7640, opened 2021-10-18
[PNI] - Subcategories - Improve subcategory sorting in admin


Pomax:
Followup to https://github.com/mozilla/foundation.mozilla.org/issues/7617 to figure out how to order the snippets nicely in the admin interface.

PR #7641, opened 2021-10-18
Add subcategory functionality to PNI categories.

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7617

Proper admin interface sorting will be tackled in https://github.com/mozilla/foundation.mozilla.org/issues/7640

Note: this can't be testing in the review app, because the only way to verify the values work for templating purposes is to already have templates written, which we can't because that depends on this existing. So... local testing!

### Testing:

- create a bunch of PNI category snippets, with the "parent category" set to something.
- add this into the `catalog.html` template so you can actually see these values doing what they should be doing:

```jinja
<h1>children for {{ current_category.name }}:</h1>
<ul>
  {% for child in current_category.get_children %}
    <li>{{ child.name }}</li>
  {% endfor %}
</ul>
```

- And then navigate to the category page that you made some test subcategories for. All of those should show up.

Also note that in the admin interface, PNI snippets should be ordered as "parent categories first, alphabetically ordered", followed by "and then all sub categories, with a reverse alphabetical parent order, and regular alphabetical ordering within that. This is because of the Meta class rule `ordering = ['sort_order', '-parent__name', 'name', ]` where the `parent__name` needs to be reverse because if there is no parent, that `parent__name` is ` None` and `None` comes first, which would be super not-useful.

### Testing part 2:

- you should be able to make new categories _without_ setting a parent category.
- you should be able to assign both toplevel categories and subcategories to products.

### DIFF
--- a/network-api/networkapi/wagtailpages/migrations/0045_pni_subcategories.py
+++ b/network-api/networkapi/wagtailpages/migrations/0045_pni_subcategories.py
@@ -7,7 +7,7 @@ import django.db.models.deletion
 class Migration(migrations.Migration):
 
     dependencies = [
-        ('wagtailpages', '0043_auto_20211011_2340'),
+        ('wagtailpages', '0044_auto_20211013_0018'),
     ]
 
     operations = [


Pomax:
@richbrennan I think in this case for (1) we're okay with that, and for (2) let's land this first, and then make a snippet `save` reject the data if that introduces a cycle in follow-up. Given that we have a fixed ontology that we're going to implement with this mechanism (see link in issue) we shouldn't be running into cycles yet, giving us plenty of time to hardned the snippet's form handling/save logic.

Issue #7476, opened 2021-09-23
[PNI] Search - update search fields/ranking


Pomax:
Felds to include in filter-search:

- Currently 
- Product name
- Company name
- Blurb
- Worst case
 
Sort/ranking priority of the results:

- Product name
- Company name
- Blurb

Potential third party libraries if we want ranking and a standard `sort(fn)` doesn't cut it:

- https://github.com/nextapps-de/flexsearch
- https://github.com/olivernn/lunr.js

jamilasnell:
"One of the last things we will be doing with the search work!" - Pomax, 2021

PR #7649, opened 2021-10-20
[PNI] Search - update basic search behaviour

### LINKED ISSUE #7475: [PNI] Search - update basic search behaviour
- [ ] keep product filtering, and make sure that categories are always listing all products, with out-of-category products hidden through CSS  by default.

When the user starts searching:
- [ ] “unhighlight” the category, if we're on a category
- [ ] highlight the "all" category
- [ ] present “search” results from the full PNI product set.

fessehaye:
Closes #7475 #7476 

 keep product filtering, and make sure that categories are always listing all products, with out-of-category products hidden through CSS by default (Right now categories are prefiltered to only show in-category products)
When the user starts searching:

*  “unhighlight” the category, if we're on a category
* highlight the "all" category
* present “search” results from the full PNI product set.
 
 Sorting/ranking priority of the results:

* Product name
* Company name
* Blurb

No changes to design

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/products.py
@@ -169,7 +169,6 @@ class BuyersGuideProductCategory(TranslatableMixin, LocalizedSnippet, models.Mod
         FieldPanel('description'),
         SnippetChooserPanel('parent'),
         FieldPanel('featured'),
-        SnippetChooserPanel('parent'),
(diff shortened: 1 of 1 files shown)


Pomax:
Hm, interesting problem: it looks like search.js does not load until all images have loaded, which means you can't search for quite a while, particularly noticable when running a stage copy without S3 settings, which causes all product images to 404, and searchign doesn't work until _all_ of those 404s have finished

Issue #6934, opened 2021-06-24
Convert ListBlock of Quotes to a Single Quote Block


KalobTaulien:
ListBlocks aren't supported in wagtail-localize yet. There's an RFC open in wagtail core to support ListBlocks by giving them a UUID, but it hasn't been worked on yet. 

Our ListBlock of Quotes accepts more than one quote (because its a ListBlock) but the template doesn't support more than one quote, so we can safely migrate the first quote in a ListBlock to a single quote.

## Acceptance criteria
- [ ] ListBlock needs to be removed
- [ ] A single quote block with the _first_ quote in the ListBlock needs to be created 
- [ ] A data migration should upload all ListBlock quotes automagically 
- [ ] Template needs to be updated (its currently a loop, but won't need to be anymore) 

## Ideas:
* Re-purpose the StructBlock that houses the ListBlock

KalobTaulien:
Putting this ticket on the backburner. Most of the work is done, but it's going to add a level of complexity to the localization work that we aren't willing to jeopardize at the moment. So we'll tackle this in the future. 

PR #6936 for reference. 

Moving back to Backlog and updating the estimate to a 2.

jamilasnell:
#ForRich?

richbrennan:
New pull request: https://github.com/mozilla/foundation.mozilla.org/pull/7660

PR #7660, opened 2021-10-21
[6934] Convert ListBlock of Quotes to a Single Quote Block

richbrennan:
This PR replaces https://github.com/mozilla/foundation.mozilla.org/pull/6936

Closes ticket #6934

**Link to sample test page**:

This is tricky to test on review apps as they contain different test data, but the same page across two review apps has a quote at the bottom of the StreamField:

- With quote list block (may not work if merged, but another review app should show similar): https://foundation-s-dependabot-i8nxyq.herokuapp.com/cms/pages/13/edit/
- With single block: https://foundation-s-feature-69-zmrc3j.herokuapp.com/cms/pages/13/edit/

With a local production backup the local page can be compared to the same on production:

- https://foundation.mozilla.org/cms/pages/684/edit/
- http://localhost:8000/cms/pages/684/edit/

**Outstanding questions:** 

1. Should I remove the Quote ListBlock from the code entirely? Its currently removed from page StreamFields?
2. The [original PR](https://github.com/mozilla/foundation.mozilla.org/pull/6936/commits/f92b8175ce710aa06d6e578791a0a35483853c79#diff-84c6d75c9e6ad48fbeca0b1706c1aa92c58769cfd850e299055911c162521777R47) only included pages with the default locale (English), however pages in other locales still have the Quote ListBlock in the page body data. The single quote is populated when the page is translated manually, but I suspect it'd be best to include pages for other locales in the data migration?


tasks.py
```diff
@@ -221,6 +221,12 @@ def makemigrations(ctx):
     manage(ctx, "makemigrations")
 
 
+@task(aliases=["docker-makemigrations-dryrun"])
```
richbrennan, line 224: Not sure if there's a better way to achieve this, but I use the `--dry-run` regularly

Pomax:
@richbrennan is this still a draft, though?

richbrennan:
Hey @Pomax 

Doh, it's review-ready, bearing in mind the outstanding questions that may lead to some additional changes.

Thank you!

TheoChevalier (changes requested):
Hey @richbrennan, thanks, that the new block works well with wagtail-localize :)

> The original PR only included pages with the default locale (English), however pages in other locales still have the Quote ListBlock in the page body data. The single quote is populated when the page is translated manually, but I suspect it'd be best to include pages for other locales in the data migration?

Yeah, we’d need to migrate alias pages for sure, manually translated pages would be "good to have".
For alias pages it’s really needed as right now the quote gets removed from the page and we have no way to add it back from the admin.

For manually translated pages, it disappears as well, but it gets added back by syncing the translations — which I plan to do as soon as this PR gets deployed. With the only caveat that I need to click sync on all pages that are translated + have a quote. It shouldn’t be a lot, but the risk is I miss one or two.

richbrennan:
Thanks for the review @TheoChevalier 

If possible, would you be kind enough to add some links to a couple of manually translated pages on production please, ideally ones that currently include the quote list? It'll help me investigate the manually translated pages.

Also, can I just check, by alias pages you're referring to untranslated versions of a page?

TheoChevalier:
> Thanks for the review @TheoChevalier
> 
> If possible, would you be kind enough to add some links to a couple of manually translated pages on production please, ideally ones that currently include the quote list? It'll help me investigate the manually translated pages.

Hey @richbrennan, here’s a translated page with a quote: https://foundation.mozilla.org/fr/who-we-are/
However I don’t think we have any translated page using the quote list…

> Also, can I just check, by alias pages you're referring to untranslated versions of a page?

Correct

richbrennan:
Thanks @TheoChevalier 

I did some additional local testing of alias and manual translated pages now I've removed the locale filter. The example https://foundation.mozilla.org/en/who-we-are/ does have a quote list, which does show in the French version of the page once the migration has been run and the page has been manually synced:
Screenshot from 2021-11-08 15-57-03

I added a quote list item to the English version a page locally, ran the migration and checked the alias version in the terminal, which showed that it has been converted to a single quote:

English page before:
Screenshot from 2021-11-08 16-33-21

French alias page after:
Screenshot from 2021-11-08 16-24-31

TheoChevalier (approved):
Thanks @richbrennan! Looks good to me now

tasks.py
Pomax: I don't think I ever do, but if this makes someone's life easier: \o/

Pomax (changes requested):
Looks good to me! However, the migration numbers are a bit out of date now, and will need an update before this can get final approval.

Pomax (changes requested):
looks like percy is failing, with quote blocks having disappeared

Pomax (approved):
good to go, land it land it land it O_O

PR #7709, opened 2021-10-27
load search when the page loads, rather than when react is done

### LINKED ISSUE #7667: [PNI] Search - make search.js its own file, so we can load it independently of the React bundle
This is a followup to #7649, which updates the search behaviour, but suffers from having to wait for the React bundle to start up, which is itself constrained by all product card images having to load in (or error out).

The PNI search functionality does not depend on React, instead relying purely on the DOM being available, so we should refactor it so that it gets built as its own file (similar to `directory-filter-listing.js` or `callpower.js`) and then load it in the `catalog.html` template as a `<script src="{% static "search.js" %}" async defer></script>`.

Pomax:
Closes #7667

This updates the build so that search.js ends up as "its own file" in the frontend/js dir, getting loaded as part of the `<head>` payload for catalog pages (PNI homepage + category pages) with `async defer` so that it can be used as soon as the page DOM is ready for reading.

Note that you'll want to hide whitespace for reviewing file changes in this case: https://github.com/mozilla/foundation.mozilla.org/pull/7709/files?diff=unified&w=1

Also note that this will need a "redo" for the search.js file itself, once #7649 has landed.

Test URL: https://foundation-s-externaliz-0hxqyx.herokuapp.com/en/privacynotincluded/

### DIFF
--- a/esbuild.config.js
+++ b/esbuild.config.js
@@ -25,6 +25,7 @@ const sources = {
   "callpower": `foundation/pages/callpower.js`,
   "directory-listing-filters": `foundation/pages/directory-listing-filters.js`,
   "bg-main": `buyers-guide/bg-main.js`,
+  "bg-search": `buyers-guide/search.js`,
   polyfills: `polyfills.js`,
 };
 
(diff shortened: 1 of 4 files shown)


Pomax:
Superceded by https://github.com/mozilla/foundation.mozilla.org/pull/7741

PR #7710, opened 2021-10-28
WIP: updated some fields to rich text, and updated criterion files

### LINKED ISSUE #7694: [PNI CMS QA] Update multiple "plain text" fields to "rich text"
More fields now need to allow for url links. Please make the follow fields richtext:

- [x] Know track record description
- [x] How does the company use this data?
- [x] How can you control your data?
- [x] General AI description
- [x] Manages security vulnerabilities Description

Reminder to also do what will be done in https://github.com/mozilla/foundation.mozilla.org/issues/7692

danielfmiranda:
Closes #7694 
Related PRs/issues #

**Link to sample test page**:


network-api/networkapi/wagtailpages/migrations/0049_pni_criterion_richtext_fields.py
```diff
@@ -0,0 +1,72 @@
...
+            print(f"Saved product: {product.title}")
+
+
+
+
+def gather_products_and_update_fields(apps, schema):
+    GeneralProductPage = apps.get_model('wagtailpages', 'GeneralProductPage')
+    SoftwareProductPage = apps.get_model('wagtailpages', 'SoftwareProductPage')
+
+    LOCALE_ID=1
```
Pomax, line 50: while this will (probably) work, we shouldn't hardcode this number. There is a `get_default_locale` function that you can use from https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/utils.py#L427 here instead. so you can do:

```python
    (DEFAULT_LOCALE, DEFAULT_LOCALE_ID) = get_default_locale()
```

and then use that default local id value.

network-api/networkapi/wagtailpages/templates/buyersguide/product_page.html
```diff
@@ -160,19 +160,19 @@ <h3 class="pni-product-smaller-body-bold mb-1 mb-md-2">{% trans "What data does
 
 
           {% trans "How does it use this data?" as how_does_it_use_data_collected %}
-          {% include "fragments/product_criterion.html"  with  label=how_does_it_use_data_collected  value=product.how_does_it_use_data_collected  ding=product.data_collection_policy_is_bad %}
+          {% include "fragments/product_criterion.html"  with  label=how_does_it_use_data_collected  value=product.how_does_it_use_data_collected  ding=product.data_collection_policy_is_bad rich_text_value=True %}
```
Pomax: while this will work, let's file a follow-up issue to see if can determine this "from the value itself" rather than assuming it's `True`, and then we can make that something the template can take care of for us, rather than needing us to hardcode it.

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -534,10 +534,11 @@ class ProductPage(AirtableMixin, FoundationMetadataPageMixin, Page):
     )
 
     # How does it use this data?
-    how_does_it_use_data_collected = models.TextField(
+    how_does_it_use_data_collected = RichTextField(
         max_length=5000,
+        features=['bold', 'italic', 'link'],
```
Pomax, line 539: let's file a follow-up to unify this set somewhere, too, because if more than two fields all need it, that's "a property of a set of fields" rather than "hardcoded values".

network-api/networkapi/wagtailpages/templates/fragments/product_criterion_primary_info.html
```diff
@@ -1,42 +1,47 @@
...
+    </h3>
+    {% if rich_text_value %}
+      <div class="criterion-richtext-value mb-0">
+        {{ value|richtext }}
+      </div>
     {% else %}
-      {% if show_value_as_symbol %}
-      <img src="{% get_static_prefix %}_images/buyers-guide/score/icon-{{value|lower}}.svg" alt="{{value}}">
+    <p class="rating pni-product-smaller-body mb-0">
+      {% if value == "Can’t Determine" or value == "CD"%}
```
Pomax, line 18: Why do we need the dual check here? This should never be the full expanded string, so if it is, that's... a problem.

Pomax (changes requested):

danielfmiranda:
Closing in place of #7754

PR #7712, opened 2021-10-28
7623 - Subcategories - front-end styling for desktop

### LINKED ISSUE #7623: [PNI] - Subcategories - front-end styling for desktop
- [x] add category title on a category view
- [x] add a subcategories "pill" section when on a category view, underneath title
- [x] recolor bg on hover
- [x] click pill to filter out all products not in that category
  - [x] recolor bg on selected
  - [x] click selected pill to undo the filtering

default:

> image

hover:

> image

selected:

> image

Figma comps: https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=1581%3A125095
Prototype: https://www.figma.com/proto/GwdoUiY4fzVyeLfXsAJGPq/Privacy-Not-Included-Website?page-id=1173%3A101577&node-id=1505%3A119992&viewport=241%2C48%2C0.25&scaling=scale-down-width&starting-point-node-id=1505%3A119992

fessehaye:
Closes #7623 

Also includes some config changes for eslint and prettier for our node tests

Does not include mobile subcategory tasks

https://foundation-s-7620-pills-bdzrnb.herokuapp.com/en/privacynotincluded/

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/buyersguide/bg_base.html
+++ b/network-api/networkapi/wagtailpages/templates/buyersguide/bg_base.html
@@ -94,9 +94,11 @@
 
           {% for cat in categories %}
             {% with original=cat.original %}
-            {% if original.published_product_page_count > 0 %}
-              {% localizedroutablepageurl home_page 'category-view' original.slug as cat_url %}
-              <a class="multipage-link {% check_active_category current_category cat %}{% if original.featured is True %} featured{% endif %}" href="{{ cat_url }}" data-name="{{ cat.name }}">{{ cat.name }}</a>
+            {% if cat.parent == None %}
+              {% if original.published_product_page_count > 0 %}
+                {% localizedroutablepageurl home_page 'category-view' original.slug as cat_url %}
+                <a class="multipage-link {% check_active_category current_category cat %}{% if original.featured is True %} featured{% endif %}" href="{{ cat_url }}" data-name="{{ cat.name }}">{{ cat.name }}</a>
+              {% endif %}
             {% endif %}
             {% endwith %}
           {% endfor %}
--- a/network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
+++ b/network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
@@ -1,13 +1,15 @@
 {% extends "buyersguide/bg_base.html" %}
 
-{% load env i18n static wagtailimages_tags cache bg_nav_tags %}
+{% load env i18n static wagtailimages_tags cache bg_nav_tags localization %}
 
 {% block bodyclass %}pni catalog{% endblock %}
 
 {% block main_content_class %}{% endblock %}
 
 {% block hero %}
-<input type="hidden" class="category-title" value="{{current_category.localized.name}}">
+<input type="hidden" class="category-title" value="{% if current_category %}{{current_category.localized.name}}{% else %}None{% endif %}">
+<input type="hidden" class="parent-title" value="{{current_category.parent.localized.name}}">
+
 <header
   class="container-fluid text-center page-header bg-gray"
   {% if page.get_banner != None %}style="background-image: url({% image page.get_banner max-1920x1280 as img %}{{ img.url }})"{% endif %}
@@ -36,10 +38,9 @@
 {% block guts %}
 <div class="project-list-section">
   <div class="container">
-    <div class="row px-3 px-sm-0">
+    <div class="row px-0 px-sm-0">
       <div class="col-12">
         <div id="sticky-bar" class="creepiness-slider bg-white text-center">
-
           <!-- temporary location for dev work -->
 
           <span class="creep-o-meter-information">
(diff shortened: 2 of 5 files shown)


sabrinang (changes requested):
- add a bit more space above creep-o-meter introduction (~32px)
image


- From the mockups, it looks like the heading of the main category doesn't change and is clickable (dark-blue on hover) to reset the category (like as if you clicked the category from the navigation bar). See [Figma mockup](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/?node-id=548%3A72128) 
image


- Some subcategory pages don't show the state where there isn't a product:
image
- Some others have it:
image

network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
```diff
@@ -36,10 +38,10 @@ <h1 class="h1-heading text-center mt-0">{{ page.header }}</h1>
 {% block guts %}
 <div class="project-list-section">
   <div class="container">
-    <div class="row px-3 px-sm-0">
+    <div class="row px-0 px-sm-0">
       <div class="col-12">
         <div id="sticky-bar" class="creepiness-slider bg-white text-center">
-
+
```
Pomax, line 44: rogue whitespace

network-api/networkapi/wagtailpages/templates/buyersguide/bg_base.html
```diff
@@ -94,9 +94,11 @@
 
           {% for cat in categories %}
             {% with original=cat.original %}
-            {% if original.published_product_page_count > 0 %}
-              {% localizedroutablepageurl home_page 'category-view' original.slug as cat_url %}
-              <a class="multipage-link {% check_active_category current_category cat %}{% if original.featured is True %} featured{% endif %}" href="{{ cat_url }}" data-name="{{ cat.name }}">{{ cat.name }}</a>
+            {% if cat.parent == None %}
+              {% if original.published_product_page_count > 0 %}
+                {% localizedroutablepageurl home_page 'category-view' original.slug as cat_url %}
+                <a class="multipage-link {% check_active_category current_category cat %}{% if original.featured is True %} featured{% endif %}" href="{{ cat_url }}" data-name="{{ cat.name }}">{{ cat.name }}</a>
```
Pomax, line 100: we'll want to extend this tag so that it returns `active` for "this category matches, _or_ this category's `.parent` matches". That way, the nav will be correctly preselected already even if the JS doesn't kick in immediately

Pomax:
```python
# Determine if a category nav link should be marked active
@register.simple_tag(name='check_active_category')
def check_active_category(current_category, target_category):
    # because we're working with potentially localized data,
    # make sure to compare the linguistic originals.
    current_category = getattr(current_category, 'original', current_category)
    target_category = getattr(target_category, 'original', target_category)
    match = current_category == target_category or current_category.parent == target_category
    return 'active' if match else ''
```

network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
```diff
@@ -52,12 +54,39 @@ <h1 class="h1-heading text-center mt-0">{{ page.header }}</h1>
...
+    </div>
+    <div class="row px-0 tw-mb-4">
+      <div class="tw-w-full">
+        <div class="tw-flex tw-space-x-2">
+          {% for cat in categories %}
+            {% with original=cat.original %}
+            {% if cat.parent != None %}
+              {% if original.published_product_page_count > 0 %}
+                {% localizedroutablepageurl home_page 'category-view' original.slug as cat_url %}
+                <a class="{% if current_category.name != cat.parent.name and current_category.parent.name != cat.parent.name %} tw-hidden {% endif %} subcategories tw-no-underline tw-text-gray-60 border tw-border-gray-20 tw-bg-white tw-px-2 tw-py-1 tw-font-sans tw-rounded-3xl tw-font-normal tw-text-[12px] tw-leading-[1.3] hover:tw-bg-pni-lilac hover:tw-border-pni-lilac "
```
Pomax, line 80: let's capture this giant tailwind string as a var =)
```
            {% with original=cat.original tailwind_css="tw-no-underline tw-text-gray-60 border tw-border-gray-20 tw-bg-white tw-px-2 tw-py-1 tw-font-sans tw-rounded-3xl tw-font-normal tw-text-[12px] tw-leading-[1.3] hover:tw-bg-pni-lilac hover:tw-border-pni-lilac" %}
            ...
                <a class="{% if current_category.name != cat.parent.name and current_category.parent.name != cat.parent.name %} tw-hidden {% endif %} subcategories {{ tailwind_css }}"
```

Pomax: We should also be able to determine whether to "preselect" the subcategory here: if `current_category.localized.name == cat.localized.name` then we can use the "selected" tailwind classes instead.

Pomax (dismissed):
A few notes - also it looks like seaching does not change the title on the page to "All", even though it does highlight the "all" entry in the nav bar:

> image

sabrinang (dismissed):
Thanks for making the other changes! I found one more thing when I click on the other categories and click back to all ... the heading doesn't change back to all. For example when I visited pets and clicked all (should show All heading):
image

fessehaye:
That is because the "ALL" link still points to mofo-staging.

Tagging @Pomax  just to let him know but the all link does work properly on my dev env

network-api/networkapi/wagtailpages/templatetags/bg_nav_tags.py
```diff
@@ -11,7 +11,8 @@ def check_active_category(current_category, target_category):
     # make sure to compare the linguistic originals.
     current_category = getattr(current_category, 'original', current_category)
     target_category = getattr(target_category, 'original', target_category)
-    return 'active' if current_category == target_category else ''
+    match = current_category == target_category or current_category.parent == target_category
+    return 'active' if match else ''
```
Pomax, line 15: Turns out this could break on `None` categories, so let's make this just a touch more robust:
```suggestion
    match = current_category == target_category
    if hasattr(current_category, 'parent'):
        match = match or current_category.parent == target_category
    return 'active' if match else ''
```

Pomax (changes requested):
one code issue, but hitting "delete" in the search does not restore the category the user was on before they started searching. Instead, we stay on "all", but with the products not sorted on creepiness even though there is no search term,

Pomax:
@sabrina when a user starts to search, and then deletes their search again (either though delete/backspace, or the (X) button) should we take the user back to the category they were on before they started searching, or do we keep the user on "all"?

Pomax (changes requested):
(for some reason the review had been marked as "comment" instead of "changes requested"...)

sabrinang:
@Pomax I think it should just stay in "all" if they remove/start searching instead of going back to a previous catergory

sabrinang (approved):
Approving, Simon explained that "All" does have the right behaviour but isn't showing on staging

Pomax (approved):

Issue #7690, opened 2021-10-25
Verify callpower CTA works with a real campaign


Pomax:
We've added callpower CTA capabilities to the site, but we still need to test this with a real campaign that can actually make calls, to verify the workflow works correctly.

- [x] calling should switch to the "a call is about to happen" 
- [x] failures should trigger the correct views
  - [x]  looks like the "you called too often" error is not triggering, instead saying "something went wrong", error is `{ "description": "2 per 1 hour", "error": "Too Many Requests", "status": 429 }`
- [x] make sure the social fields exist for the callpower CTA (like petition) for putting in shareprogress identifiers

Coordinate with Kaili on making this happen.

PR #7731, opened 2021-11-01
properly hook up the callpower CTA social share buttons

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7690 

Testing:

- create a "callpower" snippe with the following SP values, and a campaign id that you ask @pomax about in slack:

> image

- create a new campaign and set its CTA to this new snippet (it'll be the last one on the last page for the cta snippet chooser).

- view the campaign page and verify that it shows the CTA with phone number and zipcode fields, and that filling those in and calling leads to:
  - a call to your phone
  - a "thank you" view change with social buttons, which
    - generate prepulated facebook/twitter/email shares
    - copy the page link to your clipboard, for the "copy" button.


network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -87,6 +105,10 @@ class Callpower(TranslatableMixin, CTA):
         TranslatableField('call_button_label'),
         TranslatableField('success_heading'),
         TranslatableField('success_text'),
+        # Shareprogress fields
+        SynchronizedField('share_twitter'),
+        SynchronizedField('share_facebook'),
+        SynchronizedField('share_email'),
```
Pomax, line 111: @TheoChevalier should these be TranslatableField, since each shareprogress id is for one block of specific content?

TheoChevalier: They should all be kept as Synchronized, as in we want to be able to override this in the admin without exposing the value to Pontoon/.po files at all. Translators — volunteers or from agencies — can’t deal with this and guess the right shareprogress ID to use here

danielfmiranda (approved):
Hi @Pomax! 

Thanks for the easy to follow instructions, I have followed them and can confirm that I recieved a phone call.

I also entered a bad phone number and can confirm that the error messages appear as expected as well.

Approving!

Issue #7736, opened 2021-11-02
Update the robots.txt route to generate "don't index anything" for the staging domain


Pomax:
https://foundation.mofostaging.net/robots.txt current says this:
```
User-Agent: *
Disallow: /*/artifacts/thimble
Disallow: /artifacts/thimble
```
When instead it should be saying this:
```
User-Agent: *
Disallow: /
```

PR #7738, opened 2021-11-02
update the route for robots.txt

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/7736

This makes it so that running on anything except prod yields a locked down `robots.txt` response.

Testing: 
- check out this branch
- make sure your `.env` file has `ASSET_DOMAIN=` (without a value)
- `docker-compose up`
- visit http://localhost:8000/robots.txt

This should yield:
```
User-Agent: *
Disallow: /
```

- `docker-compose down`
- make sure your `.env` file has `ASSET_DOMAIN=foundation.mozilla.org`
- `docker-compose up`
- visit http://localhost:8000/robots.txt

This should yield:
```
User-Agent: *
Disallow: /*/artifacts/thimble
Disallow: /artifacts/thimble
```


danielfmiranda (commented):
Hi @Pomax, 

I am having trouble on testing this ticket correctly. 

I have pulled the branch on my local machine and have followed the instructions as well as updated the environment variable. However when I visit the URL mentioned in the ticket, the page times out every time instead of returning what it should.

If you would like I can share my screen and show you what I am doing for testing

danielfmiranda (approved):
Hi Pomax! Just followed the instructions and it is working well! 

Here is without anything set:
image

And here are the results after updating .env
image


Approving!

PR #7767, opened 2021-11-03
move 'manage block_inventory' to a cron job

### LINKED ISSUE #7760: releases take *forever*, let's figure out why
I have no idea what heroku's doing, but releases seem to take _forever_ to finish, so let's figure out what's causing that. Making `release-steps.sh` show output would probably be a solid first step.

Pomax:
Closes #7760 by removing the `block_inventory` step from the release script.


tomusher (approved):
Looks good to me!

PR #7777, opened 2021-11-04
[WIP] Translatable ctas

Pomax:
Do not merge


TheoChevalier (commented):
I tested with the prod DB, it’s indeed fixing Banner pages, but I’m getting the following if I sync a page that has a petition snippet :/

Capture d’écran 2021-11-05 à 18 38 07

I guess we will need help from Karl

zerolab:
@TheoChevalier after many rounds of using the foundation build (start off main > `inv new-env`, switch to this branch and run migrations; start off this branch > `inv new-env`) and even a more barebones setup using bakerydemo, still can't reproduce the integrity error.

I am hitting a different issue (`Cannot assign "<CTA: Signup snippet>": "BanneredCampaignPage.signup" must be a "Signup" instance`) which relates to the inheritance model.

Specifically: wagtail-localize looks at the various fields and tries qualifies them. In this case, this is a related object so
https://github.com/wagtail/wagtail-localize/blob/7b4788a31bbf75e1a950fccc58c31eb7e80d0fa0/wagtail_localize/segments/types.py#L280-L283 calls https://github.com/wagtail/wagtail/blob/main/wagtail/core/models/i18n.py#L225 to initialize a value from the instance. However it get the wrong one

Before this PR:
```python
In [1]: from bakerydemo.localize_experiments.models import *
In [2]: s = Signup.objects.last()
In [3]: s.get_translation_model()
Out[3]: bakerydemo.localize_experiments.models.Signup
```

After the PR and migration
```python
In [1]: from bakerydemo.localize_experiments.models import *
In [2]: s = Signup.objects.last()
In [3]: s.get_translation_model()
Out[3]: bakerydemo.localize_experiments.models.CTA
```

note: I have a copy of the `CTA` and `Signup` snippets and use them on a simple page

I'd suggest moving to something like

```python
class TranslatableCTAMixin(TranslatableMixin, models.Model):
    name = models.CharField(
        default='',
        max_length=100,
        help_text='Identify this component for other editors',
    )

    header = models.CharField(
        max_length=500,
        help_text='Heading that will display on page for this component',
        blank=True
    )

    description = RichTextField(
        help_text='Body (richtext) of component',
        blank=True
    )

    translatable_fields = [
        TranslatableField('header'),
        TranslatableField('description'),
    ]

    def __str__(self):
        return self.name

    class Meta(TranslatableMixin.Meta):
        abstract = True


@register_snippet
class CTA(TranslatableCTAMixin):

    class Meta(TranslatableCTAMixin.Meta):
        verbose_name_plural = 'CTA'


@register_snippet
class Signup(TranslatableCTAMixin):
    campaign_id = models.CharField(
        max_length=20,
        help_text='Which campaign identifier should this petition be tied to?',
        null=True,
        blank=True,
    )

    class Meta(TranslatableCTAMixin.Meta):
        verbose_name = 'signup snippet'
```

to keep the locale fields on the concrete models.

This means the migration would change from saving the locale data back to the `CTA` to caching the regular fields from the CTA and saving them on the concrete snippet models (e.g. `Signup`)

Do give me a shout if you need a hand with the migration.

zerolab:
@TheoChevalier @Pomax 

ok, finally dug down to crux of the issue here. As this PR moves the `locale` and `translation_key` fields up to the `CTA` class the content_type is now that of the `CTA`.

Any existing pages that used any of the CTA types and were sent for translation will have corresponding `TranslatableObject` and `TranslationSource` objects that reference the content type (`TranslatableObject.content_type` and `TranslationSource.specific_content_type`, accordingly), so at the very minimum the references would need to be updated.

However, I would still advise migrating to using the `TranslatableCTAMixin` approach as the CTA snippet chooser does show *all* CTA instances (ie. those for CTA and the derived classes, such as Signup) which at a minimum adds loads of confusion when using it. Not to mention the `Cannot assign "<CTA: Signup snippet>": "BanneredCampaignPage.signup" must be a "Signup" instance` I encountered working with fresh content

TheoChevalier:
@Pomax Based on Dan’s findings above do you feel this could get fixed & landed before Wed 9th? We have a localized petition launching that day.

If we’re not confident this can be fixed on time, I’ll get a contingency plan lined up, which should consist in [commenting out this line](https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/campaigns.py#L316), so that we will at least have the page translated.

tbrlpld:
@zerolab shout 😅 

What's your advise on how to tackle this? I have created an abstract base/mixin CTA. But, when I try to create the migrations I do get errors because of missing non-nullable fields (`id`) in this case. I guess this happens because with the multi-table inheritance the id is only defined on the parent. So I guess we would need a data migration to set the id on each subclass instance. 

Are there other obvious fields that we need to move from the parent to the subclass?  What's the best order to do these things? Do I need temporary fields that move the information from the parent to the subclass and only adds and populates the actual field in a second step?

PR #7836, opened 2021-11-11
Feature/7599 locale fixtures

### LINKED ISSUE #7559: Add a second locale to both the Django unit test suite, and our load_fake_data factory code
Follow-up to https://github.com/mozilla/foundation.mozilla.org/issues/7502

We need to add a secondary locale as part of the test framework before can write tests that can verify that changes are correct across locales. (For PNI specifically, we also want to know that votes on localized objects make it to the original object, etc).

richbrennan:
Closes #7559

- Adds a new Locale to the initial data that's created when build is set up or the database refreshed.
- Adds a new Locale to PNI localisation test and updates test to pass

This does not include adding translated data or updating all PNI tests.

@Pomax 
Is there anything else other than the above that needs to fall within the scope of this ticket?

**Question**
I'm not sure how to run the `sync_locale_trees` command when creating a review app. I tried adding it to `app.json` but it doesn't seem to work: https://github.com/mozilla/foundation.mozilla.org/pull/7836/files#diff-96677aa35debfefd031d9d34d9c70369754ee3acb2d9a9d4090e98612efee6f5R48


network-api/networkapi/wagtailpages/tests.py
```diff
@@ -43,35 +44,42 @@ def setUp(self):
...
+            '/en/privacynotincluded/',
+            'Homepage should be forwarded to /en/ by default'
+        )
 
-    @skip("TODO: REENABLE: THIS HAS BEEN TESTED MANUALLY BUT FAILS IN THIS CODE FORM ATM")
     def test_localised_homepage(self):
         """
         Test that the homepage redirects properly under different locale configurations.
         """
-        response = self.client.get('/privacynotincluded/')
```
richbrennan, line 65: Removed as its a duplicate of `test_homepage`

network-api/networkapi/wagtailpages/tests.py
```diff
@@ -43,35 +43,42 @@ def setUp(self):
...
-        self.assertEqual(response.status_code, 302, 'simple locale gets redirected')
-
         response = self.client.get('/privacynotincluded', follow=True, HTTP_ACCEPT_LANGUAGE='fr')
         self.assertEqual(
             response.redirect_chain[0][0],
             '/fr/privacynotincluded/',
             'redirects according to HTTP_ACCEPT_LANGUAGE'
         )
-        self.assertEqual(response.status_code, 302)
+        self.assertEqual(response.status_code, 200)
```
Pomax, line 81: The point of this test was to make sure that "nonexistent locales" still sends you to a page, so if french now works, we'll need to retain what this test did for a different locale in addition to making sure French passes =)

tasks.py
```diff
@@ -164,6 +166,8 @@ def setup(ctx):
         migrate(ctx)
         print("* Creating fake data.")
         manage(ctx, "load_fake_data")
+        print("* Sync locales")
```
Pomax, line 169: hm, this is shared between new-db and new-env, if we're adding more shared functionality, let's make this its own `def` to capture the code starting at "applying database migrations" and then have both invoke tasks call that.

Pomax (changes requested):
Nice! just a few small things but this looks pretty great!

network-api/networkapi/wagtailpages/tests.py
richbrennan: Done :+1:

richbrennan: I've copied this test for a valid language that doesn't have a locale, but the `status_code` is still `200` when following the redirect chain, as it ends up at `'/sw/privacynotincluded/'`, which shows the English version of the page.

richbrennan:
@Pomax 

I've addressed the points you raised.

The only outstanding thing now is this question:

I'm not sure how to run the sync_locale_trees command when creating a review app. I tried adding it to app.json but it doesn't seem to work: https://github.com/mozilla/foundation.mozilla.org/pull/7836/files#diff-96677aa35debfefd031d9d34d9c70369754ee3acb2d9a9d4090e98612efee6f5R48

network-api/networkapi/wagtailpages/tests.py
Pomax: Interesting, I wonder if the behaviour of `redirect_chain` changed across python/wagtail versions O_o

Pomax (approved):

Pomax:
I've filed https://github.com/mozilla/foundation.mozilla.org/issues/7935 to get the review apps to run `sync_locale_trees`

PR #7841, opened 2021-11-13
Fix filter for loop error

fessehaye:
Fix error that happens when trying to filter products.

1. Go to /privacynotincluded
2. Type text in search bar
3. Double check that `field not defined` error is gone


danielfmiranda (approved):
Hey Simon, just followed the instructions. The error is gone, approved!

PR #7881, opened 2021-11-17
7834 - Fix AI column Problems

### LINKED ISSUE #7834: [PNI QA] A.I. section question and answer display
The question `Is the company transparent about how the AI works?` doesn't display but the answer appears with `Does the user have control over the AI features?` while that answer does not appear.

Content on the spreadsheet I want to show on the website but doesn't:
Screen Shot 2021-11-10 at 1 49 53 PM
What I have entered in the CMS:
image

fessehaye:
Closes #7834

`Is the company transparent about how the AI works?` & `Does the user have control over the AI features?` should work correctly now


sabrinang (approved):
The question now appears with both having the correct corresponding answer — thanks for this fix Simon!

PR #7885, opened 2021-11-17
take CSRF out of the page template for PNI products

### LINKED ISSUE #7880: switch PNI product voting CSRF over to just-in-time fetching
- load product page without CSRF token so that it's the same content for all users
- make the voting JS fetch the token from a dedicated CSRF endpoint on the server.
- then add that into the vote POST instead of querying the DOM for it.

this will allow us to cache-excempt the CSRF endpoint, without PNI product pages breaking the caching.

Pomax:
Closes #7880 by moving the csrf token to its own API endpoint, that the PNI voting JS can fetch, parse, extract the text out of (I opted to keep it `{% csrf_token %}` rather than `{{csrf_token}}` just because the former makes it just that tiny bit more annoying to write a script that gets the value out, compared to just responding with an immediately-usable string) and then add into the voting POST request.

With this, PNI products should no longer need cache exemption, nor should they cause cloudflare to set a "this user wants uncached content" cookie.

Testing wise: voting should work =)

Hit up https://foundation-s-separate-v-zw6hz3.herokuapp.com/en/privacynotincluded, pick a product, and verify that the voting slider part renders, and that you can vote after moving it, with the vote count going up as you pile on the votes.

### DIFF
--- a/source/js/buyers-guide/components/creep-vote/creep-vote.jsx
+++ b/source/js/buyers-guide/components/creep-vote/creep-vote.jsx
@@ -44,15 +44,14 @@ class CreepVote extends Component {
     sessionStorage.setItem("subscribed", subscribed);
 
     return {
-      totalVotes,
       creepiness: 50,
-      hasMoved: false,
+      csrfToken: ``,
       didVote: false,
-      majority: {
-        creepiness: creepinessId,
-      },
-      subscribed,
+      hasMoved: false,
+      majority: { creepiness: creepinessId },
       showNewsletter: false,
+      subscribed,
+      totalVotes,
       voteCount,
     };
   }
@@ -61,6 +60,17 @@ class CreepVote extends Component {
     if (this.props.whenLoaded) {
       this.props.whenLoaded();
     }
+
+    // voting requires a CSRF token, which we can request on a dedicated API route
+    fetch(`/api/csrf/`)
+      .then((res) => res.text())
+      .then((html) => {
+        const d = document.createElement(`div`);
+        d.innerHTML = html
+        this.setState({
+          csrfToken: d.querySelector(`input`).value,
+        });
+      });
   }
 
   showVoteResult() {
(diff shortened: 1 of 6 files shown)


source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -44,15 +44,14 @@ class CreepVote extends Component {
...
-      totalVotes,
       creepiness: 50,
-      hasMoved: false,
+      csrfToken: ``,
       didVote: false,
-      majority: {
-        creepiness: creepinessId,
-      },
-      subscribed,
+      hasMoved: false,
```
Pomax, line 50: sorted the keys alphabetically.

source/js/buyers-guide/components/creep-vote/creep-vote.jsx
```diff
@@ -198,7 +208,6 @@ class CreepVote extends Component {
           flowText={getText(
             `Now that you’re on a roll, why not join Mozilla? We’re not creepy (we promise). We actually fight back against creepy. And we need more people like you.`
           )}
-          csrfToken={this.props.joinUsCSRF}
```
Pomax, line 201: this.. is nothing, and was nothing. the JoinUs component doesn't use a CSRT token.

source/js/buyers-guide/inject-react/creep-vote.js
```diff
@@ -35,7 +32,7 @@ export default (apps, siteUrl) => {
     }
 
     apps.push(
-      new Promise((resolve) => {
+      new Promise(async (resolve) => {
```
Pomax, line 35:
```suggestion
      new Promise((resolve) => {
```

source/js/components/join/join.jsx
```diff
@@ -259,8 +259,8 @@ class JoinUs extends Component {
    */
   renderFlowHeading() {
     return [
-      <h2 className="h3-heading text-center">{this.props.flowHeading}</h2>,
-      <p className="text-center">{this.props.flowText}</p>,
+      <h2 className="h3-heading text-center" key={`flowheading`}>{this.props.flowHeading}</h2>,
+      <p className="text-center" key={`flowheadingpara`}>{this.props.flowText}</p>,
```
Pomax, line 263: react-dev was throwing a "please use keys!" warning, so... might as well add those now.

fessehaye (approved):

Issue #7796, opened 2021-11-05
Implement email sending for reminder button


danielfmiranda:
Depends on #7795 
As mentioned in the previous ticket, if the user is not viewing on a browser that allows extension, we want to render a button that says "Send Reminder":
<img width="920" alt="image" src="https://user-images.githubusercontent.com/18314510/140569586-0f4d1201-3386-4557-98e6-0aeb39c45edd.png">

We need to: 

- [x] Reach out to design to get a copy of what the process should look like (user clicks -> form renders -> success message)
- [ ] Add the functionality for when the button is clicked on

benhohner:
Blocked:
- waiting on form design and confirmation design from designers
- waiting on copy
- email design

Ben's comments: I think this feature needs to be fleshed out more before we move ahead, we may need to delay until designers are back

kristinashu:
Rebecca mentioned in Slack that the button should work like the email share button on our blog posts e.g. https://foundation.mozilla.org/en/blog/how-to-curb-election-disinformation-on-social-media/

Clicking on the button should
- open the person's email 
- pre-populate the email with [title] [body text] (I have asked the Brandi/Audrey for this)

I don't think it needs a form with custom styling or anything like that. I also don't think the button copy needs to be updated.

Does that make sense?

benhohner:
@danielfmiranda to further clarify from eng side, I was mistaken about the interaction. Just needs a `mailto:` link with the correct subject filled in, should be pretty straighforward and no additional screens or email sending.

kristinashu:
Brandi has shared this copy for the email:

### Subject
> Reminder to download RegretsReporter on your desktop

### Body
> Mozilla built RegretsReporter to help you take back control of your YouTube recommendations and make YouTube better for everyone else. 

> [Add to Firefox](https://addons.mozilla.org/addon/regretsreporter/) | [Add to Chrome](https://chrome.google.com/webstore/detail/regretsreporter/obpoeflheeknapimliioeoefbfaakefn)

> [Visit our website to learn more.](https://foundation.mozilla.org/youtube/regretsreporter)

benhohner:
Here are some docs on the mailto format @danielfmiranda https://css-tricks.com/snippets/html/mailto-links/
Not sure if we can include sub-links, but we can probably include URLs as text. e.g. `Add to Firefox: https://addons.mozilla.org/addon/regretsreporter/`

danielfmiranda:
[...]

PR #7899, opened 2021-11-19
7795 - Added Email Button and Browser sensing

### LINKED ISSUE #7795: Add the browser-sensing buttons to the top half of the page
Depends on #7793 
In the top half of the page, there can be two or one buttons found depending on the window width.

However, we want to try and find a way to also make it so the page knows whether the user is viewing on a desktop or mobile browser. 

If viewing on desktop browser, have two buttons render which link to extension download page for firefox and chrome. 

If viewing on mobile browser, have one button render which would allow them to send an email reminder to download the extension. 

@Pomax mentioned the ability to be able to view what browser the user is using. It might be beneficial to reach out to him early this sprint to gather more details on how this can be achieved.

Figma design can be found here: https://www.figma.com/file/A0TYlvLzKxnQLZr3unohQy/YouTube-RegretsReporter---product-page-test?node-id=232%3A1216
To see how this page should format in desktop/tablet/mobile breakpoints.

fessehaye:
Closes #7795 
Related PRs/issues #7796 

* Added Send Reminder Button
* Added JS to check if user is on mobile device and show correct button

Link: 
https://foundation-s-7795-email-mpjivy.herokuapp.com/en/youtube/regretsreporter/

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/pages/regrets-reporter-landing-page/extension_hero.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/pages/regrets-reporter-landing-page/extension_hero.html
@@ -21,6 +21,11 @@
                         target="_blank">
                         {% trans "Add to Chrome" %}
                     </a>
+                    <a class="btn download-link-button download-link-button--email tw-bg-white tw-text-youtube-regrets-red tw-hidden"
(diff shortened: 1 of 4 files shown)


danielfmiranda (commented):
Hey Simon! 

Thanks for taking this on, from what I see on my Iphone and also on my computers, this seems to be working great!

One thing I wanted to bring up however is that at least on xcode simulated devices, it does not seem to be catching Ipads. However, I know this may be a simulator issue and not a device issue.

One more question I had about this PR is that the email doesnt exactly follow the copy in terms of the links. I was wondering, is there any way to have the links say "Add to Firefox" and "Add to chrome"? 

I tried to look this up myself however, and it does seem a little tricky. So if @beccaklam is OK with the email content, I will approve this PR!

Thanks again!

beccaklam (approved):
This is working on my iPhone and my iPad!
Re: Daniel's comment about the email content, we can tackle that in another ticket if it does become an issue? I'm good to approve this for now though!

fessehaye:
I think if we want a more complicated email templates that probably goes past the support of using mailto links and we would have to look into other larger solutions

PR #7900, opened 2021-11-20
7787 - drag to scroll on desktop

### LINKED ISSUE #7787: [PNI QA] Subcategories — drag to scroll on desktop
For people who aren't using a track pad or wheeled mouse, it would be good to be able to drag-to-scroll to see the additional subcategories
sTpcOsTW8K

fessehaye:
Closes #7787 

Test URL: https://foundation-s-7787-drag-vpxuans.herokuapp.com/en/privacynotincluded/categories/smart-home/


source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,54 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+        x: event.clientX,
+      };
+
+      document.addEventListener("mousemove", mouseMoveHandler);
+      document.addEventListener("mouseup", mouseUpHandler);
+    };
+
+    const mouseMoveHandler = function (event) {
+      event.preventDefault();
+      Array.from(subcategories).forEach((subcategory) => {
```
Pomax, line 42: the result of `querySelectorAll` has .forEach defined (just not `.map` or any of the other array functions)

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,54 @@ const categoryTitle = document.querySelector(`.category-title`);
...
 // TODO: turn this into a static class rather than plain JS object.
 const SearchFilter = {
   init: () => {
+    let pos = { top: 0, left: 0, x: 0, y: 0 };
+
+    const mouseDownHandler = function (event) {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+      subContainer.style.cursor = "grabbing";
+      subContainer.style.userSelect = "none";
```
Pomax, line 29: it'd be nicer if this was a class, rather than inline-style-by-proxy

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,54 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+
+      pos = {
+        left: subContainer.scrollLeft,
+        x: event.clientX,
+      };
+
+      document.addEventListener("mousemove", mouseMoveHandler);
+      document.addEventListener("mouseup", mouseUpHandler);
```
Pomax, line 37: this is missing the equivalent touch events (lets of touch-capable laptops on the market these days)

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,54 @@ const categoryTitle = document.querySelector(`.category-title`);
...
 const toggle = document.querySelector(`#product-filter-pni-toggle`);
 const subcategories = document.querySelectorAll(`.subcategories`);
+const subContainer = document.querySelector(`.subcategory-header`);
 
 // TODO: turn this into a static class rather than plain JS object.
 const SearchFilter = {
   init: () => {
+    let pos = { top: 0, left: 0, x: 0, y: 0 };
+
+    const mouseDownHandler = function (event) {
```
Pomax, line 25: rather than `mouse down`, let's name this based on its functionality (the fact that we tie it to a mousedown is basically an irrelevant detail for the function's code =).

E.g `markScrollStart` or something?

Also, `const` makes sense for arrow functions, but is unnecessary for regular functions, which already get locally scoped by name. So if this function doesn't technically rely on any scoped vars: let's move it out `init`, it'll just be a plain function. If it does, let's make it an arrow function.

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,54 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+      pos = {
+        left: subContainer.scrollLeft,
+        x: event.clientX,
+      };
+
+      document.addEventListener("mousemove", mouseMoveHandler);
+      document.addEventListener("mouseup", mouseUpHandler);
+    };
+
+    const mouseMoveHandler = function (event) {
```
Pomax, line 40: same naming note as before. Let's call this and all other event handlers based on what they do, not the event they're tied to.

Pomax (changes requested):

sabrinang (approved):
The functionality to drag to scroll across subcategories are working for me across browsers and viewports, thanks Simon!

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,63 @@ const categoryTitle = document.querySelector(`.category-title`);
...
 
 // TODO: turn this into a static class rather than plain JS object.
 const SearchFilter = {
   init: () => {
+    let pos = { left: 0, x: 0 };
+
+    const markScrollStart = (event) => {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+      subContainer.classList.add("cursor-grabbing");
```
Pomax, line 28: let's cache this classList at the start of init, such as `const subClasses = subContainer.classList`, so we don't need `subContainer.classList` everywhere. Also, `classList.add()` is an infinite-argument function, so you only need a single `.add()` with as many classes as you need added.

Pomax, line 28: let's cache this classList at the start of init, such as const subClasses = subContainer.classList, so we don't need subContainer.classList everywhere. Also, classList.add() is an infinite-argument function, so you only need a single .add() with as many classes as you need added.

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,63 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+      pos = {
+        left: subContainer.scrollLeft,
+        x: event.clientX,
+      };
+
+      document.addEventListener("mousemove", markScrollMove);
+      document.addEventListener("mouseup", markScrollEnd);
+      document.addEventListener("touchmove", markScrollMove);
+      document.addEventListener("touchend", markScrollEnd);
+      document.addEventListener("touchcancel", markScrollEnd);
```
Pomax, line 40: let's compact this into only two `addListener` calls, but for lists of event names. That way, there's fewer places the event binding can go wrong, and updating the events is just a matter of adding/removing event names:

```suggestion
      [`mousemove`, `touchmove`].forEach(type =>
        document.addEventListener(type, markScrollMove)
      );
            
      [`mouseup`, `touchend`, `touchcancel`].forEach(type =>
        document.addEventListener(type, markScrollEnd)
      );
```

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,63 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+      });
+      const dx = event.clientX - pos.x;
+      subContainer.scrollLeft = pos.left - dx;
+    };
+
+    const markScrollEnd = (event) => {
+      event.preventDefault();
+      event.stopImmediatePropagation();
+      subContainer.classList.remove("cursor-grabbing");
+      subContainer.classList.remove("select-none");
```
Pomax, line 56: remove is the same "inifnite argument" type function as `add`

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,63 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+
+      subcategories.forEach((subcategory) => {
+        subcategory.classList.remove("pointer-events-none");
+      });
+
+      document.removeEventListener("mousemove", markScrollMove);
+      document.removeEventListener("mouseup", markScrollEnd);
+      document.removeEventListener("touchmove", markScrollMove);
+      document.removeEventListener("touchend", markScrollEnd);
+      document.removeEventListener("touchcancel", markScrollEnd);
```
Pomax, line 66: let's compact this the same as above

source/js/buyers-guide/search.js
```diff
@@ -15,10 +15,63 @@ const categoryTitle = document.querySelector(`.category-title`);
...
+    };
+
+    subContainer.addEventListener("mousedown", markScrollStart);
+    subContainer.addEventListener("mouseleave", markScrollEnd);
+    subContainer.addEventListener("touchstart", markScrollStart);
+    subContainer.addEventListener("touchend", markScrollEnd);
+    subContainer.addEventListener("touchcancel", markScrollEnd);
```
Pomax, line 73: let's compact these, too.

Pomax (dismissed):

Pomax:
When testing on https://foundation-s-7787-drag-vpxuans.herokuapp.com/en/privacynotincluded/categories/smart-thermostats-smoke-detectors/ It looks like this works once (I can click and drag to move the subcats around) but after a mouse release/mouse up it stops working.

Pomax:
it seems to not always work though, sometimes on a page load nothing seems to happen (other than text-selecting)

Pomax (changes requested):
marking as "request changes" for now.

Pomax:
It looks like the highlighted subcategory doesn't quite make it into view all the time, but if @sabrinang's okay with it we can land this first, then fix that in a follow-up issue.

> 
image

Pomax:
@nancyt1 could you have a look and see if more work is required (although not necessarily as part of this PR itself)?

nancyt1 (approved):
Drag to scroll works for me across viewports and selected subcategories snapped into viewport when clicked for me. I think it looks good! 

(Not sure if it's just the review app but drag to scroll doesn't work after I refresh the page, only when I first land on it)

Pomax (approved):

Issue #7828, opened 2021-11-10
See if we move some of the metadata overrides/exceptions/etcs into mixin classes


Pomax:
we currently have the `FoundationMetadataPageMixin` that sets the relevant metadata for pages, but also have custom meta tag code in place as overrides on specific pages - it would be good to see if we can make those overrides "additional mixins" that can be used to override the behaviour of `FoundationMetadataPageMixin` (or implement overrides that it does not cover) so that the wagtail-metadata library generates the data we want to see used, without us having to inject our own `<meta>` code.

Pomax:
invalidating due to ongoing SEO work

PR #7914, opened 2021-11-22
[7778] make rss description use page get_meta_description()

### LINKED ISSUE #7778: RSS feed description should output meta description instead of encoded html
Currently the RSS feed outputs encoded page html instead of the page's meta description.
The RSS feed description fields should be the corresponding article's meta description.

See: https://foundation.mozilla.org/en/blog/rss/

The meta description behaviour should follow the existing meta description behaviour (if Meta Description is filled in on the promote tab, display that, otherwise display the first 153 characters of the article with an ellipsis at the end)

richbrennan:
Closes #7778

**Link to sample test page**: https://foundation-s-fix-7778-r-pg8gel.herokuapp.com/en/blog/rss/
The last item is using the first 153 characters of the page body text: https://foundation-s-fix-7778-r-pg8gel.herokuapp.com/cms/pages/14/edit/

This follows the same pattern as the meta description for the [product page](https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/products.py#L781). The plan is to [audit and refactor](https://github.com/mozilla/foundation.mozilla.org/issues/7828) the code for meta fields so this introduces a small amount of technical debt, however this highlights a similarity in the pattern in the meta description logic.


Pomax (approved):
Sweet. And I didn't know about `get_text_for_indexing`, that's super good to know about!

PR #7927, opened 2021-11-24
7918 - Add youtube regrets page extension to parent

fessehaye:
Add youtube-regrets-extension page type to campaigns

For testing copying the staging db would help!

1.  find youtube regrets banner page on the CMS
2. Click `Add child page`
3.  Should be able to add extension page


danielfmiranda (approved):
Hey Simon, this looks good to me! We can then add a YouTube extension page under /YouTube, and thanks to the Max limit we don't have to worry about users creating a duplicate.

If we did want to make sure to avoid confusion, we can also remove it from this list once we create the page in the DB!

PR #7832, opened 2021-11-10
Yt extension page base

danielfmiranda:
Closes #7792 
Related PRs/issues #

Using https://github.com/mozilla/foundation.mozilla.org/issues/6891 for reference, I have created a new page in the CMS which is the youtube regrets reporter extension page. 

The page can now be found in the CMS under the "homepage -> Youtube" section.

image

**Link to sample test page**: https://foundation-s-yt-extensi-srkoua.herokuapp.com/en/youtube/regretsreporter/

**Steps to test**: 

1. Please visit the test URL above, it should take you to a page with a nav bar and footer, but the content will have a placeholder for now.
2. If the URL looks correct, in terms of where we want the page to live, testing is complete!

PR #7936, opened 2021-11-25
Localize YoutubeRegretsReporterExtensionPage

TheoChevalier:
Followup from #7832

We need to override `translatable_fields` in order to make sure the slug is not translated


danielfmiranda (approved):
Looks good to me Theo! Approved

PR #7961, opened 2021-11-30
removed blue outline from creepometer when active on safari

### LINKED ISSUE #7874: [PNI] Creep-o-meter shows blue outline in Safari (small bug)
### Describe the bug
Blue select outline when I click on the creep-o-meter in Safari.

### To Reproduce
1. Go to any product page https://foundation.mozilla.org/en/privacynotincluded/netatmo-smart-video-doorbell/ 
2. Click on the creep-o-meter (click, don't slide)
3. See blue outline

### Expected behavior
No blue outline.

### Screenshots
Screen Shot 2021-11-15 at 3 52 45 PM

### Desktop (please complete the following information):
- macOS Monterey Version 12.0.1
- Safari Version 15.1 (17612.2.9.1.20)

danielfmiranda:
Closes #7874
Related PRs/issues #

Link to test page: https://foundation-s-creepomete-virg9i.mofostaging.net/en/privacynotincluded/real-body-piece-better-notice/

**Steps to test**:

1. Please visit the test link above using safari/chrome
2. Scroll down to the creepometer and click on the face/draggable bar
3. Note that there is no longer a blue outline surrounding the area like before
4. If everything looks as expected, testing is complete!


danielfmiranda:
Hi Everyone! Wanted to leave a note of something I found regarding accessibility as I know that was a concern for this ticket.

It appears that since we set up the creepometer to only allow the user to submit a vote after they tap or touch it, navigating it using a keyboard does not permit the user to submit a vote anyways.

fessehaye (approved):

kristinashu:
Thank you for this @danielfmiranda!

Could you please elaborate on 
> It appears that since we set up the creepometer to only allow the user to submit a vote after they tap or touch it, navigating it using a keyboard does not permit the user to submit a vote anyways. 

Do you think we should open a new ticket to make creepometer voting doable by tabbing? Or do you mean that by removing the blue outline we are making it so the creepometer isn't tabbable? I'm just wondering if this PR is making things worse for accessibility or if it's a separate issue? 

It seems like on production, I can tab and use the arrows to move the creepometer but I cannot submit it. While in this PR, I can no longer tab into the creepometer to move it or submit it. Is that right? Sorry, I am not super familiar with tabbing.

danielfmiranda:
Hi @kristinashu! No problem, the creepometer in this PR is still able to be selected by tab and controlled by the arrow keys, we just removed the visual cue to let the user know that the creepometer is selected. 

I believe that is making accesibility worse, however, users who were using the keyboard to navigate the page were not able to make a vote anyways since the creepometer is waiting to be tapped or clicked to un-disable the "submit" button. 

I think there are multiple solutions to this issue: 

1.  We can leave the blue outline invisible, and leave the submit button as is.
2. We can restore the blue outline, and let the submit button be un-disabled when the user moves the creepometer, VS tapped/touched
3. We can leave the blue outline invisible, and let the submit button be un-disabled when the user moves the creepometer, VS tapped/touched

Any of these shouldnt be a significant pull, as we are just changing the event that the creepometer is waiting for to enable the submit button. 

Thanks!

kristinashu:
Thank you for the notes @danielfmiranda! Option 3 sounds good, right? My preference is to keep the blue outline invisible and if you can make the creepometer move and the submit button work by tabbing/touch then that would be amazing.

danielfmiranda:
Sounds great @kristinashu! Will do

danielfmiranda:
Hi @kristinashu I have now updated the code to allow the user to vote after making a change with the keyboard.
Here are some steps to test if you would like:

1. Scroll down to show the creepometer
2. Click on whitespace in the page, can be anywhere next to the creepometer!
3. Press "tab" once
4. Navigate the creepometer left or right using arrow keys
5. Note that the button becomes enabled
6. Press Tab again and then enter
7. It should vote as expected!

If you have any questions please let me know, thank you!

kristinashu (approved):
It's working! It's very cool to tab be able to tab through everything!

Issue #7288, opened 2021-08-30
Version Upgrades


benhohner:
- [x] Fo'mo
- [x] Donate/Thunderbird

jamilasnell:
Hey @richbrennan! Would you be willing to handle the wagtail version upgrades? Would be greatly appreciated! :)

richbrennan:
- Update Wagtail to [2.15.1](https://docs.wagtail.io/en/stable/releases/2.15.html#upgrade-considerations)
- Dependabot PRs: Cypress has been basically impossible to use for local dev
- Make Cypress work for local dev

richbrennan:
Hi @Pomax 

The Python-related Dependabot PR's are failing in CI due to incompatible requirements of `importlib-metadata`. It seems as though `flake8` has it [pinned](https://github.com/PyCQA/flake8/blob/6e5600f8ec3f98fd052bd532470297e0a360750e/setup.cfg#L45) to the older `<4.3` for Python < 3.8, whereas other packages are using a newer version.

- https://github.com/mozilla/foundation.mozilla.org/pull/7930
- https://github.com/mozilla/foundation.mozilla.org/pull/7911
- https://github.com/mozilla/foundation.mozilla.org/pull/7909

Is updating to a newer version of Python a viable option?

richbrennan:
Cannot update to Wagtail 2.15.1 until `wagtail-localize-git` is updated: https://github.com/wagtail/wagtail-localize-git/blob/main/setup.py#L51

richbrennan:
@Pomax 

Also, would you be able to confirm what `Donate/Thunderbird` in the first comment refers to please?

Pomax:
"Donate/Thunderbird" refers to the https://github.com/mozilla/donate-wagtail repo and its dependencies.

Pomax:
I've filed https://github.com/wagtail/wagtail-localize-git/issues/16 to hopefully get wagtail-localize version bumped, so we can uplilft that and then get wagtail onto 2.15

Pomax:
@richbrennan it might be a good idea to see if we uplift our python versions to 3.10 for the foundation and donate repos. We don't have an issue filed for that atm, but if you want to tackle that feel free to create one.

richbrennan:
[...]

PR #7969, opened 2021-12-01
Feature/update python to 3.9.9

### LINKED ISSUE #7953: Update Python versions
The Python-related Dependabot PR's are failing in CI due to incompatible requirements of importlib-metadata. It seems as though flake8 has it pinned to the older <4.3 for Python < 3.8, whereas other packages are using a newer version.

https://github.com/mozilla/foundation.mozilla.org/pull/7930
https://github.com/mozilla/foundation.mozilla.org/pull/7911
https://github.com/mozilla/foundation.mozilla.org/pull/7909

We may not be able to go as high as Python 3.10 due to dependency limitations, so maybe aim for 3.9

richbrennan:
Closes #7953
Related PRs/issues #7288

This updates to Python 3.9.9. Ideally we'd be going up to 3.10 but there's still a dependency, `wagtail-airtable`, that [doesn't support ](https://github.com/wagtail/wagtail-airtable/blob/master/pyproject.toml)it. I can look into creating a PR so it does, but going to 3.9 should unblock some of the Dependabot PRs in the meantime.


Pomax:
We wouldn't be able to switch to 3.10 yet anyway, as the Heroku buildpack is only on 3.9.9 (as of 2 weeks ago)

Pomax (changes requested):
we'll also need to update the python requirement in our .`github/workflows` yml files.

Pomax (approved):
this works well as far as I can tell!

PR #7979, opened 2021-12-02
7919 - Add lazy loading to PNI images on catalog/product page

### LINKED ISSUE #7919: Add lazy loading to PNI images on catalog page
- [x] Add `loading="lazy"` to image/picture tags
- [x] ~~add JS to remove fixed width and heights after load to prevent layout shifts~~
- [x] ~~potentially use `intersectionobserver` to see how close to the viewport images are, and start loading them before they've scrolled into view~~

fessehaye:
Closes #7919

1. go to `/privacynotincluded`
2. Open dev tools and imitate mobile device size and disable cache on the network tab and filter for only images
3. Scroll to the top of the page and refresh
4. Should notice that some of the images are not loaded until scrolling further into the page

calculated distance from the viewport, as defined by the browser so it will load images that are outside the viewport


danielfmiranda (approved):
Hey Simon! Thanks for the easy to follow instructions, just tested it out and this is working great! Approved

Issue #7772, opened 2021-11-04
[PNI] subcategories - you can save new categories with the same name as old ones


Pomax:
This will cause a slug resolution conflict, so we should make sure that when people try to save a new category it checks to see if that name isn't already taken.

Pomax:
this got merged already

PR #7997, opened 2021-12-09
[7644] Add product category validation

### LINKED ISSUE #7644: [PNI] - Subcategories - harden the subcategory form/save logic
The initial subcategory code from #7641 allows for some things we may want to "lock down";

- categories can be nested infinitely deep
- categories can technically be linked up into a cycle

Let's harden that code a little so that at the very least that last option isn't possible anymore, and then possibly lock down the first option, too.

From @richbrennan: 
> In case it might be useful, I've been meaning to investigate this for nested categories: https://posts-by.lb.ee/building-a-configurable-taxonomy-in-wagtail-django-94ca1080fb28

richbrennan:
Closes #7644 and #7772

Adds validation to avoid:

- Categories with the same name
- Category being a descendent of itself
- Categories being nested too deep

**Link to sample test page**: https://foundation-s-feature-76-cx7ws7.herokuapp.com/cms/snippets/wagtailpages/buyersguideproductcategory/

Review site password: https://mozilla.slack.com/archives/CBKP2SG94/p1639070578008300

To test, go to the page above and try the following:

- Creating or editing a category to have the same name as an existing one
- Making a category a descendent of itself by setting it to be its own parent or a child of one if it's children
- Creating categories more than 2 levels deep

## Checklist

**Tests**


network-api/networkapi/wagtailpages/forms.py
```diff
@@ -0,0 +1,23 @@
+from wagtail.admin.forms import WagtailAdminModelForm
+
+
+class BuyersGuideProductCategoryForm(WagtailAdminModelForm):
+    def clean_name(self):
+        name = self.cleaned_data["name"]
+        if (
+            self.instance.__class__.objects.filter(name=name)
```
Pomax, line 8: let's just make this `name__iexact=name` for maximum cleanliness =D

network-api/networkapi/wagtailpages/pagemodels/products.py
```diff
@@ -202,6 +203,14 @@ def get_parent(self):
...
         return BuyersGuideProductCategory.objects.filter(parent=self)
 
+    def get_ancestors(self, inclusive=False):
+        ancestors = []
+        node = self if inclusive else self.parent
+        while node:
+            ancestors.insert(0, node)
+            node = node.parent
+        return ancestors
+
```
Pomax, line 213: It feels like we shouldn't need this: given the level of nesting we're allowing atm, if a parent has a parent, that's a signal that something's going wrong already.

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -0,0 +1,23 @@
...
+        return name
+
+    def clean_parent(self):
+        parent = self.cleaned_data["parent"]
+        if parent:
+            ancestors = parent.get_ancestors(inclusive=True)
+            if len(ancestors) > 2:
+                self.add_error("parent", "Categories can only be three levels deep.")
+            if self.instance in ancestors:
+                self.add_error("parent", "A category cannot be a decendent of itself.")
```
Pomax, line 22: We probably don't need an ancestor chain here: if a parent has a parent, this category's already in too deep.

Pomax (changes requested):

network-api/networkapi/wagtailpages/pagemodels/products.py
richbrennan: I think that's answered [this question](https://github.com/mozilla/foundation.mozilla.org/issues/7644#issuecomment-989003970) :+1:

Pomax: ah, missed that comment. Replied, but for now let's go with depth 1.

Pomax (approved):

PR #7999, opened 2021-12-09
Replace cypress with playwright

### LINKED ISSUE #7711: Replace Cypress with Playwright
The combination of docker and cypress has led to a situation where we can run cypress in CI context, where node and python run side by side, but not easy in local dev context, where different docker containers run node and python codebases. We need to rewrite how the cypress tasks work for local testing so that they fire up the correct Docker containers for us to do visual/integration testing with.

### update

Given how much headaches cypress is giving, this issue was morphed into swapping Cypress out for Playwright

Pomax:
Closes #7711

This replaces Cypress with Playwright, with updated Percy integration (using the new percy/cli rather than the older now obsolete percy-exec github action).

### To run this locally:

- grab branch
- `inv catchup`
- `npm install`
- run `docker-compose up`
- run `npm run tailwind` in a second terminal (for now)
- run `npm run playwright` in a THIRD terminal - this will run the integration tests. Of which there is all of one right now.
- or, run `npm run playwright:urls` to check our slew of URLs for 200/404-ness
- or, run `npm run playwright:visual` to get page screenshots
  - this last one will generate a timestamped screenshots dir inside `./tests`

### To run this on github:

it already does, as part of the CI for this PR.


tests/wait-for-images.js
```diff
@@ -0,0 +1,41 @@
...
+  page.on(`console`, console.log);
+
+  const images = page.locator("img");
+
+  return new Promise(async (resolve, reject) => {
+    let cutoff = 20;
+    (async function testLoaded() {
+      // force-load all lazy content
+      await images.evaluateAll((imgs) => {
+        const visible = Array.from(imgs).filter(i => i.offsetParent !== null)
```
Pomax, line 17: If an element isn't visible due to `display:none` anywhere on it or its ancestors, `e.offsetParent` will be null

danielfmiranda (approved):
Hi Pomax, thanks for providing easy-to-follow instructions. 

Everything seems to be working as expected on my local machine, and when I look at the CI tests I see there are mentions of playwright. 

Approving!

fessehaye (changes requested):
```
$ npm run playwright:urls

> playwright:urls
> wait-on -i 3000 http://localhost:8000/cms && playwright test ./tests/urls.spec.js     

Using config at C:\Users\Simon\Documents\GitHub\foundation.mozilla.org\playwright.config.js
ReferenceError: FoundationURLs is not defined
```

looks like it is missing an import?

Pomax:
@fessehaye indeed, looks like the url file was missing the fomo and mozfest URL imports - updated =D

fessehaye (approved):
Thanks for the update!

PR #8008, opened 2021-12-14
7768 - PNI - add animations for categories and products

### LINKED ISSUE #7768: PNI - add animations for categories and products
Animations related to the product search page. Including navigating the category header and the transitions for hiding and showing products.

fessehaye:
Closes #7768 

https://foundation-s-7768-anima-yqou7g.herokuapp.com/en/privacynotincluded/

* Applied animations for product home page.
* 'All' category has a scrolling animation that reveals more cards as the cards are visible in the viewport.
* All other categories will have their transitions applied immediately regardless if they are visible
* Transitions should not be applied to checkbox changes.
* Animations should still be preserved when going back/forward

Add `https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/gsap.min.js https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/ScrollTrigger.min.js` to `SCP_SCRIPT_SRC` in `.env`


Pomax:
@fessehaye taking the reviewers off until this is ready for review again, just so it's out of people's work queues.

fessehaye:
@Pomax I will need to bug you about how to go about fixing the tests with the new animation changes!

.github/workflows/continous-integration.yml
```diff
@@ -196,6 +196,9 @@ jobs:
       USE_S3: False
       X_FRAME_OPTIONS: DENY
       XSS_PROTECTION: True
+      CSP_FONT_SRC: 'self' https://fonts.gstatic.com https://fonts.googleapis.com https://code.cdn.mozilla.net
```
Pomax, line 199: remember to also update the `env.default`

network-api/networkapi/wagtailpages/templates/buyersguide/catalog.html
```diff
@@ -141,5 +141,7 @@ <h2 class="h3-heading">{% trans "Don’t see the product you’re looking for?"
  </div>
 </div>
 
+<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/gsap.min.js" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
+<script src="https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/ScrollTrigger.min.js" crossorigin="anonymous" referrerpolicy="no-referrer"></script>
```
Pomax, line 145: can these run async/deferred or do they have to load synchronously?

source/js/buyers-guide/search/search-filter.js
```diff
@@ -21,6 +24,27 @@ export class SearchFilter {
...
+    Promise.all(
+      Array.from(
+        document.querySelectorAll(".product-box.d-flex img.product-thumbnail")
+      )
+        .slice(0, 8)
+        .filter((img) => !img.complete)
+        .map(
+          (img) =>
+            new Promise((resolve) => {
+              img.onload = img.onerror = resolve;
```
Pomax, line 38: surely this should be `img.onload = resolve; img.onerror = reject`?

(and then using [Promise.allSettled](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/allSettled) rather than `Promise.all`)

source/js/buyers-guide/search/utils.js
```diff
@@ -32,8 +32,10 @@ export class Utils {
    * @param {*} parent
    */
   static updateHeader(category, parent) {
+    let headerText = document.querySelector(".category-header");
```
Pomax, line 35:
```suggestion
    const headerText = document.querySelector(".category-header");
```

source/sass/buyers-guide/views/catalog.scss
```diff
@@ -215,6 +215,9 @@ body.catalog {
         $recommendation-width: 17px;
         $recommendation-height: 15.8px;
 
+        // Needed for animation state while JS loads
+        opacity: 0;
+
```
Pomax, line 220: This will also mean that people with JS disabled see nothing, not even a "please turn on javascript to use this guide", which is not great: is there a way we can make sure this property gets preset only when JS is enabled?

Pomax (changes requested):

Pomax (approved):
aww yeah

PR #8077, opened 2022-01-07
updated use of extended_yes_no / removed old 'show_value_as_symbol' flag

### LINKED ISSUE #7757: [PNI] - Investigate "|extended_yes_no" flag
In "product_page.html" we are sending a number of items with a "|extended_yes_no" flag. 

If you take a look at the file "bg_selector_tags.py", you will see that this is used to return a human readable string from the extended yes no choices. 

However, "product_criterion.html" the template we are sending these values to, is actually expecting one of the 4 original choices (CD/NA/YES/NO). 

We should update the template and or remove the flag from some instances of "product_criterion" so we can make sure that they display the values correctly.

danielfmiranda:
Closes #7757
Related PRs/issues #

**Link to sample test page**: https://foundation-s-updated-ex-igw1dr.mofostaging.net/en/privacynotincluded/consumer-home-light-carry-contain-force/

**Steps to test**:

1. Visit the test site linked above
2. Scroll down to the "security/privacy/ai" tabs.
3. Click around, everything should look the same as before
4. If everything is behaving as expected, testing is complete!

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/fragments/product_criterion_primary_info.html
+++ b/network-api/networkapi/wagtailpages/templates/fragments/product_criterion_primary_info.html
@@ -1,4 +1,4 @@
-{% load i18n static wagtailcore_tags %}
+{% load i18n static wagtailcore_tags bg_selector_tags %}
 
 <div class="primary-info">
     <h3 class="tw-mb-4 tw-font-zilla tw-text-2xl tw-leading-7 tw-flex tw-items-center tw-flex-wrap">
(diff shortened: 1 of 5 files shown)


network-api/networkapi/wagtailpages/templates/fragments/product_criterion_primary_info.html
```diff
@@ -14,34 +14,8 @@ <h3 class="tw-mb-4 tw-font-zilla tw-text-2xl tw-leading-7 tw-flex tw-items-cente
...
-        {% endif %}
-      {% else %}
-        {% if show_value_as_symbol %}
-        <img src="{% get_static_prefix %}_images/buyers-guide/score/icon-{{value|lower}}.svg" alt="{{value}}">
-        {% else %}
-        {{ value }}
-        {% endif %}
-      {% endif %}
-    </p>
+      <p class="rating pni-product-smaller-body mb-0">
```
danielfmiranda, line 17: After taking a look through the code base, it appears that "show_value_as_symbol" is set to false or not set at all on all the fields using "product_criterion". 

To further simplify the template, I have removed the logic for "show_value_as_symbol" as it is no longer needed, and parsing the value using "extended_yes_no"

network-api/networkapi/wagtailpages/templatetags/bg_selector_tags.py
```diff
@@ -14,17 +14,6 @@ def yes_no(value):
...
-        return gettext('Yes')
-    if value == 'No':
-        return gettext('No')
-    if value == 'U':
-        return gettext('Unknown')
-    return value
-
-
 @register.filter
 def extended_yes_no(value):
```
danielfmiranda, line 18: Since the only different option in extended_boolean is "U" we can use add it to "extended_yes_no" instead, and this will allow us to parse all values using the extended_yes_no tag in "product_criterion_primary"

network-api/networkapi/wagtailpages/templates/fragments/product_tab.html
```diff
@@ -107,31 +107,30 @@ <h3 class="tw-mb-4 tw-font-zilla tw-text-2xl tw-leading-7">{% trans "What data d
                     {% blocktrans with url=min_sec_url|add:"#minimum-security-standards" asvar minimum_security_standards trimmed %}
                     Does this product meet our <span class="tw-ml-1"></span> <a id="mss-link" class="" href="{{ url }}">Minimum Security Standards</a>?
                     {% endblocktrans %}
-                    {% include "fragments/product_criterion.html"  with  show_value_as_symbol=False   value=product.meets_minimum_security_standards|yes_no   help=""   label=minimum_security_standards|safe   ding=product.show_ding_for_minimum_security_standards info=min_sec_url  %}
```
danielfmiranda, line 110: Since every field has "show_value_as_symbol" set to false, i removed them as they are no longer needed!

danielfmiranda:
Hi @fessehaye! Since you worked on this template during the PNI work last year, I tagged you with the hope you can give the review app a quick look to double check that everything still appears as it should. Thanks!

fessehaye (approved):

network-api/networkapi/wagtailpages/templatetags/bg_selector_tags.py
Pomax: good thinking!

Pomax (approved):

Issue #8092, opened 2022-01-10
PNI has duplicate categories in the prod db


Pomax:
Running wagtailpages migration 0064 on the prod db crashes on this line:

```python
    BuyersGuideProductCategory = apps.get_model("wagtailpages", "BuyersGuideProductCategory")
    healthAndExercise = BuyersGuideProductCategory.objects.get(name="Health & Exercise") <--
```

because apparently more than one category has this name. Which should not be the case. It looks like wagtailpages/buyersguideproductcategory id:5 and wagtailpages/buyersguideproductcategory id:61 both have the "Health & Exercise" name, despite #61 supposedly being a Polish translation (and so shouldn't be the same string)

TheoChevalier:
Hm, translation has been started for this snippet, but there’s currently no Polish translation here: https://pontoon.mozilla.org/pl/foundation-website-content/templates/snippets/wagtailpages.BuyersGuideProductCategory/5-health-exercise.pot/?string=230845

It’s not clear if we’re doing something wrong in our code or if it’s a wagtail-localize bug, but all I can say is this is a valid use case (translation of the page/snippet started, waiting for actual translation of the string)

TheoChevalier:
FWIW, I’ve synced these snippets, it seems the share image was missing on translations, but I doubt it has anything to do with the error above

Pomax:
I see, yes then that's probably a code issue - it uses `get` rather than `filter` so it's getting that alias atm without filtering for the english-locale-only objects.

PR #8093, opened 2022-01-11
fix locale-related migration

Pomax:
Closes https://github.com/mozilla/foundation.mozilla.org/issues/8092

- check out main
- run `inv new-env`
- run `inv copy-stage-db`
- notice it crashes while running migrations
- run `inv migrate` and observe that the error is about getting more than one product category in migration 0064
- check out this branch
- rerun `inv copy-stage-db`
- observe that there are no errors
- run `docker-compose up` and open http://localhost:8000/en/privacynotincluded/
- click on  "health & wellness" and observe that it has three subcategories.


danielfmiranda (changes requested):
Hi Pomax, 

When following the instructions, I noticed that on my local machine the subcategory pills do not seem to appear. However, I can see that the migration has ran and that the subcategories have been created.

Screenshot below:  

image

danielfmiranda (approved):
Hi Pomax! The workflow itself works great and I am able to do everything from recreating the migration errors, to running inv copy-stage-db using your branch and seeing the error go away. 

However, I am still experiencing the issue where the pills are not appearing for me on either firefox/chrome. 

That said, I know you mentioned that everything seems to be working on your machine, so I am approving in case you believe this issue might be isolated to my machine (as I know the sooner we get this fix in the better)

Please let me know if you have any questions, thanks!

PR #8094, opened 2022-01-12
tailwind classes for buttons, updated js

### LINKED ISSUE #8037: Clean up Share-buttons.js
Share-buttons.js is the file we use for the functionality and animation of the share buttons on the youtube regrets reporter extension landing page. 

While it is functional at the time I think it might be useful to clean it up a little using some suggestions from Simon such as:

- Creating a timeline and then using the .play and .reverse functionailty
- Cleaning up the SCSS to use tailwind classes as much as possible

danielfmiranda:
Closes #8037

**Steps to test**:

1. Visit https://foundation-s-share-butt-p0ashv.herokuapp.com/en/youtube/regretsreporter/
2.  Click on the share button
3. Other share buttons should animate/hide as expected
4. If code and animation look ok, testing is complete!


source/sass/views/regrets-reporter-landing-page.scss
```diff
@@ -49,44 +49,17 @@
...
       }
 
       &::before {
-        filter: brightness(1);
-        content: " ";
-        display: block;
-        width: 36px;
-        height: 36px;
-        background: no-repeat 0 0 / contain;
         transition: $button-icon-transition-speed;
```
fessehaye, line 60: Do you think this transiton variable value might be reused in a lot of the site? If so maybe we should add it to the tailwind config

fessehaye (approved):
This looks way better! Thanks for taking the time to do this

PR #8095, opened 2022-01-12
removed all if else statements from switchover to basket

### LINKED ISSUE #6596: Fo'mo Cinchy/CRM Migration Code Cleanup
Clean up code for migration on donate and foundation (do After production launch on May 25th)

 - [x] Look through code and environment variables changed on fo'mo because of migration
- [x] Update fo'mo to remove branching and method env variable

danielfmiranda:
Closes #6596 
Related PRs/issues:
#6552(regarding removal of "MOFO_NEWSLETTER_SUBSCRIBE_METHOD" env variables and related if/else statements), #6601 (regarding removal of "petition_method" env variable) 

### Details about this PR:

The goal of this pull request is to remove the "if else" statements  and env variables that we have added to the codebase during the switchover from cinchy and SQS to basket.

We added these statements so based on an environment variable, the site could know which service to use.

If you would like to see the changes I am undoing, #6552 is the PR in which I added the following environment variables:

- BASKET_URL
- MOFO_NEWSLETTER_SUBSCRIBE_METHOD

This PR removes those, and also removes the if/else statement, as the site now uses basket 100% of the time since the switchover is complete.

#6601 is where I added a if else statement for basket regarding petition newsletter sign ups. It also appears I added an environment variable "PETITION_DATA_SUBMISSION_METHOD", however this is not used anywhere in the codebase so it can be removed too.


env.default
```diff
@@ -33,12 +33,6 @@ XROBOTSTAG_ENABLED=False
 # See the Basket client docs https://basket-client.readthedocs.io/en/latest/install.html
 BASKET_URL=https://basket-dev.allizom.org/
 
-# Newsletter config:
```
danielfmiranda, line 36: Added in [#6552](https://github.com/mozilla/foundation.mozilla.org/pull/6552/files#diff-14d0b30702e03f0c4431c6bb68320bc3b0d1660ab34d9e501d28b503d424d42aR34) - no longer needed

network-api/networkapi/campaign/views.py
```diff
@@ -148,26 +148,12 @@ def signup_submission(request, signup):
     if cid is not None and cid != '':
         data['campaign_id'] = cid
 
-    # If we want to subscribe them through basket, do so, if not, subscribe them using SQS.
```
danielfmiranda, line 151: Added in [6552](https://github.com/mozilla/foundation.mozilla.org/pull/6552/files#diff-5f0534998c1054cb2971d3f838cf3a37334e8d4c546b13088a734cee6213bde5R124) - no longer needed

network-api/networkapi/campaign/views.py
```diff
@@ -148,26 +148,12 @@ def signup_submission(request, signup):
...
-        message = json.dumps({
-            'app': settings.HEROKU_APP_NAME,
-            'timestamp': datetime.now().isoformat(),
-            'data': {
-                'json': True,
-                'form': data,
-                'event_type': 'newsletter_signup_data'
-            }
-        })
-        return send_to_sqs(crm_sqs['client'], crm_queue_url, message, type='signup')
```
Pomax, line 170: does this mean we're no longer directly using sqs? If so, we can probably also clean up that code.

Pomax (changes requested):
not really a "change" request, but foremost a question that _might_ turn into additional removal =)

network-api/networkapi/campaign/views.py
danielfmiranda: Hi @Pomax ! We are still using SQS for petition sign ups [here](https://github.com/mozilla/foundation.mozilla.org/blob/8cb61055409304639ec1f3cb04e1ed8679fd8140/network-api/networkapi/campaign/views.py#L227)

Pomax: I see, in that case let's not delete this function =D

Pomax (approved):
this looks good to go

PR #8100, opened 2022-01-13
added caching to CI

### LINKED ISSUE #6728: Add caching to our workflows
There are github actions for caching things like python and node dependencies, which may speed up our CI tasks considerably after first run. We should look into applying https://github.com/actions/cache to our workflow files.

danielfmiranda:
Closes #6728

Docs Referenced:
https://docs.github.com/en/actions/advanced-guides/caching-dependencies-to-speed-up-workflows
https://github.com/actions/setup-python
https://github.com/actions/setup-node


.github/workflows/continous-integration.yml
```diff
@@ -71,6 +72,7 @@ jobs:
     - uses: actions/setup-python@v2
       with:
         python-version: 3.8
+        cache: 'pip'
```
danielfmiranda, line 75: According to https://github.com/actions/setup-python, we can set the "cache" flag, which uses https://github.com/actions/toolkit/tree/main/packages/cache under the hood!

.github/workflows/continous-integration.yml
```diff
@@ -25,6 +25,7 @@ jobs:
     - uses: actions/setup-node@v2
       with:
         node-version: 14
+        cache: 'npm'
```
danielfmiranda, line 28: According to https://github.com/actions/setup-node, we can set the "cache" flag, which uses https://github.com/actions/toolkit/tree/main/packages/cache under the hood!

Pomax (approved):
sweet

PR #8112, opened 2022-01-14
7257 MultipageNavMobile component should not be a react component

### LINKED ISSUE #7257: MultipageNavMobile component should be a Django template/partial instead of a React component
MultipageNavMobile component should honestly be a more simple component with minimum js

fessehaye:
Closes #7257 

Originally the goal was to make this as a wagtail template but I learned that the secondary nav links get implemented a few different ways through out the project. Instead, of making the several different templates of each usage I decided to still use JS to generate most of the HTML. The main difference is that we do not use react anymore.

How to test:
- check `/en/initiatives/` and adjust your screen for a mobile display and the mobile navigation should look and function the same!


danielfmiranda (approved):
Hi Simon! Sorry it took me a while to get to this, I just tested it out and it works great. Approved!

PR #8125, opened 2022-01-19
Ensure related posts are of the same locale

### LINKED ISSUE #7991: Autofilled related posts are picking translated pages
STR:
- Edit a blog post
- Make sure only one "related post" is picked, with two empty slots
- Hit save

→ Edit the page again, the two empty slots are now (likely) filled with translated pages.

image(3)

tbrlpld:
Before, it could happen that related pages would be returned that had a
different locale than the reference page.

If a reference page would have a fairly unique combination of tags, it could
happen that all related pages would be different locales of the
reference page.

This is now prevented by filtering related pages to be of the same
locale as the reference page.

Closes #7991 
Related PRs/issues #

**Link to sample test page**:

## Checklist

_Remove unnecessary checks_

**Tests**
- [x] Is the code I'm adding covered by tests? No

**Changes in Models:**
- [x] Did I update or add new fake data? No
- [x] Did I squash my migration? No migrations added
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team? Yes (no migrations)

**Documentation:**
- [x] Is my code documented? The comment is updated. 
- [x] Did I update the READMEs or wagtail documentation? No

### DIFF
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -179,9 +179,7 @@ def get_content_related_by_tag(page, result_count=3):
 
     for page_type in page_models_with_tags:
         # Get all pages that share tags with this page
-        related_pages = (
-            page_type.objects.filter(tags__in=own_tags).live()
-        )
+        related_pages = page_type.objects.filter(tags__in=own_tags).live()
 
         # Filter related pages to be of same locale as reference page
         if own_locale:
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -172,7 +172,6 @@ def get_content_related_by_tag(page, result_count=3):
 
     results = False
     own_tags = page.tags.all()
-
     own_locale = getattr(page, 'locale', None)
 
     for page_type in page_models_with_tags:
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -174,8 +174,10 @@ def get_content_related_by_tag(page, result_count=3):
     own_tags = page.tags.all()
 
     for page_type in page_models_with_tags:
-        # get all pages that share tags with this page
-        related_pages = page_type.objects.filter(tags__in=own_tags).live()
+        # Get all pages that share tags and locale  with this page
+        related_pages = (
+            page_type.objects.filter(tags__in=own_tags).filter(locale=page.locale).live()
+        )
 
         # Exlude "this page" from the result set for the page's own page type
         # so that we don't end up with "this page is most similar to itself".
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -172,11 +172,12 @@ def get_content_related_by_tag(page, result_count=3):
 
     results = False
     own_tags = page.tags.all()
+    own_locale = page.locale
 
     for page_type in page_models_with_tags:
         # Get all pages that share tags and locale  with this page
         related_pages = (
-            page_type.objects.filter(tags__in=own_tags).filter(locale=page.locale).live()
+            page_type.objects.filter(tags__in=own_tags).filter(locale=own_locale).live()
         )
 
         # Exlude "this page" from the result set for the page's own page type
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -173,9 +173,7 @@ def get_content_related_by_tag(page, result_count=3):
     results = False
     own_tags = page.tags.all()
 
-    own_locale = None
-    if hasattr(page, 'locale'):
-        own_locale = page.locale
+    own_locale = getattr(page, 'locale', None)
 
     for page_type in page_models_with_tags:
         # Get all pages that share tags with this page
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -172,7 +172,7 @@ def get_content_related_by_tag(page, result_count=3):
 
     results = False
     own_tags = page.tags.all()
-    own_locale = getattr(page, 'locale', None)
+    own_locale = page.locale
 
     for page_type in page_models_with_tags:
         # Get all pages that share tags with this page
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -172,13 +172,20 @@ def get_content_related_by_tag(page, result_count=3):
 
     results = False
     own_tags = page.tags.all()
-    own_locale = page.locale
+
+    own_locale = None
+    if hasattr(page, "locale"):
+        own_locale = page.locale
 
     for page_type in page_models_with_tags:
-        # Get all pages that share tags and locale  with this page
+        # Get all pages that share tags with this page
         related_pages = (
-            page_type.objects.filter(tags__in=own_tags).filter(locale=own_locale).live()
+            page_type.objects.filter(tags__in=own_tags).live()
         )
+
+        # Filter related pages to be of same locale as reference page
+        if own_locale:
+            related_pages = related_pages.filter(locale=own_locale)
 
         # Exlude "this page" from the result set for the page's own page type
         # so that we don't end up with "this page is most similar to itself".
--- a/network-api/networkapi/wagtailpages/utils.py
+++ b/network-api/networkapi/wagtailpages/utils.py
@@ -174,7 +174,7 @@ def get_content_related_by_tag(page, result_count=3):
     own_tags = page.tags.all()
 
     own_locale = None
-    if hasattr(page, "locale"):
+    if hasattr(page, 'locale'):
         own_locale = page.locale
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/utils.py
```diff
@@ -172,11 +172,16 @@ def get_content_related_by_tag(page, result_count=3):
 
     results = False
     own_tags = page.tags.all()
+    own_locale = getattr(page, 'locale', None)
```
Pomax, line 175: According to https://docs.wagtail.io/en/stable/reference/pages/model_reference.html this should be a well defined property, so `page.locale` should work.

Pomax (changes requested):
let's see if we can do this without getattr

network-api/networkapi/wagtailpages/utils.py
tbrlpld: The pipeline was failing because some page did not have the `locale` property. Let me see if I can find it.

tbrlpld: There it is: https://github.com/mozilla/foundation.mozilla.org/runs/4873707413?check_suite_focus=true#step:10:662

tbrlpld: Ha, weird. 

I totally missed https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.locale

I thought is only on `TranslatableMixin`: https://docs.wagtail.io/en/stable/reference/pages/model_reference.html#wagtail.core.models.TranslatableMixin.locale

Pomax:
The code looks good, but visual regression doesn't ssem to like this change, so we'll want to do some debugging to find out why.

Pomax:
Looking at the regression logs, I see: 
```
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/manage.py", line 34, in <module>
    execute_from_command_line(sys.argv)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/django/core/management/__init__.py", line 401, in execute_from_command_line
    utility.execute()
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/django/core/management/__init__.py", line 395, in execute
    self.fetch_command(subcommand).run_from_argv(self.argv)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/django/core/management/base.py", line 330, in run_from_argv
    self.execute(*args, **cmd_options)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/django/core/management/base.py", line 371, in execute
    output = self.handle(*args, **options)
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/management/commands/load_fake_data.py", line 83, in handle
    [app_factory.generate(seed) for app_factory in [
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/management/commands/load_fake_data.py", line 83, in <listcomp>
    [app_factory.generate(seed) for app_factory in [
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/factory/__init__.py", line 32, in generate
    blog.generate(seed)
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/factory/blog.py", line 113, in generate
    post = BlogPageFactory.create(parent=blog_namespace, title=title)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/factory/base.py", line 528, in create
    return cls._generate(enums.CREATE_STRATEGY, kwargs)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/factory/django.py", line 117, in _generate
    return super()._generate(strategy, params)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/factory/base.py", line 465, in _generate
    return step.build()
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/factory/builder.py", line 262, in build
    instance = self.factory_meta.instantiate(
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/factory/base.py", line 317, in instantiate
    return self.factory._create(model, *args, **kwargs)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/wagtail_factories/factories.py", line 69, in _create
    instance = cls._create_instance(model_class, parent, kwargs)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/wagtail_factories/factories.py", line 77, in _create_instance
    parent.add_child(instance=instance)
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/treebeard/mp_tree.py", line 1083, in add_child
    return MP_AddChildHandler(self, **kwargs).process()
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/treebeard/mp_tree.py", line 387, in process
    newobj.save()
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py", line 282, in save
    self.ensure_related_posts()
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py", line 269, in ensure_related_posts
    for post in self.get_missing_related_posts():
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py", line 247, in get_missing_related_posts
    related_posts = get_content_related_by_tag(self)
  File "/home/runner/work/foundation.mozilla.org/foundation.mozilla.org/network-api/networkapi/wagtailpages/utils.py", line 175, in get_content_related_by_tag
    own_locale = page.locale
  File "/opt/hostedtoolcache/Python/3.9.9/x64/lib/python3.9/site-packages/django/db/models/fields/related_descriptors.py", line 197, in __get__
    raise self.RelatedObjectDoesNotExist(
wagtail.core.models.Page.locale.RelatedObjectDoesNotExist: Page has no locale.
```
so I guess we'll have to go with `getattr` for now, but it's probably a good idea for us to file a bug over on the wagtail repo to report that as a Page model field, `locale` should probably default to `None` rather than throwing an access error.

Pomax:
Also, since the visual tests caught this, but the regular tests didn't, let's also add a small test in `wagtailpages/tests.py`, where we create a blog page and try to save it in a way that this codepath kicks in. (we can probably make that a direct test rather than faking url requests_

Pomax (changes requested):
apparently page.locale can throw, which is... not so great. Let's revert to getattr, write a small test for this codepath, and let's also file an issue with wagtail to report that this property can throw and that it probably should instead universally report `None` if there is no localization, or the default locale if localization is active.

tbrlpld:
> apparently `page.locale` can throw, which is... not so great. Let's revert to getattr, write a small test for this codepath, and let's also file an issue with wagtail to report that this property can throw and that it probably should instead universally report `None` if there is no localization, or the default locale if localization is active.

I am actually confused as to why `getattr` worked.

tbrlpld:
Re: exception vs. `None`: I think this comed from Django and not Wagtail. The issue seems to be that the field is not nullable but there is no value set apparently. 

https://github.com/wagtail/wagtail/blob/main/wagtail/core/models/i18n.py#L89

tbrlpld:
It appears the issue is not the visual test itself, but the loading of the fake data: https://github.com/mozilla/foundation.mozilla.org/runs/4874848441?check_suite_focus=true#step:10:620

tbrlpld:
Ok, I am able to recreate locally with `inv new-db`. It appears the blog post generation is failing.

tbrlpld:
@Pomax I think I finally got it. 

The locale is only set when the page is first saved ([cleaned](https://github.com/wagtail/wagtail/blob/main/wagtail/core/models/__init__.py#L437) really). But, the blog post save is overridden to add the related posts. Adding the related posts requires the locale. This is why we get the `DoesNotExist` exception on a field that is not allowed to be empty.

Only including the locale in the relation when it exists does not quite work, because all three spots will be filled with related posts potentially of other locales. This is the same problem that is being reported.

The `ensure_related_posts` seems like a validation step. We could probably move that to one of the cleaning methods. `clean` is already being used. 

It also appears that the factory it not expecting for the save method to add the related posts, because the method is explicitly called before the save method is called.

[https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/factory/blog.py#L132-L134](https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/factory/blog.py#L132-L134)

```python
for post in BlogPage.objects.all():
        post.ensure_related_posts()
        post.save()
```

I have just tested it locally and that seems to work.

Pomax (approved):
such a good find.

PR #8129, opened 2022-01-20
implemented coil on mozfest site

### LINKED ISSUE #8124: Implement coil web monetization on MozFest site
Since we are going/have added the web monetization message to the MozFest site, we would like to now implement web monetization through coil https://coil.com/.

More details to follow

danielfmiranda:
Closes #8124

Using https://github.com/mozilla/foundation.mozilla.org/pull/7704 for reference, it appears that we just needed to add the monetization meta tag to the mozfest-base page in order to enable it.

### DIFF
--- a/network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
+++ b/network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
@@ -2,7 +2,9 @@
 {% load wagtailcore_tags wagtailimages_tags homepage_tags card_tags wagtailmetadata_tags custom_image_tags i18n static %}
 
 
-{% block web_monetization %}{% endblock %}
+{% block web_monetization %}
+  <meta name="monetization" content="$ilp.uphold.com/aKpM9kwnGBy4">
+{% endblock %}
 
 
 {% block ga_identifier %}


network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
```diff
@@ -2,7 +2,9 @@
 {% load wagtailcore_tags wagtailimages_tags homepage_tags card_tags wagtailmetadata_tags custom_image_tags i18n static %}
 
 
-{% block web_monetization %}{% endblock %}
+{% block web_monetization %}
+  <meta name="monetization" content="$ilp.uphold.com/aKpM9kwnGBy4">
```
Pomax, line 6: Is there somewhere that this value can be found for future engineering work? Also, would it make sense to turn this into an environment variable like `COIL_WALLET_URL` or is this essentially a "permanent" public URL?

Pomax (changes requested):

network-api/networkapi/mozfest/templates/mozfest/mozfest-base.html
danielfmiranda: Hi @Pomax!

From Anil: It looks like the URL can be found on the coil site when logged in, and does not appear to change.

image (3)

Pomax (approved):

PR #8183, opened 2022-02-03
8173- PNI card animations to "for transitions only"

### LINKED ISSUE #8173: Switch the PNI card animations to "for transitions only"
Based on https://github.com/mozilla/foundation.mozilla.org/issues/8172:  as folks will currently see nothing if JS doesn't load (specifically, if the js library needed for transitions doesn't load for whatever reason), let's switch our PNI card animations on the homepage/category pages to show all relevant products on initial page load without transitions for the "above the fold" content, and transitions once we start scrolling, and have transitions for "all cards" on subsequent JS-based view changes (including a check to see if gsap has loaded, and if not, using our previous vanilla JS/CSS approach rather than triggering a transition)

fessehaye:
Closes #8173 

Undo animation on first 8 products when opening the privacynotincluded page
We do not have to check for gsap anymore since it's imported on build!

### DIFF
--- a/source/js/buyers-guide/search/search-filter.js
+++ b/source/js/buyers-guide/search/search-filter.js
@@ -42,9 +42,9 @@ export class SearchFilter {
         )
     ).then(() => {
       if (this.categoryTitle.value === "None") {
-        Utils.toggleScrollAnimation();
+        Utils.toggleScrollAnimation(true);
       } else {
-        Utils.toggleCategoryAnimation();
+        Utils.toggleCategoryAnimation(true);
       }
     });
   }
--- a/source/js/buyers-guide/search/utils.js
+++ b/source/js/buyers-guide/search/utils.js
@@ -158,39 +158,63 @@ export class Utils {
   /**
    * Scroll Animation used solely for the 'All Products' section
    */
-  static toggleScrollAnimation() {
+  static toggleScrollAnimation(initialLoad = false) {
     ScrollTrigger.clearScrollMemory();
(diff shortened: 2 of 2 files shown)


source/js/buyers-guide/search/search-filter.js
```diff
@@ -42,9 +42,9 @@ export class SearchFilter {
         )
     ).then(() => {
       if (this.categoryTitle.value === "None") {
-        Utils.toggleScrollAnimation();
+        Utils.toggleScrollAnimation(true);
       } else {
-        Utils.toggleCategoryAnimation();
+        Utils.toggleCategoryAnimation(true);
       }
```
Pomax, line 48: This isn't true, though. Initial load is not base don "is there a category", it's `true` by default, and then needs to get set to `false`. So in the constructor, `this.initialLoad = true`, and then in this `then` handler:
```js
    .then(() => {
      if (this.categoryTitle.value === "None") {
        Utils.toggleScrollAnimation(this.initialLoad);
      } else {
        Utils.toggleCategoryAnimation(this.initialLoad);
      }
      this.initialLoad = false;
    });
```

Pomax (changes requested):
Let's change this just a little, because right now the code logic doesn't actually toggle based on load, but purely looks at the category name.

Pomax (approved):
nope, I was wrong: unrelated review-app issue getting in the way. local testing shows this working beautifully!

nancyt1 (approved):

PR #8210, opened 2022-02-07
8131 - Fix current event sliders width

### LINKED ISSUE #8131: Current Events Slider no longer full width on MozFest homepage
The 'Current Events Slider' on mozillafestival.org is now constrained to the width of the content in the left column where previously it would be the full width of the page - even with the newsletter snippet on the page. This is across FF/Chrome/Safari on desktop. The mobile version is not affected.

## Current View:
Screenshot 2022-01-21 at 15 56 54

## Previous View:
Screenshot 2021-11-04 at 17 44 39

fessehaye:
Closes #8131

### DIFF
--- a/network-api/networkapi/mozfest/templates/fragments/hero/carousel_hero.html
+++ b/network-api/networkapi/mozfest/templates/fragments/hero/carousel_hero.html
@@ -1,6 +1,6 @@
 {% load i18n wagtailimages_tags %}
 <div data-mozfest-hero-carousel>
-  <div class="tw-relative tw-pt-[120px] tw-flex tw-flex-col tw-justify-end{% if not page.banner_video %} tw-pb-6 medium:tw-pb-[120px]{% endif %}">
+  <div class="tw-relative tw-pt-[120px] tw-flex tw-flex-col tw-justify-end{% if not page.banner_video %} tw-pb-6 medium:tw-pb-[120px] {% endif %}">
 
     {# Background images slider #}
     <div class="tw-absolute tw-inset-0">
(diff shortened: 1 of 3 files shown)


sabrinang (approved):
It appears to be the width of the container now and tested adding additional cards and looks as expected on desktop and mobile with functional links. 👍 Thanks for fixing this @fessehaye !

danielfmiranda (approved):
Hey Simon, thanks for tagging me in this, was really interesting to see how you fixed this using tailwind classes in the template. 

Also, after looking at the test app, everything looks good! Approving 👍

network-api/networkapi/mozfest/templates/fragments/hero/carousel_hero.html
```diff
@@ -1,6 +1,6 @@
 {% load i18n wagtailimages_tags %}
 <div data-mozfest-hero-carousel>
-  <div class="tw-relative tw-pt-[120px] tw-flex tw-flex-col tw-justify-end{% if not page.banner_video %} tw-pb-6 medium:tw-pb-[120px]{% endif %}">
+  <div class="tw-relative tw-pt-[120px] tw-flex tw-flex-col tw-justify-end{% if not page.banner_video %} tw-pb-6 medium:tw-pb-[120px] {% endif %}">
```
tbrlpld, line 3: @fessehaye Just spotted this `tw-justify-end{% ...`. I think this could also lead to issues if this class is not used elsewhere because is might get purged is the curly braces are touching the class name.

Issue #8216, opened 2022-02-09
Fix fomo migrations


Pomax:
- [ ] squash mozfest
- [ ] squash wagtailpages
- [ ] update dependencies

bheasman:
@Pomax to file follow ups as dedicated issue (stretch goal)

Pomax:
Closing this issue out, with the outstanding db cleanup filed as new issue

PR #8247, opened 2022-02-15
Respun wagtailpages migrations

Pomax:
Part of https://github.com/mozilla/foundation.mozilla.org/issues/8216

This squashes all `wagtailpages` migrations in preparation of regenerating the migration file as a single "from scratch" file. Right now, the migration is still "huge" and takes quite a few seconds to apply, but at least it's a single file now, with all obsolete between-migrations-code removed.

## review:

- there's not much to review in terms of code, this is the result of `manage.py squashmigrations wagtailpages 0077`, with all the data bootstrapping preserved, but all the custom "intermediate" migration code removed because it's not longer relevant (a new db has no data to uplift, and the existing prod data already has all uplifts applied)

## testing:

- [x] CI should pass
- [ ] `inv new-db` followed by `inv test-python` should pass
- [ ] `inv copy-prod-db` (if you have an old prod db file, remember to delete that first) followed by `inv migrate` should not throw any errors. Running `inv test-python` afterwards should still pass
- [ ] manual step: running the site locally after the previous step should (as far as you can tell) work just fine. Localized data should still be there, too.


tbrlpld (approved):
@Pomax I went through all the testing steps and did not hit any issues 👍

fessehaye (approved):

danielfmiranda (approved):
Hi Pomax, apologies for the delay, however this also worked for me, approved! 👍

Issue #8176, opened 2022-02-01
Resolve author-related preview crash


Pomax:
Viewing https://foundation.mozilla.org/en/blog/why-our-tech-community-is-most-excited-about-mozfest-2022/ and previewing https://foundation.mozilla.org/cms/pages/16811/edit/ yields a server error 500, checking the logs shows:

<details>
<summary>code log</summary>

```bash
2022-02-01T17:15:39.110165+00:00 app[web.4]: return self.render(context)
2022-02-01T17:15:39.110166+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/scout_apm/core/stacktracer.py", line 25, in wrapper
2022-02-01T17:15:39.110166+00:00 app[web.4]: return wrapped(*args, **kwargs)
2022-02-01T17:15:39.110166+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/django/template/loader_tags.py", line 62, in render
2022-02-01T17:15:39.110166+00:00 app[web.4]: result = block.nodelist.render(context)
2022-02-01T17:15:39.110167+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/django/template/base.py", line 938, in render
2022-02-01T17:15:39.110167+00:00 app[web.4]: bit = node.render_annotated(context)
2022-02-01T17:15:39.110168+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/django/template/base.py", line 905, in render_annotated
2022-02-01T17:15:39.110172+00:00 app[web.4]: return self.render(context)
2022-02-01T17:15:39.110172+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/django/template/loader_tags.py", line 192, in render
2022-02-01T17:15:39.110172+00:00 app[web.4]: return template.render(context)
2022-02-01T17:15:39.110173+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/scout_apm/core/stacktracer.py", line 25, in wrapper
2022-02-01T17:15:39.110173+00:00 app[web.4]: return wrapped(*args, **kwargs)
2022-02-01T17:15:39.110173+00:00 app[web.4]: File "/app/.heroku/python/lib/python3.9/site-packages/django/template/base.py", line 172, in render
2022-02-01T17:15:39.110173+00:00 app[web.4]: return self._render(context)
[...]

PR #8250, opened 2022-02-15
make sure there are no `None` entries in related posts

Pomax:
# Edit

Temporary but immediate fix for https://github.com/mozilla/foundation.mozilla.org/issues/8176

Note: this PR does not close the issue, it just stops it from being a `p1` by plugging the hole rather than fixing the boat.

# Original comment

Problem STR:

- check out `main`
- `inv copy-stage-db`
- try to load http://localhost:8000/en/blog/why-our-tech-community-is-most-excited-about-mozfest-2022/

The reason this crashes is because `blog_cards.html` tries to apply `blog_card.html` for each entry in a blog page's `page.related_posts.all()`, which is a context variable called `related_posts` that is added in the blog page's `get_context` function, but that list may contain `None` entries if the code was unable to find enough related posts during a CMS save action to pad the related posts set to three entries.

Debug STR:
- open `homepage_tags.py` and update the `get_page_authors` tag to read:
```python
@register.simple_tag
def get_page_authors(page):
    if page is None:
        return []
    else:
        try:
            all_authors = page.authors.all()
            return [{"image": author.author.image, "name": author.author.name} for author in all_authors]
        except:
            return []
```

This will stop the tag from crashing, and will load the blog page - scroll down to the "Related posts" section, and notice that there are three blog cards: two real, and one that is completely devoid of actual content.

Solution STR:

- check out this branch
- load http://localhost:8000/en/blog/why-our-tech-community-is-most-excited-about-mozfest-2022/


tbrlpld (changes requested):
[comment moved to [https://github.com/mozilla/foundation.mozilla.org/issues/8176#issue-1121084954](https://github.com/mozilla/foundation.mozilla.org/issues/8176#issue-1121084954)

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -223,7 +223,11 @@ def get_context(self, request):
             # see that same padded list during preview, but *without* actually updating
             # the model, so we control this property at render context retrieval time:
             related_posts = related_posts + self.get_missing_related_posts()
-        context['related_posts'] = related_posts
+
+        # Make sure to filter out any None entries, which may happen if the
+        # self.ensure_related_posts() function was unable to find any related
+        # posts to pad the related posts set with.
+        context['related_posts'] = list(filter(None, related_posts))
```
tbrlpld, line 230: This is mostly matter of taste I guess, but I was wondering if we should use list comprehension rather than the list -> filter -> list nesting. I find them a bit more readble 😅 

(I can't find proof right now, but I think they are also supposed to be faster)

Pomax: The reason I'd use this rather than `[x for x in related_posts if x is not None]` is the increased potential-for-bugs surface: the list comprehention has three places to typo the anonymous variable, as opposed to "there isn't even a variable to typo" in the filter call.

tbrlpld: Fair 👍

tbrlpld (approved):

PR #8252, opened 2022-02-15
regenerated the wagtailpages migrations

Pomax:
part of https://github.com/mozilla/foundation.mozilla.org/issues/8216

## review:

There's not much to review in terms of code. The approach taken was to remove the wagtailpages migrations, and call a `makemigrations` for the wagtailpages app specifically, to builds new migrations for "what the models look like, right now" without all the historical baggage.

The resulting file as then marked as replacing the previous with-historical-baggage squash migration (0001) and associated data bootstrap migration (0002), with the mozfest 0001-0015 migration updated to point to the new, clean migration.

## testing:

- [x] CI should pass
- [ ] `inv new-db` followed by `inv test-python` should pass
- [ ] `inv copy-prod-db` (if you have an old prod db file, that will have to be deleted again first!) should not throw any errors. Running `inv test-python` afterwards should still pass
- [ ] manual step: running the site locally after the previous step should (as far as you can tell) work just fine. Localized data should still be there, too.


danielfmiranda (approved):
Just tested and works great, approved! Thanks Pomax 👍

fessehaye (approved):

tbrlpld (approved):
Tested all the steps without issue. And wow are the tests fast now 🚀 🎉 Thanks for this @Pomax

PR #8258, opened 2022-02-16
Squashed mozfest migrations

Pomax:
part of https://github.com/mozilla/foundation.mozilla.org/issues/8216

This squashes the mozfest migrations into an "with historical bagage" single migration for the models, and secondary migration for the single data bootstrap operation we need.

STR (You know the drill by now):
- [x] CI should pass
- [ ] `inv new-db` with `inv test-python` should pass
- [ ] `inv copy-prod-db` with `inv test-python` should pass
- [ ] manual testing the mozfest pages should not show anything weird (including localised pages)


danielfmiranda (approved):
No errors for me when following the testing steps, approved!

tbrlpld (approved):

Issue #7282, opened 2021-08-30
Missing publish date on alias blog pages


TheoChevalier:
Page aliases for blog posts are currently showing `not published yet` while the original /en page is published. E.g:
- Page  alias: https://foundation.mozilla.org/es/blog/announcing-common-voices-kiswahili-community-event-this-august/ 
- Original: https://foundation.mozilla.org/en/blog/announcing-common-voices-kiswahili-community-event-this-august/

Wagtail is looking for the published date of the alias, we should make it look up the one from the /en page instead

Pomax:
Hm, I'm having a hard time reproducing this, here's what I've done:

1. new DB
2. created a new locale for German, synced to English
3. ran `sync_locale_trees`
4. confirmed that there is a german alias for the english "test blog page": 
![image](https://user-images.githubusercontent.com/177243/132412842-4118b820-c8e7-4409-b4e0-78ce5655f5f7.png)
5. visted http://localhost:8000/en/blog/initial-test-blog-post-with-fixed-title/: 
![image](https://user-images.githubusercontent.com/177243/132412909-70614cab-b441-4dfa-91f8-9407526a9301.png)
6. visited http://localhost:8000/de/blog/initial-test-blog-post-with-fixed-title/: 
![image](https://user-images.githubusercontent.com/177243/132412956-0752d62d-530e-43a9-ba07-2014029741a8.png)

This shows a different date format, but it does show a date. Would you happen to have additional details that lets me reproduce this locally?

TheoChevalier:
I played around, and indeed, we were missing a step and it’s the weirdest…

As step 3.5, translate the /blog (blog index) page in German. You don’t need actual translations, just starting the translation.

Now create and publish your test blog post in English, and the German blog post will show "not published yet" 🤷‍♂️🤷‍♂️🤷‍♂️

Pomax:
I still can't seem to reproduce this - I run through steps 1 through 4, then for the blog index page:

![{14189427-8A14-48C1-89D5-FC6F9E8ACD3D}](https://user-images.githubusercontent.com/177243/132564426-3b47caf0-4a98-4b32-8654-4a6010e09688.png)

I click "translate this page", then:

![{40B2938F-1C63-4356-9275-213A35683C8B}](https://user-images.githubusercontent.com/177243/132564451-5549526a-8950-4ea0-bf0f-1970530d9ea6.png)

I click "submit", then try the german blog page:

[...]

PR #8277, opened 2022-02-21
Command to update first/last published date on alias pages

TheoChevalier:
This management command written by @zerolab will fix issue #7282, once we upgrade to Wagtail 2.16 (happening in #8295).

I successfully ran it locally on a prod DB dump with Wagtail 2.16.1, the easiest way to notice its effects is that now /en/blog and /fr/blog are showing the same blog posts, while previously older alias pages were showing up at the top of localized blog homepages.


network-api/networkapi/management/commands/fix_alias_page_first_last_published.py
```diff
@@ -0,0 +1,26 @@
...
+    def handle(self, **options):
+        aliases_without_first_last_published = (
+            Page.objects
+            .filter(alias_of__isnull=False)
+            .filter(Q(alias_of__first_published_at__isnull=False) | Q(alias_of__last_published_at__isnull=False))
+            .select_related('alias_of')
+        )
+        found = aliases_without_first_last_published.count()
+        if not found:
+            return self.stdout.write(self.style.WARNING("Found no alias pages to update"))
```
Pomax, line 17: Can we use a normal print statement here instead? This doesn't feel like a warning so much as the actual desired case.

network-api/networkapi/management/commands/fix_alias_page_first_last_published.py
```diff
@@ -0,0 +1,26 @@
...
+
+        self.stdout.write(f"Found {found} alias page(s) to update")
+        for alias in aliases_without_first_last_published:
+            alias.first_published_at = alias.alias_of.first_published_at
+            alias.last_published_at = alias.alias_of.last_published_at
+            alias.save()
```
Pomax, line 23: since we're doing with a published alias, does this need a save_revision, too?

Pomax (changes requested):

network-api/networkapi/management/commands/fix_alias_page_first_last_published.py
zerolab: Alias pages don't have revisions. The first revision gets created when the alias is converted to a full page - https://github.com/wagtail/wagtail/blob/716bf92c2dc2da2aca5e8f5aa6768b5b087cd4b0/wagtail/core/actions/convert_alias.py#L46

network-api/networkapi/management/commands/fix_alias_page_first_last_published.py
TheoChevalier: Done!

Pomax (approved):

PR #8291, opened 2022-02-22
Remove moment

### LINKED ISSUE #7310: Remove moment.js dependency and update news.jsx to use plain JS
We use moment.js for a single date formatting operation in `news.jsx`, let's refactor this to plain JS instead so we can get rid of this dependency (not to mention that moment.js itself tells folks to not use it anymore =)

This will also require updating `local_development.md`

1. refactor the code that _calls_ moment functions to not do that (but leave moment itself in place)
2. update the docs
3. remove the `moment` import and dependency

danielfmiranda:
Closes #7310 

**Steps to test**:

1. Visit https://foundation-s-remove-mom-q4bh4f.mofostaging.net/en/news/
2. The dates for the publications should look the same as before, rendered as a full month name and year:
Screen Shot 2022-02-22 at 3 56 57 PM
3. If everything is looking as it should, testing is complete!


source/js/components/news/news.jsx
```diff
@@ -46,7 +46,10 @@ class News extends Component {
   }
 
   render() {
+    const currentLanguage = getCurrentLanguage();
```
danielfmiranda, line 49: Hi @TheoChevalier! Thank you very much for taking a look at this PR, I was wondering if I can get some feedack on whether or not there is a better way to localize the publish date here? 

What I am doing is getting the locale using the getCurrentLanguage function, and then sending it to [.getLocaleDateString()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString) in order to return a date in the format "January 2022". If the toLocaleDateString cannot find a matching locale, it defaults to the users preferred locale.

TheoChevalier: Thanks @danielfmiranda! I think that’s the best way of doing it, yup

TheoChevalier (approved):

docs/local_development.md
```diff
@@ -51,7 +51,7 @@ For instance, you can run also run common Django commands via invoke, such as `i
 **A few examples:**
 
 - `invoke manage load_fake_data`: add more fake data to your project,
-- `invoke npm "install moment"`: install moment, add it to your `package.json` and lock it.
+- `invoke npm "install gsap"`: install moment, add it to your `package.json` and lock it.
```
Pomax, line 54:
```suggestion
- `invoke npm "install gsap"`: install gsap, add it to your `package.json` and lock it.
```

Pomax (changes requested):

danielfmiranda:
closing in place of #8312

PR #8294, opened 2022-02-22
minor improvements to copy script

Pomax:
This makes sure we first set up site bindings and an admin password, _then_ try to migrate. That way, if migration fails, the database is still fully usable.

(No associated issue)


tbrlpld (approved):
Thanks @Pomax This looks like good changes to me 👍

PR #8295, opened 2022-02-23
Upgrade django, wagtail and wagtail-localize

### LINKED ISSUE #7282: Missing publish date on alias blog pages

TheoChevalier:
Closes #7282

With wagtail-localize 1.1rc1 now supporting wagtail 2.16, we can finally upgrade.
Wagtail 2.16 itself drops support for Django 3.1, so this PR is also upgrading to Django 3.2, which is the new LTS release.

Documentation:
https://docs.wagtail.org/en/stable/releases/2.16.html#upgrade-considerations
https://docs.djangoproject.com/en/4.0/releases/3.2/


Pomax:
I've updated the settings to disable the "slim sidebar" functionality, as it's pretty wonky and not necessary for our admin use.

Pomax (approved):

PR #8302, opened 2022-02-23
removing moment.js

### LINKED ISSUE #7310: Remove moment.js dependency and update news.jsx to use plain JS

danielfmiranda:
Closes #7310 
Related PRs/issues #8291

### DIFF
--- a/docs/local_development.md
+++ b/docs/local_development.md
@@ -51,7 +51,7 @@ For instance, you can run also run common Django commands via invoke, such as `i
 **A few examples:**
 
 - `invoke manage load_fake_data`: add more fake data to your project,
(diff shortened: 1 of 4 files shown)


Pomax:
superceded by https://github.com/mozilla/foundation.mozilla.org/issues/8312

PR #8308, opened 2022-02-24
7151 - Convert bootstrap components to tailwind

### LINKED ISSUE #7151: 3.5 Convert bootstrap components to tailwind components
Some classes like `.btn` and `.form-control` should probably be remade using the tailwind plugin components since they would too many seperate tailwind classes involved and making them reusage since not every button uses React or JS.

https://tailwindcss.com/docs/plugins

fessehaye:
Closes #7151

Recreating dropdowns and buttons in style-guide without using bootstrap
 ~~Not these buttons do not have a dark mode theme until we replace the dark-theme classes with `tw-dark`~~ #8329 is now included
* Added examples of the tailwind implementations on the style guide so it easier to visually check the differences.

https://foundation-s-7151-boots-7cd5os.mofostaging.net/en/style-guide/#dropdown-tailwind
https://foundation-s-7151-boots-7cd5os.mofostaging.net/en/style-guide/#button-tailwind

classes created:
* `tw-btn`
* `tw-btn-primary`
* `tw-btn-secondary`
* `tw-btn-popup`
* `tw-form-control`
* `tw-form-control-lg`


danielfmiranda (commented):
Hi @fessehaye!

Thank you for taking this ticket on, it was a big one but I think it is a great next step towards our migration from bootstrap to tailwind and is very much appreciated! 🙌

After taking a look at the style guide you provided, I can confirm that the buttons look great.

There is just one small thing I noticed, according to Percy it looks like nothing has changed visually except the form input for the newsletter at the bottom of the site. It appears that something is rounding out the corners where it used to be square:

Screenshot showing visual diff flagged:
Screen Shot 2022-02-28 at 5 12 52 PM

What they look like:
Screen Shot 2022-02-28 at 5 12 59 PM

Would it be possible to fix that? If not, can we tag some designers to make sure they are OK with it? 

Thank you!

danielfmiranda (approved):
Looks great Simon! Thanks for implementing my feedback 👍 approved!

network-api/networkapi/wagtailpages/templates/wagtailpages/static/styleguide.html
```diff
@@ -713,11 +713,11 @@ <h3 class="h2-heading tw-mt-7 tw-group" id="dropdown-dark">
         <a href="#dropdown-dark" aria-label="Anchor" class="after:tw-hash tw-opacity-0 group-hover:tw-opacity-100 tw-text-gray-40"></a>
       </h3>
       <hr>
-
-      <div class="tw-p-4 dark-theme tw-bg-black">
+
```
Pomax, line 716: rogue whitespace O_O

tailwind.button.js
```diff
@@ -0,0 +1,206 @@
+const plugin = require("tailwindcss/plugin");
+
+// TODO refactor with theme references after removing SASS/Bootstrap
+// https://tailwindcss.com/docs/theme#referencing-other-values
+
+const btnDefault = {
```
Pomax, line 6: Is there a way to specify this in real CSS rather than JS? (e.g. const btnDefault = require("btnDefault.twcss" or something?)

Pomax (changes requested):
one tiny issue, and one question

tailwind.button.js
fessehaye: We could add it to the sass files using `@layer components` and `@apply/theme()` 

```
@layer components {
  .btn {
    background-color: theme("colors.white");
    border-radius: theme("borderRadius.lg");
    padding: theme("spacing.6");
    box-shadow: theme("boxShadow.xl");
  }
}
```
this solution is a little worse because:
* no prefix added automatically
* have to wait for sass to load when doing changes
* no intellsense
* have to add to both main.scss and buyer-guide.scss

The tailwind plugin system uses CSS in JS but I think the only way we could import css files like that is if we included css modules in our build. Maybe it would be overkill if we used CSS modules for just this part?

fessehaye: A few button stylings also includes styling outside of our `tailwind.config.js` so we would have to get some more design feedback if we did it use the theme variables too

https://tailwindcss.com/docs/functions-and-directives

Pomax: Hm, yeah the first one is pretty worse, and the second isn't very appealing: if we need to create an additional transpile step, let's not do that for now and first get tailwind working at all, before making it "cleaner".

Pomax (approved):

PR #8377, opened 2022-03-02
make related blog post links get deleted

Pomax:
specifically, when  when the associated blog post gets deleted.

Closes https://github.com/mozilla/foundation.mozilla.org/issues/8176 by following Tibor's advice and using `models.CASCADE` rather than `models.SET_NULL` for the on_delete action for related posts. You can see this in action by opening the edit page for the `initial test` blog page, http://localhost:8000/cms/pages/13/edit, and scrolling all the way down, which should show three related posts. Hit up http://localhost:8000/cms/pages/15/edit/ and delete the page, then reload the "initial test" edit page and scroll all the way down again: there should now be two related posts, not three related posts of which one is empty.


tbrlpld (approved):
Thanks @Pomax. This is looking good.

PR #8381, opened 2022-03-02
[Mock] search exploratory for research hub

Pomax:
This is an experiment to see what we might need to implement search for the research hub (as well as blog). Since we don't have research models atm, the code just uses `BlogPage` instead, and requires a few things:

- `inv new-db`, followed by
- `inv manage update_index`, followed by
- `docker-compose up`

Then in the CMS, create a new child under the homepage of the `Mock Search Blog Index` type. Any title will do, then publish that. Access the resulting live page will give the barest of bones search form, with locale-aware searching:

- when searching on `/en/`, only `/en/` results will be shown. E.g. `test` will yield (at least) one result, `Initial test blog post with fixed title`
- when searching on any non-`/en/` locale, that locale's results will be shown first, followed by English results as well. E.g. `test` while on `/fr/` will yield (at least) two results, one for `/fr/`, and one for `/en/`.

This is mostly because `en` has all the data, whereas other lo1cales may only selectively hold data. This is 100% just a mock and in no way indicative of what searching will _actually_ do in the real implementation.

This PR is based on https://www.accordbox.com/blog/add-full-text-search-wagtail, with updates for the new (as of Wagtail v2.15) search backend.

Review app test URL: https://foundation-s-mock-searc-ovrd6m.mofostaging.net/en/mock-search/

Questions:
- how should searching work given multiple locales?
- how much of this makes sense to wrap up as a mixin, vs. just repeating for each page we want an index for?
- how easily can we swap out the current (convoluted) index page for campaigns and blogs?


network-api/networkapi/settings.py
```diff
@@ -492,6 +492,12 @@
 WAGTAIL_USAGE_COUNT_ENABLED = True
 WAGTAIL_I18N_ENABLED = True
 
+WAGTAILSEARCH_BACKENDS = {
```
Pomax, line 495: because of search changes in wagtail 2.15

network-api/networkapi/wagtailpages/migrations/0005_alter_relatedblogposts_related_post.py
```diff
@@ -0,0 +1,19 @@
+# Generated by Django 3.2.12 on 2022-03-02 00:40
```
Pomax, line 1: ignore this one, it's from a different branch.

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
```diff
@@ -0,0 +1,29 @@
...
+
+class MockSearchBlogIndex(RoutablePageMixin, Page):
+
+    template = 'wagtailpages/mock_search_blog_index.html'
+
+    def search_pages(self, locale, search_query):
+        return BlogPage.objects.live().public().filter(locale=locale).search(search_query)
+
+    @route(r"^search/$")
+    def blog_page_search(self, request, *args, **kwargs):
```
tbrlpld, line 15: **Suggestion:** I am not sure we need the `RoutablePageMixin`. We could have the same behaviour on the blog index page directly. This is assuming that the blog index displays all child pages by default and the "search" acts as a type of filter. In that case, we could move the exact functionality to [`get_context`](https://docs.wagtail.org/en/stable/reference/pages/model_reference.html#wagtail.core.models.Page.get_context) method.

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
```diff
@@ -0,0 +1,29 @@
...
+        search_query = request.GET.get("q", None)
+
+        if search_query:
+            self.search_term = search_query
+            self.search_type = 'search'
```
tbrlpld, line 20: **Question:** What other values might the `search_type` have?

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
```diff
@@ -0,0 +1,29 @@
...
+from wagtail.contrib.routable_page.models import RoutablePageMixin, route
+from networkapi.wagtailpages.utils import get_locale_from_request, get_default_locale
+from .blog import BlogPage
+
+
+class MockSearchBlogIndex(RoutablePageMixin, Page):
+
+    template = 'wagtailpages/mock_search_blog_index.html'
+
+    def search_pages(self, locale, search_query):
```
tbrlpld, line 11: **Suggestion:** We could have another argument that would allow passing the base page QuerySet. That way it would be possible to search on a pre-filtered QuerySet (something we are going to want for the research hub I think).

tbrlpld (commented):
This looks good @Pomax. I think we won't need the `RoutablePageMixin`, but it looks good otherwise 👍

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
tbrlpld: Actually, I does not even need the assumption that the "unsearched" index shows all pages. That's just how it's designed in the research hub.

Pomax: it's nice to have a dedicated URL with the word `search` in it, though

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
Pomax: ah, good point that line should not have been there anymore. It'd be useful for things like /tags or /topics for the future blog work, but not really relevant for RH

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
Pomax: We could do that through an arg, but it might make more sense to use django's cache mechanism instead.

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
tbrlpld: Could call the query parameter search 😉

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
tbrlpld: I am not sure I can follow. Can you outline how the cache would be involved here?

TheoChevalier (commented):
> how should searching work given multiple locales?

It would seem very weird to me to surface anything else besides what’s accessible by browsing, so from /fr/ I’d expect results to include pages in French and alias pages in English when there’s no translation available. Would this make sense?

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
Pomax: similar to how we're caching PNI pages so we don't have to constantly hit the db for data that changes probably less than once a day: 

https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/products.py#L111 

paired with

https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/products.py#L1641-L1654

Pomax:
> It would seem very weird to me to surface anything else besides what’s accessible by browsing, so from /fr/ I’d expect results to include pages in French and alias pages in English when there’s no translation available. Would this make sense?

yep, that sounds good to me.

network-api/networkapi/wagtailpages/pagemodels/blog/mock_search_blog_index.py
tbrlpld: Ah ok. Yea that makes sense. 

I think that might actually be best done outside of the "search" method. E.g. in the "get_context" we get the base queryset of  "research detail pages". This can be from DB or cache. Then we pass that base queryset to the "search" method.

We could even have a second method that encapulates the "getting". That should be beneficial for the testing.

PR #8383, opened 2022-03-03
pip compile lock

Pomax:
supercedes a bunch of dependabot PRs, too.


Pomax:
@tbrlpld this mostly just needs a sanity check. Testing passes, a local dev with stage db seems to work fine, and the only huge version jumps are cryptography and django-filter, both of which completely changed versioning schemes so they only "look" like huge version changes.

tbrlpld (approved):

PR #8424, opened 2022-03-08
Feature/8382 profile snippet

tbrlpld:
This PR renames the existing `ContentAuthor` model to `Profile` and extends it. 

During testing we want to ensure two things:

1. Exiting instances of the "content author" model and relations to it remain in place during the migrations

    * Checkout out the `main` branch
    * Download the staging database `inv copy-stage-db`
    * Start the containers `docker compose up`
    * Navigate to a blog page on the frontend, also open the same page in the Wagtail admin
    * Stop the container <kbd>CTRL</kbd> + <kbd>C</kbd>
    * Checkout the feature branch `feature/8382-profile-snippet`
    * Start the container `docker compose up`
    * Run the migrations `inv migrate`. This should pass without errors. 
    * Reload the blog page in frontend and admin. Check that the previously linked author is still linked to the page. In the admin you will see the button text has changed from "content author" to "profile". 
    * Try editing a profile.
    * That should be working without issue

2. When setting up a new environment, the "profile" model instances are populated correctly. 

    * Keep the feature branch checked out.
    * Stop the container if running (<kbd>CTRL</kbd> + <kbd>C</kbd>)
    * Create a new database with fake data `inv new-db`. This should pass without error. 
    * Start the server `docker compose up`
    * Check the "profile" snippets in the Wagtail admin. For each profile, the "name", "tagline" and "introduction" field should be populated. 

***

The migration here might be backward incompatible. I am actually not sure. There is a special migration type `RenameModel`. But yea... I guess code that is referencing the old model name will break. @Pomax I am interested in your take here.

***

Closes #8382
Related PRs/issues #

**Link to sample test page**: https://foundation-s-feature-83-ffxzqj.herokuapp.com/cms/snippets/wagtailpages/profile/edit/1/

## Checklist

_Remove unnecessary checks_

**Tests**
- [x] Is the code I'm adding covered by tests? -> No

**Changes in Models:**
- [x] Did I update or add new fake data? -> Yes
- [x] Did I squash my migration? -> No. Because of the renaming, it feels safer to keep the rename and the extra fields in different migrations with a proper migration dependency definition.
- [ ] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team? 

**Documentation:**
- [x] Is my code documented? -> No
- [x] Did I update the READMEs or wagtail documentation? -> No

PR #8431, opened 2022-03-10
Localize profile snippet

TheoChevalier:
Follow-up from #8424. This snippet now includes fields that should be translated, so I’ve started the work to make it localizable.
I’m a bit struggling with this message when running `makemigrations`:

Capture d’écran 2022-03-10 à 17 24 53

@tbrlpld @Pomax Would you mind having a look please? (Feel free to take over the PR if it’s easier)


tbrlpld:
@TheoChevalier You are seeing this issue because we already have data in the database and now your adding a new field to the models that is not allowed to be empty. This causes an issue because Django does not know what to fill the new field in the existing row with. There is some explanation in the Wagtail docs in how you get around this, specifically for making existing data translatable: https://docs.wagtail.org/en/stable/advanced_topics/i18n.html#making-snippets-with-existing-data-translatable

TheoChevalier:
Nice, thanks @tbrlpld! I remembered there was a migration thingy to perform when translating snippets, but couldn’t find this page back! Looking

TheoChevalier:
@tbrlpld I’ve attempted the migration here https://github.com/mozilla/foundation.mozilla.org/pull/8431/commits/282e05d90939e014440e75e792162053814201a5 but as I move to step 3 and switch to TranslatableMixin, I still get the same error as above. Is there any obvious mistake in the commit above?

TheoChevalier:
@tbrlpld Thanks for the guidance!
I’ve updated the PR accordingly, but still getting the same error unfortunately.

> I was wondering something about making the through models (ArticleAuthor and PublicationAuthor) translatable. Are you making these translatable because you want to select different authors when a page is translated (e.g. the person translating the page is used as the author instead of the original author)?

No, the intent here is just to get the content translated (tagline/intro), and have that snippet translation started if it hasn’t already been translated from a blog post or another page

tbrlpld:
@

> No, the intent here is just to get the content translated (tagline/intro), and have that snippet translation started if it hasn’t already been translated from a blog post or another page

@TheoChevalier In that case, no work is necessary. The `Profile` is translatable already. This is automatic when a model inherits from the `TranslatableMixin` (which `Profile` already does). The only thing that might make sense there is to mark `name` and `image` as sync fields and only the `tagline` and `introduction` as translation field. 

There is no need to make the through models (`ArticleAuthor` and `PublicationAuthor` translatable). These only create the connection from the page to the profile. To make sure that the displayed information on a potential "author page" uses the correct locale of the "profile" would be up to that pages presentation logic. There is no need to create different connections in the admin.

TheoChevalier:
> > No, the intent here is just to get the content translated (tagline/intro), and have that snippet translation started if it hasn’t already been translated from a blog post or another page
> 
> @TheoChevalier In that case, no work is necessary. The `Profile` is translatable already. This is automatic when a model inherits from the `TranslatableMixin` (which `Profile` already does). The only thing that might make sense there is to mark `name` and `image` as sync fields and only the `tagline` and `introduction` as translation field.
> 
> There is no need to make the through models (`ArticleAuthor` and `PublicationAuthor`) translatable. These only create the connection from the page to the profile. To make sure that the displayed information on a potential "author page" uses the correct locale of the "profile" would be up to that pages presentation logic. There is no need to create different connections in the admin.

Thanks, that makes sense! Updated the PR
This was, however, the first thing I tried, and it just doesn’t translate the snippet :(

tbrlpld:
> it just doesn’t translate the snippet

I am not sure I know what you mean by that. The snippet needs to be translated in the "snippets" section of the admin, not "inline" in the page. Also, as of now, the translatable fields are not used anywhere in the "frontend" so it would not be possible to see any changes.

tbrlpld:
I think what is still missing is that the "frontend" (the page that the site visitor sees) needs to make use of the `get_translation(locale)` method and then display the translated information.

TheoChevalier:
> > it just doesn’t translate the snippet
> 
> I am not sure I know what you mean by that. The snippet needs to be translated in the "snippets" section of the admin, not "inline" in the page. Also, as of now, the translatable fields are not used anywhere in the "frontend" so it would not be possible to see any changes.

Hmm, that’s not how we designed wagtail-localize, snippet translation should be triggered when we enable translation for the page, and it should be kept in sync.

Here’s an example on https://foundation.mozilla.org/cms/pages/3785/edit/ with Highlight snippets:

Capture d’écran 2022-03-15 à 17 42 48

Could we do something similar here? This is the desired behavior

tbrlpld:
Ah ok. Thanks for pointing that out @TheoChevalier. That makes sense. TIL: So both, the snippet and the through-model need to be translated. 

Then I guess: Ignore my above comments 😅 Sorry about that. Looks like you were on the right track all along.

tbrlpld:
@TheoChevalier if you need further help on this, could you open up a ticket for us so we can include this in the next sprint planning? This feels like it will take a bit more time

TheoChevalier:
@tbrlpld Thanks, filed #8452

Pomax:
@TheoChevalier this is still showing up in my pending PRs list, should this stay open or can it be closed? =)

PR #8450, opened 2022-03-16
8289 - Replace type.scss with tailwind plugins

### LINKED ISSUE #8289: convert big classes (h1-heading, etc) to tailwind components (source\sass\type.scss)

fessehaye:
Closes #8289 

New Classes:
* `tw-[h1,h2,h3,h4,h5,h6]-heading`
*  `tw-quote`
* `tw-type-accent`
* `tw-body-large`
* `tw-body-small`
* `tw-body`
* `tw-cta-link`

Replacing all file references to the new tailwind classes

Warning: once this PR merging, double check your current branches if you are using classes from the type.scss file and update the relevant classes

Test:
* the style guide but this change affects the whole site so if you have more time maybe check the other pages as well!


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/quote.html
```diff
@@ -1,6 +1,6 @@
 {% load wagtailimages_tags %}
 
-<div class="row section-quote section-padding">
+<div class="row section-tw-quote section-padding">
```
tbrlpld, line 3: I think this one is off. The state list of classes only contains a `tw-quote` but not `section-tw-quote`. This might be a find and replace issue.

source/sass/views/home.scss
```diff
@@ -153,7 +153,7 @@
   .section-news-you-can-use {
     background: $gray-05;
 
-    h2.h4-heading {
+    h2.tw-h4-heading {
```
tbrlpld, line 156: This seems like it was weird before, but maybe we can correct this.

tbrlpld (commented):
@fessehaye Looks good to me outside of two minor things where I am not sure we need to change something.

Also, I am not familiar with Tailwind custom classes (as in `type.js`). So I can't really judge that. 

I have check the site and the style guide and it looks all good to me.

danielfmiranda (approved):
Hey Simon! First and foremost, thanks so much for taking this work on! I can see that there were a ton of files to be updated and I appreciate you updating all of the type fonts to now use tailwind. This is going to help immensely with the effort to move to tailwind.

After taking a look at the file change,the percy comparison, and poking around the review app, everything is looking good to me! 

Approving + thanks and great job! 👍

Pomax (approved):
A herculean effort, and as far as I can tell (with the followup issues already provisioned) everything's looking good. Nice work!

tbrlpld (approved):
@fessehaye This looks good to me. 

I left comments where the element heading level did not match the heading level of the class, but that was the case before this PR already.

PR #8501, opened 2022-03-30
Hardcode admin strings in Research Hub detail page

TheoChevalier:
Follow up from #8448

Sorry if my comments lead to confusion @tbrlpld, we should translate _only_ the public facing strings, and leave the admin-facing content hardcoded.


tbrlpld (approved):
Sorry for causing this extra work. Thanks for the clarification and the fix @TheoChevalier.

PR #8528, opened 2022-04-06
8040 - Fix Privacy Policies generated per product

### LINKED ISSUE #8040: Update Wagtail Factories to include updates from PNI
* ~Include subcategories to factories~
* ~Add subcategories to products~
* ~Possibly include additional products for animation and filter testing~
* Change Privacy Policies generated per product from 5->3 to stop breaking updates to products

fessehaye:
Closes #8040 

Most of this was already done in a previous migration PR (#8252) so I just added the last fix of generating only 3 privacy policies instead of 4.

1. Go to /cms
2. Open one of the buyers guide products
3.  Should only have 3 policies

### DIFF
--- a/network-api/networkapi/wagtailpages/factory/buyersguide.py
+++ b/network-api/networkapi/wagtailpages/factory/buyersguide.py
@@ -238,8 +238,6 @@ def generate(seed):
     for i in range(52):
         # General products
         general_page = GeneralProductPageFactory.create(parent=pni_homepage,)
-        fake_privacy_policy = ProductPagePrivacyPolicyLinkFactory(page=general_page)
-        general_page.privacy_policy_links.add(fake_privacy_policy)
         general_page.save_revision().publish()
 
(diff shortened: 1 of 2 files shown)


Pomax:
This changes quite a lot more than just privacy policy links, through. `for i in range(3)` will run three times (with i=0, i=1, and i=2) and so should lead to three related products, three product updates, and three policy links. If we're seeing four for any of those, then it feels like something is generating data _outside_ of this loop, with the fix being to find where that happens and removing the data generation there.

Pomax (dismissed):
It feels like the actual problem is in a different part of the code, but also looking at #8040 this PR is only one of several tasks to be completed before it can be considered closed?

Pomax (approved):

Pomax (dismissed):
great find! Note that we'll need to update the intergration tests too though, because this changes random value sequence that gets used.

Pomax (approved):

PR #8548, opened 2022-04-11
Switch generation of review app domains to Cloudflare

tomusher:
Previously, on creation of a Heroku review app, a DNS record would be created in Route53 and a relevant Sites entry would be generated.

As we have moved the domain we used (mofostaging.net) to Cloudflare, this script now needs to be updated to create DNS records there instead.


Pomax (approved):

tbrlpld:
Thanks @Pomax

PR #8559, opened 2022-04-13
8477 - Fix width of Article content container

### LINKED ISSUE #8477: [Interactive Report] Fix width of Article content container
The tablet view for content on Article pages is currently too narrow:

It should look more similar to what we have on Blog:

Investigate, design in Figma and create dev implementation ticket.

fessehaye:
Closes #8477 

Link: https://foundation-s-8477-fix-w-wmmqa5.herokuapp.com/en/publication-page-with-chapter-pages/fixed-title-chapter-page/fixed-title-article-page/

### DIFF
--- a/source/sass/wagtail/blocks/article-blocks.scss
+++ b/source/sass/wagtail/blocks/article-blocks.scss
@@ -21,7 +21,7 @@
     max-width: 80%;
     margin: 0 auto;
 
-    @media (max-width: $bp-md) {
+    @media (max-width: $bp-lg) {
       max-width: 100%;
     }
   }


beccaklam (dismissed):
Morning @fessehaye! The tablet/medium view looks good but now the mobile/small and desktop/large screens are broken 😅 
- [ ] Would we be able to make sure the new rules only apply to the medium width screens and the old rules for small and large screens remain the same as what is currently live?

beccaklam:
Hi @fessehaye the width of pages is looking great! Thank you for those changes!
I noticed two things though that I have some questions about:
- [ ] On prod, the font-weight for body main or 'main p' is 300 but on staging it's 400, is there a reason it's different on staging? The font weight should remain 300.
- [ ] Also noticed that on staging the call out is less wide than it is on prod? Do you know why this is happening? The callout block width should remain like how it currently is on prod?
Staging:
Screen Shot 2022-04-14 at 11 53 45 AM
Prod:
Screen Shot 2022-04-14 at 11 54 23 AM

danielfmiranda (dismissed):
Hey Simon! After taking a look at the code and review site everything is looking great so far minus one thing:

image


It appears that the images that used to render on the left and the right of the screen with text wrapping around it are now just displaying as blocks that are left aligned!

I believe the css for this can be found under .article-blocks or .article-page-content, if I remember correctly, there are two classes housed within those called .left and .right which do something to align these images correctly.

Please let me know if you need any help, I would be happy to work together on this through a call!

Thanks and let me know if you have any questions :D

danielfmiranda (approved):
Looks good to me! Thanks Simon 👍

Issue #7771, opened 2021-11-04
Can’t translate/sync petition & banner pages


TheoChevalier:
Translating/Syncing translations for Banner pages crashes.

Example with [this page](https://foundation.mozilla.org/cms/pages/1935/): https://foundation.mozilla.org/cms/localize/update/122/

```
django.core.exceptions.ImproperlyConfigured: The foreign key `wagtailpages.BanneredCampaignPage.cta` was registered as a translatable field but the model it points to `wagtailpages.CTA` is not translatable`
```

Wagtail is complaining about `wagtailpages.CTA` not being translatable, but after @Pomax ’s investigation, it seems if  we do make it localizable, wagtail-localize is not happy about class inheritance.

Note for @TheoChevalier  to remember to sync changes from these pages once the issue is fixed:

- [ ] https://foundation.mozilla.org/en/what-we-fund/
- [ ] https://foundation.mozilla.org/en/campaigns/tell-tinder-its-time-for-meaningful-transparency-into-your-pricing-algorithms/
- [ ] https://foundation.mozilla.org/en/campaigns/election-integrity-and-platform-accountability/
- [ ] https://foundation.mozilla.org/en/campaigns/german-election/

Pages to translate:
- [ ] https://foundation.mozilla.org/en/youtube/

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-357)

Pomax:
[From Slack](https://mozilla.slack.com/archives/CCZGQC1T4/p1636045098002600?thread_ts=1636041695.000700&cid=CCZGQC1T4):

---

Karl:

```
@register_snippet
class Signup(CTA):
    ...

    translatable_fields = CTA.translatable_fields

    class Meta(TranslatableMixin.Meta):
        verbose_name = 'signup snippet'
```
Inheriting `TranslatableMixin.Meta` is not required here because CTA is a concrete model (it's only required when inheriting from an abstract model like `TranslatableMixin` itself)

The locale field gets added on to the first concrete model that inherits it, any models that inherit from CTA will use the locale field on CTA rather than create its own one

---

Pomax:

Ah. Let me take that off the meta and see if things work. And thank you for having a look

So I'm getting this, when running `makemigrations`:

```
You are trying to add a non-nullable field 'locale' to cta without a default; we can't do that (the database needs something to populate existing rows).
Please select a fix:
[...]

PR #8566, opened 2022-04-15
Contingency fix for banner page localization

TheoChevalier:
Related issue #7771

Since #7771 is taking more time than anticipated to get fixed, I’d like to unblock the situation for banner pages.


tbrlpld (changes requested):
Changes look good 👍 Just need to update the branch before merging.

tbrlpld (approved):

PR #8610, opened 2022-04-28
8550 - Added sitemap references

### LINKED ISSUE #8550: SEO | Create a sitemap index file
# Description 

Create a sitemap index file, which should be findable at https://foundation.mozilla.org/sitemap.xml. 

This index would link to the sitemaps for all of the languages that the website supports.

# Acceptance criteria

- [x] /sitemap.xml should be accessible on staging/production/and review apps
- [x]  sitemap should include all 10 languages included on foundation site translations

# Dev tasks

- [x] Generate Sitemap using wagtail [sitemap generator](https://docs.wagtail.org/en/stable/reference/contrib/sitemaps.html)

fessehaye:
Closes #8550 

Adds /sitemap.xml to foundation site that includes all language mappings we support!

### DIFF
--- a/network-api/networkapi/sitemaps.py
+++ b/network-api/networkapi/sitemaps.py
@@ -1,9 +1,9 @@
 # Solution came from Aleksi44 on Github:
 # https://github.com/wagtail/wagtail/issues/6583#issuecomment-798960446
 from django.contrib.sitemaps import views as sitemap_views
-from wagtail.contrib.sitemaps.sitemap_generator import Sitemap
-from django.shortcuts import render
 from django.conf import settings
+from django.shortcuts import render
+from wagtail.contrib.sitemaps.sitemap_generator import Sitemap
 
 
 class CustomSitemap(Sitemap):
@@ -30,4 +30,4 @@ def sitemap_index(request):
     context["languages"] = [lang[0] for lang in settings.LANGUAGES]
     context["domain"] = "https://foundation.mozilla.org"
 
-    return render(request, 'sitemap-index.xml',context, content_type='text/xml')
+    return render(request, 'sitemap-index.xml', context, content_type='text/xml')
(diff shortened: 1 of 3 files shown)


tbrlpld (approved):
I guess this should work but it's definitely not my area of expertise.

network-api/networkapi/templates/sitemap-index.xml
```diff
@@ -0,0 +1,59 @@
...
+
+      <loc>https://foundation.mozilla.org/pt-BR/sitemap.xml</loc>
+
+   </sitemap>
+
+   <sitemap>
+
+      <loc>https://foundation.mozilla.org/sw/sitemap.xml</loc>
+
+   </sitemap>
```
tbrlpld, line 57: Since this is very repetitive and we already have a list of the language codes in the settings, it would be nice to reuse the list we have and loop over it. That way, we avoid having to update to locations for the same reason.

tbrlpld (changes requested):
@fessehaye Just one suggestion regarding the repetition.

network-api/networkapi/sitemaps.py
```diff
@@ -2,6 +2,8 @@
 # https://github.com/wagtail/wagtail/issues/6583#issuecomment-798960446
 from django.contrib.sitemaps import views as sitemap_views
 from wagtail.contrib.sitemaps.sitemap_generator import Sitemap
+from django.shortcuts import render
+from django.conf import settings
```
tbrlpld, line 6: **Suggestion** I know we don't have any checks for it in place, but I find it usually best to keep imports sorted alphabetically.

tbrlpld (dismissed):
@fessehaye No blocker, just a small suggestion regarding import sorting from my end. 

But, there are failing checks that would need to be resolved before merging. I think that's also a formatting thing.

tbrlpld (approved):
Thanks @fessehaye. Looks good 👍

PR #8625, opened 2022-05-04
8586 - Remove newlines from sticky nav titles

### LINKED ISSUE #8586: Investigate why Sticky Nav appears broken on specific report
For some reason the sticky nav appears broken for one report I've found so far. To see bug:

1. Go to: https://foundation.mozilla.org/en/campaigns/uninvited-guests/
2. Scroll past the 'Summary Box'
3. See broken sticky nav

Screen shot:

fessehaye:
Closes #8586 

This is caused by the titles getting <br/> from within the h2 tags. Easiest solution is just trimming additional whitespace. 

I would just check some articles and see if the sticky navigation looks normal!

### DIFF
--- a/source/js/foundation/template-js-handler/publication-summary-bar.js
+++ b/source/js/foundation/template-js-handler/publication-summary-bar.js
@@ -26,7 +26,7 @@ export default () => {
     // update the title based on the header in the article in past by
     const titleObserver = new IntersectionObserver((entries) => {
       entries.forEach((entry) => {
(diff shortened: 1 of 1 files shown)


danielfmiranda (approved):
Hi Simon! Thanks for tagging me on this, after taking a look through the code and testing it out on my local machine it looks good to me! Below is a screenshot of what I see 
image
Which after reading the bug ticket, appears to be what we are expecting

Approved!

PR #8628, opened 2022-05-04
breadcrumb component for research hub pages

### LINKED ISSUE #8498: [RH] Breadcumbs component
# Description

Implement the bread crumbs component for the research hub pages. 

NB: Breadcrumbs will be on all research hub pages however there are some pages that are not created via Wagtail.

# Acceptance criteria

- [x] Breadcrumbs according to the design are displayed on all pages of the research hub. 
- [x] The first page in the bread crumbs is the research landing page. 
- [x] On the research author detail "page", the research author index is part of the bread crumbs (although the detail is actually just a route on the index and not really it's own page).

# Dev tasks

- [x] Create breadcrumb component
- [x] Use built in wagtail functionality to get info of all parent pages for breadcrumbs
- [x] Style component
- [x] Add if statement for the author detail/index page

danielfmiranda:
Closes #8498 
Related PRs/issues #

This PR introduces a new template for the research hub breadcrumbs component, and then adds that component to every research hub page. 

The template does not have much in terms of x/y spacing, as this differs between each page in the research hub. 
My thinking being that when we do the styling tickets of the pages, we can simply wrap this component in tailwind divs of our desired spacing. 

**Steps to test**:
1. Check out this branch on your local machine
2. visit localhost:8000/research, check that the title looks OK.
3. Visit localhost:8000/research/library/, and the title should now have breadcrumbs
4. Click on a research detail link, and the breadcrumbs should still render correctly.
5. Visit the authors/author detail page and check the breadcrumbs.
6. If everything is appearing as expected, testing is complete!


network-api/networkapi/wagtailpages/templates/wagtailpages/research_breadcrumb.html
```diff
@@ -0,0 +1,32 @@
+{% comment %}
```
tbrlpld, line 1: **Suggestion** since this is a fragment and not a "page", I think it would best be in the "fragments" directory. 

I guess the page templates should possibly be moved into the "pages" directory too... but that is out of scope for this PR.

tbrlpld (changes requested):
Thanks for this @danielfmiranda. This looks and works as expected. 

I do have a couple of comments around the implementation though. 

In general, I think it's best to keep templates as "'dumb" as possible and have the pages/models provide the necessary data into the context as needed. See my comment about the page method to return the breadcrumbs. 

Also, it is usually best to pass only the necessary parameters to template fragments, rather than letting the fragment pick up the page template's context and pull values out of it.

network-api/networkapi/wagtailpages/templates/wagtailpages/research_breadcrumb.html
danielfmiranda: Moved! Going to mark this as resolved

network-api/networkapi/wagtailpages/templates/wagtailpages/research_breadcrumb.html
```diff
@@ -0,0 +1,38 @@
...
+    {% endif %}
+    {% endwith %}
+
+    
+    {% if author_profile %}
+      <h1>{{author_profile.name}}</h1>
+    {% else %}
+      <h1>{{page.title}}</h1>
+    {% endif %}
+
```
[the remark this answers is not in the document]
danielfmiranda: Hi Tibor, thanks for the great feedback, that makes sense to me! So I have removed the conditional and the title from the breadcrumb component and put the title back in the pages, thanks!

network-api/networkapi/wagtailpages/pagemodels/research_hub/authors_index.py
```diff
@@ -81,3 +81,12 @@ def get_author_detail_context(self, profile_id: int):
             'author_profile': author_profile,
             'latest_research': latest_research,
         }
+
+    def get_breadcrumbs(self):
```
danielfmiranda, line 85: Hi @tbrlpld! Per our conversation on slack, I have updated the pages to have their own "get_breadcrumbs" method, which gets the pages ancestors, and then trims backwards by reversing the list before trimming. 

One thing I am still working on, is getting the author detail page to render the "authors" breadcrumb while not having the author index page display it.

Also, I was testing this method that works for all pages without having to trim the query, in case you might prefer using this instead. Though I am not sure if calling get_ancestors twice is best practice:
```
    def get_breadcrumbs(self):
        research_landing_page = self.get_ancestors().type(landing_page.ResearchLandingPage).first()
        page_parents = self.get_ancestors().descendant_of(research_landing_page, True)
        breadcrumb_list = [{"title": parent_page.title, "url": parent_page.url} for parent_page in page_parents]

        return breadcrumb_list
```

tbrlpld: You don't need to reverse and then trim. In Python you can trim from the end by using negative indexes. 

```python
>>> ["a", "b", "c"][-2:]
['b', 'c']

```

tbrlpld: Since we actually never want the "author route" (aka. author detail page) in the breadcrumb, but only the author index page if we are on a "author route", I think you can solve this by adding a `inclusive` parameter to the `get_breadcrumbs` method which you pass to the `get_ancestors` method. 

This parameter determines if the page it self is included in the ancestors. When on a "author route" you set this to `True` which will include the author index page in the breadcrumbs.

tbrlpld: **Kudos** I really like this approach of filtering for descendants of the landing page rather than trimming the list  :+1:

tbrlpld: **Suggestion** If you rebase, you will see that there is now a ResearchHubBasePage. You can add the `get_breadcrumbs` method to that so we don't have to keep these 3 page types in sync if anything changes. If you need to invoke the functions differently, you can do that in the `get_context` method of each page.

tbrlpld: **Nitpick** Since we are calling the `get_ancestors` method, lets pass it to a variable called `ancestors` or `page_ancestors` since parents are usually only one level up (but we are dealing with multiple levels).

tbrlpld: **Suggestion** I think it would be great if there were some tests around the breadcrumbs generation. That will help us detect changes of breakage in the future.

network-api/networkapi/wagtailpages/tests/research_hub/test_author_index.py
```diff
@@ -151,17 +151,15 @@ def test_get_author_detail_context_multiple_detail_pages(self):
         )
 
         # author, locale, detail pages.
-        with self.assertNumQueries(3):
-            context = self.author_index.get_author_detail_context(
-                profile_id=self.research_profile.id,
+        with self.assertNumQueries(2):
```
danielfmiranda, line 154: Hi @tbrlpld! 

I updated this because once we updated the method to use "get_latest_research" instead of the get_author_context, the number of queries seem to have dropped from 3 to 2? 

I was wondering if you could double check if this is correct, other than that, no problems adding tests for the breadcrumbs!

tbrlpld (changes requested):
Thanks @danielfmiranda This is coming along very nicely. Good job on the tests :+1:  

Mostly formatting nitpicks.

network-api/networkapi/wagtailpages/tests/research_hub/test_author_index.py
```diff
@@ -150,16 +150,16 @@ def test_get_author_detail_context_multiple_detail_pages(self):
...
-            self.assertEqual(len(context), 3)
-            self.assertIn(detail_page_1, context)
-            self.assertIn(detail_page_2, context)
-            self.assertIn(detail_page_3, context)
-            self.assertNotIn(self.detail_page, context)
+            self.assertEqual(len(latest_research), 3)
+            self.assertIn(detail_page_1, latest_research)
+            self.assertIn(detail_page_2, latest_research)
+            self.assertIn(detail_page_3, latest_research)
+            self.assertNotIn(self.detail_page, latest_research)
```
tbrlpld, line 162: Sorry I missed this before, but the test method should be renamed to reflect what is tested. I guess `test_get_latest_research` should work.

tbrlpld (changes requested):
Thanks Daniel, this is looking good. Just one tiny change I have missed and we are ready to merge.

tbrlpld:
And of course rebase on `main`

tbrlpld (approved):
Nice one. Thanks @danielfmiranda

PR #8660, opened 2022-05-12
8653 - Update PNI logo to be Title Case

### LINKED ISSUE #8653: Update PNI logo to be Title Case
We now consistently use "*Privacy Not Included" in Title Case but the header on privacynotincluded.org/ uses all lower case. 

_Screen cap from prod:_
image

### Todo:
- [ ] Capitalize the `P`, `N`, and `I` in "*Privacy Not Included" in the header

fessehaye:
Closes #8653 

Go to Privacy not included page and check the header!


TheoChevalier (changes requested):
Hm, let’s not use `tw-capitalize`, as this will apply English capitalization rules to all translations and will end up with wrong results in all locales. We should rather update the string itself

fessehaye:
Thank you for the suggestion! Never thought about that, I will change that!

tbrlpld (approved):
Thanks @fessehaye. This looks good. 

Would it be worth extracting this into a component? Seems like we have two navs that are practically the same (maybe I am missing something).

PR #8661, opened 2022-05-12
8656 - Add border to primary button

### LINKED ISSUE #8656: Primary and Secondary buttons are different heights
### Describe the bug
There is a small height difference between primary and seconodary buttons, and it seems to be caused by the border.

### Screenshots
Screen Shot 2022-05-11 at 13 56 16

fessehaye:
Closes #8656 

Examine Changes in `/en/style-guide/#button-tailwind`

### DIFF
--- a/tailwind-plugins/button.js
+++ b/tailwind-plugins/button.js
@@ -51,7 +51,7 @@ module.exports = [
         ".btn": { ...btnDefault },
         ".btn-primary": {
           ...btnDefault,
-          border: "none",
+          border: "2px solid #595cf3",
           color: "#ffffff",
           backgroundColor: "#595cf3",
           borderColor: "transparent",


tailwind-plugins/button.js
```diff
@@ -51,7 +51,7 @@ module.exports = [
         ".btn": { ...btnDefault },
         ".btn-primary": {
           ...btnDefault,
-          border: "none",
+          border: "2px solid #595cf3",
           color: "#ffffff",
           backgroundColor: "#595cf3",
           borderColor: "transparent",
```
danielfmiranda, line 57: Hi Simon! I think this line might be causing some issue. When I take a look at two buttons next to eachother, the primary button is still slightly smaller then secondary. However, when I look at browser tools I see that the border is still set to transparent. When I add 2px solid #595cf3 to through dev tools, they look to be the same size.

So it appears that this is the correction solution but something is causing it to be overriden!

danielfmiranda (dismissed):
Hi Simon! Something seems to be preventing the change from taking place, I left a comment on what I think it might be :D

PR #8727, opened 2022-05-19
8567 - Add debounce on input field

### LINKED ISSUE #8567: Buyers guide search bar crashes on iPhone
### Describe the bug

When I try to use the search bar, I can type the search term on the keyboard, but the first letters only only appear with quite some delay and after 5 or so letters no further ones appear and the the page becomes completely unresponsive.

### To Reproduce

Use an iPhone 6 with iOS 15.3

1. Go to https://foundation.mozilla.org/en/privacynotincluded/
2. Click on search bar
3. Start typing a longer term, eg "galaxy watch 3"
4. See only the first couple letters appear, eg "galax" and notice no further letters appear. I can also not clear the search bar. The page becomes completely unresponsive to the point that I have to close the browser window.

### Expected behavior

I can type search terms and see results below.

### Screenshots

image

### Smartphone (please complete the following information):
 - Device:  iPhone6 
 - OS: iOS 15.3
 - Browser safari

## Acceptance criteria

- [ ] Privacynotincluded search bar should work for adding search bar items on iphones
- [ ] Make sure routing from categories and using history navigation is still working without JS errors
- [ ] The above tests should still work on other browsers/devices

## Dev Tasks

- [ ] Emulate iphone with local dev enviornment
- [ ] Investigate JS error that crashes the entire page
- [ ] Might possibly have disable search bar until images load
- [ ] Might have to add a debounce on the search bar

fessehaye:
Closes #8567 

There was a bug when you refresh the screen with something inside the search bar it would cause an error so that is resolved now.

I do not think safari in general like a bunch of keys on inputs with event listeners compared to other browsers so we are adding an debounce on the input field that delays invoking the provided function until at least 500ms milliseconds have elapsed since its last invocation

Hopefully that resolves all of the iphone issues reported but it is still hard to test in general without a physical device

### DIFF
--- a/source/js/buyers-guide/search/search-filter.js
+++ b/source/js/buyers-guide/search/search-filter.js
@@ -12,6 +12,11 @@ export class SearchFilter {
   constructor() {
     gsap.registerPlugin(ScrollTrigger);
     gsap.defaults({ ease: "power3" });
+    gsap.config({
+      nullTargetWarn: false,
+    });
+    this.allProducts = document.querySelectorAll(`figure.product-box`);
+    this.categoryTitle = document.querySelector(`.category-title`);
 
     const { searchBar, searchInput } = this.setupSearchBar();
     setupNavLinks(this);
@@ -22,9 +27,6 @@ export class SearchFilter {
     [`mousedown`, `touchstart`].forEach((type) =>
       subContainer.addEventListener(type, markScrollStart)
     );
-
-    this.allProducts = document.querySelectorAll(`figure.product-box`);
-    this.categoryTitle = document.querySelector(`.category-title`);
 
     // we want the animation to start when the first eight products images are loaded
     Promise.allSettled(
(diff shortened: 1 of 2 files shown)


source/js/buyers-guide/search/search-filter.js
```diff
@@ -65,19 +67,30 @@ export class SearchFilter {
       );
     }
 
-    const searchInput = (this.searchInput = searchBar.querySelector(`input`));
+    const debounce = (fn, ms = 0) => {
+      let timeoutId;
+      return function (...args) {
+        clearTimeout(timeoutId);
+        timeoutId = setTimeout(() => fn.apply(this, args), ms);
+      };
```
tbrlpld, line 75: **Praise** These scope mixes of variables are still like dark magic to me 😅

tbrlpld (approved):
Thanks for this @fessehaye . It does prevent the previously experienced lock-up. 

However, the input still "feels" slow somehow. And I noticed that fast changes of the values get it close to the lock up. It does not quite lock up, but feels laggy. I was already able to observe the effect on a MacBook, so it is not necessarily an iPhone issue. Might be a general Safari issue as you are hinting at.

PR #8831, opened 2022-06-01
8281 - Remove @set-text-size mixin

### LINKED ISSUE #8281: Remove @set-text-size mixin
All SASS files are using the `@set-text-size` mixin. 

Since the move to Tailwind, we would need that mixin anymore. 

Refactoring the sass files will be a lot easier if we do not have to think about how the mixin transforms even if it makes the sass files temporarily more dry code

fessehaye:
Closes #8281 

Goal was to remove the set-text-size mixin so the scss files will be easier to cleanup in future tickets


danielfmiranda (changes requested):
Hey Simon!

Thanks for taking this work on, everything is looking great to me, though I did notice one thing on percy for the youtube regrets reporter original page:

image
It appears that the line height for the <p> element is a little tight.

However, it appears that we do not even use this template anymore so I am not sure if this is a non-issue? 

If so, please re-request review and ping me and Ill approve. Thanks! 👍

danielfmiranda (approved):
Hey Simon, thanks for taking care of that! 

Approved 😄

PR #8859, opened 2022-06-03
Blogcard multiple topics

### LINKED ISSUE #8770: [Blog] - [Blog Card] - Render multiple Topics
At the moment there is "technically" no limit to how many Topics could be set on a blog page.

 However, even if we selected multiple, we would only render the first one due to this logic on the "**blog_page.html**" "**blog_authors.html**" templates:

`      {% with category=page.specific.topics.first %}`

The ask for this ticket is to add a limit of 2 to the "topics" field, and then render each selected Topic instead of just the first of the choices.

Dev Steps
---
If I were to tackle this ticket, this is how I would do it:

- [x] Add a max_num of 2 to the [Topics field panel](https://github.com/mozilla/foundation.mozilla.org/blob/d949f21f6076be37feb95371868f376468782a2a/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py#L167) in the blog page model.
- [x] Add help_text to the [Topics field panel](https://github.com/mozilla/foundation.mozilla.org/blob/d949f21f6076be37feb95371868f376468782a2a/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py#L167) that tells them they can select multiple using ctrl + click, or, update the panel to be one of the ones that allow the user to add a new one using the + icon.
- [x] Once that is complete, update the template from #8765 to render the 2 topics chosen. They should also be links to their respective "topic" pages.

danielfmiranda:
Closes #8770

This PR implements a small change in the blog card template, where instead of just picking the first item in the blog pages list of related topics, it is instead looping through all of them. 

Design in Figma:
---
image

What this PR introduces:
---
image


tbrlpld:
@danielfmiranda the issue also asks for limiting the number of topics to 2. This does not appear to be part of this PR. Is this on purpose? Is that work done else where?

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -194,7 +194,6 @@ class BlogPage(FoundationMetadataPageMixin, Page):
         FieldPanel('first_published_at'),
         PrivacyModalPanel(),
     ]
-
     translatable_fields = [
```
tbrlpld, line 197: **Nitpick** I think the blank lines between these long lists help readability.

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -288,6 +287,12 @@ def ensure_related_posts(self):
             )
 
     def clean(self):
+
```
tbrlpld, line 290: **Nitpick** A blank line after the start of the function on the other hand is more confusing to me 😅

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -311,3 +316,7 @@ def get_meta_description(self):
                 return truncatechars(text, 153)
 
         return super().get_meta_description()
+
+    def save(self, **kwargs):
+        self.clean()
+        return super(BlogPage, self).save(**kwargs)
```
tbrlpld, line 322: **Question** Is this necessary? This should be handled by Wagtail already. 

I know it's a bit of an odd behavior of Django that it does not automatically validate models on save, but mixing these things now might cause other issues. Typically the call to the `clean` or rather `full_clean` will be done in the view of the form.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_card.html
```diff
@@ -12,13 +12,18 @@
 each blog page's dominant tag)
 
 {% block tags %}
-  {% with topic=page.specific.topics.first %}
-    {% if topic %}
-      {% localized_version topic as localized_topic %}
-      {% get_root_or_page as parent_page %}
+  {% with topics=page.specific.topics.all %}
+    {% if topics %}
```
tbrlpld, line 16: **Suggestion** Since `page.specific.topics.all` is a `QuerySet` you don't need the `if` block. `all` will always return a `QuerySet`. That `QuerySet` may be empty but it will be a `QuerySet`. Iterating over an empty `QuerySet` will just do nothing. 

Usually you only need the `if` before a `for` block if you want to use a `<ul>` around the `<li>` elements that you are rendering in the loop.

tbrlpld (changes requested):
Nice. Glad you found an easy solution to this. 

One question around the save override, the rest is just non-blocking nitpicks.

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
danielfmiranda: No longer using! Going to mark this as resolved

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
danielfmiranda: Updated! Marking as resolved 👍

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_card.html
danielfmiranda: Nice catch Tibor! I think this will help readability. I have updated this and then marked it as resolved

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -24,3 +24,16 @@ def clean_parent(self):
             if self.instance == parent:
                 self.add_error("parent", "A category cannot be a parent of itself.")
         return parent
+
+
+# Max number validation for blog page topics. Since this is a parental m2m relationship,
```
danielfmiranda, line 29: Hi @tbrlpld here is a link to the SA article I was referencing earlier, https://stackoverflow.com/a/37865151, though I am not sure if it is worth adding this link inside the codes comments. So I updated the comment to try and explain why we are doing this as best as possible

danielfmiranda:
Thanks for the great suggestions @tbrlpld! I have implemented your feedback and re-requested review. Thanks!

network-api/networkapi/wagtailpages/forms.py
tbrlpld: @danielfmiranda The SO thread you are linking is about the default Django `ManyToManyField` and revolves around the `ValueError` with message "instance needs to have a primary key value before a many-to-many relationship can be used.". 

The topics are not using the default Django `ManyToManyField`. They are using the `modelcluster.fields.ParentalManyToManyField`. The point of `django-modelcluster` is exactly to address the above mentioned `ValueError` that requires all models to be saved before a many-to-many relation can be established. 

This is not the error we have been seeing though. When you demoed it to me (and I have done some tests myself and been able to reproduce the issue), we have seen the `ValidationError` that we raised in the `BlogPage.clean()` method was not caught and turned into a form error. Instead, our `ValidationError` led to a 500 error, which is unexpected. 

I have raised that with the Wagtail team and am waiting for a response. 

Since the comment and SO thread are about a different kind of error and not the reason we are going with the form, it might be best to remove the comment to avoid confusion in the future.

tailwind.config.js
```diff
@@ -53,7 +53,7 @@ module.exports = {
     ...typePlugins,
     ...glyphPlugins,
     require("@tailwindcss/forms")({ strategy: "class" }),
-    require("@tailwindcss/line-clamp"),
+    require('@tailwindcss/line-clamp'),
```
tbrlpld, line 56: **Suggestion** It looks like the rest of the file is using `"`. I think we should keep that consistent.

tbrlpld (changes requested):
Almost there. Just two tiny suggestions.

tailwind.config.js
danielfmiranda: Updated!

danielfmiranda:
Hi @tbrlpld! Thanks for raising the issue with the Wagtail team and also for the other feedback. I have implemented what you suggested and re-requested review. Thanks!

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -24,3 +24,16 @@ def clean_parent(self):
             if self.instance == parent:
                 self.add_error("parent", "A category cannot be a parent of itself.")
         return parent
+
+
+# Max number validation for blog page topics. We are using a custom form as if
+# we do this validation in the BlogPage models clean() method, it will return the
+# error as a 500, and not as an error in the admin panel.
```
tbrlpld, line 31: I am still confused by this comment. 

How about this: 
```suggestion
# Max number validation for blog page topics. We are using a custom form to avoid
# an issue where the `ValidationError` raised in the page's `clean` method is
# not caught and turned into a form error by Wagtail.
```

tbrlpld (changes requested):
I am still abit confused about the comment, but we are basically there.

tbrlpld:
@danielfmiranda I am also just noticing that on the blog page we are still only showing one of the topics. Is there a separate ticket to address that?

danielfmiranda:
Hi @tbrlpld!

Yes, that work will take place in the "blog page" sub epic, I will take note and include it in the future "style the blog page" ticket. Thanks!

tbrlpld (approved):

tbrlpld:
Thanks @danielfmiranda Sorry for all the back and forth

PR #8862, opened 2022-06-03
New template for search bar, with styles and placeholder text as parameters

### LINKED ISSUE #8778: [Blog] - Global Search Bar
image

The goal of this ticket is to create a new template under the /fragments directory, which houses a "global search bar" that will live on the following pages:

- Blog Homepage
- Author Detail Page
- Topic Detail Page
- Search Page

Expected Behavior:
---
The expected behavior of this component will be the same across all pages. 
The user should be able to type in a search query and click "enter". They will then be sent to the Search page, with the search bar pre-populated and the search results filtered for that query.

We have something similar to this already complete in the research hub, so we should be able to use that for reference.

Please note that the design of the search bar is pretty much the same on all pages, except for the search page where it is longer.

What we can do is have the search bar take 100% of the width of its container, and then position it as needed using outer divs and tailwind classs.

danielfmiranda:
Closes #8778 
Related PRs/issues #

Instead of defining a new template for the blog search bar with nearly identical code as the research hub search bar, I instead created a new template just titled "search_bar.html", where I can then pass in parameters for things such as the icon styling, input styling, and also placeholder text. 

Screenshots:
---

Research Hub using new template, and passing styles and placeholder text as params:

Screen Shot 2022-06-09 at 4 39 22 PM

Blog Index Page using new template, and passing styles and placeholder text as params (not exactly following Figma, this was just meant to show that we can use the template and pass in different styles as needed):

Screen Shot 2022-06-09 at 4 39 50 PM


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/search_bar.html
```diff
@@ -0,0 +1,8 @@
+{% load i18n %}
+<div class="tw-relative">
+  <label for="search" class="tw-sr-only">{% translate 'Search' context 'Button' %}</label>
+  <input type="text" value="{{ search_query }}" name="search" id="search" placeholder="{% translate placeholder_text context 'Search input' %}" class="tw-w-full form-control tw-h-auto tw-p-4 tw-pr-7 tw-rounded-none {{input_styles}}" />
```
danielfmiranda, line 4: Hi @TheoChevalier! I was wondering if you could take a look and let me know if this is proper use of the translate block, as I am passing in the text as a parameter from "research_search_bar" and "blog_search_bar".

Thanks in advance!

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/research_search_bar.html
```diff
@@ -1,8 +1 @@
-{% load i18n %}
-<div class="tw-relative">
-  <label for="search" class="tw-sr-only">{% translate 'Search' context 'Button' %}</label>
-  <input type="text" value="{{ search_query }}" name="search" id="search" placeholder="{% translate 'Search our research…' context 'Search input' %}" class="tw-w-full form-control tw-h-auto tw-p-4 tw-pr-7 tw-border-gray-40 tw-rounded-none" />
-  <button type="submit" class="tw-absolute -tw-translate-y-1/2 tw-top-1/2 tw-right-3 tw-text-blue-80 tw-bg-transparent">
-    <svg fill="none" height="24" viewBox="0 0 24 24" width="24" xmlns="http://www.w3.org/2000/svg" stroke="currentColor"><g stroke-linecap="round" stroke-linejoin="round" stroke-width="2"><circle cx="-14.5" cy="9" r="8" transform="scale(-1 1)"/><path d="m8.682 14.818-7.682 7.682" transform="matrix(-1 0 0 -1 9.682 37.318)"/></g></svg>
-  </button>
-</div>
+{% include "wagtailpages/fragments/search_bar.html" with placeholder_text="Search our research..." input_styles="tw-border-gray-40" %}
```
danielfmiranda, line 1: The reason for this new template is so we do not have to update classes on every template where the research hub search bar is being used, if there is a future design change

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/search_bar.html
tbrlpld: @danielfmiranda I would be surprised if this works, because you can pass any text dynamically to the template. I think you will need to move the translation into the template where the text is actually defined. There you can use `{% translate "My new placeholder text" context "Search input placeholder text" as placeholder_text %}` and then pass the `placeholder_text` variable to the included template, like so `{% include "wagtailpages/fragments/search_bar.html" with placeholder_text=placeholder_text input_styles="tw-border-gray-40" %}`

tbrlpld (changes requested):
@danielfmiranda I think this is going great. I would suggest to make the blog styles the default ones, because that seems to be more in line with the rest of the Mozilla design language. Override what you need for the research hub variant. And I would honestly say, just use the icon direction from the blog design and just keep that for the research hub too. Makes it more consistent.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/search_bar.html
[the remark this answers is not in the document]
danielfmiranda: Great Idea, I have added this!

danielfmiranda:
Hi @tbrlpld!

Thank you for the great suggestions, and for the tip on how to get translated strings working with the search bar.

I have implemented the request to pass the translated strings as a parameter, and also updated the direction of the icon so both templates match. I also updated the master search bar template to have the blog styling as a default, but users can still add custom styles as a parameter.

Thanks again!

tbrlpld (approved):

PR #8863, opened 2022-06-04
Add a slow option for the new-db command

tbrlpld:
I have been continuously experiencing "too many clients" errors
everytime I run `inv new-db`. I have made sure to stop containers before
everytime to no change.

What does apparently help is to make sure that the containers are
stopped between the managememnt commands.

This commit adds an optional flag that can be passed to enable the
stopping of the containers after each management command.

To create a new db in this fashion run `inv new-db --slow`.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -86,27 +86,53 @@ def createsuperuser(ctx):
     print("\nCreated superuser `admin` with password `admin`.")
 
 
-def initialize_database(ctx):
+def initialize_database(ctx, slow=False):
+    """
+    Initialize the database.
+
+    To stop all containers after each management command, pass `slow=True`.
+    """
     print("* Applying database migrations.")
     migrate(ctx)
+    if slow:
+        ctx.run("docker-compose stop")
+
     print("* Creating fake data")
     manage(ctx, "load_fake_data")
+    if slow:
+        ctx.run("docker-compose stop")
+
     print("* Sync locales")
     manage(ctx, "sync_locale_trees")
+    if slow:
+        ctx.run("docker-compose stop")
+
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -86,27 +86,53 @@ def createsuperuser(ctx):
...
-def initialize_database(ctx):
+def initialize_database(ctx, slow=False):
+    """
+    Initialize the database.
+
+    To stop all containers after each management command, pass `slow=True`.
+    """
     print("* Applying database migrations.")
     migrate(ctx)
+    if slow:
```
fessehaye, line 97: I feel like we should be able to wrap this if statement into a function so we do not have to copy and paste this every few lines

fessehaye (changes requested):

danielfmiranda (approved):
Hey Tibor, just tested this out and it works for me! Approved

tasks.py
Pomax: It also feels like this should not be necessary in so many places: each operation runs sequentially, so a better solution would be to first make sure the containers are stopped all the way at the top of new-db, after which things should work fine. (the copy-db script does that in Node code, but should be trivial to port to python)

tbrlpld: I have tried it only at the top with a simple `docker compose down` but that does not solve the issue. It is also not reproducible at which step it happens. 

I have tried a bunch of solutions that would have been more elegant solutions, but none of them worked as I had hoped.

tbrlpld: @fessehaye I think I have found a solution that is a bit more elegant.

fessehaye (approved):

PR #8982, opened 2022-06-16
Initial work for blog authors route

### LINKED ISSUE #8796: [Blog] - [Author] - Author Detail Page Template
The goal of this ticket is to create the "Author Detail" template that will render whenever a user visits an Author's Detail page.

A good example of a page touring and rendering detail pages is the [research hub author detail page](https://github.com/mozilla/foundation.mozilla.org/blob/7f955673295ff6d593bbdb9cc407a531d308747d/network-api/networkapi/wagtailpages/pagemodels/research_hub/authors_index.py#L51)

Acceptance Criteria:
---

- [x] There is a new route, along the lines of "blog/author/<author_name>  that will render the new template when visited.
- [x] Based on the author in the URL params, it should pull their data. At the moment, just rendering name and bio without styling will work.
- [x] Blog cards and blog posts are updated so that the author names are linking to the authors detail page.

Dev Steps:
---

- [ ] Create a new route for this template, using the link above for reference
- [ ] Figure out what behavior we would like for the "/authors/" route, should we render a list of authors? Or, should we redirect back to the blog page.
- [ ] Create a new template for the author detail page
- [ ] Render with minimal styling, the authors name and bio when visiting the author detail page

kevinhowbrook:
Closes #8796
Related PRs/issues #8796

**Link to sample test page**:
- /en/blog/authors/
- /en/blog/authors/[profile_slug]

## Checklist

**Tests**

**Changes in Models:**
- [ ] ~~Did I squash my migration?~~ This contains a manual migration to call save on Profile objects so slugs are populated.
- [ ] ~~[Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~~ I don't _think_ this is relevant as it's new functionality with a new Profile.slug field, no field changes or removals.

**Documentation:**
- [x] Is my code documented?


network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -238,6 +241,39 @@ def entries_by_topic(self, request, topic, *args, **kwargs):
...
+        return self.render(
+            request,
+            context_overrides={
+                'title': 'Blog authors',
+                'author_profiles': author_profiles
+            },
+            template="wagtailpages/blog_author_index_page.html"
+        )
+
+    @route(r'^authors/(?P<profile_id>.+)/', name='blog-author-detail')
```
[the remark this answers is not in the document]
tbrlpld: Ah yea, that is true. We have very similar functionality on the [research hub author index page](https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/research_hub/authors_index.py#L51-L70). There I have also added the sluggified author name, but really only to make the URL look nicer when shared... 

I was actually contemplating this lately if we should just update the profile model with a slug field so be able to simplify these things.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -238,6 +241,39 @@ def entries_by_topic(self, request, topic, *args, **kwargs):
...
+    def blog_author_detail(self, request: HttpRequest, profile_id: str, *args, **kwargs) -> 'HttpResponse':
+        try:
+            author_profile = get_object_or_404(Profile, id=profile_id)
+        except ValueError:
+            author_profile = None
+        if not author_profile:
+            raise Http404()
```
tbrlpld, line 266: **Question** Isn't this the same as only `author_profile = get_object_or_404(Profile, id=profile_id)`?

tbrlpld (commented):
This is looking great @kevinhowbrook! Thank you for this. 

I agree, slugs on the profile would make the URL nicer, more readable. If we include this here, I will make a ticket to update it in the research hub accordingly.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
kevinhowbrook: you're right - I added this before - will fix

kevinhowbrook: I've simplified this in 6ab09672c8f9b526af8e4bf8f7731df8632d968a

I still want to catch the ValueError just in case something like `string` makes its way through as a query, because `get_object_or_404` will raise it [ref](https://github.com/django/django/blob/main/django/shortcuts.py#L80) Another option would be to try casting `int(profile_id)` but I prefer catching what `get_object_or_404` is specifically raising

kevinhowbrook:
@tbrlpld this has been reworked to use profile slugs like we suggested.

network-api/networkapi/wagtailpages/migrations/0030_populate_profile_slug.py
```diff
@@ -0,0 +1,23 @@
+# Generated by Kevin on 2022-06-17 09:55
+
+import django.db.models.deletion
+import modelcluster.fields
+from django.db import migrations, models
+from networkapi.wagtailpages.models import Profile
```
tbrlpld, line 6: **Question** Isn't this one of those cases where we need to use `apps.get_model` rather than importing directly? Or would this work because we are not setting the field directly by are relying on the save method to do the migration for us? 

https://docs.djangoproject.com/en/4.0/topics/migrations/#data-migrations

network-api/networkapi/wagtailpages/migrations/0030_populate_profile_slug.py
```diff
@@ -0,0 +1,23 @@
...
+    ]
+
+    def populate_profile_slugs(apps, schema_editor):
+        """Call save on the Profile snippets to populate
+        slug values"""
+        for profile in Profile.objects.all():
+            profile.save()
+
+    operations = [
+        migrations.RunPython(populate_profile_slugs),
```
tbrlpld, line 22: **Suggestion** I believe we should be passing `RunPython.noop` as the `reverse_code` to allow this migration to be reversed. Otherwise, I think, backward migrations will fail at this step. 

https://docs.djangoproject.com/en/4.0/ref/migration-operations/#django.db.migrations.operations.RunPython.noop

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -238,6 +241,53 @@ def entries_by_topic(self, request, topic, *args, **kwargs):
 
         return IndexPage.serve(self, request, *args, **kwargs)
 
+    @route(r'^authors/$')
+    def blog_author_index(self, request: HttpRequest, *args, **kwargs) -> 'HttpResponse':
+        """If the page is called with /authors/, render a list of Profile
+        objects that have been reference in blog pages.
```
tbrlpld, line 247: **Typo**

```suggestion
        objects that have been referenced in blog pages.
```

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -238,6 +241,53 @@ def entries_by_topic(self, request, topic, *args, **kwargs):
...
+
+        Args:
+            request (HttpRequest)
+        Returns:
+            HttpResponse: Response containing new template and a queryset of all
+            Profiles used as blog authors.
+        """
```
tbrlpld, line 254: **Kudos** For adding a docstring! We should do more of those 😅

network-api/networkapi/wagtailpages/pagemodels/profiles.py
```diff
@@ -41,3 +45,7 @@ class Profile(TranslatableMixin, models.Model):
 
     def __str__(self):
         return self.name
+
+    def save(self, *args, **kwargs):
+        self.slug = slugify(f'{self.name}-{str(self.id)}')
+        super(Profile, self).save(*args, **kwargs)
```
tbrlpld, line 51: **Issue** I believe the missing blank line at the end is failing CI.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html
```diff
@@ -10,7 +10,9 @@
     {% spaceless %}
     {% if show_full_info %}{% trans "By" %} {% endif %}
     {% for author in authors %}
-    <span>{% if authors_length > 1 and forloop.last %}{% trans "and" %} {% endif %}<b class="tw-font-extrabold">{{ author.name }}{% if not forloop.last %}{% if forloop.counter != authors_length|add:-1%},{% endif %} {% endif %}</b></span>
+        <a href="{% routablepageurl page=blog_index url_name='blog-author-detail' profile_slug=author.slug  %}">
+            <span>{% if authors_length > 1 and forloop.last %}{% trans "and" %} {% endif %}<b class="tw-font-extrabold">{{ author.name }}{% if not forloop.last %}{% if forloop.counter != authors_length|add:-1%},{% endif %} {% endif %}</b></span>
+        </a>
```
tbrlpld, line 15: **Suggestion** Sorry, I missed this before. This currently will make the author texts blue and on the last name also have the link on the "and". I think it might be better to wrap only `{{ author.name }}` in the link and also add `class="tw-text-black"` to the anchor tag.

network-api/networkapi/wagtailpages/tests/test_blog.py
```diff
@@ -50,11 +53,65 @@ def test_no_query(self):
...
+        response = self.client.get(path=self.blog_index_url)
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+
+    def test_authors_rendered(self):
+        # Test that the blog index page renders blog authors
+        response = self.client.get(path=self.blog_index_url)
+        self.assertTemplateUsed(response, "wagtailpages/blog_author_index_page.html")
+        self.assertContains(response, self.profile_1.name)
+        self.assertContains(response, self.profile_2.name)
+        self.assertNotIn(self.profile_3.name, str(response.content))
```
tbrlpld, line 95: **Question** Would `self.assertNotContains` not work here?

tbrlpld (changes requested):
Thanks @kevinhowbrook This is looking really nice. I left mostly minor questions.

network-api/networkapi/wagtailpages/migrations/0030_populate_profile_slug.py
kevinhowbrook: Have altered this in 1331e9dd04eb6152f3eb60f9dfb1cc51d77fe48c

Good spot - it will work, be we probably don't want to (can't) rely on the model save method here so I've updated it to mimic what we expect on save

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
kevinhowbrook: fixed with 28ca028809da4157e7410cc11a1baa7f9e67c6bf

network-api/networkapi/wagtailpages/pagemodels/profiles.py
kevinhowbrook: addressed in f331eca11ff6765f9e2e574c803de1f8e96a8a9b

network-api/networkapi/wagtailpages/tests/test_blog.py
kevinhowbrook: fixed in 022ab55e9ea317868ecb197b36199671add2f5f5 - thanks!

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_authors.html
kevinhowbrook: Thanks :) fixed with 626e6c8adc402eee1a2ae92420f915d3070b9563

tbrlpld (approved):
Thanks @kevinhowbrook Looks great. Approved.

PR #8992, opened 2022-06-17
Blog landing featured posts section

### LINKED ISSUE #8782: [Blog] - [Landing] - Featured Posts Component
image

The goal of this ticket is to create a new fragment to house the HTML of the "top 5 featured blog bost" section of the Landing Page. 

In terms of backend updates, there is not much to be done as we already have a "feaured pages" related in blog_index.py :D 

Dev Tasks:
---
- [ ] In blog_index.py, update the limit for "featured_pages" from 2 to 5.
- [x] Update the factory file to now populate 5 instead of 2 as well. 
- [x] Create a new template, probably under the fragments directory
- [ ] Update the template to pull the 5 featured pages (we can probably use blog_index_page.html) for reference
- [ ] Style the component to match the design using tailwind classes as much as possible (We can probably use the new card from #8726 to render the 3 featured blog posts in the middle)

NOTE: If using grid, we can probably just leave a placeholder in the bottom right where the #8683 is going to be

Acceptance Criteria:
---
- [ ] Users can now choose 5 posts to feature instead of 2.
- [ ] The posts render as designed, with 2 of them being wide, and then the 3 in the middle using the new blog card component.

danielfmiranda:
Closes #8782 

This PR introduces the ability for users to pick 5 total "featured posts" for the top half of the blog landing page.

**[NOTE]:** The 5th featured item is looking different than figma as the "featured callout box" has not been implemented yet. Once that has been implemented, the 5th item will appear as designed in figma. But the design in this PR would be the default, for example if the user does not add anything to the featured callout box.

### Screenshots:

Desktop:
---

Screenshot 2022-06-21 at 17-27-11 Blog

Mobile:
---


nancyt1 (approved):

fessehaye (approved):

tbrlpld:
@danielfmiranda I think something got quite broken when you pulled in the latest changes from `main`.




There are a couple of other issue I see. 

- Since we don't force 5 posts to be features, how should we handle less than 5 features? This might be one for @nancyt1 too.
- Why not merge before the call out is merged? Similar question as above, if we don't force it, how can we handle the call out box not being present? 
- The factories should be updated to set featured articles. 
- You screenshot shows the larger features image being a little larger than 2 columns of cards. In the design the large feature is exactly 2 cards wide which means that the gaps are lining up nicely.

danielfmiranda:
Hi @tbrlpld thanks for the great feedback! 

I have added the following:
- Updates to the factory to populate the blog index page with 5 featured posts
- After confirming with @nancyt1, we updated the min/max count of featured posts to 5, so no other designs are needed.
- Updated the index page's featured section to use grid like you mentioned, which now has the 1st and 5th featured post image being the same width as two cards. Thank you for this suggestion as the grid worked great! 


If everything looks good to you, this PR is ready to be merged, and then I can update the grid accordingly to fit the "callout box" in a different PR.


Here is a new screenshot of what they look like now, on mobile breakpoints, they look the same as the screenshots above. 

Screenshot 2022-07-05 at 15-42-15 Blog

danielfmiranda:
Confirming that I coordinated with Xavier Harding to populate this section with the 5 latest blog posts once this change goes live

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -100,10 +100,10 @@ class BlogIndexPage(IndexPage):
     content_panels = IndexPage.content_panels + [
         InlinePanel(
             'featured_pages',
-            label='Featured Posts',
-            help_text='Choose two blog pages to feature',
-            min_num=0,
-            max_num=2,
+            label='Featured',
```
danielfmiranda, line 103: Hi @TheoChevalier!

I was wondering if you could please take a look at how we are setting up this "featured_pages" field to be a SynchronizedField [here](https://github.com/mozilla/foundation.mozilla.org/blob/db8245e82d4736663567cf11900da64876993b19/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L118) and let me know if its missing anything?

Asking because I am experiencing a discrepancy in the ordering between the english page and the french page when trying to translate. 

For example, lets say on the english page I had the posts ordered [1,2,3,4,5], and translated the page in french.

If I reorder the posts in the english page to [5,4,3,2,1] and save, the french page still continues to have the original ordering.

Thank you very much in advance!

TheoChevalier: Hey @danielfmiranda, I’ve found that everything works here, you just need to sync translations after publishing the changes in English:

Capture d’écran 2022-07-07 à 09 50 23
(Menu label is "Sync translated pages")


Then check the "Publish changes now" on the next screen
Capture d’écran 2022-07-07 à 09 50 33


Let me know if you need more help testing this!

danielfmiranda: thanks @TheoChevalier for checking!!

tbrlpld (approved):

PR #9008, opened 2022-06-21
blog index page related topics box

### LINKED ISSUE #8688: [Blog] - [Landing] - Topic Callout Box
### Design:

image

This ticket is to create the new "Related Topics" callout box fragment. Once this is created we can then add it to the "topic landing page" template.

Dev Tasks:
---
Backend:

- [x] Add a new relation to the blog index page, similar to the "FeaturedBlogPages" model in blog_index.py, called something like "featured_topics"
- [x] Make these new featured topics selectable by the user in the CMS. 

Once this is done, we can then move on to the front end:

Frontend:

- [x] Create a new template under fragments directory
- [x] This template should pull all of these "featured_topics" and render a badge that links to their respective "Topic Detail Page"

Acceptance Criteria:
---

- [x] Users can select topics to be displayed in this new component
- [ ] Once selected, they should render as designed in figma.

danielfmiranda:
Closes #8688, #8785

### Screenshots:

Desktop:
---

Desktop (Hover State):
---
image

Mobile:
---


network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -119,6 +133,11 @@ class BlogIndexPage(IndexPage):
 
     template = 'wagtailpages/blog_index_page.html'
 
+    def get_context(self, request):
+        context = super().get_context(request)
```
danielfmiranda, line 137: @tbrlpld  The reason for adding the related topics to the context is because it appears that design would like us to also add "related topics" to the actual topics themselves, and to render those on the topic filter pages.

I figured since this is a routable page, we can tap into the "topic" route and say "if we are filtering based on x topic, grab its "related topics" and update the context so we can render those in the component"

nancyt1 (approved):

network-api/networkapi/wagtailpages/templates/wagtailpages/index_page.html
```diff
@@ -57,6 +57,11 @@ <h3 class="tw-body result-info">
 
     <div class="index-entries row">
     {% else %}
+
+    {% if not filtered %}
+      {% block entry_list_title %}{% endblock%}
+    {% endif %}
+
```
tbrlpld, line 64: **Suggestion** I think we should avoid introducing special cases in parent templates that are only needed for one specific child. You can extended the content of a block by using something like this in the child template:

```html+django
{% block theblockname %}
     <h2>The block did not have a heading before</h2>
    {{ block.super }}
{% endblock theblockname %}
```

See also: https://docs.djangoproject.com/en/4.0/ref/templates/language/#template-inheritance

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -91,10 +94,20 @@ class BlogIndexPage(IndexPage):
     with additional logic to explore topics.
     """
 
+    related_topics = ParentalManyToManyField(
+        BlogPageTopic,
+        help_text='Which topics would you like to feature on the page? '
+                  'Please select a max of 7.',
+        blank=True,
+        verbose_name='Topics',
```
tbrlpld, line 102: **Question** Why change the heading here? The field is called "related_topics" and is displayed on the page as "Related topics". Not sure what the use of the different name here would be.

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -39,3 +39,15 @@ def clean(self):
             self.add_error('topics', 'Please select 2 topics max.')
 
         return cleaned_data
+
+
+# Similar validation to the BlogPageForm, but for the Blog Index page
+# and its field 'related_topics'.
+class BlogIndexTopicCountValidationForm(WagtailAdminPageForm):
```
tbrlpld, line 46: **Suggestion** Since you can only define one form per page type, I think `BlogIndexPageForm` would be the more appropriate form name.

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -39,3 +39,15 @@ def clean(self):
...
+    def clean(self):
+        cleaned_data = super(BlogIndexTopicCountValidationForm, self).clean()
```
tbrlpld, line 48: **Suggestion** I think you don't need to repeat the class name in the `super()` call since Python 3.6. `super().clean()` should be enough.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -91,10 +94,20 @@ class BlogIndexPage(IndexPage):
     with additional logic to explore topics.
     """
 
+    related_topics = ParentalManyToManyField(
```
tbrlpld, line 97: **Nitpick** I like to define the config options (like `subpage_types` and `base_form_class`) before the fields, but that's only a matter of personal taste (so consider this totally optional).

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
tbrlpld: Yup, makes sense. I like to be explicit about what I am adding to the context anyhow.

tbrlpld (changes requested):
@danielfmiranda This is coming a long nicely. I think there might be some conflict with latest changes on main due to the blog card work, but nothing too bad. 

One main thing I am missing at the moment is that we create some related topics in the factories. That would be nice I think.

network-api/networkapi/wagtailpages/templates/wagtailpages/blog_index_page.html
```diff
@@ -27,6 +27,10 @@
   {% endif %}
 {% endblock %}
 
+{% block entry_list_title %}
+  <h1 class="tw-uppercase tw-text-3xl tw-font-bold tw-font-sans tw-text-gray-80 tw-mb-5">{% trans "Latest Blog Posts" %}</h1>
```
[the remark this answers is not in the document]
danielfmiranda: Thanks for the suggestion Tibor! I have added this

network-api/networkapi/wagtailpages/templates/wagtailpages/index_page.html
danielfmiranda: I did not know this and appreciate that you sent me the documentation around this, I gave it a look and found it very useful. Thanks!

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: Good point! Going to remove

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: I agree that this sounds like a good idea! I have added this to the latest commit

network-api/networkapi/wagtailpages/forms.py
danielfmiranda: This does work! So I updated both of the custom form classes

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/related_topics_cutout.html
```diff
@@ -0,0 +1,19 @@
+{% load blog_tags localization i18n %}
+    
+<div class="col-12 col-md-6 col-lg-4 mb-5">
```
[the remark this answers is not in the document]
danielfmiranda: Good catch Tibor, updated!

tbrlpld (approved):

PR #9030, opened 2022-06-29
Add pyright type checking

tbrlpld:
At Torchbox we are currently working toward adding type checking to Python in all our projects. Since this is also on our maintain roadmap the foundation site, I have added PyRight to the project. If you are interested in getting type information on your code, you can run PyRight with `npm run pyright` (or `inv npm "run pyright"`). 

As most of the foundation site's code is not types at the moment this will show plenty errors. I hope over time to add more type annotations to the code I am working on. At Torchbox, we are also working on a codemod script to add ignore comments to the existing type errors so that existing code does not need to be changed immediately and type annotations are added to the code sections when we are working on them anyhow. Once that codemode script is available I would be happy to apply it to this project too, to make the type check output a bit less intimidating. 

@fessehaye @danielfmiranda Since both of you are using VS Code, I believe you should already see type hints in your editor without any changes necessary. 

I have also added a CI step that does run `npm run pyright`. This step fails at the moment, because of the many type checking errors that we still have in the project. To not fail every CI pipeline, I have marked that step as "allow to fail". This means we won't be notified of type errors in CI at the moment. However, with this check already in place, we simply have to remove the `continue-on-error: true` line to fully activate type checking in CI. Of course, we want to wait with activating this till the codemod script has been run to ignore the currently exiting type errors.


danielfmiranda (approved):
Hi @tbrlpld!

Thanks for taking this work on, I just pulled the branch and tested out if pyright was working for me and can confirm that it is! Though like you mentioned at the moment it does return quite a lot of errors as expected.

Approved 👍

tbrlpld:
Thanks for reviewing and updating the branch  @danielfmiranda 🙌

PR #9043, opened 2022-07-01
8775 blog search load more

### LINKED ISSUE #8775: [Blog] - [Search] - Load More Results
image

After #8772 and #8843 are complete, the search page template will return the first 6 blog posts, filtered by a search query or not.

The goal of this ticket is to add a button at the bottom which reads  "Load more results" and when clicked, appends another 6 entries to the results.

Acceptance Criteria:
---

- [ ] If there are more than 6 entries in the list of results, the "load more" button should render
- [ ] When clicked, it should append 6 more results to the list

tbrlpld:
This PR adds the "load more" functionality for the blog search page. Because we are returning pre-rendered HTML that is inserted into the DOM, I had to reorganize some of the templates. 

Also, the JS is updated from a string that defines the URL to the `URL` and `URLSearchParams` APIs. This makes it easier to keep existing query parameters around (as is necessary for the search) without much string manipulation. 

I have provided tests for the backend functionality. To test the frontend manually I found it convenient to use the existing `inv new-db` script and then just duplicate one of the page with a distinct word in the title several times. You are also going to make testing easier for yourself by setting the "page size" on the blog index to 4 (which is the lowest option).

Closes #8775
Related PRs/issues #8775

**Link to sample test page**:  http://localhost:8000/en/blog/search/

## Checklist

_Remove unnecessary checks_

**Tests**

**Changes in Models:**
- [x] [Are my changes backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [x] Is my code documented? Through the tests and comments where necessary. 


tbrlpld:
cc @kevinhowbrook

kevinhowbrook:
> cc @kevinhowbrook

Thanks @tbrlpld code looks fine to me and I've tested this locally by changing the page size.
The only thing I didn't check was when a query is performed, the load more button still works.

One thing I did notice was that I can't see a way to hit the `/search` page via the navigation, not sure if there is a plan to implement that?

danielfmiranda (approved):
Hey @tbrlpld!

Apologies for the delay, I just pulled this into my local machine and tested it out. I can confirm that its working as intended! Approved 👍

danielfmiranda (commented):
Hi @tbrlpld! Just pasting my findings that I mentioned on slack here as well:

It appears that the "featured posts" are being hidden from the search page/results, for example if I add "initial blog post with fixed title" to the "featured posts" section and visit the search page, it will not appear on the landing page or even when I look that title up.

Like you mentioned, I believe this is due to the blog_index [get_all_entries](https://github.com/mozilla/foundation.mozilla.org/blob/857ee7399a555969a5c62df1fcdba0d20a9ca7ed/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L143) function which takes all featured posts and removes them from the list.

If you need any help with this/want to pair through it, let me know! 👍

danielfmiranda (approved):
Hi @tbrlpld! Thanks for all the work you did in this PR, after testing again, I can confirm that everything is working as expected! 

Approved 👍

tbrlpld:
Thanks for the review @danielfmiranda

PR #9061, opened 2022-07-06
[Blog] Localization fixes

### LINKED ISSUE #9060: [Blog] Localization fixes
Improving localization for recently landed changes for blog pages, blog landing page, authors, etc.

TheoChevalier:
Closes #9060

This PR contains multiple small improvements/fixes to the recent features landed related to blog posts.

The biggest change here is it’s enabling localization of blog author snippets, since we’ve added two fields (tagline/introduction) that need to be localized. I’ve followed [these instructions](https://docs.wagtail.org/en/v2.16.2/advanced_topics/i18n.html#translatable-snippets), and managed to get it work with both the test and prod DBs, but would appreciate a closer look from someone used to perforn these migrations to make sure mine is clean.


source/js/components/petition/locale-strings.jsx
```diff
@@ -34,6 +34,7 @@ export default {
     "Yes, I want to receive email updates about Mozilla's campaigns.": `Yes, I want to receive email updates about Mozilla’s campaigns.`,
     "Add my name": `Add my name`,
     "Sign up": `Sign up`,
+    Subscribe: `Subscribe`,
```
TheoChevalier, line 37: This string was introduced as the submit button in the newsletter sign up component added at the end of blog posts, but the actual string + translations were never registered in this file

PR #9079, opened 2022-07-13
9006 - Fix PNI subcategories scrolling on mobile

### LINKED ISSUE #9006: PNI tag buttons not scrolling on tablet or mobile
How to reproduce:
1. Go to privacynotincluded.org on a tablet or mobile device and select a category page where the tags are cut off on the screen
2. Try to swipe to see more tags
3. See error

fessehaye:
Closes #9006 

subcategories can now be clicked and dragged horizontally on mobile. You need enough categories to test this but prod/staging DB should work!


tbrlpld:
@fessehaye Can you provide a URL of an affected page? I am also not sure what the "tags" are. 

I am also curious, why did you have to remove these touch events listeners for this to work?

tbrlpld:
Via Slack: 

> If you use prod db - the mental health app category will work for testing.
>
> The reason why we got rid of the touch events is that I replaced with dragging with JS with this CSS property https://developer.mozilla.org/en-US/docs/Web/CSS/touch-action

tbrlpld (approved):
@fessehaye Thanks for the explanation on Slack and the example link. This works and looks fine to me. 

One question: Why are we using `touch-action` instead of `overflow-x: scroll`?

PR #991, opened 2018-01-24
added django-debug-toolbar

anweshknayak:
fixes #966

PR #9102, opened 2022-07-14
Add django-debug-toolbar

tbrlpld:
This PR adds the Django Debug Toolbar to the project. 

It seems like it used to be installed but then disappeared. I don't find a PR that did that on purpose. So I am not sure why that happened. There were some attempts on doing it that failed for some reason: #5665, #991. 

The Debug Toolbar is an incredibly useful tool for debugging and development of Django projects, so I am adding it back.

The Debug Toolbar is only installed with the dev dependencies. Also, it needs to be enabled with `DEBUG_TOOLBAR_ENABLED` environment variable. I have changed this from the previously existing `DISABLE_DEBUG_TOOLBAR` to `DEBUG_TOOLBAR_ENABLED`, because I find it easier to think in terms of activating something rather than deactivating, especially when it is off/disabled by default and I need to do the enabling. "Not disabled" is just one negation too many. 

***

To get the Debug Toolbar installed create a `.env` file with `DEBUG_TOOLBAR_ENABLED = True` and run `inv catchup`. Load any page and see the "DjDT" icon in the top right.


danielfmiranda (approved):
Hi Tibor! Just tested this out on my local machine and can confirm that it works for me. Thanks for taking this work on! 

Approved 👍

PR #9145, opened 2022-07-26
Add mypy drop pyright

tbrlpld:
The Torchbox internal approach to type checking has changed and the code mode script to deal with the large amount of errors that Pyright produces is unlikely to be created. 

To still allow type checking of Python code I would propose the switch from Pyright to mypy. Mypy is more designed for slow adoption of type checking into existing projects than Pyright. This is apparent in the different number of error that the two tools generate. While Pyright generated 380 errors, mypy only generates 27 errors. This difference is due to a difference in philosophy. Mypy, by default, only checks functions that have a typed signature (meaning that the types of the arguments and the returns are specified in the `def` line). Pyright on the other hand tries to type the entire codebase by leaning heavily on type inference (it tries to guess the types of variables, arguments and returns). Mypy can work similarly, by running it in strict mode, but will be more forgiving by default. I think therefore adding mypy to an existing project is quite a bit easier. 

I have changed the CI step from Pyright to mypy, but still allow it to fail. As part of my personal development effort at Torchbox I will try and fix the existing type errors that mypy is throwing, but that will be a separate PR. 

I have also added `inv mypy` as a quick way to run mypy locally.


danielfmiranda (approved):
Hi @tbrlpld! Thanks for explaining the difference between the two in the description of this PR. This makes sense to me, and after pulling the changes in to my local machine and testing them out, I can confirm that it works! Approved 👍

PR #9085, opened 2022-07-13
Blog index callout box

danielfmiranda:
Closes #8783 

This PR introduces a new block to be used on the blog index page called the "Blog Index Callout Block". 

We made this field required, with a maximum count of 1.

It features the following fields:
- [x] Title
- [x] Body
- [x] Related Topics to be displayed at the top
- [x] Yes/No Value for displaying the headphones icon next to the topics
- [x] Audio field for optional Audio player within the box
- [x] Link Button Text/URL

I have also added this block and fields to the factory, so whenever a user creates a  new local instance of our repo, this field will be set up and populated.

Design in Figma:
---
**Design in Figma**:
image

Screenshots of PR:
---

<table>
  <tr>
    <td>With Audio</td>
     <td>Without Audio</td>
     <td>Without Button</td>
  </tr>
  <tr>
    <td>Screen Shot 2022-07-14 at 4 14 46 PM</td>
    <td>Screen Shot 2022-07-14 at 4 16 15 PM</td>
    <td>Screen Shot 2022-07-14 at 4 17 02 PM</td>
  </tr>
 </table>

**Desktop (XL):**
Screenshot 2022-07-14 at 16-19-02 Blog

**Desktop (L):**
Screenshot 2022-07-14 at 16-19-17 Blog

**Desktop (M):**

**Mobile:**

PR #9157, opened 2022-07-28
Localize blog callout

TheoChevalier:
Related PR #9085

Localize blog index callout box


danielfmiranda (approved):
Thank you very much @TheoChevalier! Looks good to me 👍

PR #9161, opened 2022-07-28
Run CI on all PRs regardless of branch

tbrlpld:
When ever we are making PRs between branches, it is still a good idea to
make sure that the code passes CI. Otherwise this can lead to bad merges
that break other branches.

This is particularly important when we work on longer projects with
integration branches (e.g. PNI).


danielfmiranda (approved):
Hi @tbrlpld,

This makes sense to me, especially since we have a lot of work coming up that is probably gong to be out of an integration branch such as PNI.

Approved!

PR #9163, opened 2022-07-29
Remove the `close` keyword from the PR template

tbrlpld:
Our tickets are not closed only because the PR gets merged. We still
need to perform QA on staging and confirm that all the acceptance
criteria are met.

Therefore, we should not suggest adding the `close` keyword. Rather,
link the issue by normal mention (that line still exists).

This commit also contains tiny cosmetic changes to the PR template.


danielfmiranda (approved):
Hi @tbrlpld!

Thanks for taking this on, this makes sense to me especially as we do not do QA testing in review apps at the moment.

Approved!

mtdenton (approved):

fessehaye (approved):

PR #9165, opened 2022-07-29
Fix the createsuperuser task

tbrlpld:
When setting up the `pyrun` task I introduced another layer through
which all the strings are passing. That broke the `createsuperuser` task
that was again trying to pass raw Python strings as a shell string...
which was already inside of the tasks.py and had to pass through the new
pyrun methods...

This was just one layer of string escaping too many.

I have now moved the actual creation of the admin to a management
comand and now we only need to run that command with a the task.

### DIFF
--- a/network-api/networkapi/management/commands/create_admin.py
+++ b/network-api/networkapi/management/commands/create_admin.py
@@ -1,0 +1,14 @@
+from django.contrib.auth import models as auth_models
+from django.core.management import base as management_base
+
+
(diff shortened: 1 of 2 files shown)


danielfmiranda (approved):
Hi @tbrlpld!

Great catch, I just tested this PR out and can confirm that it was originally broken, but the fixes introduced do solve the issue.

Thanks for taking this work on! Approved 👍

Issue #9170, opened 2022-07-30
[PNI] - "In the Press"/Popular Component [Backend]


fessehaye:
[Figma Component Link
](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3240%3A32961)

![image](https://user-images.githubusercontent.com/2374073/181871499-121b0f39-3899-4c12-8bc6-cf3d2f9cc76e.png)
![image](https://user-images.githubusercontent.com/2374073/181875714-971acfcb-420e-4b17-9769-b5330ba59d5e.png)

In the Press/Popular Component will only be used for Homepage 

Acceptance Criteria:
* [x] Should be able to add up to 3 news updates (product updates) via CMS for "in the press" on homepage
* [x] Should be able to add up to 3 articles via CMS for "Popular" on homepage
* [x] Both new fields should be orderable

Dev Tasks: 
* [ ] Add CMS field that allows you to add related PNI articles for in the Press (limit 3)
* [ ] limit to only adding PNI articles with "In the Press" tag
* [ ] Add CMS field that allows you to add related PNI articles for "Popular" (limit 3)
* [ ] Update wagtail factories to included new CMS changes
* [ ] Factories updated to reflect new CMS changes

tbrlpld:
@cdanfon This is ready to test on the preview app: https://foundation-pni-2022.herokuapp.com/cms/pages/1444/edit/

cdanfon:
All ACs satisfied, good to go @tbrlpld !

PR #9209, opened 2022-08-08
9170 refactor orderable relations

tbrlpld:
# Description

This PR does not add any functionality, it only reorganizes the implementation of a common pattern I have observed. This is  in preparation of #9170 which is going to add additional relations. 

Often we want to create related objects on pages, and we want them to be orderable. This means we create through model between the parent page and the related object. In the past we had to make sure that the templates handle this case and pull the actual model of interest out of the through model. 

To simplify this, I have been using methods on the page that return the related models directly. But this also has been  proven repetitive. I have now created a `OrderableRelationQuerySet` that can be used as the manager for the through models and adds the retrieval method `related_items` which returns the items that the page is related to. 

Link to sample test page: https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/
Related PRs/issues: #9170

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/article_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/article_page.html
@@ -11,14 +11,14 @@
 {% endblock hero %}
 
 {% block guts %}
-  {% with content_category_relations=page.content_category_relations.all  %}
-    {% if content_category_relations %}
+  {% with content_categories=page.content_category_relations.related_items  %}
+    {% if content_categories %}
       <div>
         <strong>Categories</strong>
 
         <ul>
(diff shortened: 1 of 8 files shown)


tbrlpld:
cc @kevinhowbrook

danielfmiranda (approved):
Hi @tbrlpld! Thank you very much for taking on this work, I too have noticed that we were starting to use orderables repeatedly, and this seems like a great solution to keep things simpler.

Approved!

kevinhowbrook (approved):
Nice cleanup :+1:

tbrlpld:
Thanks for reviewing @kevinhowbrook and @danielfmiranda 👍

Issue #9168, opened 2022-07-30
[PNI] - Newsletter CTA stylings


fessehaye:
[Desktop comp](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3406%3A38682)

Use existing newsletter CTA react component as a base! Does not require back end additions!

Dev Tasks:

- [x] Style Newsletter CTA block
- [x] SVG asterisk background
- [x] Gradient background
- [x] Rounded corners

Acceptance Criteria:
* [x] PNI Article Page Styled Correctly

![image](https://user-images.githubusercontent.com/2374073/181864801-8dc5f04f-2f56-4690-b6c3-24eea32e84a6.png)

cdanfon:
Hi @fessehaye Thanks for adding this to the PNI milestone. Does this ticket have any dependencies? For example, does it makes sense to complete this one before other tickets in the PNI milestone or is it blocking anything?

@tbrlpld for reference

cdanfon:
Hey @fessehaye Does this ACs refer to the ability to add a CTA to an PNI article page as a Wagtail user?

Sorry I'm a bit confused about what I need to do to be able to test is working!

fessehaye:
https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/articles/this-is-a-test/

cdanfon:
All good @fessehaye, thanks for adding the test link

PR #9213, opened 2022-08-10
9168 - Newsletter CTA stylings

fessehaye:
Related PRs/issues: #9168 
Adding PNI newsletter styling in `pni_newsletter_box.html`

since this block doesn't exist in any current page type yet, the best way to test this is just go on `blog_page.html` and replace the existing newsletter template on L68 with `{% include "fragments/buyersguide/pni_newsletter_box.html" %}`


danielfmiranda (approved):
Great Job Simon! This looks amazing 👍 also tested out that commands still work after updating the tasks.py file and can confirm that works as well.

Approved!

Issue #9190, opened 2022-08-02
Blog Index page's "Latest Blog Posts" section has extra item


danielfmiranda:
### Describe the bug
Now that we have added the "Related Topics" box to the "Latest Blog Post" section on the Blog Index page, it is pushing one extra item to a new row:

This issue persists even after clicking the "Load More Results" Button

### To Reproduce
1. Go to the Blog Index Page
2. Scroll down to latest blog posts
3. Note the extra item

### Expected behavior
If possible, the initial list of items should end with a full row, instead of a row of one.

### Screenshots

![Screenshot 2022-08-02 at 16-21-40 Blog](https://user-images.githubusercontent.com/18314510/182493143-2d9baf38-a4d9-4b18-a530-03baf92d1f94.png)

danielfmiranda:
Closed via #9214

Issue #12, opened 2017-01-11
Research Django interfaces


kristinashu:
Do a web search to find popular/viable django interface bolt-ons/etc that make working with Django easier from a user/admin point of view.

Evaluate selected django interface(s) from a UX perspective.

Compare django to wordpress (UI, number of steps to complete a task…)

kristinashu:
We can use the use cases gathered in https://github.com/mozilla/network/issues/9 and see how these might work in django.

kristinashu:
https://getgrav.org/ looks great!

Pomax:
There's also https://www.django-cms.org/en/

kristinashu:
Yes! That's a good looking one.

kristinashu:
Opened in our repo https://github.com/mozilla/missy-elephant/issues/30

PR #9214, opened 2022-08-10
removing extra item on initial load in for blog index page

danielfmiranda:
Related PRs/issues: #9190
# Description

In this PR I introduced a if statement on the BlogIndexPage model, that will remove the extra item from the initial entry set. 

The if statement checks:
- If the page is filtered (meaning we wouldnt be on the landing page with the callout box, meaning no update is necessary)
- If the page size is set to 12 or 24 (the page sizes where this problem is occuring)

I only check for 12 or 24 because these are the page sizes that are divisible by 3, which is what the columns are set to. 4 and 8 always have extra items on the row, though the page size of 8 loads in nicely since the empty space is taken by the callout box. 

# Screenshots:

 **Screenshot of bug:**

Screen Shot 2022-08-08 at 1 09 22 PM

**Page Size 4:**

Note that the initial entry count is unaffected, as page size of 4 or 8 are going to have overflowing cards anyway.
Screen Shot 2022-08-08 at 1 09 22 PM

Screen Shot 2022-08-08 at 1 09 22 PM

**Page Size 12:**

Note that the initial entry count is - 1 so there is no extra card, and then loading more cards renders 4 rows as expected
Screen Shot 2022-08-08 at 1 09 22 PM
Screen Shot 2022-08-08 at 1 09 22 PM

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -155,6 +155,13 @@ class BlogIndexPage(IndexPage):
     def get_context(self, request):
         context = super().get_context(request)
         context["related_topics"] = self.get_related_topics()
+
+        if not hasattr(self, 'filtered') and self.page_size == 12 or self.page_size == 24:
```
danielfmiranda, line 159: Making sure we are not on a filtered page, which do not render the callout box at all.

Also making sure that the page size is 12/24, which are the only page sizes divisible by 3 and having this issue.

network-api/networkapi/wagtailpages/pagemodels/index.py
```diff
@@ -202,6 +203,9 @@ def generate_entries_set_html(self, request, *args, **kwargs):
                 pass
 
         start = page * page_size
+        if 'entries_offset_for_callout' in context:
+            start = start - 1
```
danielfmiranda, line 207: If the BlogIndexPages initial entries were offset by the callout box (for example setting initial entry list size from 12 -> 11), we then need to update the starting point of the next page.

If we dont, the next page is going to start at 13, leaving entry #12 out of the rendered list.

tbrlpld: **Question:** Do we use callout boxes in other index pages than the blog index? If not, we should not make this change on the index page, but rather keep it to the blog index.

tbrlpld: Moving this to the blog index would also mean you don't need to pull in the context that you don't really need here.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
tbrlpld: It would be good to move this check into a helper method that clearly states what this check is supposed to mean. Something like `is_showing_callout_box_in_entries`. 

Also, we probably don't want to show the callout box if there are not topics.

network-api/networkapi/wagtailpages/pagemodels/index.py
tbrlpld: I think it would be good to add some tests around this to cover all the edge cases, but I think in general this approach (of adjusting the start point) is fine.

tbrlpld (dismissed):

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: Great Idea! Updated and also added the check for the topics

network-api/networkapi/wagtailpages/templates/wagtailpages/blog_index_page.html
```diff
@@ -75,7 +75,7 @@ <h2 class="tw-hidden tw-uppercase tw-text-3xl tw-mb-0 tw-font-bold tw-font-sans
     <ul id="index-entries" class="tw-grid medium:tw-grid-cols-2 large:tw-grid-cols-3 tw-gap-x-5 tw-gap-y-4 tw-list-none tw-divide-y tw-divide-gray-20 tw-px-[15px] medium:tw-divide-y-0">
       {% include "./fragments/entry_cards_item_loop.html" %}   
 
-      {% if not filtered %}
+      {% if not filtered and related_topics %}
```
danielfmiranda, line 78: If there are no topics set, do not render the topics box

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -306,6 +313,68 @@ def extract_topic_information(self, topic_slug):
             'topic': topic_object
         }
 
+    def get_related_topics(self):
+        related_topics = self.related_topics.all()
+        return related_topics
+
+    # Helper function to show if entry list is offset by topic box
+    def is_showing_topics_box_in_entry_list(self):
```
danielfmiranda, line 321: New helper function as suggested from feedback

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -306,6 +313,68 @@ def extract_topic_information(self, topic_slug):
             'topic': topic_object
         }
 
+    def get_related_topics(self):
```
danielfmiranda, line 316: Moved this topic up here because I thought it would make sense to have all the routes at the end of the file.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -306,6 +313,68 @@ def extract_topic_information(self, topic_slug):
...
+        if (
+            not hasattr(self, "filtered")
+            and self.get_related_topics().count()
+            and (self.page_size == 12 or self.page_size == 24)
+        ):
+            return True
+        else:
+            return False
+
+    @route('^entries/')
```
danielfmiranda, line 331: Per request from feedback, I have moved the function for rendering new entries here, and simplified a little bit by removing any unneeded logic + adding the check for whether or not we need to update the index by 1.

danielfmiranda:
Thanks for the great feedback @tbrlpld!

Per request I have:
- Added some tests
- Created a helper function for checking whether or not the list is offset
- Updated blog index to load its new entry pages, instead of relying on index.py's

fessehaye (approved):
tests look good to me

Issue #9187, opened 2022-08-02
Only show English topics in admin panel


TheoChevalier:
I’m closing https://github.com/mozilla/foundation.mozilla.org/issues/7437 and filing a new issue, because it’s been a while and there has been changes, starting with categories renamed as topics.

## The issue:
- Localized topics are showing up in the admin panel in addition to the English ones (Notice "Mozilla explique", the French translation of "Mozilla Explains")
<img width="386" alt="Capture d’écran 2022-08-02 à 18 20 39" src="https://user-images.githubusercontent.com/1294206/182424061-db721c5d-d093-4b70-bdb6-64c6608dacb4.png">

## The fix:

- Moving from checkboxes to snippet chooser, that now have locale selectors
<img width="891" alt="Capture d’écran 2022-04-18 à 16 52 23" src="https://user-images.githubusercontent.com/1294206/163826614-b5642ed2-4e55-4f4f-a8be-7c5ad0133884.png">

- Also note that there is a workaround/hack available here too: https://github.com/wagtail/wagtail-localize/issues/534
I’d suggest exploring this, in case it can get implemented faster so that we can finally translate topics.

TheoChevalier:
I’ve implemented the workaround in https://github.com/mozilla/foundation.mozilla.org/issues/7437, which may even be a better solution, since localized topics are not just hidden behind a language selector, they can’t get picked at all

simont-cr:
Closing as obsolete issue as part of 2024 Clean-up.

PR #9220, opened 2022-08-11
Remove non-English topics from Blog index admin

TheoChevalier:
Linked issue #9187 

Hey @danielfmiranda, sorry I realized afterwards that topics were recently added to the Blog index page as well, so we will want to hide the non-EN entries here too


danielfmiranda (approved):
Great catch @TheoChevalier! Approved 👍

PR #9221, opened 2022-08-11
Add deploy to heroku gitub action 

fessehaye:
Test to see if there is a way to deploy to heroku via github actions

Should be manually triggered for testing purposes then changed to trigger on main git push


fessehaye:
not needed anymore

Issue #9104, opened 2022-07-14
[PNI] - Create PNI Article Page [Frontend]


fessehaye:
[Desktop comp](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3406%3A38682)
[Mobile Comp](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3471%3A46724)
[Tablet Comp](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3738%3A45437)

* This can basically be a copy of the blog page type
* Authors – already exists on the blog page
* adding related reading are apart of a seperate ticket that will be added later

Dev Tasks:

- [x] Style Article page
- [x] Include mobile styling
- [x] Include tablet styling
- [x] Add Newsletter CTA

Acceptance Criteria:
* [ ] PNI Article Page Styled Correctly
* [ ] PNI Article Page Works for Mobile and Tablet
* [ ] Newsletter CTA is visible

![Article Page Type.jpg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/126ebb64-31c7-4a95-8d01-d3c1519c1065)

tbrlpld:
@fessehaye The newsletter CTA is a frontend thing. Might need a separate ticket for that.

fessehaye:
* Related reading card is the same as homepage
* Current blog is MVP for stylings

fessehaye:
https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/articles/this-is-a-test/

cdanfon:
Hey @fessehaye I've checked the link, thanks for adding it. 

A couple of notes: 

1. In the desktop version, on Chrome, looks like there's too much padding-top & bottom compared to the Figma design. 
2 In the mobile version, using inspect tool and Samsung Galaxy S8 the shape of the box is horizontal instead of landscape as per Figma file. 
3. In the table version, using inspect tool and iPad Air, there's too much padding top & bottom.

See attached examples: 

![desktop.png](https://images.zenhubusercontent.com/624c0bfa994c73e54a68130b/2e6cdfdc-00e6-4f61-a08a-950b6420f852)

fessehaye:
@nancyt1 reviewed the updated changes and approved so I will move this to the next column 

[...]

PR #9238, opened 2022-08-16
9104 - Create PNI Article Page [Frontend]

fessehaye:
Related issue: #9104 
Frontend stylings for PNI article
*related readings are excluded due to it being a part of a later ticket

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/blog_authors.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/blog_authors.html
@@ -1,0 +1,45 @@
+{% load wagtailimages_tags static i18n localization homepage_tags wagtailroutablepage_tags %}
+
+{% with authors=page.author_profile_relations.related_items %}
+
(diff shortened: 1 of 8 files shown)


danielfmiranda (dismissed):
Hey SImon! 

Thanks for tagging me on this, after pulling this into my local machine and testing, everything is looking pretty good to me! I just have 3 small suggested changes to make the template match figma a little closer. The first two can be found in this review, the third is just updating this line https://github.com/mozilla/foundation.mozilla.org/blob/e27fe16eeb0ee62eccef8a50a0e0de60647b722e/source/js/components/join/join.jsx#L292 from `24ch` to `30ch`

Before:
Screen Shot 2022-08-16 at 1 54 27 PM


After:

Screen Shot 2022-08-16 at 1 58 00 PM

danielfmiranda (approved):
Thanks for the updates Simon! Approved 👍

Issue #9172, opened 2022-07-30
[PNI] - Popular articles component - [Frontend]


fessehaye:
[Figma Component Link
](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3240%3A32961)

![image](https://user-images.githubusercontent.com/2374073/181875410-9550836f-64c0-4dd2-92f3-e4c68a40394a.png)

Popular articles Component will only be used for Homepage 

Dev Tasks: 
* [x] Create Wagtail template for Popular block
* [x] Allows max 3 articles
* [x] include article thumbnail
* [x] All Articles leads to Editorial Content index page

Acceptance Criteria:
* [x] Component is created and ready to be used on homepage
* [x] "All Articles" - leads to content index page

tbrlpld:
The related backend ticket #9170 is a p1 and currently being worked on, so I presume this is also p1.

fessehaye:
https://foundation-pni-2022.herokuapp.com/en/privacynotincluded

cdanfon:
Hey @fessehaye I had a look at this but doesn't seem to be styled on my end? 

<img width="965" alt="Screenshot 2022-08-26 at 11 51 50" src="https://user-images.githubusercontent.com/102978346/186888443-f4a132b4-67b2-4197-bd85-d3a628df7787.png">

I cannot see the 'All articles ->' link either

Am I doing something wrong? I've checked @danielfmiranda comments on #9242 and looks fine on this machine. 

Thanks!

tbrlpld:
@cdanfon Your browser might have cached the old non styled version of the page. Does it show up styles when you visit is with a private browser window?

cdanfon:
Hey @tbrlpld I've just checked in incognito but getting the same result?

![Screenshot 2022-08-30 at 12.21.54.png](https://images.zenhubusercontent.com/624c0bfa994c73e54a68130b/1040a386-1709-47cc-870c-732b64dbf915n)

cdanfon:
Also checked on safari and the same:

![Screenshot 2022-08-30 at 12.25.13.png](https://images.zenhubusercontent.com/624c0bfa994c73e54a68130b/73bbe561-f432-48df-8157-4bd684a4722d)

tbrlpld:
@cdanfon I just notice that it's the same for me too. I don't see any styling on the preview app. (It does work locally)

[...]

PR #9242, opened 2022-08-17
9172 - popular articles frontend component

fessehaye:
Related: #9172 

Added some templates to make some of the frontend for related articles easier in the future

Changes are visible from PNI homepage


danielfmiranda (approved):
Looks great @fessehaye! 
Here is a screenshot of what it looks like on my local machine while testing: 
image


Amazing job, approved! 👍

Issue #9086, opened 2022-07-14
[PNI] CTA box component [Frontend]


fessehaye:
[Link to Figma component](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3231%3A33672)

![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/885100bc-a55d-4d52-b7e2-a73eccbf9032)

Create a wagtail component for the upcoming event promotion.

Dev Tasks:
* [ ] Title 
* [ ] Description
* [ ] Emoji PNG
* [ ] Apply Background gradient
* [ ] Link & Link text

Acceptance Criteria:

To be tested as part of  #9188

cdanfon:
Design is being updated, @beccaklam to update the ticket once design is complete and for @fessehaye to re-groom the ticket then,

tbrlpld:
Backend ticket: #9188

tbrlpld:
@fessehaye You mentioned in planning you might have some references for me regarding the gradient background. Would love to hear what you have 😄

tbrlpld:
@cdanfon I can not release this to the preview app because a bad PR was merged that breaks the PNI homepage (it's not pushed to the preview app yet).

cdanfon:
Thanks @tbrlpld 

Do we know when the next deployment is likely to happen?

tbrlpld:
@cdanfon That should happen after the issue is fixed. No idea when that will be worked on.

cdanfon:
Hey @tibor is this now unblocked based on this [convo](https://torchbox.slack.com/archives/CG07HH4G5/p1662726609164439?thread_ts=1662721872.105359&cid=CG07HH4G5)?

tbrlpld:
@cdanfon The frontend component can now be tested in the styleguide on the preview app: https://foundation-pni-2022.herokuapp.com/en/style-guide/#buyersguide

cdanfon:
Hey @tbrlpld thank you!

It looks good to me, the only thing to mention is that the Figma file has 2 variations of an icon: heart & eyes - see below:

<img width="960" alt="Screenshot 2022-09-12 at 13 22 45" src="https://user-images.githubusercontent.com/102978346/189652485-d8c214e6-3966-44c7-a8af-b43aa1490d01.png">

Is it worth including the eyes too? 
@beccaklam thoughts?

[...]

PR #9326, opened 2022-09-07
9086 CTA box component

tbrlpld:
# Description

This PR adds the template for the Buyer's Guide call to action component. 

Because the backend is still in the works, I opted to add the component to the styleguide (which I guess is a good idea anyhow to document our components). 

The component does not set it's outer width. This makes the component flexible for different use cases. The parent element / including template should define the width of the element, e.g. through a wrapping element if necessary. 

Link to sample test page: http://localhost:8000/en/style-guide/#buyersguide
Related PRs/issues: #9086 

# Checklist

**Tests**
- [ ] ~~Is the code I'm adding covered by tests?~~ No.

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? Through the style guide.


tbrlpld:
For visibility @danielfmiranda

source/sass/buyers-guide/bg-main.scss
```diff
@@ -65,6 +65,12 @@ html {
   background: $pni-product-image-background;
 }
 
+.bg-call-to-action-content,
+.bg-call-to-action-content > .rich-text > p {
+  // Overriding the default styling for a paragraph.
```
fessehaye, line 70: If you wanted an tailwind only solution, you could always use arbitrary variants! https://tailwindcss.com/blog/tailwindcss-v3-1#arbitrary-values-but-for-variants we would need to update tailwind but it would be an easy upgrade

tbrlpld: Ah awesome. I was looking for that but did not find it in the docs for some reason. I will update since this is much easier for this onetime use 👍

Issue #9253, opened 2022-08-23
YTR 2022  | Top hero intro section and moving stats upwards


cdanfon:
Link to design - https://www.figma.com/file/XtEAabujYbKaP5mLnCtvVS/YouTube-User-Control-Report-Landing-Page?node-id=0%3A1

ACs:

- [ ] Copy matches content in this [document](https://docs.google.com/document/d/1-QU-Yv17WjAa788tNFmAxWbTJT7NGY2snY5qPSEGlWk/edit)
- [x] Content is broken down as per the new design
- [x] Stats appear at the end of this section and continue to be animated and match the new colour
- [x] Two illustrations to appear on each side of the stats as per the new design 

**TO DO**: clarify the use of illustration on tablet & mobile

cdanfon:
@beccaklam to confirm where the button should be getting the content (pdf) from.

cdanfon:
@kristinashu This is now ready for MoFo to review, please note my comments on Content.

cdanfon:
I can't satisfy AC #1 as the content has been changed since Steve completed this ticket. 

Content changes would need to be re-factored towards the estimate.

kristinashu:
This looks amazing, thank you so much! And sorry for those last minute copy changes.

There was also a stakeholder request to make the subtitle stand out more. Nancy mocked it up (see screep cap from Figma below). Should this be done here or in a new ticket? 
<img width="709" alt="image" src="https://user-images.githubusercontent.com/1068783/188724300-4a31c191-36a0-450b-b1dc-d6217a80b2d2.png">

cc @nancyt1 in case you see anything else

cdanfon:
NB Removed AC the following AC which will be done as part of a separate ticket: 

- [ ] Copy matches content in this [document](https://docs.google.com/document/d/1-QU-Yv17WjAa788tNFmAxWbTJT7NGY2snY5qPSEGlWk/edit)

cdanfon:
Hey @kristinashu if it's a one-off change we can update it within this ticket, @siimonevans are you OK to pick up this small change request please? 

Thanks!

PR #9323, opened 2022-09-07
YouTube Regrets l10n fixes

TheoChevalier:
I did a first pass on localization for the yt regrets 2022 landing page

PR #9328, opened 2022-09-08
Youtube Regrets 2022 - Top hero intro section #9253

siimonevans:
Addresses comment [here](https://github.com/mozilla/foundation.mozilla.org/issues/9253#issuecomment-1238578701) on 9253.

The spacing isn't quite right between the three text elements (when comparing with the Figma file), but the only way to overcome that would be to override this class, which isn't easy to do in a nice way:

```
@media (min-width: 768px)
.medium\:tw-space-y-6 > :not([hidden]) ~ :not([hidden]) {
   ...
    margin-top: calc(2rem * calc(1 - var(--tw-space-y-reverse))) !important;
    margin-bottom: calc(2rem * var(--tw-space-y-reverse)) !important;
}
```

Should I create a special class to make those overrides, or is the above snipper in place to prevent people from using incorrect spacing?

Screenshots:
Screenshot 2022-09-08 at 11 04 35
Screenshot 2022-09-08 at 11 05 01

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/pages/youtube-regrets-2022/fragments/hero.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/pages/youtube-regrets-2022/fragments/hero.html
@@ -13,7 +13,7 @@
       <p class="mb-1 tw-text-3xl tw-font-bold mt-0">{% trans "Investigating YouTube's ineffective user controls" %}</p>
       <p class="tw-body-large large:tw-text-2xl mt-0">
         {% blocktrans trimmed %}
-          powered by 22,722 volunteers, Mozilla scrutinized YouTube to determine how much control people actually have over the platform’s recommendation algorithm.
+          Powered by 22,722 volunteers, Mozilla scrutinized YouTube to determine how much control people actually have over the platform’s recommendation algorithm.
           <span class="tw-italic">This is what we learned</span>.
         {% endblocktrans %}
       </p>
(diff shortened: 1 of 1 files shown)


siimonevans:
> Hey @siimonevans, if it’s okay I have imported changes from this PR into mine here as I was already splitting the string (but was doing it wrong by using a ) #9323
> 
> Also I think it’s pretty important to capitalize "Powered", as they are meant to be read as separate sentences

Thanks @TheoChevalier - sorted.

tbrlpld (changes requested):
@siimonevans This does not solve the spacing issue for me and uses the wrong font face for the subtitle.

The override is not necessary. It seems like you are trying to override the equal spacing between the children of the flex parent. You should be able to control the spacing between title and subtitle by adding a wrapper around them (then the wrapper will have the y spacing to the next paragraph, but title and subtitle won't to each  other).

siimonevans:
> @siimonevans This does not solve the spacing issue for me and uses the wrong font face for the subtitle.
> 
> The override is not necessary. It seems like you are trying to override the equal spacing between the children of the flex parent. You should be able to control the spacing between title and subtitle by adding a wrapper around them (then the wrapper will have the y spacing to the next paragraph, but title and subtitle won't to each other).
> 
> Also, please stick with the Tailwind spacing utilities (`tw-`) over the Bootstrap ones.

@tbrlpld sorry, I wasn't clear enough. The code snippet I linked is what currently exists and controls the spacing - I was looking for advice on how to override that rather than suggesting a solution 😄 I'll take a look at the wrapper classes and replace the bootstrap - I'm so used to using vanilla tailwind classes like `mb-0` that I forget the prefixing…

siimonevans:
Hey @tbrlpld I've pushed some fixes, hopefully this is more inline with project standards now - thanks for the help!

tbrlpld (approved):

PR #9361, opened 2022-09-13
Replace Scrolltrigger for bottom of page with scroll eventlistener

fessehaye:
* Replaced some of scrolltrigger usage with basic event listener
* Added a scroll buffer in case the page can not 100% reach the bottom incase we need to revisit this.


danielfmiranda (approved):
Approved!

Issue #9349, opened 2022-09-09
[YTR-USC] Slider is cropping images improperly


beccaklam:
The slider appears to cropping images improperly. The resulting image displayed on a published webpage is 429 x 241 pixels:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/cd38705e-9205-4d42-87b5-674ddea9daf0)

But for some reason the source image has been cropped to 445 x 185 pixels:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/2048ea8b-4255-4f79-8e36-0963a74f2b47)
 
The original image that was uploaded to wagtail had dimensions with the ratio of 429 x 241 pixels however. 
![Screen Shot 2022-09-09 at 3.08.34 PM.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/6191c4ab-8b24-4e6e-9745-dcf4fc66edc2)

Not quite sure what is happening here but this appears to make all images stretched no matter what ratio is uploaded.

mtdenton:
Did some investigating. Seems like this slider block resizes the image to 445x185:
https://github.com/mozilla/foundation.mozilla.org/blob/f2fdcd1de0ae46ec689712aa9f223c1689d0e02b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/foundation_slider_block.html#L19

...and then forces it to be 16:9 (445x185 is not 16:9) via tw-aspect-video here:
https://github.com/mozilla/foundation.mozilla.org/blob/f72fa97c740b40dc9f783ba55e0fdf8a60cc2c18/network-api/networkapi/mozfest/templates/fragments/slider_card.html#L3

@fessehaye Legit bug here or was there a reason?

mtdenton:
Side note as well, noticing that all of the original source images uploaded into the CMS have smaller dimensions than the final product which is causing the image to be upsized in Wagtail, so you're losing fidelity. As much as I don't care for Tucker Carlson's face, if you wanted to avoid pixelation on the final image, you could pull much larger thumbnail images using this:

https://youtube-thumbnail-grabber.com/

...and Wagtail will then downscale the image instead of upscale.

beccaklam:
Thanks for the tip @mtdenton, I've updated the images.

PR #9370, opened 2022-09-15
Fix Slider img scaling and speed up animation

fessehaye:
Related PRs/issues: #9349 

Slider images were not using the correct aspect ratio.
Continuous scrolling animation is reduced by ~1s.


danielfmiranda (approved):
Thanks for the fix Simon! Tested and Approved 👍

PR #9408, opened 2022-09-20
updated category_page.html to show cta whether or not it has related articles

danielfmiranda:
The featured_cta section on the category page was inside an if statement, that only rendered if a category had related articles.

This PR takes the cta section out of that if statement, so it can show based on whether or not the category has "show_cta" set to true, but does not depend on related articles.

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/category_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/category_page.html
@@ -26,22 +26,23 @@
 
           </div>
         {% endif %}
-
-        <div
-        class="
-          {% if category != current_category %}
-            d-none
-          {% endif %}
-        "
-        >
-          {% with cta=featured_cta %}
-            {% if cta %}
-              {% include "fragments/buyersguide/call_to_action_box.html" with icon=cta.sticker_image heading=cta.title body=cta.content link_text=cta.link_label link_href=cta.get_target_url %}
-            {% endif %}
-          {% endwith %}
-        </div>
-
       {% endwith %}
+
+      <div
+      class="
+        {% if category != current_category %}
+          d-none
+        {% endif %}
+      "
+      >
+        {% with cta=featured_cta %}
+          {% if current_category.show_cta %}
+            {% include "fragments/buyersguide/call_to_action_box.html" with icon=cta.sticker_image heading=cta.title body=cta.content link_text=cta.link_label link_href=cta.get_target_url %}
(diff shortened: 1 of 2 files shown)


network-api/networkapi/templates/pages/buyersguide/category_page.html
```diff
@@ -24,14 +24,23 @@ <h2>({{ category.name }}) Primary related articles (First 3 of list of 6)</h2>
...
+          {% endif %}
+        "
+        >
+          {% with cta=featured_cta %}
+            {% if cta %}
+              {% include "fragments/buyersguide/call_to_action_box.html" with icon=cta.sticker_image heading=cta.title body=cta.content link_text=cta.link_label link_href=cta.get_target_url %}
+            {% endif %}
+          {% endwith %}
+        </div>
+
```
tbrlpld, line 43: Have you considered also moving it outside of the `{% with primary_related_articles=category.get_primary_related_articles  %}` block? Seems like this is no information we need.

Also, it seems to me the `{% if cta %}` should be `{% if category.show_cta %}`.

danielfmiranda: Hi Tibor, I have moved it out of the with block, as well as updated it to use category.show_cta.

Originally, we were checking if this was set to true in the backend and updating the context, but I think your solution makes more sense to me!

network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
```diff
@@ -314,9 +314,6 @@ def categories_page(self, request, slug):
                                f' | Mozilla Foundation'
         context['template_cache_key_fragment'] = f'{category.slug}_{request.LANGUAGE_CODE}'
 
-        if not category.show_cta:
```
danielfmiranda, line 317: As mentioned in my previous comment, we were originally updating the context, however, I think using show_cta  in the template is more readable.

tbrlpld: This change is necessary anyhow. When the page loads, it already needs to contain the information for all the categories. JS is then used to switch between which category is shown. That means we can not control if the CTA should be shown based on the first category that the template is loaded with. 

(I am not really a fan of how this works and I think this is also a big contributor to how slow the page is).

tbrlpld (approved):

Issue #9336, opened 2022-09-08
 GA4 | Add data layer event for commento comment submission


danielfmiranda:
This is a follow up ticket to #9271 as we had to reach out to commento.io support before starting this work.

The following  is a populated example of a  data layer push that should take place whenever a comment is submitted

Example form: https://foundation.mozilla.org/en/privacynotincluded/threema/

```
window.dataLayer = window.dataLayer || [];
window.dataLayer.push({
	'event' : 'form_submission',
	'form_name' : 'Threema',
	'form_location' : 'foundation.mozilla.org/en/privacynotincluded/threema/',
	'form_type' : 'petition-form',
         'form_id' : '8611'
});

```

danielfmiranda:
Just providing a quick update, I have reached out to commento.io support and have added @cdanfon to the BCC list, however, after ~ 7 days and a follow up there has been no response.

@cdanfon I am wondering if this means you would like me to work on implementing this through the javascript method I mentioned in #9271?

cdanfon:
Hey @danielfmiranda thanks for the update! Let's go with option 1.

danielfmiranda:
Providing an update on this ticket:
- I have tested out my code through the google analytics preview tool using the steps found in ticket #9387, and can confirm that it works when testing using that URL.
- However, when testing on staging, the event does not seem to fire.

I am going to continue to investigate this

danielfmiranda:
One possible problem could be that since these commento elements are being load through commentos JS, our minified JS including the file that adds the event handlers runs before these objects even exist.

For example, here is a screenshot where I set a breakpoint where we do a querySelector for the commento submit button:
<img width="1493" alt="image" src="https://user-images.githubusercontent.com/18314510/191384486-5f7d7a4b-c17f-4396-aff9-848fbdb59da0.png">
Note that the commento box has not loaded in yet.

danielfmiranda:
PR #9414 now provides a fix! :fingers-crossed:

danielfmiranda:
Hi @cdanfon, this ticket is now merged into staging and ready for testing from the torchbox team. 
When testing myself, everything seems to be working as expected:
![image.png](https://images.zenhubusercontent.com/6046b7f1bc36adc9ce95a16a/c31728fd-0f94-4d34-86a4-e5b3f67018bc)

Here are the following test cases that I have checked:
- Logged in, and submitting comment
- Attempt to submit a comment but not logged in
- Log in, then submit a comment

[...]

PR #9414, opened 2022-09-22
Updated commento GA tracking

danielfmiranda:
Link to sample test page:
Related PRs/issues: #9336

# Description
This PR is an update to #9387 as that PR's code would break if the html elements did not load in time.

This file now does the following:
- Creates a mutation observer `commentListObserver` which "listens" to the commento div for any mutations to the childList. (EX: Added nodes, removed nodes, etc.)
- If any mutation fits the criteria of a new comment, a datalayer event is pushed.
  - The current criteria for a new comment is:
    A new div, which houses one child div with the class `commento-card`.

# Why this method?

The reason for using the mutationObserver, instead of just saying "is the user logged in? Did they fill out the box and click submit? Then push to datalayer." is because I believe the user flow that is most likely is going to the be the following for the majority of people:

- User tries to comment for the first time without logging in, and clicks submit
- They are then shown a window to log in through social networks (google, twitter, etc)
- Once they select, they are automatically logged in and the comment submitted, which means that event would be lost.

# Testing as a dev:
Testing this on your local machine can be a little tricky, as commento does not allow us to render the comment box of instances of localhost. To get around this, I had to do the following:

1. Install https://ngrok.com/download
2. Run localhost instance using docker-compose up
3. In your terminal, run ngrok http 8000, take note of the domain that is pointing to localhost
4. Log into https://commento.io/ using credentials from 1password
5. Click "new domain" and add the domain that was returned from ngrok
6. Visit PNI on this domain, and commento should now render the comment box.

**However:** I am not sure if this needs to be tested on local hosts, as the torchbox team has asked to test this on staging, where commento works without problem. Also, I have tested this using the google tag manager preview tool and can confirm it works!

### DIFF
--- a/source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
+++ b/source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
@@ -10,11 +10,13 @@ export default () => {
   );
 
   const commentListUpdateHandler = (mutations) => {
+    // The div which new comments get appended to.
     let commentList = document.querySelector("#commento-main-area");
 
     mutations.forEach(function (mutation) {
-      // If there was a added node to the comment list container div, push the datalayer event.
-      if ((mutation.target == commentList) && mutation.addedNodes.length) {
+      // New comment mutations come in the form of a single added node, with no attributes.
+      let newNodeAttributes = mutation.addedNodes[0].attributes;
+      if ((newNodeAttributes.length == 0) && (mutation.target == commentList)) {
 
         window.dataLayer = window.dataLayer || [];
         window.dataLayer.push({
@@ -28,24 +30,12 @@ export default () => {
     });
   };
 
-  const commentoFormHasBeenLoadedIn = () => {
-    // Stop listening to the parent div for any activity.
-    commentoParentDivObserver.disconnect();
 
-    // Instead create a new observer that listens for updates to the comments list.
-    const commentListObserver = new MutationObserver(commentListUpdateHandler);
-    commentListObserver.observe(commentoContainer, {
-      childList: true,
-      subtree:true
-    });
-  };
 
   // Listen to the Commento container div for any updates within.
-  const commentoParentDivObserver = new MutationObserver(commentoFormHasBeenLoadedIn);
-  commentoParentDivObserver.observe(commentoContainer, {
+  const commentListObserver = new MutationObserver(commentListUpdateHandler);
+  commentListObserver.observe(commentoContainer, {
     childList: true,
-    subtree: true,
-    attributes: true,
-    characterData: true,
+    subtree: true
   });
 };
--- a/source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
+++ b/source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
@@ -3,53 +3,49 @@
  * Since at the moment there is no official callback for comment submission through commento itself,
  * we have this file which will push a datalayer event whenever a user submits a comment.
  */
- export default () => {
+export default () => {
   // Commento's parent div, in which it loads all of its content through JS.
   const commentoContainer = document.querySelector(
     "#view-product-page #commento"
   );
-  // The products initial count of comments, which we will use to compare against the new number
-  let commentCount;
 
+  const commentListUpdateHandler = (mutations) => {
+    let commentList = document.querySelector("#commento-main-area");
 
-  const mutationHandler = (mutations) => {
-  // Since commento sometimes does not load its contents before we run this JS file,
-  // We are using this mutation handler to wait for a change in its parent div, 
-  // whether it be an item loading in, the user typing in the text box, etc, 
-  // so we know the content has loaded in, and its safe to add the event handlers 
-  // and define the initial count of comments. Then disconnecting the mutation handler.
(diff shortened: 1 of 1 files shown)


source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
```diff
@@ -4,34 +4,48 @@
...
   const commentoContainer = document.querySelector(
     "#view-product-page #commento"
   );
-  const submitButton = document.querySelector("#commento-submit-button-root");
 
-  if (commentoContainer && submitButton) {
-    submitButton.addEventListener(`click`, () => {
-      commentSubmitted();
+  const commentListUpdateHandler = (mutations) => {
+    let commentList = document.querySelector("#commento-main-area");
```
danielfmiranda, line 13: the reason for defining this variable here, is because this div gets deleted/re-rendered during the following scenarios, meaning the observer would then have nothing to point at and would no longer push the datalayer event:
- If the user is not logged in and enters a comment, they get prompted to log in, after logging in this div gets re-rendered and the existing comment gets immediately submitted.
- If a user is not logged in and logs in, the div gets re-rendered.

danielfmiranda, line 14: The reason for this line is that whenever a user logs in/out, the "#commento-main-area" div gets deleted and re-rendered.
 So if we were to define this once somewhere else in the code and then reference it here, and the user had to log in before submitting a comment (which I think is going to be the case for most users), this object would no longer exist.

source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
```diff
@@ -4,34 +4,38 @@
...
-    let textBoxValue = document.querySelector("#commento-textarea-root").value;
-
-    if (loggedInContainer && textBoxValue) {
-      window.dataLayer = window.dataLayer || [];
+  const commentListUpdateHandler = (mutations) => {
+    mutations.forEach(function (mutation) {
+      // New comments are appended in the form of: <div><div class='commento-card'/></div>
+      // If comment tracking ever breaks, this logic be a good place to check first.
+      if (
+        mutation.addedNodes[0].firstChild.classList.contains("commento-card")
```
danielfmiranda, line 21: This is what a "new comment" event currently looks like:
- A mutation with 1 added node, which is a div.
Screen Shot 2022-10-03 at 12 19 56 PM

- This new div has one child, which is the new comment div with the classname of 'commento-card'

Screen Shot 2022-10-03 at 12 20 57 PM


This If statement is saying "if the first child of the added node has the class 'commento-card', its safe to say a comment has been submitted and we can push the datalayer event.

source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
```diff
@@ -4,34 +4,38 @@
...
-    );
-    let textBoxValue = document.querySelector("#commento-textarea-root").value;
-
-    if (loggedInContainer && textBoxValue) {
-      window.dataLayer = window.dataLayer || [];
+  const commentListUpdateHandler = (mutations) => {
+    mutations.forEach(function (mutation) {
+      // New comments are appended in the form of: <div><div class='commento-card'/></div>
+      // If comment tracking ever breaks, this logic be a good place to check first.
+      if (
```
danielfmiranda, line 20: This is what a "new comment" event currently looks like:
- A mutation with 1 added node, which is a div.
Screen Shot 2022-10-03 at 12 19 56 PM

- This new div has one child, which is the new comment div, with the classname of 'commento-card'

Screen Shot 2022-10-03 at 12 20 57 PM


This If statement is saying "if the first child of the added node matches `<div class='commento-card'>`, its safe to say a comment has been submitted and we can push the datalayer event.

source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
```diff
@@ -4,34 +4,38 @@
...
+        mutation.addedNodes[0].firstChild.matches(".commento-card")
+      ) {
```
danielfmiranda, line 22: This is what a "new comment" event currently looks like:
- A mutation with 1 added node, which is a div.
Screen Shot 2022-10-03 at 12 19 56 PM

- This new div has one child, which is the new comment div, with the classname of 'commento-card'

Screen Shot 2022-10-03 at 12 20 57 PM


This If statement is saying "if the first child of the added node matches `<div class='commento-card'>`, its safe to say a comment has been submitted and we can push the datalayer event.

fessehaye (approved):

source/js/buyers-guide/template-js-handler/product-page-comment-handler.js
```diff
@@ -4,34 +4,38 @@
...
-      "#commento-logged-container"
-    );
-    let textBoxValue = document.querySelector("#commento-textarea-root").value;
-
-    if (loggedInContainer && textBoxValue) {
-      window.dataLayer = window.dataLayer || [];
+  const commentListUpdateHandler = (mutations) => {
+    mutations.forEach(function (mutation) {
+      // New comments are appended in the form of: <div><div class='commento-card'/></div>
+      // If comment tracking ever breaks, this logic would be a good place to check first.
```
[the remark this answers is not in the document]
danielfmiranda: Updated!

Issue #9097, opened 2022-07-14
[PNI] - Product Page Editorial Content


fessehaye:
https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3083%3A30133

![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/c8e88813-fd0d-4487-b337-854788bfcc7b)

Dev Tasks:
* [ ] Add Dive Deeper Component to layout
* [ ] What to read next Component to layout
* [ ]  Full Width Newsletter
* [ ]  Related Product Style changes

Acceptance Criteria: 
* [x] Editorial Content is visible on the bottom of any specific product page
* [x] Editorial content is styled correctly and has correct grid layout
* [x] Editorial content is styled correctly on mobile

cdanfon:
@beccaklam to update design
@fessehaye update description

beccaklam:
Design has been updated.

tbrlpld:
@fessehaye I am not sure how to understand this AC "Editorial Content is visible on the bottom of any specific product page"

tbrlpld:
@cdanfon This is ready for QA on the preview app: https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/threema/

cdanfon:
Looks good to me @tbrlpld Closing this, thanks!

PR #9421, opened 2022-09-23
9097 Product page editorial content

tbrlpld:
# Description

This PR adds the relation and links sections to the bottom of the product page. This was mostly achieved by fitting previously created components into a new layout. 

I did have to adjust the styling of the related products section though, because no component for this was created before hand. 

Link to sample test page: http://localhost:8000/en/privacynotincluded/general-percy-product/
Related PRs/issues: #9097 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/related_product.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/related_product.html
@@ -2,7 +2,7 @@
 
 <div class="related-product">
   <a class="tw-block {% if related_product.adult_content %} adult-content{% endif %}" href="{% relocalized_url related_product.url %}">
-    <div class="img-container">
+    <div class="img-container tw-p-5">
       {% image related_product.image width-600 as img %}
       <img
         class="product-thumbnail"
--- a/network-api/networkapi/templates/fragments/buyersguide/related_product.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/related_product.html
@@ -8,12 +8,12 @@
         class="product-thumbnail"
         width="600"
         loading="lazy"
-        src="{{img.url}}"
+        src="{{ img.url }}"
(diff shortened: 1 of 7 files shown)


network-api/networkapi/templates/fragments/buyersguide/related_product.html
```diff
@@ -0,0 +1,21 @@
...
+<div class="related-product">
+  <a class="tw-block {% if related_product.adult_content %} adult-content{% endif %}" href="{% relocalized_url related_product.url %}">
+    <div class="img-container tw-p-5">
+      {% image related_product.image width-600 as img %}
+      <img
+        class="product-thumbnail"
+        width="600"
+        loading="lazy"
+        src="{{img.url}}"
+        alt="{{related_product.title}}"
```
kevinhowbrook, line 12: **suggestion - non blocking**

I'm not privy to the `alt` system on the site, however, I think the images alt attribute should come from `related_product.image.alt` (if it's there) - but you might be intentionally using the title here down to publishing styles, just thought it worth mentioning.

kevinhowbrook (approved):

network-api/networkapi/templates/fragments/buyersguide/related_product.html
tbrlpld: Good point. I have extracted this nearly as is from the product page. Looks like we are using the default Wagtail image model and don't have an alt field. The `alt` property would only return the image title and from what I can see on prod, this is not always fit for use as alt text (often just the filename). I guess using the product title might make more sense here, because all images I have seen do display the product in question (or a depiction there of -- e.g. logo of an app). I am actually wondering if we should just remove it, since it does not add any information..

Issue #9099, opened 2022-07-14
[PNI] - Newsletter component in product grid


fessehaye:
<img width="868" alt="Screen Shot 2022-08-09 at 9 39 21 AM" src="https://user-images.githubusercontent.com/74557207/183677738-8fdc6b2d-dd6d-49f8-ab7f-0a4e9ea875b8.png">

[Figma: Component States](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3231%3A33672)
[Figma: Design in Context](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3096%3A30793)

Dev Tasks: 
* [ ] Add newsletter block after 10 Products (2x width)
* [ ] Switch from using Flexbox -> CSS Grid

Acceptance Criteria:
* [ ] Newsletter is visible in product review grid

cdanfon:
@beccaklam to update the design

cdanfon:
@fessehaye to update description

tbrlpld:
@cdanfon This is ready for testing on the preview app: https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/

cdanfon:
Thanks @tbrlpld looks good to me

PR #9435, opened 2022-09-27
9099 newsletter in product grid

tbrlpld:
# Description

This PR adds the newsletter component into the product grid. I was not able to keep the centering of the last row of cards and combine it with a sensible positioning of the newsletter box when the category filters on the homepage are used. As a compromise, I switched from flex to grid. This allows the newsletter box to stick to the right hand side of the grid when the category filters are applied. 

Also, I was not able to get a ordering in place in which more cards are placed above the newsletter signup when the category filters are applied. It appears like the `nth-...` selectors do not update if classes are toggled with CSS. 

Since I also had to fix some indenting in the `catalog.html` template, it might be easier to follow the changes in the commits. 

Link to sample test page: http://localhost:8000/en/privacynotincluded/
Related PRs/issues: #9099

# Checklist

**Tests**
- [ ] ~~Is the code I'm adding covered by tests?~~ No. 

**Changes in Models:**

**Documentation:**

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/item.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/item.html
@@ -7,7 +7,6 @@
     tw-flex-col
     tw-justify-between
     tw-m-0
-    tw-w-[var(--item-width)]
 
     {# We need to unset the order rule at the next breakpoint, because the nth-child pseudo-class changes. #}
     tw-order-none
--- a/network-api/networkapi/templates/fragments/buyersguide/item.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/item.html
@@ -6,10 +6,18 @@
     {% if matched == False %}d-none{% else %}d-flex{% endif %}
     tw-flex-col
     tw-justify-between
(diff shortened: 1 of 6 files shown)


kevinhowbrook:
Is there a sample test page I can check this on @tbrlpld ?

tbrlpld:
> Is there a sample test page I can check this on @tbrlpld ?

Whoops. Sorry about that. You should be able to test on http://localhost:8000/en/privacynotincluded/

kevinhowbrook:
Code wise looks good to me.
Looks like the newsletter cta is rendered after the products and the designs look like it's supposed to be in the flow of the grid items:

image

Not sure if that's desired or intentional?

tbrlpld:
@kevinhowbrook That should be correctly in the grid. I think you might need to fully restart the containers, maybe even `inv catchup`. I have been running into this myself... we really need to do something about this. I am afraid this has to do with the super slow CSS build and how Docker is used in the project in general.

kevinhowbrook (approved):

Issue #7556, opened 2021-10-04
Numbers in templates are still getting localized


Pomax:
This is caused by `L10N=True`: even though we removed `USE_THOUSAND_SEPARATOR`, which turns this behaviour off for English/default, the `L10N` setting "overrules" that and turns on silent number reformatting with even less control. There is the `{{ value | unlocalize }}` template filter from the `l10n` tags that we can use, but it would be fantastic if we can find out how to actually turns this off for every locale, except where `intcomma` is explicitly used.

TheoChevalier:
@Pomax I’m curious where you see this happening?

TheoChevalier:
Ah, the data-creepiness attribute, just caught up with the PNI PR.

Pomax:
yup

richbrennan:
I've audited the templates by searching for instances of `data-*` or `<script>` in order to identify variables that may need the `unlocalize` filter:

mozfest_footer.html
- `data-signup-id="{{ mozfest_footer.id }}"`

signup.html
- `data-signup-id="{{ page.signup.id }}"`

buyersguide_item.html - ALREADY ADDED FOR PNI
- `data-creepiness="{{ product.creepiness }}">`

bannered_campaign_page.html
- `data-signup-id="{{ page.signup.id }}"`

index_page.html
- `data-page-size="{{ page.specific.page_size }}"`

profile_blocks.html
- `data-program-year="{{ value.program_year }}"` - APPLY DATE FILTER?

pulse_project_list.html
- `data-max="{{ value.size }}"`

petition.html
  - `data-petition-id="{{ cta.id }}"`

load-more-regrets.html
- `{{ page.specific.batch_size }}`
- `{{ page.specific.max_batch_size }}`

Pomax:
this would be good to revisit

mtdenton:
@cdanfon Let's move this up the backlog too and see if it's still relevant, might be a fairly easy win.

tbrlpld:
> yup. Apparently we missed this part in https://docs.djangoproject.com/en/3.2/ref/settings/#use-thousand-separator:
> 
> > These settings may also be dictated by the locale, which takes precedence.
> 
> Without any clues as to how it does that, and how you stop it (that I can find) beyond "you don't, you have to use that template tag".


This is the full documentation of the `USE_THOUSAND_SEPARATOR` setting:
[...]

PR #9439, opened 2022-09-29
7556 localized numbers

tbrlpld:
# Description

It was previously observed that numbers entered into the template have been localized broadly. As a response we had turned off the `USE_THOUSAND_SEPARATOR` setting. But, this still left the decimal separators to be localized (e.g. the German locale would use a comma instead of a dot). This can lead to issues when the values are inserted into the template for computational use, rather than enduser consumption (think values of inline scripts and style, but also HTML attributes like `width` and `height` on the `img` tag). It would be possible to deactivate that feature with `L10N = False`, but that is not a future proof change since the setting will be deprecated in Django 4.0 and ignored starting with Django 5.0. 

Therefore, we will need to make sure that when ever we insert a value (especially float values) into the template for programmatic use, that we use the `unlocalize` filter to make sure the format of the numbers follows English standards and is interpretable in a programmatic way. 

See also: https://docs.djangoproject.com/en/4.1/topics/i18n/formatting/#controlling-localization-in-templates

Link to sample test page:
Related PRs/issues: #7556

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented?
- [x] Did I update the READMEs or wagtail documentation?


tbrlpld:
@fessehaye @danielfmiranda @TheoChevalier @kevinhowbrook Tagging all of you for visibility. This is probably something we all need to keep an eye out for going forward.

TheoChevalier (approved):
Thanks for going through all this @tbrlpld!

I also suspect disabling `USE_L10N` would have broken the date localization (blog post publication dates, for instance), so this seem to be the right (albeit painful) way

danielfmiranda (approved):
Thanks for the tag Tibor! will keep an eye out for this moving forward like you mentioned, also, kudos for updating the readme 👍 approved!

PR #9451, opened 2022-10-03
[PNI] Update translation files management

TheoChevalier:
Refactoring the bash script I’m using to extract strings (as documented here https://github.com/mozilla/foundation.mozilla.org#translations) that’s running locally when using `inv makemessages`, as there are a couple changes needed in the PNI branch:

I’m moving back the templates used in the About section into a /about subdirectory in order to extract them into their own .po files, as this content is only translated into a subset of locales.

I’ve merged the PNI UI strings (previously in `network-api/networkapi/wagtailpages/templates/buyersguide/`) into the main .po file, since most PNI UI strings were already there, and I would have to migrate those translations anyway due to the template location change. Having them in the main .po file is probably safer in case templates move again in the future.

This is also addressing another issue: Django >=3.2 stopped processing locales in ab-CD format, the script is now moving the files around automatically.

I know all the translation management in the project is a bit obscure, but I’m happy to clarify anything!


tbrlpld (commented):
@TheoChevalier Thanks for updating this. 

I am not quite sure I understand this comment: 
> I’m moving back the templates used in the About section into a /about subdirectory in order to extract them into their own .po files, as this content is only translated into a subset of locales.

Moving the templates is also breaking the tests. Those would need to be updated too. Can you add some context why we need the extra nesting level?

TheoChevalier:
I *think* I’ve fixed the tests, but don’t manage to get them running here 🤨

tbrlpld:
Thanks @TheoChevalier! That is super helpful.

tbrlpld (approved):

Issue #9215, opened 2022-08-10
SPIKE | Automatic deployment from main to staging


cdanfon:
This spike is to investigate options for automatic deployments from main to staging. 

These are a few initial thoughts/questions: 

- What problem/pain point are we trying to solve?
- What are the technical options?
- What things do we have to consider around internal stakeholders? 
- How does the workflow look like?

[Time box to 2 hours]

tbrlpld:
@cdanfon @mtdenton @danielfmiranda @fessehaye 


I have been looking into this now for 2h and here are my notes: 


- What problem/pain point are we trying to solve?
    - Current process:
        - A feature or bug is developed on a dedicated branch. The work receives code review. Once the work is approved and passes all tests on CI, it is merged into the `main` branch.
        - After the merge to `main` a developer needs to check that CI on `main` still passes (maybe we are missing merge migrations etc.)?
        - Once CI passes on `main` the developer needs to pull the latest changes from GitHub to their local machine.
        - Once the developer has the latest changes on their machine, they can push the changes to staging.
    - Pain points:
        - We often have to wait for CI to pass twice, first on the feature branch before the merge and then on main. The CI on the feature branch is needed because often the branch needs to be updated before the merge which triggers CI. Each CI run takes like 10-15min to pass.
        - Did we check that the merged result on `main` passes  This needs to be checked manually. I could accidentally fetch and deploy without waiting for CI to pass.
        - Did something else get merged into `main` between checking CI to pass and pulling the changes locally. This could lead to accidentally pushing unknown changes.
        - The manual pushing could deploy a wrong branch or wrong changes on accident. The local branch is not protected and could differ from what other devs have or be broken.
        - We sometimes forget to deploy our changes to staging right after merging. This leads to a single deployment releasing multiple possibly unrelated changes to staging.
        - If merged work is not deployed to staging right after it is merged and CI passes we are unnecessarily delaying QA, which delays deliver to stakeholders / users.
- What are the technical options?
    - Activate the Heroku GitHub integration: [https://devcenter.heroku.com/articles/github-integration](https://devcenter.heroku.com/articles/github-integration)
        - This is what we have used in the past until the Heroku-Github security breach: [https://status.heroku.com/incidents/2413](https://status.heroku.com/incidents/2413)
[...]

PR #9494, opened 2022-10-14
Custom GitHub Actions workflow to continuously deploy `main` to the Heroku staging app

tbrlpld:
This PR enables a custom GitHub Actions workflow to deploy the `main` branch to the staging app on Heroku. 

The custom workflow checks that the CI pipline has finished successfully, checks out the `main` branch and pushes it to Heroku via the traditional Git push deployment.

The docs are updated to reflect this change and contain information on which credentials need to be set up and how to get them. 

I have tested that the Heroku access works with a simple `git fetch heroku main`. This is the workflow run of the test: https://github.com/mozilla/foundation.mozilla.org/actions/runs/3492453124/jobs/5846247852

Related PRs/issues: #9215

### DIFF
--- a/.github/workflows/continuous-deployment.yaml
+++ b/.github/workflows/continuous-deployment.yaml
@@ -28,7 +28,9 @@ jobs:
       - name: Deploy to staging
         # run: git push heroku main
         # Test this workflow with a benign action at first
-        run: git fetch heroku main
+        run:
+          - cat ~/.netrc
+          - git fetch heroku main
 
 #   on-failure:
 #     runs-on: ubuntu-latest
--- a/.github/workflows/continuous-deployment.yaml
+++ b/.github/workflows/continuous-deployment.yaml
@@ -28,9 +28,9 @@ jobs:
       - name: Deploy to staging
         # run: git push heroku main
         # Test this workflow with a benign action at first
-        run:
-          - cat ~/.netrc
-          - git fetch heroku main
+        run: |
+          cat ~/.netrc
+          git fetch heroku main
 
 #   on-failure:
 #     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


mtdenton:
This looks good to me, but leaving this for someone else to approve so that I'm not the only one checking this out. Thanks @tbrlpld

danielfmiranda (approved):
Hi @tbrlpld,

This PR looks good to me. Thanks again for taking this work on and updating the documentation as well. This will be very helpful in the future.

Approved!

.github/workflows/continuous-deployment.yaml
```diff
@@ -0,0 +1,43 @@
+name: Continuous Deployment
+# Based on https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-a-workflow-based-on-the-conclusion-of-another-workflow
+
+on:
+  # Trigger this workflow (continuous deployment) when the "Continous Integration" workflow
+  # had completed on the main branch.
+  workflow_run:
+    workflows: [Continuous Integration]
+    types: [completed]
+    branches: ["main"]
```
fessehaye, line 10: Is there a reason we want this after contious intergration vs `on:[workflow_dispatch]` ?

fessehaye (approved):
I had one small question but its not a blocker!

mmmavis (approved):
Thanks for taking this on @tbrlpld ! I don't have a way to test this out but code and documentation both look great to me! 👍🏻

.github/workflows/continuous-deployment.yaml
tbrlpld: If we don't wait for CI to pass we might be deploying broken code

Issue #9194, opened 2022-08-03
[PNI] Translate article page with related articles


tbrlpld:
# Describe the bug

When trying to translate a page of the new PNI article page type (#9083) which has related articles defined, the admin crashes. 

```
Traceback (most recent call last):
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/urls/__init__.py", line 125, in wrapper
    return view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/auth.py", line 174, in decorated_view
    response = view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/views/pages/edit.py", line 304, in dispatch
    response = self.run_hook('before_edit_page', self.request, self.page)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/views/generic/hooks.py", line 16, in run_hook
    result = fn(*args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/wagtail_hooks.py", line 259, in before_edit_page
    return edit_translation.edit_translation(request, translation, page)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/views/edit_translation.py", line 657, in edit_translation
    location_info = get_segment_location_info(
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/views/edit_translation.py", line 451, in get_segment_location_info
    "widget": widget_from_field(child_field) if widget else None,
[...]

PR #9508, opened 2022-10-19
9194 article page with related articles trans bug

tbrlpld:
# Description

This PR makes related articles relation (orderable) on the Buyer's Guide Article Page non-translatable. This is to fix the error described in #9194, but also the desired modeling of the relation (according to @TheoChevalier).

Without this PR, an error occurred when you would try to translate an article page that has related articles defined.

Now, the orderable relations are not translatable anymore. This means the same orderable from the default locale is used on the article page in a different locale. When displaying a given article in a locale (different than the default) we want all related articles to also be for that same locale. 

I have added a test to check for that described behavior and an implementation of that behavior. Unfortunately, I am currently not sure how we can avoid the 1+n query that this implementation causes.

Link to sample test page: http://localhost:8000/cms/pages/80/edit/
Related PRs/issues: #9194 

# Checklist

**Tests**

**Changes in Models:**
- [x] Did I squash my migration? Only one migration added. 

**Documentation:**
- [x] Is my code documented?


kevinhowbrook (commented):
So this looks good to me, I tested it locally and for what is intended by the code it looks like it does what we want. Whilst reviewing there was a few things, which might already be answered but i didn't delve too deep.

Have we considered updating localize to fix https://github.com/mozilla/foundation.mozilla.org/issues/9194 - there's a lot of conversation around the bug here though so I assume we are in a quite specific bug fixing territory.

There are 2 related article fields on the page model. This was just a bit confusing to me in code, I can see the get_related_articles method is used by the 2 lower methods to get the first and last items in a list of 6, I'm just confused as to where/how the field definitions for related articles in the hero and the lower related (popular) articles are as two separate panels. I'm noting this just to be sure there is nothing extra we need to consider.

Finally, when selecting articles to relate, I can't actually select any, this might be a specific QS override somewhere though.

tbrlpld:
@kevinhowbrook I am responding to your questions below. 

> Have we considered updating localize to fix #9194 - there's a lot of conversation around the bug here though so I assume we are in a quite specific bug fixing territory.

I have checked the `wagtail-localize` changelog and did not see anything that would make me think an update would fix the bug. I have not tried the later versions though. 

Either way, I also learned that @TheoChevalier (the localization manager for the project) prefers the related article relations to be synchronized instead of translated. See [this comment](https://github.com/mozilla/foundation.mozilla.org/issues/9194#issuecomment-1282119467).

> There are 2 related article fields on the page model. This was just a bit confusing to me in code, I can see the get_related_articles method is used by the 2 lower methods to get the first and last items in a list of 6, I'm just confused as to where/how the field definitions for related articles in the hero and the lower related (popular) articles are as two separate panels. I'm noting this just to be sure there is nothing extra we need to consider.

Sorry, I am not sure I follow. Where is the second field to relate articles on the page model? There should only be one field on the page model to create the connecting. That field is limited to 6 relations. When displaying the page, we split that list in half. The first half is displayed as "related articles" up top, the second half is displayed as "what to read next" in the bottom.

It does seem that this was not the most intuitive decision... But changing that is out of scope for this ticket. We might be changing that in the future based on other discussions. 

> Finally, when selecting articles to relate, I can't actually select any, this might be a specific QS override somewhere though.

Is this in the translation? If so than that is by design. This is what is preventing the error, but also the desired behaviour that @TheoChevalier prefers. The relations are only defined in the default locale (English for this site). For all other locales, we synchronize the relations, meaning we use the same ones that are defined for the default locale. That is why we need to take care of the localization when returning / displaying the related articles. This is the change in `get_related_articles`. For each related article we now return the localized version of that article.

kevinhowbrook (approved):

Issue #9088, opened 2022-07-14
[PNI] - Add Parallax visuals with background Shapes


fessehaye:
https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3403%3A38681

Used in homepage,editorial and category page

Dev Tasks:
* [x] Add background shapes as SVG on main, category and editorial page
* [x] Work with designers to fine tune parallax effect for scrolling

Acceptance Criteria:

- [x]  As a user, I should be able to see at least 4-6 svg shapes on the background of the main page.
- [x]  As a user, I should be able to see at least 4-6 svg shapes on the background of the category page.
- [x]  As a user, I should be able to see at least 4-6 svg shapes on the background of the editorial page.
- [x]  As a user, I should be notice that the shapes slightly shifts in the background as you are scrolling down or up the page.
- [x]  As a user, I should see that the asterisk shapes are rotating as you scroll up and down the page.
- [x]  As a user, I should not see the background shapes on mobile viewports, tablet is fine.
- [x]  As a user, I should not notice any scroll-jacking issues or interfere with interacting existing PNI pages.

Ideas: 
* https://www.netlify.com/blog/fun-parallax-scrolling-css-for-matterday/#parallax-visuals
* https://loremipsum.io/
* https://locomotivemtl.github.io/locomotive-scroll/
* https://www.flowdough.nl/

cdanfon:
@fessehaye Can you please give us a number of hours to timebox this issue?

nancyt1:
@fessehaye Here's the [link](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=4575%3A58497) to the background shapes in figma

cdanfon:
Hey @kristinashu @nancyt1 this is now on staging and ready for review.

@fessehaye do you mind tagging the reviewer in a comment for visibility? Otherwise it could be missed.Thanks!

fessehaye:
@beccaklam can you review it since Nancy is away. I did review this with her before the PR was pushed for timings of the animations

beccaklam:
This looks good to go! I found one thing: 
![image](https://user-images.githubusercontent.com/20977376/203850671-ec65b613-5da8-4006-9943-f86245df24e9.png)
but @fessehaye said it will be addressed in a separate PR!

PR #9516, opened 2022-10-20
9088 - Add Parallax visuals with background Shapes

fessehaye:
# Description

Adding background shapes to category pages/home page/editorial page. Shapes are only visible on tablet/desktop screen size. Asterisk symbols should be rotating as you scroll down/up the page. Symbols should not be covering existing content and should be on the back row. Parallax effect will make the svg shapes feel like they are scrolling at a different pace as the rest of the content.

Should be able to test with urls below and just scrolling up and down the page.

image

Link to sample test page: /en/privacynoincluded, en/privacynotincluded/articles/
Related PRs/issues: #9088


tbrlpld:
Hey @fessehaye, can  you add some description to the PR to give me some context about what I am looking at here? That would be very helpful. Thank you.

tbrlpld (commented):
See previous comment

network-api/networkapi/templates/pages/buyersguide/base.html
```diff
@@ -78,22 +78,26 @@
   {% include "fragments/buyersguide/primary_nav.html" %}
   </div>
 
-  {% get_bg_home_page as home_page %}
-
-  {% block hero %}{% endblock %}
+  <div class="underneath-screen-overlay">
```
tbrlpld, line 81: Can you add some context around what this class is doing?

network-api/networkapi/templates/pages/buyersguide/base.html
```diff
@@ -78,22 +78,26 @@
...
+        {% block category_nav %}
+          {% include "fragments/buyersguide/pni_mobile_nav.html" with pagetype=pagetype categories=categories current_category=current_category %}
+          {% include "fragments/buyersguide/pni_category_nav.html" with pagetype=pagetype categories=categories current_category=current_category %}
+        {% endblock category_nav %}
 
+        <main class="main-content {% block main_content_class %}pb-5{% endblock %}">
+          {% block guts %}{% endblock %}
+        </main>
+      </div>
+      {% block background-parallax %}{% endblock %}
```
tbrlpld, line 99: What do you think about using `_` in the block name, rather than `-` to be consistent with the other block names in this template?

network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -151,3 +151,7 @@ <h2 class="tw-h3-heading">{% trans "Don’t see the product you’re looking for
 
   <script src="{% static "_js/bg-search.compiled.js" %}" async type="module"></script>
 {% endblock %}
+
+{% block background-parallax %}
+  {% include "fragments/buyersguide/shape_background.html" %}
+{% endblock %}
```
tbrlpld, line 157: Can you add some context why you need to add this block again? This template inherits from the `base.html` where you also add the background. Shouldn't that background be inherited too?

network-api/networkapi/templates/pages/buyersguide/editorial_content_index_page.html
```diff
@@ -4,75 +4,79 @@
...
+              {% if related_articles %}
+                {% trans "Popular" as heading %}
+                {% include "fragments/buyersguide/related_reading.html" with articles=related_articles heading=heading %}
+              {% endif %}
+            </aside>
+          {% endif %}
+        {% endwith %}
+      </div>
     </div>
+    {% include "fragments/buyersguide/shape_background.html" %}
```
tbrlpld, line 79: Same question as above: Why is the background not inherited from the `base.html`?

source/js/buyers-guide/template-js-handler/parallax-background.js
```diff
@@ -0,0 +1,39 @@
...
+  if (parallaxBackground) {
+    const svgShapes = document.querySelectorAll("#pni-parallax-background svg");
+    svgShapes.forEach((shape) => {
+      gsap.to(shape, {
+        y: -1000,
+        ease: "none",
+        scrollTrigger: {
+          trigger: "body",
+          start: "top top", // when the top of the trigger hits the top of the viewport
+          end: "bottom bottom",
```
tbrlpld, line 17: Does this mean we don't use the animation for scrolling the bottom of the shapes wrapper to the top of the scroller/viewport? It seems like a change of the scroll behaviour while the element is in view, no?

tbrlpld (commented):
Hey @fessehaye I have left a few question. 

I have also noticed testing this locally that the background moves faster than the foreground. Just wanted to double check this is desired, because you said it would be the other way around in the meeting earlier today.

tbrlpld:
@fessehaye Just noticed this jumping background when jumping between categories on the PNI homepage. 

Screen Recording 2022-10-24 at 13.49.13.mov

tbrlpld:
@fessehaye I also just noticed that the background is jumping when you switch filters on the PNI homepage.


https://user-images.githubusercontent.com/24797493/197627365-389e9507-f423-4a1f-bbf0-27cd4305e7a1.mov

network-api/networkapi/templates/pages/buyersguide/base.html
fessehaye: old class, cant find any reference so removing

network-api/networkapi/templates/pages/buyersguide/base.html
fessehaye: done

network-api/networkapi/templates/pages/buyersguide/editorial_content_index_page.html
fessehaye: removed

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: > @fessehaye I also just noticed that the background is jumping when you switch filters on the PNI homepage.
> 
>  Screen.Recording.2022-10-24.at.13.49.13.mov

resolved

fessehaye:
> Hey @fessehaye I have left a few question.
> 
> I have also noticed testing this locally that the background moves faster than the foreground. Just wanted to double check this is desired, because you said it would be the other way around in the meeting earlier today.

the background is suppose to moves faster than the foreground, I think I get the two mixed up. Nancy and I did go over the speed and we did adjust the speed to what she wanted afterwards

source/js/buyers-guide/template-js-handler/parallax-background.js
fessehaye: scrollTrigger start/bottom is kinda confusing to explain. If you want a visual understanding of what is happening I suggest adding `markers: true` to the object to see how the trigger and scrubbing works. The goal is the start of the scroll animation is the top of the body and bottom is the body of the animation and the animation progress is based on the scroll position. like https://loremipsum.io/

source/js/buyers-guide/template-js-handler/parallax-background.js
```diff
@@ -0,0 +1,44 @@
...
+    svgShapes.forEach((shape) => {
+      gsap.to(shape, {
+        y: -1500,
+        ease: "none",
+        scrollTrigger: {
+          trigger: "body",
+          start: "top top", // when the top of the trigger hits the top of the viewport
+          end: "bottom bottom",
+          scrub: true,
+          markers: true,
```
tbrlpld, line 19: Looks like the markers are still turned on.

network-api/networkapi/templates/pages/buyersguide/editorial_content_index_page.html
```diff
@@ -2,78 +2,85 @@
 {% load i18n localization static wagtailcore_tags wagtailimages_tags %}
 
 {% block guts %}
-  <div class="tw-container tw-my-6">
-    <h1>{{ page.title }}</h1>
+  <div class="tw-relative">
+    <div class="tw-container tw-my-6 tw-relative tw-z-10">
```
tbrlpld, line 6: Can you add some context why we need these two elements to be relative? It seems like the outer is only wrapping the inner.

tbrlpld (commented):
@fessehaye I have left some small questions, but I have also noticed that this animation is not prevented when used with "reduced motion" set. I think that is definitely something we would want to address as this is an accessibility issue.

fessehaye:
Addressed all 4 comments @tbrlpld

network-api/networkapi/templates/pages/buyersguide/base.html
```diff
@@ -84,15 +84,25 @@
     {% endblock %}
   </div>
 
-  {% block hero %}{% endblock %}
+  <div class="">
+    <div class="tw-relative">
+      <div class="tw-relative tw-z-10">
```
tbrlpld, line 89: Do we need the wrapper without a class and the two nested `div`s with `tw-relative`?

tbrlpld (commented):
Thanks for those updates @fessehaye There is one more question about wrappers.

network-api/networkapi/templates/pages/buyersguide/base.html
fessehaye: Yes, we need this 100%, else the shapes will overlap will existing content that has CSS transformations applied since that effects the  stacking order on the homepage. Since we are using gsap for a few things, I would recommend we keep this part.

cdanfon:
@tbrlpld please pick this one up and if it is not ready for prod we'll leaving it out from the prod push on Tuesday 15th.

network-api/networkapi/templates/pages/buyersguide/base.html
tbrlpld: @fessehaye What do you think about adding some comment to explain this? It not immediately obvious why it is needed, at least to me. 

I also tired without `<div class="">` locally and it seemed to work, but maybe I am missing something.

tbrlpld (commented):
Just one question about `<div class="">`.

network-api/networkapi/templates/pages/buyersguide/base.html
fessehaye: Oh I misread this, we need the two relative classes but we don't need the empty class div afaik. That was just a artifact of removing `underneath-screen-overlay`. I can add a comment about the two relative classes though

fessehaye:
@tbrlpld addressed the div problem and added comments

Issue #9463, opened 2022-10-06
[DFL] Stack aside and CTA elements


tbrlpld:
I just noticed this: 

![Screen Shot 2022-10-05 at 13 26 16-fullpage](https://user-images.githubusercontent.com/24797493/194157204-a2f2e894-0e2b-43c2-a58a-857845afb0b9.png)

I think this is what Kristina had been asking about too.

_Originally posted by @tbrlpld in https://github.com/mozilla/foundation.mozilla.org/issues/9434#issuecomment-1268939419_

stevedya:
This should be solved by https://github.com/mozilla/foundation.mozilla.org/pull/9538  Will revisit once that PR is finished with

cdanfon:
@stevedya I'm not sure if #9538 has been merged to staging yet, but at the moment is not stacking under the aside as expected but neither behaving like it was previously.
 
<img width="1485" alt="Screenshot 2022-10-31 at 11 02 10" src="https://user-images.githubusercontent.com/102978346/198993875-7f2a25e5-b29f-4a6a-8536-1b4ce4de9a88.png">

<img width="1485" alt="Screenshot 2022-10-31 at 11 03 41" src="https://user-images.githubusercontent.com/102978346/198993861-46cd815f-7c4a-469c-bdf7-779c13acb58d.png">

cdanfon:
@stevedya @tbrlpld I'm a bit confused about this one, what's the current state for this ticket please?

stevedya:
@cdanfon From my last comment above this should be solved by PR #9538

tbrlpld:
@cdanfon This is now on staging.

cdanfon:
Checked and looks good to me. When using a CTA on the side the 'aside' block changes positions and stacks in the main column instead of the aside section. 

@gutermul over to you for testing, although I believe @kristinashu made a mention about this.

gutermul:
When there's body content above (and presumably below) the "block with an aside," it isn't aligned with the block- it is still full page, and pours over (see screenshot). Is there a way to remedy this? I also am now wondering whether it limits what other streamfield blocks can be used on a page with an aside block...



![Image](https://user-images.githubusercontent.com/21171146/199720124-29729e58-bdf6-4271-8213-305dee1a059c.png)

gutermul:
signed off, thanks

cdanfon:
This is now on prod. Closing as completed

Issue #9432, opened 2022-09-26
DFL 2022 | Aside block for card listing (all cards)


cdanfon:
Ticket for the creation of the aside block for https://github.com/mozilla/foundation.mozilla.org/issues/9394

### Describe the feature request

In #9393 a new block was introduced, on the designs, there is also an 'aside block'. This block is out of the flow of the main body content and sits in the right sidebar.

For this aside block we will need to do the following:

- Create a new 'aside' block consisting of title and text (link button block already exists)
- Create a new 'Aside' field on the Banner Page type where blocks can be added. (just the new block and the button link block)
- Configre the template so IF the new aside field is populated with blocks, the template is slimmer and the aside field shows on the right on desktop view and stacks at the TOP on mobile (shown in the screenshot s the purple outlined area). 

### Screenshots / mock-ups

![Image](https://user-images.githubusercontent.com/2128707/192495492-da8a985d-66ee-4874-bab7-eafe7fa89cf3.png)

### Additional context
Link to [designs](https://www.figma.com/file/goD9bRqPoedQAxi8OsInkR/Data-Futures-Lab%3A-Designs?node-id=626%3A9963)

---

**Dev Tasks:** 
* [ ] Create new aside block
* [ ] Configure new field for aside content
* [ ] Configure template logic for new aside area

**Acceptance Criteria:**
- [ ] Design to look as per Figma files
- [x] As an editor creating a banner page I can add link and text blocks to the aside area
- [x] If an aside block is not added, the template layout is full width

cdanfon:
After discussing in stand up this is ready to be picked up

kevinhowbrook:
Lisa to check if we can float the aside block to the top of the sidebar so we can re-work this to support mutiples

kevinhowbrook:
I've re-worked the PR for this https://github.com/mozilla/foundation.mozilla.org/pull/9437
FE attention is needed because I can't fix whats going on with the link block in the aside area.

tbrlpld:
@kevinhowbrook Left some small questions.

tbrlpld:
@cdanfon This is now on staging

cdanfon:
Hey @kevinhowbrook tested on stage, a couple of things on this one, although looks like Simon Evans will be picking this on up:

1) The aside section doesn't quite level as per design (don't know if this is a dev constrain)
2) The aside button in the design is left aligned, whilst in the staging site is centred 


See attached 
[...]

Issue #9393, opened 2022-09-20
DFL 2022 | Landing Listing card


kevinhowbrook:
### Describe the feature request

We need to make a new listing card to be added on the landing section (see screenshot)

This should include:
- Image
- Title (text field)
- body (text field)
- optional page link (for title). (page chooser field)
- No footer/author. 

### Screenshots / mock-ups
![Screenshot(13)](https://user-images.githubusercontent.com/2128707/192488865-285d0567-6fa4-430a-8e8e-f4ffd6430eae.png)

### Additional context
Link to [designs](https://www.figma.com/file/goD9bRqPoedQAxi8OsInkR/Data-Futures-Lab%3A-Designs?node-id=626%3A9963)

---

**Dev Tasks:** 
* [ ] Create new listing card this will consist of title, body, optional metadata and optional page link
* [ ] Configure new text block template

**Acceptance Criteria:**
- [x] Design to look as per Figma files
- [x] As an editor creating a banner page I can add listing cards
- [x] Template stacks verticle on mobile
- [x] Title for cards to be black instead of blue as per outdated design

kristinashu:
Design should follow existing styles and convention on the site i.e. the headline should be black not blue.

kristinashu:
The listing card is working very well! 

The hover state for the heading doesn't seem to match our [style guide](https://foundation.mozilla.org/en/style-guide/#headers-links). 

Screen cap from staging (the underline looks blue and possibly too thin, it should be black and thicker)
<img width="395" alt="image" src="https://user-images.githubusercontent.com/1068783/195136063-ce5f17d4-3155-4ff8-8f9e-0ac2d067ea2f.png">

cdanfon:
### Amend needed: 

- [ ]Hover state for the header to be thicker and black to match style guide as per Kristina's comment above

cdanfon:
Steps for MoFo testing

1. Go to https://foundation.mofostaging.net/cms/pages/1769/add_subpage/
2. Create a new 'banner page'
3. Fill the 'title' field (this is compulsory and won't let you preview the page unless it's been filled out)
4. Add a 'listing' block in the body section (screenshot attached)
5. Check Acceptance criteria 2 & 3 (1 is part of the FE task)

NOTE: 
- To be tested alongside #9394 
- Hover state will be fixed as part of the FE work

cdanfon:
**FE Amend needed:**

[...]

PR #9538, opened 2022-10-24
#9432 [DFL] Aside block styling fixes 

stevedya:
# Description

This PR addresses button style issues for #9432 as well as adds a new streamfield for allowing blocks with asides

**Modifications made:**

- Left align aside button
- Add tailwindcss/container-queries plugin (this is used to determine when the container has gotten smaller by the CTA sidebar or signup form and renders the aside in a single column above the left content side). This is needed because when the CTA or signup form is on desktop we can tell how wide the left column is and be responsive with the content to avoid the aside rendering on the same row. 

### Example when paragraph is used  with no aside and listing block is used with an aside
Screen Shot 2022-10-28 at 12 44 43 PM

### How this content looks in wagtail
Screen Shot 2022-10-28 at 12 42 27 PM

### Before entering in content in wagtail
Screen Shot 2022-10-28 at 12 41 25 PM

### Example when cta or signup form is on the page
Screen Shot 2022-10-27 at 10 50 33 PM

Related PRs/issues:  #9432, #9463, #9393

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


tbrlpld (changes requested):
Thanks for giving this a shot @stevedya. While it does work in Chrome and Safari, the container queries are not yet supported in Firefox, which I think we need before using something on this site.

https://caniuse.com/css-container-queries

stevedya:
@tbrlpld Thats alright, the container queries aren't a make or break for this to work properly. This layout will still work with regular media queries as well. 

**Note:** The only issue that the container queries were solving was for the instances when you have a CTA or Signup block active on the site (causes the content to be broken into its narrow width layout and this block on desktop would still be in its 3 column layout because its relying on the screen size). 

So another option i forgot to consider, which is arguably a cleaner approach is to add a parent class when the CTA or Signup is active, that we can use to set this block to a single column. 

I'll quickly modify it to work this way

stevedya:
@tbrlpld This one is ready for you to look at again 😄

network-api/networkapi/wagtailpages/pagemodels/customblocks/block_with_aside.py
```diff
@@ -0,0 +1,19 @@
...
+
+class BlockWithAside(blocks.StructBlock):
+    content = blocks.StreamBlock([
+        ("listing", ListingBlock()),
+        ('paragraph', blocks.RichTextBlock(
+            features=full_content_rich_text_options
+        ))
+    ], max_num=1, help_text='The wider block that appears on the left on desktop', icon='doc-full')
+
+    aside = blocks.StreamBlock(aside_fields, required=False, icon='doc-full', help_text='Elements here will appear in the column on the right side of the page on desktop. This can be left blank if you would just like a 2/3 column on the left')
```
tbrlpld, line 15: What do you think about adding some line breaks here?

network-api/networkapi/wagtailpages/pagemodels/customblocks/block_with_aside.py
```diff
@@ -0,0 +1,19 @@
...
+from .full_content_rich_text_options import full_content_rich_text_options
+
+
+class BlockWithAside(blocks.StructBlock):
+    content = blocks.StreamBlock([
+        ("listing", ListingBlock()),
+        ('paragraph', blocks.RichTextBlock(
+            features=full_content_rich_text_options
+        ))
+    ], max_num=1, help_text='The wider block that appears on the left on desktop', icon='doc-full')
```
tbrlpld, line 13: What do you thing about using the same indentation for all the attributes? I think that is a bit more common. 


I know we need to set up a formatter for this repo ... 😓

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/block_with_aside.html
```diff
@@ -0,0 +1,17 @@
...
+  [&.block-with-aside_+_.block-with-aside] - Will add a margin to space out these specific blocks from each other
+  The blocks inside content will determine their own margin bottom for vertical spacing
+
+  group-[.has-sidebar] - If the parent has a sidebar like CTA or Signup Block then apply styles to convert this into a single column
+{% endcomment %}
+
+<section class="block-with-aside [&.block-with-aside_+_.block-with-aside]:tw-mt-7">
+  <div class="tw-grid tw-grid-cols-1 large:tw-grid-cols-3 tw-gap-4 large:tw-gap-7 group-[.has-sidebar]:tw-grid-cols-1 group-[.has-sidebar]:tw-gap-4 ">
+    <div class="tw-col-span-1 large:tw-col-span-2 group-[.has-sidebar]:tw-col-span-1">
+      {{ value.content }}
```
tbrlpld, line 11: This works 😲 I would have suspected we need `{% include_block %}` or something

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/listing_block.html
```diff
@@ -2,26 +2,29 @@
...
-  <div class="tw-row tw-py-4">
+  <ul class="tw-mb-7 tw-list-none tw-pl-0 tw-grid tw-gap-4 medium:tw-grid-cols-2 medium:tw-gap-7">
     {% for block in self.cards %}
       {% with card=block %}
         {% image card.image fill-350x197 as img %}
-        <div class="tw-px-4 medium:tw-w-1/2 tw-mb-5 group">
+        <li>
+          {% if card.link %}
+            <a href="{% pageurl card.link %}"
+               class="tw-group/listing hover:tw-no-underline focus:tw-no-underline hover:tw-text-black focus:tw-text-black"
```
tbrlpld, line 12: Nice. Just learned about named groups yesterday 😅

tbrlpld (commented):
I am not 100% sure, but I think the aside should sit outside of the normal width of the document. So basically in the same area where the cta/signups are shown. Like adding an aside should have the same effect as a cta/signup. 🤔 Hm... I am not sure anymore. I think this case has not been designed for. 

The body content should all have the same width, I think. 

Screen Shot 2022-10-28 at 15 47 24

cdanfon:
Let's put this onto staging and ask for Kristina's opinion @tbrlpld

network-api/networkapi/wagtailpages/pagemodels/customblocks/block_with_aside.py
stevedya: Yep thats a great idea! I've made this change now

network-api/networkapi/wagtailpages/pagemodels/customblocks/block_with_aside.py
stevedya: I've pushed the attributes to a new line now so they can be at the same indent, hopefully it's closer to what you were expecting now.

tbrlpld:
@stevedya There also seems to be an unintended change of the narrow body width that is caused by this PR. See the visual regression tests: https://percy.io/Mozilla-Foundation/foundation.mozilla.org/builds/22591214/changed/1259054366?browser=chrome&browser_ids=30%2C31&subcategories=unreviewed%2Cchanges_requested&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280

stevedya:
Hey @tbrlpld thanks for noticing those regressions, i've fixed them now so you can check this one out again. 🥲

tbrlpld (approved):

Issue #9522, opened 2022-10-21
[PNI] Update Tagline


beccaklam:
**Links to comps:**
Homepage [desktop](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=4762%3A74534) and [mobile](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3075%3A29691)
Product Reviews [desktop](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3080%3A30591), [tablet](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3467%3A41084), and [mobile](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3083%3A29539)
Creep-o-meter face [mobile](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=4893%3A105021) example

### Goals:
- Update styling on homepage and Product Reviews pages
- Creep-o-meter face animation should always begin in the bottom right hand corner instead of transitioning there
- Make text editable in the CMS, design to provide guidance on character limit
- Question for Dev: This might introduce a new type style, do we have any existing styles that are similar? Will we have too many styles after this? (devs: yes)

### Acceptance criteria:
**Content**
- [x] New tagline should say: Make smart choices to protect your privacy.
Search for products. Read expert reviews. Get tips and tricks.

**Styling**

- [x] There should be a < br > between "Make smart choices to protect your privacy." and "Search for products..."
- [x] Remove previous creepy face styling, change to text only
- [x] Creep-o-meter face animation should now always begin in the bottom right hand corner instead of transitioning there
- [x] Desktop and Tablet text styling for tagline should be Zilla Slab, Light, size 36px (new style)
- [x] Mobile text styling should be Zilla Slab, Light, size 24px (new style, technically Mobile Heading 3 with an override on font weight from regular to light, but could just set this as a new PNI tagline style instead)
- [x] Text is centered
- [x] Desktop spacing: Padding top and bottom should be 48px
- [x] Tablet and mobile padding top and bottom should be 24px

**CMS**
~Make CMS editable (stretch)~

[...]

PR #9546, opened 2022-10-25
9522 - Update Tagline

fessehaye:
# Description

Update tagline with new text/style + creepy-face only available on bottom right corner now instead of on top.

image

Link to sample test page: http://localhost:3000/en/privacynotincluded/
Related PRs/issues: #9522

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/catalog.html
+++ b/network-api/networkapi/templates/pages/buyersguide/catalog.html
@@ -11,27 +11,10 @@
   <input type="hidden" class="parent-title" value="{{current_category.parent.localized.name}}">
   <div class="tw-container">
     <div class="tw-row">
-      <div class="tw-w-full">
-        <div id="sticky-bar" class="creepiness-slider bg-white text-center tw-justify-center creep-o-meter-moved search-active tw-hidden">
-          <div class="creep-o-meter-information">
-            <p class="speech-bubble-container">
-              <span class="speech-bubble tw-bg-gradient-to-t tw-from-purple-05 tw-to-blue-05">
-                <img class="hook" src="{% static "_images/buyers-guide/icon-bubble-hook.svg" %}">
-                <span class="text"></span>
-              </span>
-            </p>
-            <span class="current-creepiness"></span>
-          </div>
-        </div>
-      </div>
-    </div>
-  </div>
-  <div class="tw-container">
-    <div class="tw-row">
       <div class="tw-px-4 tw-py-5 large:tw-py-7 tw-w-full">
-        <h1 class="tw-font-zilla tw-font-light tw-text-center tw-text-2xl large:tw-text-4xl tw-m-0">
+        <span class="tw-block tw-font-zilla tw-font-light tw-text-center tw-text-2xl large:tw-text-4xl tw-m-0">
           {% trans "Make smart choices to protect your privacy. <br /> Search for products. Read expert reviews. Get tips and tricks." %}
-        </h1>
+        </span>
       </div>
     </div>
   </div>
@@ -68,6 +51,23 @@
 {% endblock %}
 
 {% block guts %}
+  <div class="tw-container">
+    <div class="tw-row">
+      <div class="tw-w-full">
+        <div id="sticky-bar" class="creepiness-slider bg-white text-center tw-justify-center creep-o-meter-moved search-active tw-hidden">
+          <div class="creep-o-meter-information">
+            <p class="speech-bubble-container">
+              <span class="speech-bubble tw-bg-gradient-to-t tw-from-purple-05 tw-to-blue-05">
+                <img class="hook" src="{% static "_images/buyers-guide/icon-bubble-hook.svg" %}">
+                <span class="text"></span>
+              </span>
+            </p>
+            <span class="current-creepiness"></span>
+          </div>
+        </div>
+      </div>
+    </div>
+  </div>
   {% get_bg_home_page as home_page %}
   <div class="project-list-section">
     <div class="container">
--- a/network-api/networkapi/templates/pages/buyersguide/catalog.html
+++ b/network-api/networkapi/templates/pages/buyersguide/catalog.html
@@ -11,18 +11,27 @@
   <input type="hidden" class="parent-title" value="{{current_category.parent.localized.name}}">
   <div class="tw-container">
     <div class="tw-row">
-      <div class="tw-px-4 tw-w-full">
-        <div id="sticky-bar" class="creepiness-slider bg-white text-center tw-flex tw-justify-center">
+      <div class="tw-w-full">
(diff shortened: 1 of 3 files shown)


network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -11,13 +11,13 @@
   <input type="hidden" class="parent-title" value="{{current_category.parent.localized.name}}">
   <div class="tw-container">
     <div class="tw-row">
-      <div class="tw-px-4 tw-w-full">
-        <div id="sticky-bar" class="creepiness-slider bg-white text-center tw-flex tw-justify-center">
+      <div class="tw-w-full">
```
tbrlpld, line 14: Can you add some context why we need these to be two elements? Isn't a row already full width?

network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -26,6 +26,15 @@
       </div>
     </div>
   </div>
+  <div class="tw-container">
+    <div class="tw-row">
+      <div class="tw-px-4 tw-py-5 large:tw-py-7 tw-w-full">
+        <h1 class="tw-font-zilla tw-font-light tw-text-center tw-text-2xl large:tw-text-4xl tw-m-0">
+          {% trans "Make smart choices to protect your privacy. <br /> Search for products. Read expert reviews. Get tips and tricks." %}
+        </h1>
```
tbrlpld, line 34: Can you add some context why this is the `h1`? This seems rather long for the title / top level heading of a page. What do you think about using "Privacy not included" as the `h1` on the home page and something like "product reviews" on the category pages? If there is nothing in the design that could act as that `h1` we could make that for screen readers only.

network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -26,6 +26,15 @@
       </div>
     </div>
   </div>
```
tbrlpld, line 28: What do you think about moving this into the `guts` block, since we are not showing it in the header anymore?

tbrlpld (commented):
Thanks for this @fessehaye. This is coming a long nicely. I have left a few small questions.

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: Rows in bootstrap are only responsible for adding flex and the negative margin. You always need a column class or to establish width else it will be width:auto, which is not what we want in this situation.

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: changed this to a span

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: moved

network-api/networkapi/templates/pages/buyersguide/catalog.html
tbrlpld: Thanks @fessehaye this is helpful. 

Just wondering, do we still need to follow that Bootstrap convention of container-row-col? I can't really find any equivalent of the Bootstrap `col` classes for use with Tailwind.

network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -11,18 +11,10 @@
...
-                <span class="text">{% trans "We created this guide to help you shop<br> for safe, secure connected products." %}</span>
-              </span>
-            </p>
-            <span class="current-creepiness"></span>
-          </div>
-        </div>
+      <div class="tw-px-4 tw-py-5 large:tw-py-7 tw-w-full">
+        <span class="tw-block tw-font-zilla tw-font-light tw-text-center tw-text-2xl large:tw-text-4xl tw-m-0">
+          {% trans "Make smart choices to protect your privacy. <br /> Search for products. Read expert reviews. Get tips and tricks." %}
+        </span>
```
tbrlpld, line 17: Can you add some context why we use `span` with `tw-block` instead of `div`? I have seen this elsewhere too. I am just wondering if I am missing something about that.

tbrlpld (approved):
Thanks for these changes @fessehaye. I have left some questions for my understanding, the PR is approved. 

Also, when you force-push while the PR is under review it makes it really hard for me to follow the changes. It might be better to wait with force-push until the review is done.

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: Spans are not display block by default unlike other elements, so text-align does not work for inline elements! 

Good to know about the force pushes!

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: https://tailwindcss.com/docs/width Since we have `x/12` widths on tailwind. I do not really think adding the bootstrap columns back add any real value.

tbrlpld: True, but aren't those columns what creates the gutters between columns? I guess the question is, are the rows useful without the cols?

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: I do not think there is much different from div/spans but I like to use spans for text-content so its easier to use querySelector stuff in the future so I don't have to add unique ids/classes. I honestly think both are fine and it is just preference but I should look up which is more semantic.

tbrlpld: > Spans are not display block by default unlike other elements, [...]

That's why I was wondering why not just use a `div` if we want it to be a block element anyways 🤔

tbrlpld: > I do not think there is much different from div/spans but I like to use spans for text-content so its easier to use querySelector stuff in the future so I don't have to add unique ids/classes. I honestly think both are fine and it is just preference but I should look up which is more semantic.

Thanks for expanding on that 👍

network-api/networkapi/templates/pages/buyersguide/catalog.html
tbrlpld: It looks like Tailwind creates the space between row elements with `space-x-...`. 



I wonder if that negative margin + padding hack for the columns as in Bootstrap is still needed.

fessehaye: So the trick is col-4 in bootstrap is the same as px-4 w-4/12 in tailwind. If you just add px-4 to column like elements it will cover the padding/negative margin. I documented this on the tailwind epic!

Issue #8034, opened 2021-12-20
Figure out why the pulse filters don't show up in the list of permissions


Pomax:
Pulse filters show up in the list of snippets:

![image](https://user-images.githubusercontent.com/177243/146805723-3155bf75-0b0d-4b90-a77e-bb87507bd030.png)

But they are nowhere to be found in the list of items that can be assigned permissions when creating/editing a group.

Code:

https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/pulse.py#L9

richbrennan:
Findings so far:

- Pulse Filter shows in Group permissions locally with production DB backup, but not on staging or production

- Clearing the staging cache and then restarting the dynos doesn't help

This seems to be something to do with permissions not being present on staging/production. [This Group form code](https://github.com/wagtail/wagtail/blob/v2.15.1/wagtail/users/forms.py#L204) fetches the registered [permissions for snippets ](https://github.com/wagtail/wagtail/blob/v2.15.1/wagtail/snippets/wagtail_hooks.py#L39)when constructing the 'Object permissions' list on the Group admin page.

When recreating the call in the shell locally vs staging the list of permission objects is empty on staging:

Local:
```>>> from django.contrib.auth.models import Permission
>>> from django.contrib.contenttypes.models import ContentType
>>> from networkapi.wagtailpages.pagemodels.pulse import PulseFilter
>>> ct = ContentType.objects.get_for_model(PulseFilter)
>>> Permission.objects.filter(content_type=ct)
<QuerySet [<Permission: wagtailpages | pulse filter | Can add pulse filter>, <Permission: wagtailpages | pulse filter | Can change pulse filter>, <Permission: wagtailpages | pulse filter | Can delete pulse filter>, <Permission: wagtailpages | pulse filter | Can view pulse filter>]>
```

Staging:
```>>> from django.contrib.auth.models import Permission
>>> from django.contrib.contenttypes.models import ContentType
>>> from networkapi.wagtailpages.pagemodels.pulse import PulseFilter
>>> ct = ContentType.objects.get_for_model(PulseFilter)
>>> Permission.objects.filter(content_type=ct)
<QuerySet []>
```

So to progress we need to investigate why the permissions don't appear to be present on staging and production.

Pomax:
[...]

PR #9547, opened 2022-10-26
Exit release script when release commands fail

tomusher:
Prevents releases from succeeding when migrations fail. As discovered in #8034

### DIFF
--- a/release-steps.sh
+++ b/release-steps.sh
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -eo pipefail
 
 cd network-api
 


release-steps.sh
```diff
@@ -1,4 +1,5 @@
 #!/usr/bin/env bash
+set -eo pipefail
```
tbrlpld, line 2: @tomusher Does the whole file count as a pipeline?

tbrlpld (approved):

release-steps.sh
tomusher: No, the main difference here is the inclusion of `set -e`. The `o pipefail` is just future proofing in case we do introduce some pipelines to this script.

tomusher:
Thanks for reviewing @tbrlpld !

release-steps.sh
tbrlpld: 💡 thanks for the explanation @tomusher

Issue #9557, opened 2022-10-26
[PNI QA] Cannot select the category nav on a product review page (Mobile - any browser)


beccaklam:
The category nav drop down menu is not working on the product review page on mobile.

To replicate the bug:
1. Go to https://foundation-pni-2022.herokuapp.com/privacynotincluded/ on your mobile device
2. Click into any product review, such as https://foundation-pni-2022.herokuapp.com/privacynotincluded/amazon-echo-show/
3. Try to tap the drop down menu besides "All Categories" or even "All Categories" itself
4. See bug (nothing happens)

Note: This appears to work on Firefox desktop when you adjust the view width to a mobile size but not on a mobile device itself (at least on iOS)

tbrlpld:
I can even reproduce on desktop Firefox with small viewport.

cdanfon:
Currently with Tibor for code review

fessehaye:
ready on staging

beccaklam:
This is fixed! ✅

PR #9568, opened 2022-10-27
9557 - Cannot select the category nav on a product review

fessehaye:
# Description

I would like to address `source/js/buyers-guide/analytics-events.js` usage at a later date. I do not know if any of the analytics are still being used and this is not the first time this file has caused problems on your client side JS

Link to sample test page:
Related PRs/issues: #9557

### DIFF
--- a/source/js/buyers-guide/bg-main.js
+++ b/source/js/buyers-guide/bg-main.js
@@ -76,8 +76,8 @@ let main = {
         document.body.classList.add(`react-loaded`);
         this.initPageSpecificScript();
         // bind custom analytics only once everything's up and loaded
-        AnalyticsEvents.init();
         PNIMobileNav.init();
+        AnalyticsEvents.init();
       });
     });
   },
--- a/source/js/buyers-guide/bg-main.js
+++ b/source/js/buyers-guide/bg-main.js
@@ -75,8 +75,9 @@ let main = {
       Promise.all(apps).then(() => {
         document.body.classList.add(`react-loaded`);
         this.initPageSpecificScript();
-        // bind custom analytics only once everything's up and loaded
         PNIMobileNav.init();
+        // bind custom analytics only once everything's up and loaded
+        // Analytics events does give errors quite often, do not add JS after this
         AnalyticsEvents.init();
       });
     });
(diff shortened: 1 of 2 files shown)


tbrlpld:
@fessehaye Sorry, but I don't understand the PR description. What are we doing here? What does this have to do with analytics?

tbrlpld (commented):
See above

fessehaye:
@tbrlpld  analytics keeps throwing errors of missing selectors which means PNIMobileNav never gets initalized. I think for long term stuff we should look into removing or modifying the analytics-events.js but I think I suggested this last year but was told not? I also think that is past the scope of this bug.

source/js/buyers-guide/bg-main.js
```diff
@@ -76,8 +76,8 @@ let main = {
         document.body.classList.add(`react-loaded`);
         this.initPageSpecificScript();
         // bind custom analytics only once everything's up and loaded
-        AnalyticsEvents.init();
         PNIMobileNav.init();
+        AnalyticsEvents.init();
```
tbrlpld, line 80: If I understand correctly you are changing the order here because errors in `AnalyticsEvents.init()` prevent `PNIMobileNav.init()` from ever running?

source/js/buyers-guide/bg-main.js
```diff
@@ -76,8 +76,8 @@ let main = {
         document.body.classList.add(`react-loaded`);
         this.initPageSpecificScript();
         // bind custom analytics only once everything's up and loaded
```
tbrlpld, line 78: Should we move this comment with the `AnalyticsEvents.init()`?

source/js/buyers-guide/bg-main.js
tbrlpld: What do you think about adding a comment about the flakiness of the analytics initialization to avoid other scripts being added after it?

tbrlpld (commented):
@fessehaye I left some small questions.

source/js/buyers-guide/bg-main.js
fessehaye: Yeah I can do that!

tbrlpld (approved):

tbrlpld:
@fessehaye Looks like another one that got merged into the wrong branch. All PRs should be updated to point to `main`. I was expecting this to happen automatically and that `pni-2022` is delete.

Issue #9553, opened 2022-10-26
[PNI QA] Search not working on product review pages


beccaklam:
On a product review page (not to be confused with the Product Reviews page), search is not working.

To replicate bug:
1. Go to https://foundation-pni-2022.herokuapp.com/privacynotincluded/ on any device (desktop, tablet, mobile)
2. Click on any product review like https://foundation-pni-2022.herokuapp.com/en/privacynotincluded/threema/
3. Go to the search bar
4. Type something into the search bar and hit enter
5. See bug (nothing happens)

mtdenton:
@fessehaye seems like javascript errors might be causing this and #9557: 
<img width="1723" alt="Screen Shot 2022-10-26 at 6 20 30 PM" src="https://user-images.githubusercontent.com/41487159/198157399-27b949bd-0216-4cb1-8246-f4c7fb7ca7f2.png">

fessehaye:
I think Becca was referring to the desktop screen? But good to know that mobile isn't working either. This overlaps some of the nav stuff I am doing so I will grab this one. @mtdenton

mtdenton:
It adjusted viewport size because I opened the dev tools, just trying to show that there are javascript errors that need addressed.

beccaklam:
It's for all screens actually, I think @fessehaye

tbrlpld:
Mobile (Desktop Firefox with small viewport) works for me. Desktop (Firefox) does not work.

tbrlpld:
Works for me on iOS (Safari)

tbrlpld:
I have tried cleaning this up, but it turns out to be more work than expected. Looks like a refactor has been needed for a while: https://github.com/mozilla/foundation.mozilla.org/issues/8015

I am moving forward with this as is, but would propose to follow up with #8015 as soon as possible.

tbrlpld:
@beccaklam This is on staging ready for testing.

beccaklam:
This looks good to me!

Issue #8015, opened 2021-12-15
PNI - Clean up search filter code


Pomax:
Follow-up to https://github.com/mozilla/foundation.mozilla.org/issues/7988 

With the refactor complete, we should consolidate/optimize the search filter code as much as possible.

## Dev tasks: 
* [ ] Consolidate mobile and desktop
* [ ] Renaming 

(to be groomed after the team has the opp to discuss and agree on approach)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-257)

tbrlpld:
This is 100% necessary. I find it extremely hard to understand the search code.

tbrlpld:
Also follow up to #9553

tbrlpld:
I would like to propose to make the search a individual template fragment that can be included multiple times on a page (e.g. for mobile and desktop). We can pass an id into the fragment. 

Similarly, all JS should be generic and only need the id used for the fragment to instantiate a new search bar.

mtdenton:
This should belong to a PNI refactor epic.

PR #9570, opened 2022-10-27
9553 - Fix Search on Desktop

fessehaye:
# Description

Search bar on desktop should now be working on non-product search pages. Just needed to add similar JS that allows the mobile search bar to work globally.

Link to sample test page: go to /privacynotincluded and click on a page or article
Related PRs/issues: #9553


tbrlpld (commented):
@fessehaye I am confused, the PR is supposed to be for the desktop search but all the changes address the mobile search? Am I missing something here?

fessehaye:
@tbrlpld before, the search bar logic outside of product review pages only worked on the mobile-visible search bar. Since I added the regular search bar references I just renamed the existing mobile variables so it is easier to understand which are being used for each case!

tbrlpld:
@fessehaye I am still confused where the changes regarding the desktop search are. All the changes are in `/mobile-search-bar.js`

tbrlpld (commented):
See above

tbrlpld (commented):
@fessehaye Sorry, I still don't understand why are copying the mobile search functionality instead of creating a generic search bar that works regardless of the viewport sizes / location of the search bar. Can you add some context to that decision?

fessehaye:
> @fessehaye Sorry, I still don't understand why are copying the mobile search functionality instead of creating a generic search bar that works regardless of the viewport sizes / location of the search bar. Can you add some context to that decision?

@tbrlpld Are you referring to the the js files being different or the html elements?

tbrlpld:
> @tbrlpld Are you referring to the the js files being different or the html elements?

I was referring to the JS files, because that's what is updated in this PR. 

But, if there are two HTML templates I am also wondering why we have that twice. It seems to me that the search bar elements would be the same only with different widths (which could be limited by the wrapping element). But that's not something we need to address in this PR.

fessehaye:
> > @tbrlpld Are you referring to the the js files being different or the html elements?
> 
> I was referring to the JS files, because that's what is updated in this PR.
> 
> But, if there are two HTML templates I am also wondering why we have that twice. It seems to me that the search bar elements would be the same only with different widths (which could be limited by the wrapping element). But that's not something we need to address in this PR.

Okay, `mobile-search-bar.js` does have unique logic related to toggling the mobile search bar visibility that should not be included in a generic search bar js file. If that is the case, do we want to create a generic search bar js file and add that to the desktop/mobile js files?

tbrlpld:
> Okay, `mobile-search-bar.js L2-16` does have unique logic related to toggling the mobile search bar visibility that should not be included in a generic search bar js file. If that is the case, do we want to create a generic search bar js file and add that to the desktop/mobile js files?

Is the desktop search module needed if this is basically only the generic search? If you think it's valuable for some setup code we could do that. 

If there is no extra file needed for the setup code of the search, we could also just have a search module and one for the search visibility toggle.

tbrlpld (commented):
@fessehaye Sorry I am confused, I don't seem to see the generic search module you mentioned.

fessehaye:
Hey @tbrlpld, I think there is some confusion about what needs to be split into modules and what not. Can we just make a ticket to document this discussion/suggested changes, so this ticket can be pushed to production if there isn't other problems related to this task?

tbrlpld:
@fessehaye I weary of the tech debt we are accumulating within PNI already. I don't think it we should keep adding things to do in The Refactor™️ because no one knows if it ever comes. And in this case we always have to remember to duplicate all the functionality in both files. 

And it appeared like we were on the same page since you mentioned this: 

> Okay, mobile-search-bar.js L2-16 does have unique logic related to toggling the mobile search bar visibility that should not be included in a generic search bar js file. If that is the case, do we want to create a generic search bar js file and add that to the desktop/mobile js files?

From which I assumed we would put all the shared logic into the `search-bar.js` and then use that functionality in `desktop-search-bar.js` and `mobile-search-bar.js`. Additionally `mobile-search-bar.js` would have some extra logic for the toggling of the visibility. 

I was imagining something where `desktop-search-bar.js` and `mobile-search-bar.js` import something like `setupSearchBar()` from `search-bar.js` and pass in the `id`, `class`, object or what ever is that is different between the two.

tbrlpld (commented):
@fessehaye See last comment

tbrlpld (commented):
Thanks @fessehaye. This feels a lot cleaner to me. 

And I just realized where some of the confusion has come from. I have been assuming that the search bar is a template fragment itself that is included in two locations. But it turns out that this the markup is included directly in two separate locations and slightly different. 

I was wondering what you think about turning this into a template fragment that we can reused in both locations? Of course the width is different, but that could be solved by making the fragment fill 100% width and having a wrapper around the include that constrains the size/width. 

Something like 
```html+django
<div class="tw-w-96">
  {% include "fragments/buyersguide/search_bar.html" %}
</div>
```

If we need to, we could pass some id into the include as well.

tbrlpld:
Ah, looks like the tablet version does have another difference by having this inner content wrapper that restricts that width. 

image

tbrlpld (approved):
Moving forward as is, but I think we should follow up with a refactor of the search/filter code for the buyer's guide (#8015).

Issue #9392, opened 2022-09-20
DFL 2022 | Introduction changes


kevinhowbrook:
### Describe the feature request
Currently, the introduction in the banner, below the section title comes from the child; this is a bit confusing since the title comes from the section parent. The instances I have found all copy the same intro. This should be updated to use also the introduction from the parent. 

### Screenshots / mock-ups
![Screenshot](https://user-images.githubusercontent.com/2128707/191251008-6925f3f5-2037-4d21-a4b5-426724372169.png)

### Additional context
Link to [designs](https://www.figma.com/file/goD9bRqPoedQAxi8OsInkR/Data-Futures-Lab%3A-Designs?node-id=626%3A9963)

**Estimate**
1 day

kristinashu:
How will https://github.com/mozilla/foundation.mozilla.org/issues/9391 affect this? If the top isn't reloading then I'm not sure the intro text can change?

Can you please clarify "Currently, the introduction in the banner, below the section title comes from the child." I tested this by removing the intro text from the Overview here https://foundation.mofostaging.net/en/what-we-fund/ and it seems to just not show any intro text, which is good.

kristinashu:
Also please note that mobile layout should remain the same as it is currently on prod. The intro text should not disappear on mobile as it does in the Figma design.

stevedya:
Hey @tbrlpld i've added a PR for this that is ready for review here: https://github.com/mozilla/foundation.mozilla.org/pull/9573

tbrlpld:
I left some questions @stevedya

stevedya:
This one is ready for you to look at again @tbrlpld

cdanfon:
Hey @tbrlpld curious to know where we got to with this one, it was a nice to have for this project but would be good to get into on production if it's been addressed by Steve

tbrlpld:
@cdanfon I think this is basically done. We only need to update the branch with the latest changes from `main` and fix possible merge issues.

cdanfon:
Thanks @tbrlpld can I leave this with you please? 

Let me know if you have the capacity this sprint otherwise we can look into incorporating this into the upcoming one.

tbrlpld:
@cdanfon Yea, you can assign to me. Can you make sure this shows on the Zenhub board too?

cdanfon:
Sorted, I've given it a '1' as an estimate - feel free to update it if you think it's not right :)

mtdenton:
[...]

PR #9573, opened 2022-10-28
#9392 [DFP] - Allowing intro to generate from parent page on bannered campaigns

stevedya:
# Description

This PR allows BanneredCampaign pages to have an option to use the intro from their parent. 

- Added an option to the BE for use_intro_from_parent, defaulting this ensures that any previous bannered pages on the main site wont be affected. 
- Conditionally choose the intro based on if the page is using parent's intro, has its own intro or has none

Related PRs/issues: #9392 

## Screenshot
Screen Shot 2022-10-28 at 3 01 58 PM

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [X] Is my code documented?


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
```diff
@@ -47,8 +47,10 @@ <h1 class="tw-h1-heading">{{ page.hero_headline }}</h1>
     <div class="row cutout-wrapper-row">
       <div class="cutout col-12 col-lg-8 pt-4">
         <h1 class="tw-h1-heading mb-0">{% if root.title %}{{ root.title }}{% elif page.header %}{{ page.header }}{% else %}{{ page.title }}{% endif %}</h1>
-        {% if page.intro %}
-        <p class="tw-body-large mt-3 mb-0">{{ page.intro }}</p>
+        {% if page.use_intro_from_parent and root.specific.intro %}
+          <p class="tw-body-large mt-3 mb-0">{{ root.specific.intro }}</p>
+        {% elif page.intro %}
+          <p class="tw-body-large mt-3 mb-0">{{ page.intro }}</p>
```
tbrlpld, line 53: What do you think about moving this to a method on the page and adding some tests for it? There are quite a few cases to test and I think it would be easier to test all these path automatically rather than do it in the template.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -374,15 +374,29 @@ class BanneredCampaignPage(PrimaryPage):
     tags = ClusterTaggableManager(through=BanneredCampaignTag, blank=True)
     aside = StreamField(aside_fields, blank=True)
 
+    use_intro_from_parent = models.BooleanField(
+        default=False,
+        blank=False,
+        help_text='This field will overwrite the intro field above with the intro field from this pages parent'
```
tbrlpld, line 380: Since this field does not contain the override but rather a toggle, what do you think about the following change?

```suggestion
        help_text='Overwrite the intro field above with the intro field from this pages parent'
```

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -374,15 +374,29 @@ class BanneredCampaignPage(PrimaryPage):
...
+    )
+
     panel_count = len(PrimaryPage.content_panels)
-    n = panel_count - 1
+    introPanel = PrimaryPage.content_panels.pop(-2)
+
+    n = panel_count - 2
```
tbrlpld, line 386: Can you add some context what this is doing? And what do you think about this being a method with some tests?

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -374,15 +374,29 @@ class BanneredCampaignPage(PrimaryPage):
...
 
+    use_intro_from_parent = models.BooleanField(
+        default=False,
+        blank=False,
+        help_text='This field will overwrite the intro field above with the intro field from this pages parent'
+    )
+
     panel_count = len(PrimaryPage.content_panels)
-    n = panel_count - 1
+    introPanel = PrimaryPage.content_panels.pop(-2)
```
tbrlpld, line 384: I think using snake case would be more consistent with the rest of the Python code.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -374,15 +374,29 @@ class BanneredCampaignPage(PrimaryPage):
...
+        ],
+            heading='Intro options',
+            help_text='Select a parent intro, write your own intro or leave both blank for no intro'),
         SnippetChooserPanel('cta'),
         SnippetChooserPanel('signup'),
     ] + PrimaryPage.content_panels[n:] + [
-        StreamFieldPanel('aside'),
-    ]
+                         StreamFieldPanel('aside'),
+                     ]
```
tbrlpld, line 399: Can you add some context on why we are using this larg indentation here?

tbrlpld, line 399: It looks like `PrimaryPage.content_panels` is 4 lines. What do you think of just duplicating those lines here?

tbrlpld (commented):
@stevedya I have left a few questions around this. While this seems like a simple change I think there are a lot of edge cases to consider here and I think we might be better served with some testing to make sure the behavior is what we want.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: great suggestion, i've made this change now and it solves all those other issues mentioned 🎉

stevedya: Fixed!

tbrlpld:
@stevedya This one seems to have some formatting issues preventing CI from passing.

tbrlpld (commented):
Thanks for the changes @stevedya. Only the question regarding the move of the logic to a method remains.

stevedya:
@tbrlpld I think its a great idea and I can work on moving this today 🙂

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
stevedya: Excellent suggestion! I'll work on this quickly today. I noticed the campaign pages don't have their own testing setup, should this test be on the primary page?

tbrlpld: I don't think this should be affecting the primary page 🤔 You might need to create the test setup. You can might be able to use the primary page test setup as an example (maybe even share setup?)

stevedya: I've revised this a bit now, for you to look at again :)

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -436,10 +435,20 @@ class BanneredCampaignPage(PrimaryPage):
 
     show_in_menus_default = True
 
+    def get_intro(self):
+        if self.use_intro_from_parent:
+            parent = self.get_parent()
+            if parent:
+                return parent.specific.intro
```
tbrlpld, line 442: I believe this could raise an exception if the bannered page is nested under a page that does not have the `intro` field. What do you think about adding a test for that case? Should we return an empty string for that case?

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -377,26 +379,23 @@ class BanneredCampaignPage(PrimaryPage):
     use_intro_from_parent = models.BooleanField(
         default=False,
         blank=False,
-        help_text='This field will overwrite the intro field above with the intro field from this pages parent'
+        help_text='Overwrite the intro field above with the intro field from this pages parent'
     )
```
tbrlpld, line 383: What do you think about validating this field with a check to see if the pages parent does actually have an `intro` field? It might be better for the editors to have clear error message instead of surprising them with no intro showing up while it seems like they expect one, because they selected this option.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -436,10 +435,20 @@ class BanneredCampaignPage(PrimaryPage):
 
     show_in_menus_default = True
 
+    def get_intro(self):
+        if self.use_intro_from_parent:
+            parent = self.get_parent()
+            if parent:
```
tbrlpld, line 441: Can this fail? I think except for the root object all pages will have a parent.

tbrlpld (commented):
@stevedya I have left a few questions around the `get_intro` method.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Oops, guess i just got so used to my real life `parents === None`  😂 

Sorry bad joke haha

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Although, I think its better to be more explicit with whats happening so I've gone ahead an added that validation as you've suggested 😄

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Sure! I've modified this now and added a test for the case where the parent intro is None

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
```diff
@@ -47,10 +47,8 @@ <h1 class="tw-h1-heading">{{ page.hero_headline }}</h1>
     <div class="row cutout-wrapper-row">
       <div class="cutout col-12 col-lg-8 pt-4">
         <h1 class="tw-h1-heading mb-0">{% if root.title %}{{ root.title }}{% elif page.header %}{{ page.header }}{% else %}{{ page.title }}{% endif %}</h1>
-        {% if page.use_intro_from_parent and root.specific.intro %}
-          <p class="tw-body-large mt-3 mb-0">{{ root.specific.intro }}</p>
-        {% elif page.intro %}
-          <p class="tw-body-large mt-3 mb-0">{{ page.intro }}</p>
+        {% if self.intro or self.get_intro %}
+          <p class="tw-body-large mt-3 mb-0">{% firstof self.get_intro self.intro %}</p>
```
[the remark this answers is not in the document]
stevedya: Sure! It seems like a big daunting change but i've done it anyway 😅. I also had to add it to the mozfest primary page too and extracted the home banner to its own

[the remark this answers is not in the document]
stevedya: Oh actually seen now that mozfest uses a different heroguts.html file haha

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
tbrlpld: I think it's fine to be defensive about these things. Maybe we just make this more explicit by defining the return value for that add a test for the case.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
tbrlpld: Sorry, I should have been clearer. I meant the case where the banner page is a child of a different page type, specifically one that does not have the `intro` field (e.g. `Homepage`).

tbrlpld: But the test you added is good nonetheless. There should really be one test for each branch of each if statement. Which is also why I like to fail/return early if we know we can not get further.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/primary_heroguts.html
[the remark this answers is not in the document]
tbrlpld: Ha yea, I was confused about that for a minute too.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -422,11 +436,30 @@ class BanneredCampaignPage(PrimaryPage):
...
     class Meta:
         verbose_name = "Banner Page"
         verbose_name_plural = "Banner pages"
+
+    def clean(self):
+        if self.use_intro_from_parent and self.intro:
+            raise ValidationError({
+                'use_intro_from_parent': ValidationError("Please use only one intro option."),
+                'intro': ValidationError("Please use only one intro option.")
+            })
```
tbrlpld, line 465: Nice. A different check than I meant, but a good one still!


I was thinking about something like (pseudo-code):
```
if use_intro_from_parent:
    parent = get_parent
    if not has_intro_field(parent):
        raise ValidationError
```

tbrlpld (commented):
Thanks for the changes @stevedya! We are almost there. I think just some minor handling of some edge cases and we got it. 


Thanks also for fixing that hero template 🎉

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Thanks for explaining further, i've made some changes to this now that should hopefully be sufficient

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Ahh yes thanks for this. I've added validation for this as well now.

stevedya:
Hey @tbrlpld I've made some changes to the validations and logic and it's ready again for you to take a look. Thanks!

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -418,6 +433,15 @@ class BanneredCampaignPage(PrimaryPage):
 
     show_in_menus_default = True
 
+    def get_intro(self):
+        if self.use_intro_from_parent:
+            parent = self.get_parent()
+            try:
+                return parent.specific.intro
+            except AttributeError:
+                return None
```
tbrlpld, line 442: **Kudos** Nicely done, handling both cases (no parent, and a parent with no intro) in one 👌

tbrlpld (approved):
@stevedya This looks all good to me. Unfortunately I think we need to recreate the migrations again and CI also shows some issue.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
stevedya: Yas yay! Thanks! I tried to keep it simple as possible haha

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -426,3 +450,16 @@ def get_context(self, request):
...
+            raise ValidationError({
+                'use_intro_from_parent': ValidationError("Please use only one intro option."),
+                'intro': ValidationError("Please use only one intro option.")
+            })
+        if self.use_intro_from_parent:
+            parent = self.get_parent()
+            if not hasattr(parent.specific, 'intro'):
+                raise ValidationError({
+                    'use_intro_from_parent': ValidationError("Parent intro field does not exist"),
+                })
```
tbrlpld, line 465: Just noticing this can fail when the page is first created, because there is no parent that can be retrieved at that point.

tbrlpld (changes requested):

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
tbrlpld: Maybe I can do something with a  custom form. [https://docs.wagtail.org/en/stable/advanced_topics/customisation/page_editing_interface.html#custom-edit-handler-forms](https://docs.wagtail.org/en/stable/advanced_topics/customisation/page_editing_interface.html#custom-edit-handler-forms)

I probably want to add a test for this as well.

Issue #9552, opened 2022-10-26
[PNI QA] Product Reviews page loads too low on the screen


beccaklam:
Currently when one clicks "Product Reviews" in the menu it jumps to the part of the page right above the privacy ding and 'sort by creepiness' filter:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/67de65f9-b5ca-45e0-9986-f909310725c6)

The expected behaviour is that the page would load at the top of the page, including the navigation:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/9b30d57e-9b09-43e3-911f-af9db7b69633)

tbrlpld:
The title and the description don't seem to match @beccaklam

tbrlpld:
I feel like I have raised that before somewhere... 🤔

PR #9576, opened 2022-10-28
9552 - Product Reviews scrolls too low

fessehaye:
# Description

`window.scrollTo(0,0)` Seems to be missing when `product-review` link gets clicked!

Link to sample test page: /en/privacynotincluded -> click on `product review`
Related PRs/issues: #9552

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
@@ -1,6 +1,6 @@
 {% load bg_nav_tags localization i18n static wagtailroutablepage_tags wagtailmetadata_tags %}
 <nav id="multipage-nav" class="pni-category-nav text-center d-none d-md-block tw-no-scrollbar" title="{% trans "site navigation" context "Tooltip on menu items" %}">
-    <div class="container tw-py-2" id="product-review">
+    <div class="container tw-py-2" id="buyersguide-category-container">
       <div class="row">
         <div class="col">
           <div id="pni-category-wrapper" class="tw-flex tw-items-center tw-w-max tw-min-w-full">
--- a/source/js/buyers-guide/search/member-functions.js
+++ b/source/js/buyers-guide/search/member-functions.js
@@ -217,6 +217,7 @@ export function setupReviewLinks(instance) {
         nav.classList.add("active");
         location.hash = "product-review";
         editorialContent.classList.add("tw-hidden");
+        window.scrollTo(0, 0);
         if (burger && burger.classList.contains("menu-open")) {
           document.querySelector(".burger").click();
         }
--- a/source/js/buyers-guide/search/member-functions.js
+++ b/source/js/buyers-guide/search/member-functions.js
@@ -209,7 +209,9 @@ export function setupReviewLinks(instance) {
 
   for (const nav of navLinks) {
     nav.addEventListener("click", (evt) => {
-      const editorialContent = document.querySelector(".editorial-content");
+      const editorialContent = document.querySelector(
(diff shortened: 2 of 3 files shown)


tbrlpld (commented):
@fessehaye This works when clicking the link from the homepage, but there is still this scrolling down when clicking the "product reviews" link on a category page or when loading the URL (`http://localhost:8000/en/privacynotincluded/#product-review`) directly.


https://user-images.githubusercontent.com/24797493/199320895-f3d899e4-c71a-45c2-a52b-19f6478c1ac7.mov

network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
```diff
@@ -1,6 +1,6 @@
 {% load bg_nav_tags localization i18n static wagtailroutablepage_tags wagtailmetadata_tags %}
 <nav id="multipage-nav" class="pni-category-nav text-center d-none d-md-block tw-no-scrollbar" title="{% trans "site navigation" context "Tooltip on menu items" %}">
-    <div class="container tw-py-2" id="product-review">
+    <div class="container tw-py-2" id="buyersguide-category-container">
       <div class="row">
         <div class="col">
           <div id="pni-category-wrapper" class="tw-flex tw-items-center tw-w-max tw-min-w-full">
```
tbrlpld, line 6: These ids (`buyersguide-category-container` and `pni-category-wrapper`) are confusingly similar. Do we need a wrapper and a container? Could we use one in place of the other or is there a crucial difference I am missing?

tbrlpld (commented):
@fessehaye I have left a question about the similar ids. 

This is working when clicking "product reviews" on the homepage and hides the editorial content. When loading the URL directly it also generates the view without the editorial content immediately. But, when I am on a category page, it only adds the `#product-reviews` hash to the URL and stays where it is. I would have expected to get back to the "all reviews" state. 



https://user-images.githubusercontent.com/24797493/201211799-52fbbe35-e9fe-446c-9911-aba58a8d6e82.mov

network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
fessehaye: We need both of these for `pni-category-dropdown.js` and they are both essential for calculating what should be added to the dropdown.

tbrlpld: Are there any more expressive names we could use?

tbrlpld (commented):
@fessehaye There are no commits or comments since the last review? Am I missing something?

fessehaye:
Sorry! Probably accidentally clicked when I was doing the other PR! Should be working now!

tbrlpld (commented):
Can you add some context why bypassing the event handler setup when the editorial content is hidden should be the fix? It is not obvious to me.

tbrlpld:
It does look like the behavior is now changed and clicking the "product reviews" link does a full page reload (instead of nothing). But this still feel odd when we have already collapsed the editorial content. I would expect the link to have the same effect as the "all reviews" one in the category nav. 

When the editorial content is collapsed and all reviews are shown, clicking the "product reviews" link does nothing (except adding the location to the URL) -- this seems ok. But the full page reload when a category filter is active seems odd. 

There is also one odd thing in that all cards are hidden when the "all reviews" category link is clicked after the "product reviews" link.




https://user-images.githubusercontent.com/24797493/204660108-883fad37-504c-4dee-b8e8-3f1cec047cc7.mov

fessehaye:
Better to start from scratch from this PR since I am leaving so I will close this one!

Issue #9542, opened 2022-10-25
[PNI] Positioning of related articles on article page


tbrlpld:
Positioning of the related articles does not match the [design](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3406%3A38682). 

![image.png](https://images.zenhubusercontent.com/61e1e4f0d1fded7471402d12/f325f305-8459-4231-bb02-bd7623084118)

beccaklam:
Adding this mobile screen shot to this bug.
![IMG_EA8D905AE42A-1.jpeg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/3a6a1cdd-89f8-42ed-92bc-abf51e4e8323)
 Not sure why it's saying "Primary Related Articles"?

tbrlpld:
@beccaklam That was just a marker for us in the template. I don't think we expected the frontend to be not being properly worked on for so long.

beccaklam:
@fessehaye Hi Simon, thanks for working on this. I'm a bit confused about the placement of the related articles. The comp looks like this:
![image](https://user-images.githubusercontent.com/20977376/200921501-3f4d9265-c779-487d-b1fb-c17cbc7a6cf9.png)
But what I'm seeing on staging is this:
<img width="758" alt="Screenshot 2022-11-09 at 11 16 37 AM" src="https://user-images.githubusercontent.com/20977376/200921555-095068e2-6626-4982-b58d-45a817b30ec8.png">
I know that we had a conversation with Nancy where we decided we would have the related articles at the end of an article but _only_ in the niche situation you mentioned where there was very little content (e.g. one photo no paragraph block). Was there some kind of mix up or am I missing something?

beccaklam:
We also would need more spacing between the email CTA box and the "What to read next" section:
![image](https://user-images.githubusercontent.com/20977376/200922135-3040912f-0179-483c-a5fe-3628412afea0.png)

kristinashu:
Note that this is not a prod push blocker since we can just not add the related articles until this is done.

beccaklam:
Follow up ticket in #9664. Closing this ticket!

PR #9588, opened 2022-10-31
9542 - related articles on article page

fessehaye:
# Description

Added Related Content positioning on article pages.

Will use similar floating position from FIgma when there is enough content from cms.body. If JS is disabled or not enough content, related articles will appear after CMS rich text is added.

Link to sample test page: any PNI article
Related PRs/issues: #9542

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/article_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/article_page.html
@@ -49,10 +49,13 @@
 
   {% with related_articles=page.get_primary_related_articles  %}
     {% if related_articles %}
-      {% trans "PRIMARY RELATED ARTICLES" %}
-      <div class="tw-m-4">
-        {% include "fragments/buyersguide/related_reading.html" with articles=related_articles %}
+    <div class="tw-container tw-my-4">
+      <div class="tw-row">
+        <div class="tw-px-4 tw-w-8/12 tw-mx-auto" id="pni-related-content">
+          {% include "fragments/buyersguide/related_reading.html" with articles=related_articles %}
+        </div>
       </div>
+    </div>
     {% endif %}
   {% endwith %}
 
@@ -101,6 +104,5 @@
       </div>
     {% endif %}
   {% endwith %}
-
 </div>
 {% endblock guts %}
--- a/network-api/networkapi/templates/pages/buyersguide/article_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/article_page.html
@@ -55,7 +55,7 @@
     {% if related_articles %}
     <div class="tw-container tw-my-4">
       <div class="tw-row">
-        <div class="tw-px-4 tw-w-8/12 tw-mx-auto" id="pni-related-content">
+        <div class="tw-px-4 tw-w-8/12 tw-mx-auto" id="buyersguide-related-content">
           {% include "fragments/buyersguide/related_reading.html" with articles=related_articles %}
         </div>
       </div>
--- a/network-api/networkapi/templates/pages/buyersguide/article_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/article_page.html
@@ -47,6 +47,10 @@
     </div>
   </div>
 
+  {% for block in page.body %}
+      {% include_block block with parent_page=page page_type="blog" %}
+  {% endfor %}
+
   {% with related_articles=page.get_primary_related_articles  %}
     {% if related_articles %}
     <div class="tw-container tw-my-4">
@@ -58,10 +62,6 @@
     </div>
     {% endif %}
   {% endwith %}
-
-  {% for block in page.body %}
-      {% include_block block with parent_page=page page_type="blog" %}
-  {% endfor %}
 
   <div class="tw-container tw-mt-6">
     <div class="tw-row tw-justify-center">
--- a/source/js/buyers-guide/bg-main.js
+++ b/source/js/buyers-guide/bg-main.js
@@ -113,6 +113,22 @@ let main = {
     if (document.querySelector(`body.pni.catalog`)) {
       HomepageSlider.init();
     }
+    if (document.querySelector("#view-article")) {
+      const firstParagraph = document.querySelector(
+        ".paragraph-block .rich-text"
+      );
+      const relatedContainer = document.querySelector("#pni-related-content");
+      if (!firstParagraph || !relatedContainer) return;
+      const relatedContent = relatedContainer.querySelector("div");
+      const relatedPlace = firstParagraph.querySelector("*:nth-child(3)");
+      if (!relatedPlace) return;
+      firstParagraph.insertBefore(relatedContent, relatedPlace);
+      relatedContent.classList.add(
+        "large:tw-float-right",
+        "large:tw-pl-4",
+        "large:tw-mr-[-20%]"
+      );
+    }
   },
 };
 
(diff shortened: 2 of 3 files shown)


source/js/buyers-guide/bg-main.js
```diff
@@ -112,6 +112,23 @@ let main = {
...
+      const relatedPlace = firstParagraph.querySelector("*:nth-child(3)");
+      if (!relatedPlace) return;
+      firstParagraph.insertBefore(relatedContent, relatedPlace);
+      relatedContent.classList.add(
+        "large:tw-float-right",
+        "tw-p-4",
+        "tw-pr-0",
+        "large:tw-mr-[-20%]"
+      );
+    }
```
tbrlpld, line 131: What do you think about using a named function for this so it is obvious what it does without having to read through the whole function body?

network-api/networkapi/templates/pages/buyersguide/article_page.html
```diff
@@ -47,19 +47,22 @@ <h1 class="tw-h1-heading">{{ page.title }}</h1>
...
+  {% endfor %}
+
   {% with related_articles=page.get_primary_related_articles  %}
     {% if related_articles %}
-      {% trans "PRIMARY RELATED ARTICLES" %}
-      <div class="tw-m-4">
-        {% include "fragments/buyersguide/related_reading.html" with articles=related_articles %}
+    <div class="tw-container tw-my-4">
+      <div class="tw-row">
+        <div class="tw-px-4 tw-w-8/12 tw-mx-auto" id="pni-related-content">
```
tbrlpld, line 58: What do you think about using `related-articles` as the id? That would feel more consistent with the variables and template names.

tbrlpld: I am also wondering about the `pni-` prefix. Can you add some context why we need this?

tbrlpld (commented):
@fessehaye I have left a few small questions

network-api/networkapi/templates/pages/buyersguide/article_page.html
fessehaye: I do not think we should use `related-articles` over `pni-related-content` since it is referring to the parent div of the related article template.

I have done the pni- prefix for other templates on PNI just incase I think the name is too generic. I did this for some of the new dropdowns as well. I think we have had some issues in the past where over generic names for classes/ids have caused bugs and unintended changes so I think it is better to be more specific then not.

source/js/buyers-guide/bg-main.js
fessehaye: addressed

network-api/networkapi/templates/pages/buyersguide/article_page.html
tbrlpld: Thanks for expanding on this. 

What do you think about using `buyersguide` as the prefix if you think this needs to be more specific? I find the mixed use of `pni` and `buyersguide` rather confusing. (E.g. all the templates are in a directory called `buyersguide` but the id uses `pni`.)

source/js/buyers-guide/related-articles.js
```diff
@@ -0,0 +1,21 @@
...
+    relatedContent.classList.add(
+      "large:tw-float-right",
+      "tw-p-4",
+      "tw-pr-0",
+      "large:tw-mr-[-20%]"
+    );
+  },
+};
+
+export default RelatedArticles;
```
tbrlpld, line 21: Thanks for moving this out of the buyer's guide main function. This is definitely nicer.

source/js/buyers-guide/related-articles.js
```diff
@@ -0,0 +1,21 @@
...
+    const relatedContent = relatedContainer.querySelector("div");
+    const relatedPlace = firstParagraph.querySelector("*:nth-child(3)");
+    if (!relatedPlace) return;
+    firstParagraph.insertBefore(relatedContent, relatedPlace);
+    relatedContent.classList.add(
+      "large:tw-float-right",
+      "tw-p-4",
+      "tw-pr-0",
+      "large:tw-mr-[-20%]"
+    );
```
tbrlpld, line 17: What do you think about adding a method name that describes what it does? `RelatedArticles.init()` does tell me that it's about the related articles but not what to expect from happening. What do you think about something like `floatRelatedArticlesNextToFirstParagraph`?

source/js/buyers-guide/related-articles.js
```diff
@@ -0,0 +1,21 @@
+const RelatedArticles = {
+  init: () => {
+    const firstParagraph = document.querySelector(
+      ".paragraph-block .rich-text"
+    );
+    const relatedContainer = document.querySelector("#pni-related-content");
+    if (!firstParagraph || !relatedContainer) return;
+    const relatedContent = relatedContainer.querySelector("div");
+    const relatedPlace = firstParagraph.querySelector("*:nth-child(3)");
+    if (!relatedPlace) return;
```
tbrlpld, line 10: Can you add some context on what is happening here?

tbrlpld (commented):
Thanks for moving the feature out into a separate file @fessehaye. I think this is a nice separation. I have had a few more small questions.

source/js/buyers-guide/related-articles.js
fessehaye: this basically checks if there is at least 3 children inside the rich body container. Usually there is a header(h1-6) and a few paragraphs blocks to start most articles. If we do not have at least that many elements we leave the related articles at the bottom of the article page.

network-api/networkapi/templates/pages/buyersguide/article_page.html
fessehaye: Does that mean we want to change all `pni-` prefixes to `buyersguide-`? That is a quite a bit of classes names to change including tailwind color palettes? I would almost think changing the buyersguide references would be easier but either or would be more work then the original ticket. We could open a ticket about this and discuss why we should do one over the other but I do not think this large of a refactor should be included on this PR.

source/js/buyers-guide/related-articles.js
fessehaye: addressed

network-api/networkapi/templates/pages/buyersguide/article_page.html
tbrlpld: Yea, I think it would make sense to change all the `pni-` occurrences to `buyersguide-` since all files as well as the page types and other classes use that label. Of course this does not need to happen all in this PR. But I think when we are already touching something (or adding something new like in this case) we should strive toward more consistency.

source/js/buyers-guide/related-articles.js
tbrlpld: Nice. Thanks. That makes it a lot easier to understand what's happening.

source/js/buyers-guide/related-articles.js
tbrlpld: What do you think about adding this as a comment or docstring to the method?

tbrlpld (commented):
Thanks for renaming that method @fessehaye 👍

I left some follow up questions regarding the flag and the id naming.

network-api/networkapi/templates/pages/buyersguide/article_page.html
fessehaye: Okay, then I will update the ones currently added on this PR.

fessehaye:
@tbrlpld addressed both comments

tbrlpld (approved):
Thanks for these updates @fessehaye 👍 

Just noticing that the target branch is still `pni-2022`. I think we need to update that to `main`. Apart from that, this is good to go.

PR #9592, opened 2022-11-01
Add newsletter streamfield block to Mozfest pages

### LINKED ISSUE #9436: Enable Newsletter Sign Up Box on MozFest Site
Enable **Newsletter Sign Up Box** on MozFest Site by re-using #8787 however, this needs to be treated as a streamfield instead of as part of the page as per #8787 

The signup box should link to the MozFest Signup

tomusher:
Adds a block which allows inline newsletter signup forms to be embedded in Mozfest pages. Resolves #9436


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/newsletter_signup_block.html
```diff
@@ -0,0 +1,21 @@
...
+      class="join-us"
+      data-form-position="callout-box"
+      data-form-style="pop"
+      data-layout="2-column"
+      data-signup-id="{{ self.signup.id | unlocalize }}"
+      data-cta-header="{{ self.signup.header | escape }}"
+      data-cta-description="{{ self.signup.description | escape }}"
+      data-newsletter="{{ self.signup.newsletter }}"
+      data-ask-name="{{self.signup.ask_name}}">
+    >
```
tbrlpld, line 18: Can you add some context how this works? I am guessing there is some JS uses this?

tbrlpld (commented):
@tomusher I left a question on how this works. Sorry, I am not super familiar with the newsletter signup. 

I am also running into issues with missing columns when running this locally. I think we might need to update the branch from main and recreate the migrations

tbrlpld:
Thanks @tomusher looks like my migration issue is resolved.

danielfmiranda (commented):
Hi @tomusher!

In terms of code added and functionality, this looks great to me! 

I just have two questions in terms of design I was wondering either you or a design stakeholder could help answer for me.

On pages with a signup snippet, the block looks a little "squished":
Screenshot 2022-11-10 at 10-17-26 Spaces


Also on pages **without** a sign up snippet set, the block looks great, but it is wider than the rest of the content:
Screenshot 2022-11-10 at 10-17-08 Spaces



Are these two things to be expected?
If so, I will approve the PR as everything else looks good to me 👍 thanks!

tomusher:
Thanks for reviewing @danielfmiranda!

Making the block wider than content on normal pages was deliberate in order to avoid the 'squashed' effect in your first screenshot.

I'm not sure what's the best approach for pages with a signup snippet, but I expect it would be unnecessary to add one of these blocks when there is already a signup snippet.

cdanfon:
Has this been approved @danielfmiranda if so can we deploy to staging please?

danielfmiranda (approved):
Thanks for the explanation @tomusher! Approved

tomusher:
Rebased on main and resolved migration conflicts

Issue #9595, opened 2022-11-01
[PNI] Menu shaking when product grid reaches top of screen


tbrlpld:
# Description

Visit the PNI page (http://localhost:8000/en/privacynotincluded/) and scroll down. 

Notice the shaking when the product grid reaches the top of the screen.

https://user-images.githubusercontent.com/24797493/199338503-0ef0cea9-9a63-4cd2-a0da-d8d280577a8b.mov

# Expected behaviour

Smooth scrolling. 

# Acceptance criteria

- [x] On the buyer's guide (home) page I can scroll down until the product grid scrolls smoothly out of view at the top of the screen.

kristinashu:
I'm also seeing this sometimes. Seems really bad in FF and Chrome and less bad in Safari.

cdanfon:
NB Once this ticket has been completed we can do a push to production

fessehaye:
ready on staging @cdanfon

cdanfon:
@kristinash assigning to you for testing!

kristinashu:
@beccaklam could you please check, ty!

kristinashu:
I've just check this and it looks good! I'll move this to the Ready for Production column

PR #9641, opened 2022-11-08
9595 -Menu shaking when product grid reaches top of screen

fessehaye:
# Description

Looks like the creepy-face is causing the screen shaking when transitioning to `display: none` we can get a similar effect by changing the user-select and opacity.

Link to sample test page: pni homepage
Related PRs/issues: #9595


danielfmiranda (approved):
Hey Simon, just tested and works great! Approved

Issue #9642, opened 2022-11-08
Sign up snippets not syncing across translations


danielfmiranda:
## Describe the bug
Signup snippet changes on mozfest pages are not being synchronized, even after clicking "sync translations".

**From Slack:**
The sign up snippet got removed from the updated English page https://www.mozillafestival.org/en/proposals/ in order to use the full width, but syncing translations doesn’t remove it from translated pages.

## Steps to reproduce

1. Visit https://www.mozillafestival.org/en/proposals/, note that there is no signup snippet set.
![Screenshot 2022-11-08 at 14-18-01 Submit your MozFest Session before December 16th](https://user-images.githubusercontent.com/18314510/200687907-e6ae8480-cb4f-4531-98c3-7529fe4b87eb.png)

2. Visit https://www.mozillafestival.org/fr/proposals/, and note that there is **still** a signup snippet on the page, even after clicking "sync translations".
![Screenshot 2022-11-08 at 14-18-24 Proposez votre session pour le MozFest avant le 16 décembre](https://user-images.githubusercontent.com/18314510/200688015-3c04798e-ee67-4132-a19a-bc9d8056eeee.png)

## Expected behavior

If a mozfest page removes/changes it's linked signup snippet, its localized versions should show the same change after clicking "sync translations".

tbrlpld:
I think it would be good to describe the expected behaviour in a bit more detail. Are no changes carrying over?

danielfmiranda:
Hi @cdanfon this is now on staging and ready to be tested!

# Acceptance Criteria

- [ ] As a user, if I navigate to a given page in a language (such as English), it should provide the same sign-up snippet on all  other localized versions.
EX:
[mozillafestival.mofostaging.net/en/proposals/](http://mozillafestival.mofostaging.net/en/proposals/) vs. [mozillafestival.mofostaging.net/fr/proposals/](http://mozillafestival.mofostaging.net/fr/proposals/)
- [ ] As a content editor, if I remove a sign-up snippet from a page in the dashboard, then it should be removed for all languages for a given page, after selecting "Sync translated pages -> Publish Immediately" in the CMS.  
- [ ] As a content editor, if I add a sign-up snippet to a page from the dashboard, then it should appear for all languages for a given page, after selecting "Sync translated pages -> Publish Immediately" in the CMS.  


# Steps to test:

[...]

PR #9623, opened 2022-11-05
#9415 Sync translations for featured hero article (and other related article fields)

tbrlpld:
# Description

This PR sets the different fields on the Buyer's Guide homepage to be synchronized with the original version of the page. 
Actually, we would prefer them to be translated (which means they would always point to the translated version of an object automatically). 
But, there are two bugs in Wagtail Localize that prevent this. 
Specifically: 
* https://github.com/wagtail/wagtail-localize/issues/430, and
* https://github.com/wagtail/wagtail-localize/issues/640

As a workaround, we can use synchronized fields. The caveat is that we need to make sure that we still point to the translated/localized version of the articles in the frontend. 
To handle this, I have added/adjusted the getter methods for the different field. 

To test that the synchronization works:
1. Login to the admin
2. Navigate to the edit view of the **french** version of the buyer's guide homepage. 
3. Click the "translate this page" button.
4. Check the "french" locale tickbox and click the "submit" button.
1. Open the **french** version of the buyer's guide homepage: http://localhost:8000/fr/privacynotincluded/
4. See the titles of the different featured articles (hero, hero supporting, advice, popular)
5. Navigate to the edit view of the **english** version of the buyer's guide homepage. 
6. Pick different pages for the related links and publish the **english** page.
7. Synchronize the buyer's guide homepage translation (in the tree view click the "more" button)  Screen Shot 2022-11-04 at 17 01 567. 
8. Synchronize and publish the translations.8. 
9. Reload the **french** version of the buyer's guide homepage: http://localhost:8000/fr/privacynotincluded/
10. Observe that the featured articles should be updated according to the changes you have made.

Test that the localization works: 
1. Open the translated version of the buyer's guide homepage: http://localhost:8000/fr/privacynotincluded/
2. Follow the related articles links and notice that the active locale is still the translated one (you can notice this in the URL -- if you started with `http://localhost:8000/fr/privacynotincluded/` you should still see the `fr/` bit in the URL.

Link to sample test page: http://localhost:8000/fr/privacynotincluded/
Related PRs/issues: #9415

# Checklist

**Tests**
- [x] Is the code I'm adding covered by tests?

**Changes in Models:**
- [ ] ~~Did I update or add new fake data?~~
- [ ] ~~Did I squash my migration?~~ No migrations.
[...]

Issue #5712, opened 2020-11-24
mozfest: add salesforce ids to signup snippet as optional field


Pomax:
Make sure that this can be used for mozfest secondary landing pages, and ensure that that data can "go" somewhere in the correct format

PR #9644, opened 2022-11-08
updated to synchronized field, also updated template to pull localized version of signup

danielfmiranda:
# Description

This PR updates the 'signup' field from TranslatableField to SynchronizedField, as changes to the signup field on any English mozfest page would not get synchronized to its localized versions.

Link to sample test page: 
http://mozfest.localhost:8000/en/proposals/
http://mozfest.localhost:8000/fr/proposals/

Related PRs/issues: #9642

# Steps to test

1. Check this branch out and run `inv copy-prod-db` **Note**: This command can take a while!
2. Navigate to http://mozfest.localhost:8000/en/proposals/ and note that there is no "signup" set.
Screen Shot 2022-11-08 at 2 36 02 PM

3. Navigate to http://mozfest.localhost:8000/fr/proposals/ and note that there is an existing "signup" form.
Screen Shot 2022-11-08 at 2 36 02 PM

4. Log into the CMS http://mozfest.localhost:8000/ , navigate to the "Call to proposals" page, and select "Sync Translations".
Screen Shot 2022-11-08 at 2 36 02 PM

5. Go back to French version of the page http://mozfest.localhost:8000/en/proposals/ , and note that it is now the same as the English version. (No "signup" set). 
**Side note:** It appears that the use_wide_template field is also not being synced among translations, this will be taken care of in #9645
Screen Shot 2022-11-08 at 2 36 02 PM

6. To test further, make a change to "signup" field of the English version of the page, publish, and click "Sync Translations".
7. Verify that the change took effect in both the English and French versions of the page. If everything is looking as expected, testing is complete!

# Checklist

**Tests**

**Changes in Models:**
- ~~ ] Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~~

**Documentation:**


network-api/networkapi/mozfest/templates/partials/signup.html
```diff
@@ -11,8 +11,8 @@
           class="join-us mb-5"
           data-form-position="newsletter_signup_sidebar"
           data-signup-id="{{ page.signup.id|unlocalize }}"
-          data-cta-header="{{ page.signup.header | escape }}"
-          data-cta-description="{{ page.signup.description | escape }}"
+          data-cta-header="{{ page.signup.localized.header | escape }}"
```
danielfmiranda, line 14: Using https://github.com/mozilla/foundation.mozilla.org/pull/7658/files for reference, to make sure that the signup snippets are still localized.

TheoChevalier (approved):
Looks good, I assume we’re hitting https://github.com/wagtail/wagtail-localize/issues/430 or https://github.com/wagtail/wagtail-localize/issues/640 ?

If so, we may want to add a comment similar to what we did in https://github.com/mozilla/foundation.mozilla.org/pull/9623/

danielfmiranda:
Hi @TheoChevalier this does feel similar to the two tickets above, however, this differs in the sense that this relates to a direct FK relationship instead of using a orderable as the two examples above. 

I have created the ticket https://github.com/wagtail/wagtail-localize/issues/648 in the wagtail-localize repo, and added it to the comments as suggested!

TheoChevalier:
> Hi @TheoChevalier this does feel similar to the two tickets above, however, this differs in the sense that this relates to a direct FK relationship instead of using a orderable as the two examples above.
> 
> I have created the ticket [wagtail/wagtail-localize#648](https://github.com/wagtail/wagtail-localize/issues/648) in the wagtail-localize repo, and added it to the comments as suggested!

That’s perfect, thank you!

network-api/networkapi/mozfest/templates/partials/signup.html
```diff
@@ -11,8 +11,8 @@
           class="join-us mb-5"
           data-form-position="newsletter_signup_sidebar"
           data-signup-id="{{ page.signup.id|unlocalize }}"
```
tbrlpld, line 13: Should the signup ids be the same between the locales? Maybe a question for @TheoChevalier

network-api/networkapi/mozfest/templates/partials/signup.html
tbrlpld: Have you considered checking how many extra database queries it adds when we reuse the `.localized` method multiple times? I wonder if it would make sense to pull the localized signup into a variable and access the properties on that.

tbrlpld (commented):
@danielfmiranda I left some small questions for you and @TheoChevalier

danielfmiranda:
Hi @tbrlpld, great question! 

After taking a look at the queries using the django toolbar, I have found the following results:

| Language |  No localization |  .localized on both | {% with .localized %}  |
|---|---|---|---|
| English |  88 Queries | 90 Queries   | 89 Queries  |
| French | 93 Queries  | 97 Queries  | 95 Queries  |

Due to this, I have updated the template to be wrapped by with {% with %} statement like you mentioned. Thanks!

network-api/networkapi/mozfest/templates/partials/signup.html
danielfmiranda: Hey Tibor as far as I know I believe they should be, but maybe @TheoChevalier can confirm

TheoChevalier: Hmm… Sorry, it’s not super clear to me what these IDs are, unless these are the Salesforce IDs? In that case, yes, they should be the same

danielfmiranda: Thanks Theo! After doing some digging, it does look like these are in fact salesforce IDs. 

For reference:
https://github.com/mozilla/foundation.mozilla.org/issues/5712 Here is the issue asking that salesforce IDs be implemented
https://github.com/mozilla/foundation.mozilla.org/pull/5714 Here is the PR adding the ID field 👍 

Looks like they should be the same! @tbrlpld

tbrlpld: @danielfmiranda I think the `signup.id` is different from the Salesforce ID. If you look at the field name that is added in the PR you linked to, it's `campaign_id` that seems to be the Salesforce ID: 

https://github.com/mozilla/foundation.mozilla.org/pull/5714/files#diff-62de065a3de1e8fe56917af0194a4239e0575943f3b8ecac0235a496db99eb2cR57

tbrlpld:
> Hi @tbrlpld, great question!
> 
> After taking a look at the queries using the django toolbar, I have found the following results:
> 
> Language	No localization	.localized on both	{% with .localized %}
> English	88 Queries	90 Queries	89 Queries
> French	93 Queries	97 Queries	95 Queries
> Due to this, I have updated the template to be wrapped by with {% with %} statement like you mentioned. Thanks!

Thanks for this @danielfmiranda! Super insightful 👍 


Not part of this PR, but: 90 database queries for such a simple page 😱

network-api/networkapi/mozfest/templates/partials/signup.html
```diff
@@ -7,15 +7,17 @@
...
-        </div>
+        {% with localized_signup=page.signup.localized %}
+          <div
+            class="join-us mb-5"
+            data-form-position="newsletter_signup_sidebar"
+            data-signup-id="{{ page.signup.id|unlocalize }}"
+            data-cta-header="{{ localized_signup.header | escape }}"
+            data-cta-description="{{ localized_signup.description | escape }}"
+            data-newsletter="{{ page.signup.newsletter }}"
+            data-ask-name="{{page.signup.ask_name}}">
```
tbrlpld, line 18: @danielfmiranda What do you think about using all field from the localized signup in the template and rather defer the decision what translated and what is synchronized to the `Signup` model? 

Looking at the current `Signup` model we are only defining `header` and `description` as translatable. The other fields are not listed at all, which leaves them in this weird place where they are only copied from the source version when the translation is first created, but are not keep in sync thereafter. I feel like this might be a source of future bugs. 

So my idea would be to add all the other signup fields as synchronized fields on the model and use all fields from the localized signup in the template. 

See also: https://www.wagtail-localize.org/how-to/field-configuration/#neither-excluded

tbrlpld: We might also want to figure out if `signup-id` in the HTML should be the `campaign_id` field on the model 🤔

tbrlpld (commented):
@danielfmiranda Thanks for this! 

I left one more question regarding the translation / synchronization of the fields.

network-api/networkapi/mozfest/templates/partials/signup.html
danielfmiranda: Hi @tbrlpld! Thanks for the link and for the explanation, this definitely makes sense to me!  

I have now updated the `SignUp` model to have the other fields it is using in the template as `SynchronizedFields`. 
I also updated all other templates in the code base that use `SignUp` to now be wrapped with the localized version of the snippet and using those localized fields instead as discussed.



To answer the second question, it appears that all instances of the signup snippets being used in HTML templates pass in the snippets own ID (unlocalized) as the `data-signup-id` variable, and not the `campaign_id`. This variable is then used in [join-us.jsx](https://github.com/mozilla/foundation.mozilla.org/blob/main/source/js/common/inject-react/join-us.js) to create a path to the foundation site's API, and not to salesforce. So I believe they are correct as is.

```jsx

    const props = element.dataset;
    const sid = props.signupId || 0;

    props.apiUrl = `${siteUrl}/api/campaign/signups/${sid}/`;

```

tbrlpld: Looking at `network-api/networkapi/campaign/views.py` and the functions `signup_submission_view` and `signup_submission`, I think you are correct, we should be passing the object's `id` (not the `campaign_id`). But, I think we should use the localized version. 

This again is to make sure that the differentiation between the locales is done in the `Signup` model and not in the template.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -129,6 +129,9 @@ class Signup(TranslatableMixin, CTA):
         # Fields from the CTA model
```
tbrlpld, line 129: What do you think about also adding the `name` field here? 


Also, should we set the `translatable_fields` also on the `CTA` model? And only extend that list in the `Signup` model?

tbrlpld (approved):
Thanks for all the digging @danielfmiranda This looks good to me. I left an optional question about adding the `translatable_fields` to `CTA` as well, but that is totally up to you.

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
danielfmiranda: Hi @tbrlpld! This also makes sense to me. To implement the following suggestion, I have updated the `CTA` model to include the following translatable_fields:

```python
class CTA(models.Model):
...
    translatable_fields = [
        TranslatableField("name"),
        TranslatableField("header"),
        TranslatableField("description"),
        SynchronizedField("newsletter"),
    ]
```

and am extending the list in the Signup Model like so:

```python
class Signup(TranslatableMixin, CTA):
...
    translatable_fields = CTA.translatable_fields + [
        SynchronizedField("campaign_id"),
        SynchronizedField("ask_name"),
    ]
```

Wanted to double check this is what you were suggesting with your comment before I merge this in? Thanks!

network-api/networkapi/mozfest/templates/partials/signup.html
tbrlpld: Thanks for digging into this @danielfmiranda. It's a bit confusing in this context, but we need the `unlocalize`. This is unrelated to **which** number is displayed and only controls **how** the number is displayed. 

We use localized number display in general, e.g. "1,000" for one thousand in US style but that would be "1.000" in German number style. Either one of these would be wrong here though. Since this is a number we need to read computationally we need to the be plain "1000". That is what `unlocalize` does. 

See also: 
* https://docs.djangoproject.com/en/4.1/topics/i18n/formatting/#unlocalize

tbrlpld: Regarding the **which** number: I am a bit confused. `page.signup` is a synchronized field. So the french page should use the **same** signup as the english page. But `page.signup.localized` should be the translated version of the signup. Translations do have their own `id`. Did you create translations of the signup itself before these tests?

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
tbrlpld: Yup @danielfmiranda That's exactly what I meant.

PR #9661, opened 2022-11-10
Homepage extra margin removal

fessehaye:
# Description

Small fix to account for not including some of the extra homepage content

Removing this extra margin
image


danielfmiranda (approved):
Tested and Approved!

Issue #9649, opened 2022-11-09
[PNI] Newsletter signup broken without JS


tbrlpld:

fessehaye:
Lets just hide the component when JS is disabled. It is a react component so without server side rendering there is very little we can do to make this form work without JS

tbrlpld:
There is no form on the services site that we could link to?

tbrlpld:
We could also create a simple form page that handles the requests to the newsletter service sever side.

fessehaye:
Ahh I see so you would like an offline form as a replacement, that is fair. That would take a bit longer of course so I will at least hide the content for now since this is a quick fix and will make the page look less broken. I think if we did that I would like to look into doing that for our other newsletter related components as well!

tbrlpld:
Yea, that makes sense. Maybe we create a separate ticket for that since the description might need some more info. We can keep this as a bug fix ticket and create the other one as a product improvement / feature ticket.

simont-cr:
Closing as obsolete issue as part of 2024 Clean-up.

PR #9663, opened 2022-11-10
9649 - Hide Newsletter Box without JS

fessehaye:
# Description

Hiding the react component when JS is disabled using `noscript`

Link to sample test page:
Related PRs/issues: #9649

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags i18n static %}
 
-<div id="buyersguide_newsletter_box" class="tw-hidden tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
+<div id="buyersguide-newsletter-box" class="tw-hidden tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
     <img
   class="tw-absolute tw-right-0 tw-top-0 tw-hidden large:tw-block"
   src="{% static "_images/buyers-guide/asterick.svg" %}"
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags i18n static %}
 
-<div class="tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
+<div id="buyersguide_newsletter_box" class="tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
     <img
   class="tw-absolute tw-right-0 tw-top-0 tw-hidden large:tw-block"
   src="{% static "_images/buyers-guide/asterick.svg" %}"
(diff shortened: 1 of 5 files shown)


tbrlpld (commented):
@fessehaye This does work for people who disable JS, but I don't think this works when JS is slow to or fails to load (because technically JS is not disabled in that case -- e.g. embedded script would work). 

What do you think about hiding the component by default (`tw-hidden`) and the remove the hiding with JS when it is loaded and initializes the newsletter signup?

network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
```diff
@@ -1,6 +1,6 @@
 {% load wagtailcore_tags i18n static %}
 
-<div class="tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
+<div id="buyersguide_newsletter_box" class="tw-hidden tw-relative tw-py-5 tw-px-5 tw-rounded-lg medium:tw-px-7 medium:tw-py-6 tw-bg-gradient-to-r tw-from-purple-05 tw-to-blue-05">
```
tbrlpld, line 3: What do you think about using `-` instead of `_` in the HTML `id` attribute? That seems like the more common style in the code base.

source/js/components/join/join.jsx
```diff
@@ -21,6 +21,31 @@ class JoinUs extends Component {
...
+        .querySelector(
+          ".product-box-list > .tw-col-span-full.tw-col-end-5.tw-order-1.tw-hidden"
+        )
+        .classList.add("tw-flex");
+      document
+        .querySelector(
+          ".product-box-list > .tw-col-span-full.tw-col-end-5.tw-order-1.tw-hidden"
+        )
+        .classList.remove("tw-hidden");
+    }
```
tbrlpld, line 48: What do you think about moving this code into the buyer's guide modules? I seems like the `JoinUs` component is used in several other places in which the concept of product box and buyer's guide newsletter box are not known and the toggling should have no effect.

source/js/components/join/join.jsx
```diff
@@ -21,6 +21,31 @@ class JoinUs extends Component {
       privacyCheckbox: utility.generateUniqueId(`join-privacy-checkbox`),
     };
     this.state = this.getInitialState(props);
+
+    // Used for toggle visibility for buyersguide newsletter container
+    if (document.querySelector("#buyersguide_newsletter_box")) {
+      document
+        .querySelector("#buyersguide_newsletter_box")
```
tbrlpld, line 28: Have you considered storing the found element in a variable? I think this could aid readability.

source/js/components/join/join.jsx
```diff
@@ -21,6 +21,31 @@ class JoinUs extends Component {
...
+        .classList.remove("tw-hidden");
+    }
+
+    // Used for buyersguide product review grid toggle visibility
+    if (
+      document.querySelector(
+        ".product-box-list > .tw-col-span-full.tw-col-end-5.tw-order-1.tw-hidden"
+      )
```
tbrlpld, line 36: Have you considered traversing to this element from the newsletter box which we are already pulling above? I am it seems fragile to have this selection tied to the list of classes.

tbrlpld, line 36: Similar to the above, what do you think about storing the queried element in a variable?

tbrlpld (commented):
@fessehaye I have left some questions.

network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
tbrlpld: Why was this resolved?

tbrlpld (commented):
Thanks, this is looking a lot better. Just one minor question.

network-api/networkapi/templates/fragments/buyersguide/pni_newsletter_box.html
fessehaye: I must have missed that! I thought I addressed that on a previous commit. It is added now @tbrlpld

tbrlpld (approved):
Thanks for the updates @fessehaye This is looking good 👍

Issue #9691, opened 2022-11-14
[PNI] Category order in drop down should reflect desired order 


kristinashu:
**Context:** 
Jen updated the order of the categories in the CMS. The order is properly reflected in the main nav however the order is reversed in the drop down menu. 

**Acceptance Criteria:**
- [ ] As an editor I can manually define a sort order in the back end for categories
- [ ] As a site visitor I can see the manually defined category order in the dropdown (more categories) nav menu

Screen cap of CMS on staging:
<img width="554" alt="image" src="https://user-images.githubusercontent.com/1068783/201714477-68ee6c02-c629-4536-a779-ae548a939bf9.png">

Screen cap of the nav on staging showing the wrong order in the nav:
<img width="882" alt="image" src="https://user-images.githubusercontent.com/1068783/201714538-9e30ca28-5fba-4df0-9d6a-1e4892a30088.png">

mtdenton:
Looking to negotiate this into our current sprint, removing the label for now.

tbrlpld:
@kristinashu This is on staging ready for testing.

kristinashu:
This looks good on staging and is ready for prod!

PR #9692, opened 2022-11-14
updated from push to unshift to have correct order

danielfmiranda:
# Description

Link to sample test page: http://localhost:8000/privacynotincluded
Related PRs/issues: #9691

This PR updates the JS logic which creates the array used in the "More Categories" dropdown menu.

Instead of using `.push()`, we are now using `.unshift()`, as prepending the items adds them in the correct order.

# Steps to test

1. Visit the CMS and take a look at the PNI Product Category Snippet page. Note the order that they are in.
2. Visit the PNI home page, and take a look at the category navigation, they should be in the same order as the CMS, categories within the "more categories" dropdown included.
3. If you would like to test further, update the order of the categories in the CMS and revisit the navigation.
4. If everything is working as expected, testing is complete!

# Screenshots

Order in the CMS:
Screenshot 2022-11-14 at 1 45 04 PM

Order on the front end:
Screenshot 2022-11-14 at 1 59 56 PM

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


danielfmiranda:
Hi @TheoChevalier! I have a couple of quick questions that I was hoping you might be able to help me answer:

**Question 1:**
As mentioned above, this PR makes sure that the categories in the PNI category nav follow the sort order. However, I noticed that when visiting a localized version of the page, for example /fr/, the categories still follow the sort order of the English categories. 
I was wondering if this was to be expected?

**Question 2:**
I noticed that when visiting a localized version of the page, for example /fr/, clicking on a localized category pushes the english slug to the URL. Do we want this to be the localized slug instead?

tbrlpld (approved):
Looks like it was reverse after all huh? 


Looks good to me @danielfmiranda 👍

danielfmiranda:
Thanks for the review @tbrlpld! 🙌 and yes it does! After double checking that categories were getting returned in order through the backend, it looks the JS was just adding them in reverse order to the list used in the drop down menu.

PR #9702, opened 2022-11-15
Center the image vertically in an image teaser block

### LINKED ISSUE #9621: Mozfest | Reposition image in the "image teaser" block
Ticket open from the back of #9535

The image doesn't centre vertically when longer text is posted. Attached example:

Screenshot 2022-11-03 at 15 55 42

The goal of this ticket is to reposition the image to align with the content in the "image teaser" block.

tomusher:
Resolves #9621 

Before:
image-teaser-before

After:
image-teaser-after


danielfmiranda (approved):
Hi @tomusher just gave this a test and can confirm that its working as expected!

image


great job, approved 👍

PR #9703, opened 2022-11-15
Add Tito Event model to replace global Tito config

### LINKED ISSUE #8013: MozFest: Add support for multiple Tito events
Each Tito event has its own webhook API key and newsletter sign-up question ID (currently set in environment variables `TITO_SECURITY_TOKEN` and `TITO_NEWSLETTER_QUESTION_ID`). As these are currently global in the Mozilla site only one Tito event can be supported at a time.

Suggest creating a new Event snippet to store the event slug (currently set on the Tito widget block), API key and newsletter sign-up question ID (keep 'releases' specific to the Tito widget block). The snippet can then be selected in the Tito widget block. Access to the event snippets in Wagtail can be restricted with Group permissions to help protect the event API keys, if required.

The webhook Django view will also need to be updated to find retrieve the API key and question ID based on the event received by the hook.

Update the `Mozilla Festival` section of the project README to reflect the changes.

tomusher:
Adds a new `TitoEvent` snippet model and updates the `TitoWidget` block to support referencing a specific event.

Tests updated, readme update, and old environment vars removed.

Resolves #8013


network-api/networkapi/events/utils.py
```diff
@@ -1,25 +1,47 @@
...
+from .models import TitoEvent
 
 
-def is_valid_tito_request(signature, request_body):
-    secret = bytes(settings.TITO_SECURITY_TOKEN, "utf-8")
-    signed = sign_tito_request(secret, request_body)
+def _tito_event_from_request_dict(request_dict):
+    """Given the decoded and parsed body of webhook request from Tito, return
+    the appropriate TitoEvent if it exists.
+    Return False if a matching TitoEvent does not exit."""
```
danielfmiranda, line 12:
```suggestion
    Return False if a matching TitoEvent does not exist."""
```

danielfmiranda (approved):
Hi @tomusher, everything looks good to me after testing! Approved

Issue #9491, opened 2022-10-14
[PNI] Content index does not save "load more" pagination state


tbrlpld:
# Description

The PNI editorial content index page does not save the pagination state when navigation with the "load more" botton. 
This means when the page is refreshed it page starts only displaying the first page. 

This can lead to navigation issues for users. If for some reason (storage full, etc.) the pages are not cached properly, navigating back to the index might show a different state than the user left. E.g. navigating from the index (after clicking the "load more" button) to an article and back to the index would show only the first page of the index again. 

To avoid this, we should save the state of the pagination with the "load more" button. 

# Dev notes

* Updating the history and URL is possible with [`hx-push-state`](https://htmx.org/attributes/hx-push-url/). 
* We need to avoid this functionality from colliding with the "non-JS" pagination that is first loaded and only replaced with the "load more" functionality. E.g. when directly navigating to the index with `?page=2` it should still show only the second page of articles. We should be able to use a second query parameter, e.g. `expanded`, to signal that we should render all articles from all pages from the first to the requested one. E.g. `?page=2&expanded` shows all articles form the first page and all articles from the second page. The "non-JS" pagination would still go to a normal `?page=3` which would only show the articles of the third page. The "load more" button should add the articles from the thrid page to the current index (and update the active URL accordingly). 

# Acceptance criteria

- [x] As a user, when I visit the content index page, click the "load more" button I can see the URL now includes the page and expansion query parameters (`?page=2&expanded`).
- [x] As a user, when I visit the content index page, click the "load more" button and refresh the page, I find the index page in the same state (shows all the loaded articles) as before the refresh.

kristinashu:
This is working very nicely! Ready for prod!

PR #9711, opened 2022-11-16
9491 content index load more state

tbrlpld:
# Description

This PR adds functionality to store information about the "load more" pagination state in the URL.  Previously, if you visited the content index then click the "load more" button and then refresh, you would get the state back that you had before you clicked the "load more" button, only the first page is displayed.  This was a degradation of the "load more" pagination compared to normal page reload pagination. 

This PR introduces the idea of an expanded page.  When a page is loaded in expanded state, all items from the previous pages are also displayed.  An expanded page can be requested with the querystring parameter `expanded=true` in the URL. With the ability to load an expanded page we can update the URL when the "load more" button is clicked and have the URL reflect which page was expanded to. A refresh of the page will bring back the same state. 

To make it easier to facilitate this pagination change, this PR also changes how a page of items is retrieved with HTMX and how the pagination is applied. Previously, this was using a separate endpoint / route on the page and a special method existed to retrieve the paginated items. This is switched now, so that the pagination is in a separate method that received the items to paginate. The decision to respond with a full page or only the items is now based on the HTTP header that HTMX uses on its requests and `django_htmx` is used to check for that header. 

Link to sample test page: http://localhost:8000/en/privacynotincluded/articles/
Related PRs/issues: #9491 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? By comments and tests. 
- [ ] ~~id I update the READMEs or wagtail documentation?~~


danielfmiranda (approved):
Hi @tbrlpld, 

After giving the code a look through and testing this out on my local machine, I can confirm that everything is working as expected! 

Amazing work, thanks for taking this on. Approved! 👍

network-api/networkapi/wagtailpages/tests/buyersguide/test_editorial_content_index.py
```diff
@@ -156,15 +157,52 @@ def test_serve_paginated_items_page_2(self):
...
 
-        self.assertEqual(route, "items/")
+            self.assertFalse(response.context["items"].has_next())
+            self.assertFalse(response.context["items"].has_previous())
 
-    def test_items_route_template(self):
-        url = self.get_items_route_url()
+    def test_hx_request_templates(self):
+        """
+        Only the items template (not the full page) template is used for HTMX requests.
```
kevinhowbrook, line 199: **Question**

I think the comment is throwing me a bit here, but does this test need updating to check the template fragment rather than the editorial_content_index_page ?

kevinhowbrook (approved):
Thanks @tbrlpld looks good to me, just a query about a test.
Locally everything tests fine.

network-api/networkapi/wagtailpages/tests/buyersguide/test_editorial_content_index.py
```diff
@@ -156,15 +157,52 @@ def test_serve_paginated_items_page_2(self):
...
 
-        response = self.client.get(url)
+        HTMX sets a header 'HX-Request' with value set to 'true' for it's requests.
```
kevinhowbrook, line 201: Well TIL! 

```
HX-Current-URL | http://localhost:8000/en/privacynotincluded/articles/
HX-Request | true
HX-Target | index-navigation
```

Issue #9609, opened 2022-11-02
[PNI] Add ids or classes to new elements for tracking purposes


nancyt1:
**Homepage Editorial Content**
- [x] Add id to featured article div
- [x] Add id for related section associated with featured article

**Related Articles**
Add id or class to related or popular article blocks around the site. I wanted to be able to tell if clicks are coming from related blocks, these don't have to be unique to each one but some pages do have 2 places where it appears so an ID distinguishing which section (like top or bottom) would be helpful.
- [x] Article post page for top and bottom sections
- [x] Article index page
- [x] Product pages
- [x] Category pages (not implemented yet I don't think)

**Related Products Block**
- [x] Add id or class to "Related Products" section on product pages

**Newsletter Signup**
Not sure if this is possible but would be great to have!
- [x] Add id to newsletter signup in the product grid
- [x] Add id to newsletter signup at the end of article index and post pages

tbrlpld:
Left two optional questions.

danielfmiranda:
Hi @nancyt1! This is now on staging and ready to be tested if you would like to take a look. Thanks!

danielfmiranda:
This has been tested together with design and they confirm that the classes/ google tag manager is working as expected. Closing!

PR #9713, opened 2022-11-16
added classname to "article-index-popular-article" sections

danielfmiranda:
# Description

Link to sample test page:
Related PRs/issues: #9609, #9670

It appears that we missed a "Popular Articles" section to add an ID to in #9670, as there is a component for mobile screens and a component for desktop size screens as well.

However while discussing this ticket, we decided that the best step would be to use classnames instead of IDs.

This PR removes the ID's that we have added, and instead adds them as a classname to both requested elements.


nancyt1 (approved):

mmmavis (approved):
Code looks good to me! 🎉 Will need @nancyt1 to verify on the GA4 side.

Could I ask that we have the PR title and description changed (id -> class) for documentation purposes? Thank you!

Issue #9606, opened 2022-11-02
[PNI] Make the `Share image:` and `Meta description:` required fields on PNI Article pages


kristinashu:
### Context:
Raised in https://github.com/mozilla/foundation.mozilla.org/issues/9603#issuecomment-1301100267 and related to https://github.com/mozilla/foundation.mozilla.org/issues/9464

Make the `Share image:` and `Meta description:` required fields on all PNI Article pages (similar to how it is on our main blog site pages)

### Acceptance Criteria:
- [x] As a user I can't publish a PNI Article until I have added a `Share image:` 
- [x] As a user I can't publish a PNI Article until I have added a `Meta description:`

danielfmiranda:
Tested this on staging, can confirm that its working as expected!

kristinashu:
Sorry I forgot about this! Yes, confirmed it's working properly. Thank you!

PR #9725, opened 2022-11-18
Buyersguide article's "search_image" and "search_description" now required

danielfmiranda:
# Description

Related PRs/issues: #9606 

This PR adds the following:
- Updated admin form for the `BuyersGuideArticlePage`, which now makes the `search_description` and `search_image` fields required.
- Updated `BuyersGuideArticlePageFactory` to include `search_image` and `search_description`
- Some tests to make sure that these fields are required on the `BuyersGuideArticlePage` form

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? (added comments)


danielfmiranda:
For context, I am overriding the BuyersGuideArticlePage admin form, to make the “search_description” and “search_image” field required like so:


**forms.py**
```python
class BuyersGuideArticlePageForm(WagtailAdminPageForm):
        super().__init__(*args, **kwargs)

        form.fields["search_image"].required = True
        form.fields["search_description"].required = True
```

then importing to article page for use like so:

**article_page.py**
```python
class BuyersGuideArticlePage(foundation_metadata.FoundationMetadataPageMixin, wagtail_models.Page):
...
    # Custom base form for additional validation
    base_form_class = BuyersGuideArticlePageForm
```





I am now trying to test that this form does in fact validate for “search_image” in the following test, however, it seems to be giving me issues:

**test_article_page.py**
```python
    def test_article_page_requires_share_image(self):
        edit_handler = pagemodels.BuyersGuideArticlePage.get_edit_handler()
        article_page_form = edit_handler.get_form_class()

        article_page = buyersguide_factories.BuyersGuideArticlePageFactory(
            parent=self.content_index,
        )

        form = article_page_form(
            instance=article_page,
            data=({"search_description": None}),
        )

        self.assertTrue(form.is_valid())
```


However, when running the test, the `form.is_valid()` statement is returning the following error:


```
    self.assertTrue(form.is_valid())
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/forms/pages.py", line 144, in is_valid
    return super().is_valid()
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/modelcluster/forms.py", line 318, in is_valid
    form_is_valid = super(ClusterForm, self).is_valid()
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/forms/forms.py", line 175, in is_valid
    return self.is_bound and not self.errors
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/forms/forms.py", line 170, in errors
    self.full_clean()
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/forms/forms.py", line 372, in full_clean
    self._clean_fields()
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/forms/forms.py", line 384, in _clean_fields
    value = field.widget.value_from_datadict(self.data, self.files, self.add_prefix(name))
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/core/blocks/base.py", line 535, in value_from_datadict
    return self.block_def.value_from_datadict(data, files, name)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/core/blocks/stream_block.py", line 97, in value_from_datadict
    count = int(data['%s-count' % prefix])
KeyError: 'body-count'
```

danielfmiranda:
Tests are now in place and working in regards to checking the form for errors.


For example, the following test:

```python
    def test_article_page_requires_both_search_fields(self):
        """
        Test that the buyersguide article page form will return validation
        errors for both "search_image" and "search_description" fields
        if neither are updated.
        """
        article_page = buyersguide_factories.BuyersGuideArticlePageFactory(
            parent=self.content_index
        )

        test_form = self.article_page_form(
            instance=article_page,
            data=self.generate_form_data(
                {
                    "title": article_page.title,
                    "slug": article_page.slug,
                    "search_description": None,
                    "search_image": None,
                }
            )
        )

        self.assertEqual(2, len(test_form.errors))
        self.assertIn("search_image", test_form.errors)
        self.assertIn("search_description", test_form.errors)
```

returns a total of 2 errors, regarding the `search_image` and `search_description` field as expected.


The next test, which tests a valid form:

```python
    def test_article_page_with_search_fields_is_valid(self):
        """
        Test that a buyersguide article page form with
        the search fields set is valid with no errors.
        """
        article_page = buyersguide_factories.BuyersGuideArticlePageFactory(
            parent=self.content_index
        )

        test_form = self.article_page_form(
            instance=article_page,
            data=self.generate_form_data(
                {
                    "title": "New Test Title",
                    "slug": "new-test-title",
                    "search_description": article_page.search_description,
                    "search_image": article_page.search_image,
                }
            )
        )

        self.assertEqual(0, len(test_form.errors))

```

also works as expected, returning 0 errors since both required fields `search_image` and `search_description` have data.





However, I was hoping to add one more assert statement to these tests, to check whether or not `test_form.is_valid()`, similar to what is being done in [test_products.py](https://github.com/mozilla/foundation.mozilla.org/blob/d1a729ba0edf815e375325346c2d6e3ec301fd80/network-api/networkapi/wagtailpages/tests/buyersguide/test_products.py#L594-L605)

```python
        form = self.form_class(
            instance=cat1,
            data=self.generate_form_data({"name": cat1.name, "sort_order": cat1.sort_order, "parent": cat1}),
        )

        self.assertFalse(form.is_valid())
```




However when I add these to my tests, particularly to the one with **valid** data, `test_form.is_valid()` seems to always be returning false.

@tbrlpld I was wondering if you could please give me a hand by giving these a quick look to see if I may be missing something, or, if you think that the checks in these tests are sufficient? 

Thanks in advance!

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -469,3 +469,10 @@ def get_default_locale():
 def get_original_by_slug(Model, slug):
     (DEFAULT_LOCALE, DEFAULT_LOCALE_ID) = get_default_locale()
     return Model.objects.get(slug=slug, locale=DEFAULT_LOCALE_ID)
+
+
+def requireShareFields(form):
```
tbrlpld, line 474: @danielfmiranda What do you think about changing this to `snake_case` which is more common in Python?

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -206,3 +210,114 @@ def test_secondary_related_articles_no_related_articles(self):
...
+        the search fields set is valid with no errors.
+        """
+        article_page = buyersguide_factories.BuyersGuideArticlePageFactory(parent=self.content_index)
+
+        test_form = self.article_page_form(
+            instance=article_page,
+            data=self.generate_form_data(
+                {
+                    "title": "New Test Title",
+                    "slug": "new-test-title",
```
tbrlpld, line 316: This should not be the error, but why are you changing slug and title here?

tbrlpld:
@danielfmiranda I need to take a closer looks at this. 

Just a few questions:

* Why are you adding title and slug in the form data? 
* Why are you changing title and slug in one test? 
* What happens if you add `form.is_valid()` in all the tests? 
* What was the behavior before the changes? 
* Are there any non-field errors?

network-api/networkapi/wagtailpages/utils.py
danielfmiranda: good catch Tibor, updated!

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: Hi Tibor, another good catch, originally I added the title/slug fields in the test form data as they are also required fields, and leaving them out would return a validation error.

In this test, I was updating the title and slug as I believed this would be a good test of "Made a change in the form, but have valid data for the `search_image` and `search_description`", but I think that this makes the test a little hard to follow.

Instead, I have updated `generate_form_data()` to add the page's `title`,`slug`, and required `body` field to the form data, as well a comment explaining that these are required and necessary to add to the form, so we dont get any validation errors against fields we are not trying to test against.

danielfmiranda:
Hi @tbrlpld thank you very much for taking a look!

To respond to the questions:

**- Why are you adding title and slug in the form data?**
  - Originally, I added these to the form data as they are required fields, and leaving them out would return validation errors. However, I have decided to update the `generate_form_data()` function to add these to the form instead, with a comment explaining we are adding them to avoid errors against fields we are not testing against. 
  
 **- Why are you changing title and slug in one test?**
  - I had thought an update to the title and slug fields would have been a good test of "make change to the page, but have valid data for `share_image` and `share_description`", but decided to remove that as we are not testing against these fields and I think makes the test a little hard to follow.

**- What happens if you add form.is_valid() in all the tests?**
  - In tests with invalid data, `form.is_valid()` is returned as `False`, and returns a error list.
  - In the test with **valid** data, `form.is_valid()` is returned as `False`, but returns no error list.
 
**- What was the behavior before the changes?**

- Before the changes, users were able to submit/publish a BuyersGuideArticle page with empty `search_description` and `search_image` fields.

**- Are there any non-field errors?**
  - No, there are no non-field errors on any of the tests.

tbrlpld:
> **- What was the behavior before the changes?**
> 
> * Before the changes, users were able to submit/publish a BuyersGuideArticle page with empty `search_description` and `search_image` fields.
> 

Sorry, I should have been clearer. I mean if that valid test passes without your changes to the behavior.

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -26,14 +27,16 @@ def clean_parent(self):
         return parent
 
 
-class BlogPageForm(WagtailAdminPageForm):
+class BuyersGuideArticlePageForm(WagtailAdminPageForm):
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
+        require_share_fields(self)
+
```
tbrlpld, line 34: What do you think about moving this closer to the rest of the buyer's guide classes? Maybe create a `buyersguide/forms.py` module?

network-api/networkapi/wagtailpages/utils.py
```diff
@@ -469,3 +469,10 @@ def get_default_locale():
 def get_original_by_slug(Model, slug):
     (DEFAULT_LOCALE, DEFAULT_LOCALE_ID) = get_default_locale()
     return Model.objects.get(slug=slug, locale=DEFAULT_LOCALE_ID)
+
+
+def require_share_fields(form):
+    # Overrides a page's admin form in order to make promote tab's
+    # "search image" and "search description" fields required.
+    form.fields["search_description"].required = True
+    form.fields["search_image"].required = True
```
tbrlpld, line 478: What do you think about keeping this in the `forms.py`? I am not sure this can be reused anywhere else, this seems very specific. It is also only two lines and no logic but rather configuration. What do you think about inlining this instead of the function? I usually like this approach of extracting functionality into named functions, but this seems like an odd case somehow 🤔

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -206,3 +210,112 @@ def test_secondary_related_articles_no_related_articles(self):
...
+
+    def test_article_page_requires_search_image(self):
+        """
+        Test that the buyersguide article page form will
+        return a validation error if no "search_image" is set.
+        """
+        article_page = buyersguide_factories.BuyersGuideArticlePageFactory(parent=self.content_index)
+
+        test_form = self.article_page_form(
+            instance=article_page,
```
tbrlpld, line 242: Can you add some context why we need an existing `instance`? The form should be working in a create scenario too.

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -29,6 +31,8 @@ def setUpTestData(cls):
         cls.content_index = buyersguide_factories.BuyersGuideEditorialContentIndexPageFactory(
             parent=cls.buyersguide_homepage,
         )
+        edit_handler = pagemodels.BuyersGuideArticlePage.get_edit_handler()
+        cls.article_page_form = edit_handler.get_form_class()
```
tbrlpld, line 35: Can you add some context why we need to go through this `get_edit_handler` method instead of using the custom form class directly? Where did you find this method? It's not documented on the model reference: https://docs.wagtail.org/en/v2.16.3/reference/pages/model_reference.html

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -206,3 +210,112 @@ def test_secondary_related_articles_no_related_articles(self):
...
+        one StreamField in the form, or else the form will return a validation error.
+        """
+        return nested_form_data(
+            {
+                **data,
+                "title": article_page.title,
+                "slug": article_page.slug,
+                "body": streamfield([("paragraph", rich_text("Hello world!"))]),
+            }
+        )
```
tbrlpld, line 232: **Kudos** Nicely done and explained and helper method 👍

network-api/networkapi/wagtailpages/pagemodels/blog/blog.py
```diff
@@ -118,6 +118,9 @@ class Meta:
 
 
 class BlogPage(FoundationMetadataPageMixin, Page):
+    # Custom base form for additional validation
+    base_form_class = BlogPageForm
+
```
tbrlpld, line 123: **Kudos** Nice, I think this is a good idea to move configuration like this above the fields 👍

tbrlpld (commented):
@danielfmiranda This is a tough one. I don't really see any thing obvious why that test would be failing. I left some questions around potential simplifications of the setup that might help us find out what is going on.

network-api/networkapi/wagtailpages/forms.py
danielfmiranda: Hey Tibor, this makes sense to me. I have created a new `buyersguide/forms.py` file, and have moved both `BuyersGuideArticlePageForm`, as well as `BuyersGuideProductCategoryForm`, since it is BuyersGuide related as well

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: https://github.com/wagtail/wagtail/blob/5fc191b116ee835762c282e13a5829fa873cc516/wagtail/admin/tests/test_edit_handlers.py#L145-L156

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: Hi Tibor, no reason in particular. I just added this to try and test the event of "someone is editing an existing articlepage". However, after giving it some more thought, I believe that a create scenario works better, and also makes the code a little less confusing, so I have decided to remove the instance!

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: Hi Tibor! 

I found this method in the wagtail source code's test suite, in a test called [test_get_form_for_page_with_custom_base](https://github.com/wagtail/wagtail/blob/main/wagtail/admin/tests/test_edit_handlers.py#L262-L272)

```python
    def test_get_form_for_page_with_custom_base(self):
        """
        ValidatedPage sets a custom base_form_class. This should be used as the
        base class when constructing a form for ValidatedPages
        """
        edit_handler = ValidatedPage.get_edit_handler()
        GeneratedValidatedPageForm = edit_handler.get_form_class()

        # The generated form should inherit from ValidatedPageForm, because
        # ValidatedPage.base_form_class == ValidatedPageForm
        self.assertTrue(issubclass(GeneratedValidatedPageForm, ValidatedPageForm))
```

danielfmiranda: Also have seen something similar when Rich was testing the validity of forms for [BuyersGuide Products](https://github.com/mozilla/foundation.mozilla.org/blob/137337624947b0f1fc71d612e78eaed92195ab49/network-api/networkapi/wagtailpages/tests.py#L471-L484)

```python
class BuyersGuideProductCategoryTest(TestCase):
    def setUp(self):
        edit_handler = get_snippet_edit_handler(BuyersGuideProductCategory)
        self.form_class = edit_handler.get_form_class()

    def test_cannot_have_duplicate_name(self):
        BuyersGuideProductCategory.objects.create(name="Cat 1")

        form = self.form_class(
            data=self.generate_form_data({"name": "Cat 1", "sort_order": 1}),
        )

        self.assertFalse(form.is_valid())
        self.assertEqual(1, len(form.errors))
        self.assertIn("name", form.errors)
```

tbrlpld: I think it's important to know why are using certain features because it can lead to unexpected outcomes.

Looking at this for another moment, I think it is important to the `get_edit_handler` because our form is not fully featured. It does not define a model or any fields. It would therefore not be usable on it's own.

tbrlpld: Before we just more on, there is still the question of what happens in the "valid" test without your changes / the base class. Do we already see that behavior that the form is "never valid"?

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -206,3 +210,107 @@ def test_secondary_related_articles_no_related_articles(self):
...
+        required, and the form will return a validation error against them if left out.
+
+        Since we have the "body" field set as required, we need to submit at least
+        one StreamField in the form, or else the form will return a validation error.
+        """
+        return nested_form_data(
+            {
+                **data,
+                "title": article_page.title,
+                "slug": article_page.slug,
```
tbrlpld, line 229: We don't want to test anything about the `title` and `slug` fields coming from some existing article page. Have you considered setting these as strings directly? 

In that case, we would also not need to create an article page in each test.

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
```diff
@@ -206,3 +210,107 @@ def test_secondary_related_articles_no_related_articles(self):
         result = article_page.get_secondary_related_articles()
 
         self.assertListEqual(result, [])
+
```
tbrlpld, line 213: All the tests below are testing the form rather than the page type. What do you think about moving this into a separate class and potentially into a `test_forms.py` module, so that the test module structure matches the implementation module structure?

tbrlpld (commented):
Thanks for the updates @danielfmiranda I have left a few more comments and questions.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/forms.py
```diff
@@ -0,0 +1,35 @@
...
+        return parent
+
+
+class BuyersGuideArticlePageForm(WagtailAdminPageForm):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        # Overriding admin form in order to make promote tab's
+        # "search image" and "search description" fields required.
+        self.fields["search_description"].required = True
+        self.fields["search_image"].required = True
```
tbrlpld, line 35: I also wonder if we maybe should do something closer to this section in the docs: https://docs.wagtail.org/en/stable/advanced_topics/customisation/page_editing_interface.html#custom-edit-handler-forms

... I think I am just realizing what the issue might be... 

We are overriding props of the form instance in our class, but our class does not have these fields. The super class also does not have these fields. The child class that is dynamically generated from our "base form class" is the one that has these fields. 

Maybe we should rather add the fields directly to the form class in the configuration we want them to be. They won't be overridden according to the docs. 

Or we should perform the checks in the `clean` method. This would have the downside that the UI does not show the fields as required.

tbrlpld:
Also, considering the issues this is creating, could we override the fields on the model directly? If we wanted to repeat that functionality on multiple models, we can create a mixin from these field definitions.

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: Hi Tibor, thanks for the suggested. I think this makes sense because this would simplify the test further, and make sure that this function produces the same data for every test. 

I have implemented this in the latest commit!

network-api/networkapi/wagtailpages/tests/buyersguide/test_article_page.py
danielfmiranda: This definitely makes sense to me, as we are indeed making changes to the form but not the model directly. 
The latest commit now includes `buyersguide/tests/test_forms.py`

network-api/networkapi/wagtailpages/forms.py
```diff
@@ -1,35 +1,9 @@
...
-        return name
-
-    def clean_parent(self):
-        parent = self.cleaned_data["parent"]
-        if parent:
-            if parent.parent:
-                self.add_error("parent", "Categories can only be two levels deep.")
-            if self.instance == parent:
-                self.add_error("parent", "A category cannot be a parent of itself.")
-        return parent
```
danielfmiranda, line 26: Since this is related to the buyers guide, I believe this makes more sense to live in [buyersguide/forms.py](https://github.com/mozilla/foundation.mozilla.org/blob/82019588da866bec89f8b4b7d2396cf634b90f54/network-api/networkapi/wagtailpages/pagemodels/buyersguide/forms.py)

network-api/networkapi/wagtailpages/pagemodels/buyersguide/forms.py
```diff
@@ -0,0 +1,35 @@
+from wagtail.admin.forms import WagtailAdminModelForm, WagtailAdminPageForm
+
+
+class BuyersGuideProductCategoryForm(WagtailAdminModelForm):
```
danielfmiranda, line 4: Like mentioned in other comment, I made no updates to `BuyersGuideProductCategoryForm`, other than moving it to this file

danielfmiranda:
Hi Tibor, thanks again for your great feedback and suggestions! 

I have implemented the following:
- Hardcoding the article page `title`/`slug` strings
- Moving the form test to `buyersguide/test_forms.py`


I also finally got the test to work using a different method, `get_form_for_model`, I saw the wagtail/wagtailtrans team was the same method in a test for a custom WagtailAdminModelForm implementation similar to ours:

[Link to the test file](https://github.com/wagtail/wagtailtrans/blob/9899b234b2db48eca0d3953a78d1045247002434/tests/unit/test_models.py)
```python
class TestWagtailAdminLanguageForm:

    def setup(self):
        self.form_class = get_form_for_model(model=models.Language, form_class=models.WagtailAdminLanguageForm)

    def test_clean_is_default(self):
        language = LanguageFactory(is_default=True)
        form = self.form_class(instance=language, data={
            'code': language.code,
            'position': language.position,
            'is_default': True,
        })

        assert form.is_valid()
```


Now all asserts, including `assertTrue(form.is_valid()` and `assertFalse(form.is_valid()` work as expected.


**In regards to the other suggestions regarding updating the implementation of our custom form:**
- Overriding the model fields did not work
- Custom clean method validation **did** work, but like you mentioned, the fields would not show as "required"
- Defining the fields in the form outside of the Init **did** work, however, help text was being overwritten and the image field took some work to preform the same. 
  - Regarding the help text, I know we could define it in the fields, but then we would have to make sure that it stays up to date with the original help text.

Due to the reasons above, I believe that since the form tests are now working, we should stick with the for as-is, as it  is the simplest option of the three in my opinion.

However, was curious to hear what you think.

Thanks!

network-api/networkapi/wagtailpages/forms.py
tbrlpld: Absolutely

tbrlpld:
Link to the implementation of `get_form_for_model`:

https://github.com/wagtail/wagtail/blob/v2.16.3/wagtail/admin/edit_handlers.py#L41-L71

network-api/networkapi/wagtailpages/tests/buyersguide/test_forms.py
```diff
@@ -0,0 +1,117 @@
...
+        self.assertEqual(1, len(test_form.errors))
+        self.assertIn("search_image", test_form.errors)
+        self.assertIn("This field is required.", test_form.errors["search_image"])
+
+    def test_article_page_requires_search_description(self):
+        """
+        Test that the buyersguide article page form will
+        return a validation error if no "search_description" is set.
+        """
+        article_page = buyersguide_factories.BuyersGuideArticlePageFactory()
```
tbrlpld, line 62: Can you add some context why we still need to generate the article page if we don't set the `instance` anymore? If this is just to have a convenient way to pass a valid image id, what do you think about using `networkapi.wagtailpages.factory.image_factory.ImageFactory` and passing `ImageFactory().id`?

network-api/networkapi/wagtailpages/tests/buyersguide/test_forms.py
```diff
@@ -0,0 +1,117 @@
...
+        self.assertIn("search_description", test_form.errors)
+        self.assertIn("This field is required.", test_form.errors["search_image"])
+        self.assertIn("This field is required.", test_form.errors["search_description"])
+
+    def test_article_page_with_search_fields_is_valid(self):
+        """
+        Test that a buyersguide article page form with
+        the search fields set is valid with no errors.
+        """
+        article_page = buyersguide_factories.BuyersGuideArticlePageFactory()
```
tbrlpld, line 105: Basically the same question as above about why we need to generate an article page?

tbrlpld (approved):
@danielfmiranda This is looking good. I have left one question about the use of the article factory. Addressing that question is optional however.

network-api/networkapi/wagtailpages/tests/buyersguide/test_forms.py
danielfmiranda: Good question and great suggestion Tibor! I have updated the tests to no longer use the article page factory and instead use the image factory when we need it. I think this makes things simpler as we have a set string for the search description tests, and only use a factory for the image. Updated!

Issue #9541, opened 2022-10-25
[PNI] Switch related articles on category page when category changes


tbrlpld:
**Context:**
When switching between the categories, the related articles don't update unless the page is reloaded. 
https://user-images.githubusercontent.com/24797493/197841991-53d750ca-9185-4b5f-bc8a-567b1d647ad3.mov

**Acceptance Criteria:** 
- [x] As a user navigating the site, when I select a different category filter the 'Related articles' update to show related articles to that category
- [x] As a user navigating the site, when I select a category filter that doesn't have any articles, the 'Related articles' section doesn't show

NB: A prerequisite is that there are related articles created for that specific category

beccaklam:
The positioning of the related article also needs to be fixed. It should look like this:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/85c0d65f-3867-4ebf-a0a3-705f171bb72b)
[Link to design](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3080%3A30591)

tbrlpld:
@beccaklam Thanks, there already is #9540. That seems more like what you are talking about. This issue here is a about the functionality of switching which articles are displayed, not the location.

beccaklam:
@tbrlpld Oops thanks for catching that! I've moved my comment to the correct ticket.

fessehaye:
@kristinashu You might need to create some related articles for the category on staging to test this

kristinashu:
Sorry, asking @beccaklam to please review this! Ty!

beccaklam:
This looks good to go! Thank you!

Issue #9540, opened 2022-10-25
[PNI] Positioning of related articles on the category page


tbrlpld:
# Context: 

Current position does not match [design](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3080%3A30591). 

![image.png](https://images.zenhubusercontent.com/61e1e4f0d1fded7471402d12/487981be-812b-44cd-b221-a6c0c4d100f9)

# Acceptance Criteria: 

- [ ] The related article section on the categories page is placed inside the grid as per Figma design on desktop
- [x] The related article section on the categories page is placed inside the grid as per Figma design on mobile
- [ ] The related article section on the categories page is placed inside the grid as per Figma design on tablet

beccaklam:
![IMG_EA8D905AE42A-1.jpeg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/3a6a1cdd-89f8-42ed-92bc-abf51e4e8323)
Adding this screen shot to this bug. Not sure why it's saying "Primary Related Articles"?

beccaklam:
The positioning of the related article also needs to be fixed. It should look like this:
![image.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/85c0d65f-3867-4ebf-a0a3-705f171bb72b)
[Link to design](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=3080%3A30591)

kristinashu:
p3 since we can just hide (not add) articles for now

tbrlpld:
@beccaklam This is ready for QA on staging.

tbrlpld:
@beccaklam This is ready for QA on staging.

tbrlpld:
@beccaklam To QA this, you will have to add related articles to the category you want to test and then load that categories page. 

The toggle between the categories does not work yet, meaning the related articles will remain the same when you switch categories. The articles should update with the category only on page refresh. The dynamic switching of the related articles is part of #9541.

beccaklam:
Hi @tbrlpld this is working and looking mostly as expected! One small thing, can you increase the padding for desktop and mobile to 1.875rem? Currently it's 1rem but in the figma comp the padding is supposed to be 30px.

tbrlpld:
[...]

PR #9729, opened 2022-11-18
9540 positioning of related articles on category page

tbrlpld:
# Description

This PR positions the related articles (and while we are at is also the featured CTA) in the product grid on the category page. 
The previously used position was only used for backend development and was never the intended position.

Which category is actively displayed on the page can be changed with JS. 
This PR does not consider that. 
The display of the CTA and the related articles are only considering the category that is active when the page is first loaded.
Information about for which category the CTA should be displayed and which related articles are associated with which category is added to markup with data attributes. 
This is where the JS can get that data from when it is implemented in #9541. 

Link to sample test page: http://localhost:8000/en/privacynotincluded/categories/entertainment/
Related PRs/issues: #9540

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? I have added some comments where I think they are needed / appropriate. 

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/category_page.html
+++ b/network-api/networkapi/templates/pages/buyersguide/category_page.html
@@ -25,27 +25,32 @@
                 </div>
               </div>
             </div>
-
           </div>
         {% endif %}
       {% endwith %}
 
+    {% endfor %}
+
+    {% if featured_cta %}
+      {% comment %}
+        We render the featured CTA only once into the markup if one exists.
+        The categories only have a toggle to define if the featured CTA should be shown when the page is filtered for the category.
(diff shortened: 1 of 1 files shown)


danielfmiranda (approved):
Hi Tibor, just pulled this into my local machine and tested it out and can confirm its working as expected!

One thing I did want to mention was that on pages with less than 5-6 products, there is a gap in the grid: 
image

However, I dont think this is an issue once we merge it into our DB with many more products.

tbrlpld:
Thanks for catching this @danielfmiranda. Yea, I agree, let's see if this comes up with more realistic data and if Design would want us to change something if it does.

Issue #9601, opened 2022-11-02
DFL 2022 | Events - Text only card - URL field


cdanfon:
This ticket is to include a URL field to DFL 2022 | Events - Text only card #9398

Currently the _**heading link**_ comes from a page picker, we want to include an option for URL fields that opens in a new tab (as these are expected to be external links). 

The editor would have 2 options: 

- Page chooser
- URL field

**Dev tasks:** 

- Include a URL field
- Add some logic that prevents using both page picker & URL field at the same time
- Include some helper text: _'Please use one of the options below - either the internal page picker or the external URL field'_
- URL link to open in a new tab

Additional context https://github.com/mozilla/foundation.mozilla.org/issues/9398

kevinhowbrook:
This is ready for code review here https://github.com/mozilla/foundation.mozilla.org/pull/9739

Thanks

kevinhowbrook:
Thanks for the review @tbrlpld https://github.com/mozilla/foundation.mozilla.org/pull/9739 has been amended

tbrlpld:
@cdanfon This is now on staging.

cdanfon:
Thanks, looks good to me! 

@gutermul Over to you for testing

danielfmiranda:
Pushed to prod, closing!

PR #9739, opened 2022-11-21
Adds option to use URL field for teaser card links

kevinhowbrook:
# Description

We would like the option to use URLs in the text only teaser card. If used, they should open links in a new tab.

Link to sample test page: http://localhost:8000/en/youtube/ > Add text only teaser block
Related PRs/issues: #9601 

# Checklist

**Tests**
- [ ] ~Is the code I'm adding covered by tests?~ No

**Changes in Models:**
- [ ] ~Did I update or add new fake data?~ No
- [ ] ~Did I squash my migration?~ I didn't because they are single migration files in 2 apps
- [ ] ~Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~ I don't think this needs to be backwards compatible as it's a new field on an existing block

**Documentation:**
- [ ] Is my code documented?


network-api/networkapi/wagtailpages/pagemodels/customblocks/text_only_teaser.py
```diff
@@ -1,13 +1,30 @@
+from django.forms.utils import ErrorList
 from wagtail.core import blocks
+from wagtail.core.blocks.struct_block import StructBlockValidationError
 
 
 class TextOnlyTeaserCard(blocks.StructBlock):
 
     heading = blocks.CharBlock(help_text="Heading for the card.")
     heading_link = blocks.PageChooserBlock(required=False, help_text="Page that the header should link out to.")
```
tbrlpld, line 9: What do you think about renaming this field to `link_page` or something like that? I think that would feel a bit more consistent from a naming perspective.

tbrlpld (approved):
@kevinhowbrook This looks good to me. I left one optional question about the other link field name.

network-api/networkapi/wagtailpages/pagemodels/customblocks/text_only_teaser.py
```diff
@@ -1,13 +1,30 @@
...
     body = blocks.TextBlock(help_text="Body text of the card.", required=False, max_length=200)
 
+    def clean(self, value):
+        result = super().clean(value)
+        errors = {}
+
+        if value["heading_link"] and value["link_url"]:
+            errors["heading_link"] = ErrorList(["Please choose between a heading link OR a URL value."])
+        if errors:
+            raise StructBlockValidationError(errors)
```
tbrlpld, line 24: Can you add some context why are not raising the `StructBlockValidationError` directly when we detect the issue? 


```suggestion
        if value["heading_link"] and value["link_url"]:
            raise StructBlockValidationError(
              {"heading_link": ErrorList(["Please choose between a heading link OR a URL value."])}
            )
```

tbrlpld (commented):
Actually just left another question about the structure of the error raising logic.

network-api/networkapi/wagtailpages/pagemodels/customblocks/text_only_teaser.py
kevinhowbrook: Sure thing, added in bf60a832c284edc85749b846ba6e33f76d990d29

network-api/networkapi/wagtailpages/pagemodels/customblocks/text_only_teaser.py
kevinhowbrook: This is just preference I think, it allows for other errors to be added into the `errors` dict. I'll rework this to be more specific though

kevinhowbrook: Fixed in bdf625d3490912c61bdc08dee10fa43f3d077e9b

kevinhowbrook:
Rebased on main and migrations fixed in e3a222bb71008d6d1d8ee4a7a6bc8c36d01b2690

Issue #9602, opened 2022-11-02
DFL 2022 | Landing listing card - URL field


cdanfon:
This ticket is to include a URL field to the streamfield created from [DFL 2022 | Landing Listing card#9393](https://github.com/mozilla/foundation.mozilla.org/issues/9393)

Currently the _**heading link**_ comes from a page picker, we want to include an option for URL fields that opens in a new tab (as these are expected to be external links). 

The editor would have 2 options: 

- Page chooser
- URL field

**Dev tasks:** 

- Include a URL field
- Add some logic that prevents using both page picker & URL field at the same time
- Include some helper text: _'Please use one of the options below - either the internal page picker or the external URL field'_
- URL link to open in new tab

Additional context https://github.com/mozilla/foundation.mozilla.org/issues/9393

kevinhowbrook:
This is ready to code review here https://github.com/mozilla/foundation.mozilla.org/pull/9741

Thanks

kevinhowbrook:
Thanks @tbrlpld - https://github.com/mozilla/foundation.mozilla.org/pull/9741 has been amended

tbrlpld:
@cdanfon This is now on staging.

cdanfon:
Thanks, LGTM 

@gutermul over to you for testing

gutermul:
looks good to me. thanks.

danielfmiranda:
Pushed to prod! Closing

PR #9741, opened 2022-11-21
Adds URL field for listing card

kevinhowbrook:
# Description

Based off https://github.com/mozilla/foundation.mozilla.org/pull/9739

We would like the option to use URLs in the listing card. If used, they should open links in a new tab.

Link to sample test page: http://localhost:8000/en/youtube/ > Add listing block
Related PRs/issues: https://github.com/mozilla/foundation.mozilla.org/issues/9602

# Checklist

**Tests**
- [ ] ~Is the code I'm adding covered by tests?~

**Changes in Models:**
- [ ] ~Did I update or add new fake data?~ No
- [ ] ~Did I squash my migration?~ I didn't because they are single migration files in 2 apps
- [ ] ~Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~ I don't think this needs to be backwards compatible as it's a new field on an existing block

**Documentation:**
- [ ] Is my code documented?


network-api/networkapi/wagtailpages/pagemodels/customblocks/listing.py
```diff
@@ -13,6 +16,22 @@ class ListingCard(blocks.StructBlock):
 
     link = blocks.PageChooserBlock(required=False, help_text="Page that this should link out to.")
```
tbrlpld, line 17: What do you think about calling this something like `link_page`?

network-api/networkapi/wagtailpages/pagemodels/customblocks/listing.py
```diff
@@ -13,6 +16,22 @@ class ListingCard(blocks.StructBlock):
...
+    )
+
+    def clean(self, value):
+        result = super().clean(value)
+        errors = {}
+
+        if value["link"] and value["link_url"]:
+            errors["link"] = ErrorList(["Please choose between a link OR a URL value."])
+        if errors:
+            raise StructBlockValidationError(errors)
```
tbrlpld, line 31: I guess the same question as in #9739, about why we are using the `errors` instead of raising the error directly.

tbrlpld (approved):
Basically the same two questions as for #9739

network-api/networkapi/wagtailpages/pagemodels/customblocks/listing.py
kevinhowbrook: Thanks, changed in 470041f6164fa3dc18fb7a33ad8612ca82daae96

network-api/networkapi/wagtailpages/pagemodels/customblocks/listing.py
kevinhowbrook: Yep, reflected that in 470041f6164fa3dc18fb7a33ad8612ca82daae96

PR #9715, opened 2022-11-16
Fixes #9714 undefined variable error

mmmavis:
# Description

Fixes undefined variable error, following the same code style in the file.

1. Go PNI Articles page
2. Click on the search box
3. Make sure you don't see the `ReferenceError` error in dev tool console.

Link to sample test page:
Related PRs/issues: #9714

PR #9750, opened 2022-11-22
Get rid of variable ref error

mmmavis:
Code changes are exactly the same as in https://github.com/mozilla/foundation.mozilla.org/pull/9715 . I couldn't find an easy way to update my first commit message in the original PR (removing the `fixes` keyword) so I decided to just create a new PR.


danielfmiranda (approved):
Thanks Mavis! Approved 👍

Issue #9712, opened 2022-11-16
[PNI QA] Search is showing different results depending on where you search from


kristinashu:
Related to https://github.com/mozilla/foundation.mozilla.org/issues/9693 & https://github.com/mozilla/foundation.mozilla.org/issues/9553

### Describe the bug
Searching for `amazon` on the Product Reviews page shows:

<img width="600" alt="image" src="https://user-images.githubusercontent.com/1068783/202784502-c8772176-e9c4-45c6-870e-7a33d3299fa4.png">

while searching for `amazon` on the Articles page shows less accurate results:
<img width="600" alt="image" src="https://user-images.githubusercontent.com/1068783/202284099-42505ad3-a299-4a7f-ba36-499b08a2ec6f.png">

People have also noted that desktop and mobile can show different results.

### AC
- [ ] As a user, I see the same relevant search results no matter where on the site I search from or what my browser width is

@mtdenton putting this as a P1 since we know a high portion of users use search

mtdenton:
@cdanfon @fessehaye Let's make sure we get this groomed up first thing Monday.

tbrlpld:
@beccaklam I think something went wrong when uploading the example images, because the results look the same to me.

mmmavis:
> As a user, I see the same relevant search results no matter where on the site I search from or what my browser width is

Hello @kristinashu @fessehaye  I'm reviewing Simon's PR and wondering what the expected order the search results should be in. How do we define relevance? I noticed that "relevance" isn't listed in the sorting dropdown. If we always show most relevant search results first, should `relevance` be part of the dropdown options?

kristinashu:
You nailed it @mmmavis! Yes, `relevance` is `name` first, and then `company`, and then `blurb`

@fessehaye @beccaklam and I ended up talking about the dropdown sort order a bunch today and how it is confusing. We have some notes [here](https://docs.google.com/document/d/1opBAnQH_VjG5eHEe7txpkw6QBTeu79S0MWNT5-m7WnE/edit#heading=h.fs3gtczc7i18). For the scope of this ticket, we're just focusing on making things the same across all page. There will be follow up tickets to resolve the confusion caused by the sort order.

mmmavis:
@kristinashu Thanks! Will check out the doc.

fessehaye:
@kristinashu this is ready for staging

kristinashu:
Looks good, ready for prod!

PR #9760, opened 2022-11-24
9712 - Search is showing different results

fessehaye:
# Description

Main problem is not the search showing different results, its that the products are being sorted differently when the search text is apart of the starting URL

See related issue for repro steps. You might need to use the prod database to test

Related PRs/issues: #9712

### DIFF
--- a/network-api/networkapi/templates/pages/buyersguide/catalog.html
+++ b/network-api/networkapi/templates/pages/buyersguide/catalog.html
@@ -79,7 +79,7 @@
         <div class="tw-w-full tw-px-4 xlarge:tw-px-0 tw-flex tw-relative tw-flex-wrap medium:tw-flex-nowrap medium:tw-left-auto medium:tw-ml-0 medium:tw-static">
           <div class="tw-flex tw-items-end tw-space-x-2 tw-overflow-auto tw-pb-2 tw-no-scrollbar tw-touch-pan-x tw-w-full medium:tw-w-auto medium:w-auto subcategory-header">
 
-            <span id="product-filter-pni" class="tw-flex tw-cursor-pointer tw-text-gray-60 border tw-border-gray-20 tw-px-2 tw-py-1 tw-font-sans tw-rounded-3xl tw-font-normal tw-text-[12px] tw-leading-[1.3] hover:tw-border-blue-10 hover:tw-bg-blue-10">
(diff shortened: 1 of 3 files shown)


network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -79,7 +79,7 @@
         <div class="tw-w-full tw-px-4 xlarge:tw-px-0 tw-flex tw-relative tw-flex-wrap medium:tw-flex-nowrap medium:tw-left-auto medium:tw-ml-0 medium:tw-static">
           <div class="tw-flex tw-items-end tw-space-x-2 tw-overflow-auto tw-pb-2 tw-no-scrollbar tw-touch-pan-x tw-w-full medium:tw-w-auto medium:w-auto subcategory-header">
 
-            <span id="product-filter-pni" class="tw-flex tw-cursor-pointer tw-text-gray-60 border tw-border-gray-20 tw-px-2 tw-py-1 tw-font-sans tw-rounded-3xl tw-font-normal tw-text-[12px] tw-leading-[1.3] hover:tw-border-blue-10 hover:tw-bg-blue-10">
```
fessehaye, line 82: this change is just to solve this visual issue 
image

source/js/buyers-guide/search/history.js
```diff
@@ -117,7 +117,12 @@ export function performInitialHistoryReplace(
       });
   }
 
-  Utils.sortProductCards();
+  // sorting priorities change when search text is applied
+  if (searchParameter) {
+    Utils.sortFilteredProducts();
+  } else {
+    Utils.sortProductCards();
+  }
```
mmmavis, line 125: Hi @fessehaye , 

I don't know the specs and codebase well enough. Could you help clarify why we have two different sorting functions? I imagine the sorting logic would be the same whether it's a full product list or a filtered one. The only difference would be which project list we apply the same sorting function on.

fessehaye: So `sortFilteredProducts()` applies how products should be sorted from last years update https://github.com/mozilla/foundation.mozilla.org/issues/7476

Where `sortProductCards()` applies what sort is applied from the pni-category dropdown that was added this years update.

So the sorting logic is different whether it's a full product list or a filtered one. This is a little confusing so we added it for discussion for the next PNI meeting!

mmmavis: Got it! Thanks for the explanation.

mmmavis (approved):
R+!

PR #9768, opened 2022-11-25
Fix Tito event newsletter check

tomusher:
Resolves an issue with the newsletter check when receiving a Tito webhook


danielfmiranda (approved):
Looks good to me! Approved

PR #9769, opened 2022-11-25
9766 - Stop translating slugs on multiple pages

TheoChevalier:
Making sure [these slugs](https://pontoon.mozilla.org/fr/foundation-website-content/all-resources/?search=slug&status=translated) get removed from Pontoon


tbrlpld (approved):
Thanks @TheoChevalier for fixing this.

Issue #9747, opened 2022-11-22
[FRU] Add Apple Pay Verification Files


mtdenton:
For the foundation stack, we need to add two Apple Pay Verification files to the /.well-known directory in order to enable Apple Pay for foundation.mozilla.org and mozillafestival.org.

We can accomplish this by taking the contents of the files provided by Fundraise Up and saving them as environment variables in Heroku, then selecting which to serve based on the base URL of the request.

@mtdenton will provide the files upon request

### Dev Tasks
- [x] Create two production environment variables on Heroku, one for each file provided, and store the file's contents. Keep the names mnemonic for their respective base URLs
- [x] Write logic to serve the contents of these files when Apple Pay checks (base url)/.well-known/apple-developer-merchantid-domain-association

### Acceptance Criteria
- [ ] As Apple, I can navigate to foundation.mozilla.org/.well-known/apple-developer-merchantid-domain-association and see the verification code for foundation.mozilla.org
- [ ] As Apple, I can navigate to mozillafestival.org/.well-known/apple-developer-merchantid-domain-association and see the verification code for mozillafestival.org

mtdenton:
Added environment variables listed in #9779 into staging to test environment variables on foundation.mofostaging.net and mozillafestival.mofostaging.net, both were successfully verifed on Fundraise Up. Ready for production.

PR #9779, opened 2022-11-29
added apple pay domain association view

danielfmiranda:
# Description

Related PRs/issues: #9747

For the foundation stack, we need to add two Apple Pay Verification files to the /.well-known directory in order to enable Apple Pay for foundation.mozilla.org and mozillafestival.org.

It has been requested to take the contents of these 2 files provided by Fundraise Up, and save them as environment variables. Then using logic in the backend, serve the contents of the associated files when Apple Pay checks
"(base url)/.well-known/apple-developer-merchantid-domain-association"

**This PR adds:**
- Two new environment variables: 
  - `APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION` 
  - `APPLE_PAY_DOMAIN_ASSOCIATION_KEY_MOZFEST`
 - A new view `apple_pay_domain_association_view`, which is called when Apple Pay or a user visits the route `/.well-known/apple-developer-merchantid-domain-association`. This view checks whether the request is being made from "foundation.mozilla.org" or "mozillafestival.org" and will serve what is set for the associated environment variable.

# Steps to test

1. Check this branch out
2. In your .env file, add string values for `APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION`, `APPLE_PAY_DOMAIN_ASSOCIATION_KEY_MOZFEST`. (EX: "test_value_123")
3. Visit http://localhost:8000/.well-known/apple-developer-merchantid-domain-association, and the value you set for `APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION` should appear.

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


network-api/networkapi/views.py
```diff
@@ -19,3 +22,15 @@ def review_app_help_view(request):
...
     else:
         return HttpResponse(status=404)
+
+
+@require_GET
+def apple_pay_domain_association_view(request):
+    """
+    Returns string needed for Apple Pay domain association/verification.
+    """
+    if Site.find_for_request(request).hostname == "www.mozillafestival.org":
```
danielfmiranda, line 32: Hi @mtdenton!

Had a quick question about this line and one of the dev tasks in the ticket:

Dev Task: "Create two **_production_** environment variables on Heroku, one for each file provided, and store the file's contents. Keep the names mnemonic for their respective base URLs"

Based on the dev task above, I am only checking for the production URL of Mozfest. I was wondering however if we wanted to instead update this to include the staging URL (www.mozillafestival.mofostagin.net) and the local dev URL (mozfest.localhost:8000) for testing purposes?

Or, is apple pay verification something that only happens on prod?

Thanks in advance!

tbrlpld: @danielfmiranda I think hardcoding host names is not a good idea, exactly because of the different environments. It would also require code changes when the configuration changes (e.g. the domain). 

The foundation site and the Mozfest site have separate page types for their homepage. You could use that to determine which key should be served.

tbrlpld (commented):
@danielfmiranda See my comment on your question. 

Also, what do you think of a small test for this custom view?

network-api/networkapi/views.py
```diff
@@ -19,3 +22,15 @@ def review_app_help_view(request):
...
+        verification_key = settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY_MOZFEST
+    else:
+        verification_key = settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION
+    return HttpResponse(verification_key, content_type="text/plain")
```
tbrlpld, line 36: Considering that we can not guarantee in code that the environment variables are set, what do you think about responding with different error codes: 

* 400 Bad request - the site can not be matched to either of the cases (Foundation or Mozfest -- based on Site root page type)
* 501 Not implemented - no verification key can be determined while the site is valid (e.g. for the case where the environment variable is not set).

danielfmiranda:
Hi @tbrlpld thanks for some great suggestions! 

I have implemented your feedback to the view, as well as some tests in `test_views.py`, and this PR is now ready for re-review.

Thanks again!

network-api/networkapi/views.py
danielfmiranda: Great suggestion, updated!

network-api/networkapi/views.py
danielfmiranda: Updated!

network-api/networkapi/tests/tests.py
```diff
@@ -407,7 +407,7 @@ def test_general_page_export_fields(self):
 
 class TestCreateWagtailImageUtility(TestCase):
     def setUp(self):
-        self.image_path = abspath(join(dirname(__file__), "../media/images/placeholders/products/teddy.jpg"))
+        self.image_path = abspath(join(dirname(__file__), "../../media/images/placeholders/products/teddy.jpg"))
```
danielfmiranda, line 410: Since I moved this file to the newly created /tests/ directory, I had to update this path.

network-api/networkapi/views.py
```diff
@@ -19,3 +24,37 @@ def review_app_help_view(request):
...
+            status_code = 200
+        else:
+            response_contents = key_not_found_message
+            status_code = 501
+
+    else:
+        response_contents = "Request site not recognized."
+        status_code = 400
+
+    return HttpResponse(response_contents, status=status_code, content_type="text/plain; charset=utf-8")
```
tbrlpld, line 60: **Kudos** Nice. Very clear and easy to follow.

tbrlpld (approved):
Thanks for the tests and the restructuring of the view. This is very clear now 👍

Issue #9667, opened 2022-11-10
Sort Python imports consistently with `isort`


tbrlpld:
# Description

> isort is a Python utility / library to sort imports alphabetically, and automatically separated into sections and by type. It provides a command line utility, Python library and [plugins for various editors](https://github.com/pycqa/isort/wiki/isort-Plugins) to quickly sort all your imports. It requires Python 3.6+ to run but supports formatting Python 2 code too.

Without a defined sort order it can be hard to see which objects are already imported. 
Defining a sort order and checking that code adheres to it manually is pretty laborious.
A linter can check the import sorting and a formatter can do the sorting for us.
`isort` combines both.

# Dev tasks

- [x] `isort` is installed and configured for the project
- [x] There is an `inv` command to check the import sorting
- [x] There is an `inv` command to apply the import sorting
- [x] CI checks that all imports are sorted correctly with `isort`
- [x] All files are sorted correctly with `isort`

# Acceptance criteria

- [x] `isort` is installed and configured for the project
- [x] There is an `inv` command to check the import sorting
- [x] There is an `inv` command to apply the import sorting
- [x] CI checks that all imports are sorted correctly with `isort`
- [x] All files are sorted correctly with `isort`

tbrlpld:
Everything is set up and formatted in #9781

PR #9780, opened 2022-11-29
#9667 Set up `isort`

tbrlpld:
# Description

This PR adds the [isort](https://pycqa.github.io/isort/index.html) code formatter for consistent import sorting for Python and configures that all Python files are checked during CI. Currently the CI step is allowed to fail. It will be made mandatory with a follow-up PR that also formats all existing files.

This PR also adds `inv isort-check` and `inv isort` commands to run isort locally. These commands are also run then  `inv lint` or `inv format` are run respectively. 

This PR also configures which files `flake8` should ignore in the config. This makes it unnecessary to always specific the files to check on the command line and thus more reproducible and less error prone.

Related PRs/issues: #9667

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? The commands have help strings.

### DIFF
--- a/.github/workflows/continous-integration.yml
+++ b/.github/workflows/continous-integration.yml
@@ -2,7 +2,7 @@ name: Continuous Integration
 
 on:
   push:
-    branches: ["main", "pni-2022"]
+    branches: ["main"]
     paths-ignore:
       - maintenance/**/*.*
   pull_request:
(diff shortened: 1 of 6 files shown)


danielfmiranda (approved):
Hi @tbrlpld, just gave this a test and can confirm that its working for me.

Thanks for taking this work on as I can see this being very helpful in the future! 

Approved 👍

PR #9789, opened 2022-12-01
Added dependabot check to Percy CI

fessehaye:
# Description
Lets try to add a check to not run percy when dependabots make a PR to save some screenshots on our free plan


mmmavis (approved):
Code looks great based on the [GitHub doc](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/automating-dependabot-with-github-actions) you linked us @fessehaye 

Note that we probably need to re-enable Percy for dependabot PRs when we start having the capacity to go through them.

Issue #9709, opened 2022-11-16
[PNI] SPIKE | Image Deferment


beccaklam:
Placeholder ticket to address slow loading on PNI. Some recorded pain points:

> From Nov 14 QA 
> [Rebecca] Loading category pages can sometimes take a long time, wondering if we have anything in the works to speed things up? 
    [Nancy] +1, especially when you click a category page from About or Articles. Usually loading is fine from the homepage or product reviews page +1

> [Rebecca] Not sure if anyone else is noticing this but when I use PNI on mobile touch seems delayed a lot of the time? I often have to click a filter, or the search icon, three times before something happens?
    [Kristina] I’m getting this really bad on my phone in Chrome

cc: @mtdenton

kristinashu:
Could this be related to https://github.com/mozilla/foundation.mozilla.org/issues/9545?

mtdenton:
> Could this be related to #9545?

Lack of deferment may be part of it, but @tbrlpld thinks the root cause is more likely to be inefficient database querying.

tbrlpld:
@kristinashu #9545 was why I asked is the above issues were occurring in logged-in state, but @beccaklam had clarified that the issues described above happen when not-logged-in.

tbrlpld:
This might get more tricky (so good thing it's a spike), because I just had a look and `loading="lazy"` is already defined. But, it also does not appear to be respected. When I load the page all images get loaded before the JS is loaded. Something is off here. 

<img width="2030" alt="Screen Shot 2022-11-17 at 07 06 40" src="https://user-images.githubusercontent.com/24797493/202482525-62191559-aee6-498e-b3c5-7440e624abfc.png">


There might also be an opportunity here by defining a few more source options for the browsers to decide which image to load. E.g. on mobile we display the image at 170x170, but load 600x600 or even 720x720 on retina displays. 

```html
<picture class="product-thumbnail">
<source srcset="https://assets.mofoprod.net/network/images/Euki.2e16d0ba.fill-360x360.png 1x, https://assets.mofoprod.net/network/images/Euki.2e16d0ba.fill-720x720.png 2x">
<img class="product-thumbnail" loading="lazy" src="https://assets.mofoprod.net/network/images/Euki.2e16d0ba.fill-600x600.png" alt="link to Euki">
</picture>
```

beccaklam:
[...]

PR #9792, opened 2022-12-02
9709 - Fix lazy image loading + width/height attr

fessehaye:
# Description

Working on the image spike, after adding width/height to the element I reduced the numbered of images loaded on the homepage from the products from ~44 > 14. I had to play with the width/height values a bit.
```
<img
        class="product-thumbnail tw-w-full" <---- added class
        {% image product.image fill-600x600 as img %}
        loading="lazy"
        src="{{ img.url }}"
        width="500"     <---- added attr
        height="500"    <---- added attr
        alt="{% blocktrans with product=product.title %}link to {{product}}{% endblocktrans %}"
      >
 ```
The width and height are way higher then the image actually is because we just set the width to 100% but having the width/height attribute actually affects how loading="lazy" works. So by just adding these attrs we don't lose anything + we get no visual regressions. Increasing with width+height to a higher value then 500+ doesn't decrease the amount of images loaded which is strange.
I think if we want to get that 14 product images > 4 or less we should probably use a js implementation of lazy loading since with using the loading tag does not really give us control of the viewport threshold that is used to detect what gets loaded. But I am not sure that optimization is as high priority/high reward imo.
At the meantime, I am going to add this change as a PR since this is a small change without any setbacks or visual regressions I have found so far.

Link to sample test page: /en/privacynotincluded
Related PRs/issues: #9709

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/item.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/item.html
@@ -41,10 +41,12 @@
         srcset="{{ img_1x.url }} 1x, {{ img_2x.url}} 2x"
       >
       <img
-        class="product-thumbnail"
+        class="product-thumbnail tw-w-full"
         {% image product.image fill-600x600 as img %}
         loading="lazy"
         src="{{ img.url }}"
+        width="500"
+        height="500"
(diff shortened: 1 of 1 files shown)


danielfmiranda (approved):
Hey Simon, 

Looks good to me! 

Approved 👍

Issue #9034, opened 2022-06-30
Setup linting for Django templates


Pomax:
I ran across https://github.com/prettier/prettier/issues/5581#issuecomment-1169111071 and as I recall, there was no linting for the django code (yet), so some of the mentioned solutions might actually be good fits for the parts that currently left unlinted.

tbrlpld:
I agree, it would be very valuable to setup linting for Django templates. One option is: https://djlint.com/

If we find the above too much, at the very least we should make sure our templates are indented correctly. That could be achieved with https://github.com/rtts/djhtml

PR #9806, opened 2022-12-05
#9034 set up django template linting

tbrlpld:
# Description

Related PRs/issues: #9034 

This PR adds [`djlint`](https://djlint.com/) to the dev tooling. `djlint` is a linter and formatter for HTML templates. 

This PR also adds `inv` commands to run `djlint` in lint, check and format mode. 

* lint mode - shows template code issues and potential bug sources.
* check mode - checks if the template code fits the formatting style.
* format mode - changes the template format to fit the consistent style. 

Additionally, CI is updated to run format check and linting for the templates. This CI step is still allowed to fail. Format and linting will be fixed in follow up PR that then also require the respective CI step to pass. 

While testing the linting, I was running into issues with the `wagtailadmin/404.html` which caused the linting process to hang. It turns out those issues were caused by the large embedded SVG. I have moved the SVG into a separate file the is included, which is allowing the linting to finish successfully. 

Also, this PR is ordering the functions in the `tasks.py` file to make it a little more structured.

These are the statistics of the first pass of the linter:

```
D004 5   (Django) Static urls should follow {% static path/to/file %} pattern.
D018 31  (Django) Internal links should use the {% url ... %} pattern.
H005 2   Html tag should have lang attribute.
H006 96  Img tag should have height and width attributes.
H008 57  Attributes should be double quoted.
H012 2   There should be no spaces around attribute =.
H013 23  Img tag should have an alt attribute.
H014 69  Found extra blank lines.
H017 317 Tag should be self closing.
H020 4   Empty tag pair found. Consider removing.
H021 3   Inline styles should be avoided.
H023 3   Do not use entity references.
H025 52  Tag seems to be an orphan.
H026 31  Empty id and class tags can be removed.
H030 3   Consider adding a meta description.
H031 3   Consider adding meta keywords.
T001 467 Variables should be wrapped in a single whitespace.
T002 185 Double quotes should be used in tags.
T003 357 Endblock should have name. Ex: {% endblock body %}.
T032 84  Extra whitespace found in template tags.

Linted 289 files, found 1794 errors.
```

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? Tasks have docstrings.


tasks.py
```diff
@@ -437,3 +457,24 @@ def pip_sync(ctx):
             "docker-compose run --rm backend ./dockerpythonvenv/bin/pip-sync requirements.txt dev-requirements.txt",
             **PLATFORM_ARG,
         )
+
```
danielfmiranda, line 460: I think separating these into their own section is a great idea!

tbrlpld:
Thanks for reviewing @danielfmiranda

fessehaye:
Moving this for ready to prod since its a dev task

PR #9822, opened 2022-12-08
Update padding of image blocks

danielfmiranda:
# Description
Related PRs/issues: #9825

Image blocks previously had their container's x-padding removed through `blocks.scss` like so:
```scss
.image-block {
  @extend .px-0;
}
```

However, only `main.scss` was importing the `blocks.scss` module. Meaning that this rule was not being carried over to image blocks used in the buyers guide, since that instead uses `bg-main.scss`.

This PR resolves the issue by removing the SCSS above and instead adding the tailwind class `tw-px-0` to the image block's container classes. 

Adding the tailwind class to the template means that image blocks will have their container's x-padding removed regardless of where they are used, and also helps towards our effort of using as little SCSS as possible.

# Screenshots

### Before:

206596143-62e265e7-45be-444e-acf8-c050dc5e0e7a

### After:
Screenshot 2022-12-08 at 16-46-27 Privacy Not Included A Buyer’s Guide for Connected Products

# Steps to test

1. Checkout this branch, and create/edit a BuyersGuideCampaignPage. (These can be found/created under "Homepage" -> "Privacy Not Included" -> "Articles")
2. In the "Body" field, add a new "image" block, with an image of your choice.
3. Publish the page, and scroll to your image. It should be left aligned with the rest of the text content around it.
4. If everything is looking as expected, testing is complete!


fessehaye (approved):
thx for the explanation!

Issue #9761, opened 2022-11-24
[PNI] Increase padding at the bottom of 'What to read next' sections


beccaklam:
The margin bottom for 'What to read next' sections should be closer to 120px, currently it is only 3rem (48px).
![Screenshot 2022-11-24 at 10.48.44 AM.png](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/c05252ec-1ef1-4d42-9972-d11f7475d563)
This was found during QA of the category page but it should also be applied any where the 'What to read next' component is found.

Acceptance Criteria
- [ ] The bottom margin of the 'What to read next' component is closer to 120px or 7.5rem or whatever is the closest figma token.

fessehaye:
@nancyt1 Would you be able to check this since Becca is away?

fessehaye:
currently on staging

nancyt1:
Bottom padding looks good! Per slack message, it'd be great to add 1 more rem to the top margin (3rem to 4rem) but other than that looks good!

danielfmiranda:
Pushed to prod! Closing

PR #9826, opened 2022-12-09
Updated margin-top for What to read next section

fessehaye:
# Description

Just updating margin-top from 3rem-> 4rem based on Nancy Feedback on #9761


danielfmiranda (approved):
Thanks for the tag Simon! Looks good to me, approved 👍

Issue #8933, opened 2022-06-09
Update node container


tbrlpld:
# Description

When running commands like `inv new-env` or `inv catch-up` the `package-lock.json` file keeps getting updated. 

I believe this might have something to do with the node version that is used in the Dockerfile (which is 14.13.1). 

I see errors like this: 
```
npm WARN read-shrinkwrap This version of npm is compatible with lockfileVersion@1, but package-lock.json was generated for lockfileVersion@2. I'll try to do my best with it!
```

We should probably update the node version that is used in the container. We also need to make sure that running `inv new-env` or `inv catch-up` can be run without changed to the `package-lock.json` file. 

We probably want to make sure that the development node and the live version also match. Production is currently on `19.2.0`. See also: https://devcenter.heroku.com/articles/nodejs-support
We can set this to something else too in the `package.json`. 
Considering the support table, using version `18.x` seems reasonable, LTS supported until 2025. 

# Acceptance criteria

- [ ] As a developer on my local setup, when I run `inv new-env` or `inv catchup` the `package-lock.json` file is not updated but stays the same as before I ran either command.
- [ ] The development container uses the `node` version as we do in production.

tbrlpld:
@cdanfon This issue has just prevented a release to staging and now requires extra work to get `main` to a releasable state. Would be good for us to address this sooner rather than later.

tbrlpld:
Could use some ACs I think

cdanfon:
Are you OK to add some Tibor?

tbrlpld:
I'll add it for the next grooming round

tbrlpld:
Yea, happy to do this as part of grooming.

tbrlpld:
Staging is running `18.13.0`.

tbrlpld:
When creating a new environment with `inv new-env` we also get version `18.13.0`. This should also update to the latest minor and patch version based on the time of the environment generation.

PR #9835, opened 2022-12-12
#8933 Update and limit Node version to 18

tbrlpld:
# Description

Previously, we were using Node 14 for local development and CI. But, the Node version did not have an upper limit in the `package.json` file. Also, the `package-lock.json` was using the outdate version 1 format.

This lead to a few things. When installing a new dependency, a bigger change in the lock file would occur because the lockfile would change format version. Also, the lockfile format version was switching with every other PR that made any kind of changes to it, I guess based on if the dev used a local or the container node to make the change. Production was running on Node 19 while we develop and run CI on Node 14. Node 14 is soon to be deprecated.

This PR updates the local and CI versions of Node to version 18 and also sets this as a range in the `package.json`. The information in the `package.json` will be used on Heroku. I have opted to go with version 18, instead of 19 (which is already used in production) because of the longer cycle for version 18, which is until 2025, while support for 19 end next year (2023).

Additionally, this PR updates the lockfile format to version 2. Along with the version updates, this should prevent the outdated format from being used again. 

See also: 
* https://devcenter.heroku.com/articles/nodejs-support

Related PRs/issues: #8933

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


tbrlpld:
Don't merge yet, because of code freeze for winter break 🥶

danielfmiranda (approved):
Looks good to me 🙌 thanks for taking this on Tibor!

fessehaye (approved):

Issue #9603, opened 2022-11-02
[PNI QA] Hero article image should pull OG image


kristinashu:
Currently the hero image pulls the image from the Articles `Hero image` but it should pull from `Share Image`. 

Share image is better because
- It has an aspect ratio closer to what is used on the homepage and therefor it will look better
- Not all articles will have a `Hero Image`
- Our main site typically pulls from the `Share image`

Note: This work is related to related ticket https://github.com/mozilla/foundation.mozilla.org/issues/9606

### AC
- [ ] As a user, I see the `Share Image` on the homepage hero article when that article is features (for this to work the share image needs to be set on the article, which will be required in #9606). 

### Screen caps

Screen cap of homepage feature article
<img width="1199" alt="image" src="https://user-images.githubusercontent.com/1068783/199552526-cd2a669d-f3ed-4c0d-8718-ab6df4102d13.png">

Screen cap of 'Hero image' field in CMS
<img width="625" alt="image" src="https://user-images.githubusercontent.com/1068783/199553696-92a65068-d03e-4bdb-afb3-30b6e9a88a81.png">

Screen cap of 'Share image' field (image that should be used in the homepage feature)
<img width="634" alt="image" src="https://user-images.githubusercontent.com/1068783/199555245-7ebd710d-37d2-4101-9422-172c391f874b.png">

tbrlpld:
@kristinashu Is it more likely for an article to have a share image than a hero image? Would it make sense to fall back to the hero image if no share image is set? Do we need to need to warn editors when they are selecting the featured article and it does have neither share nor hero image?

kristinashu:
Ah this raises something else... the Share Image and Meta Description were recently made required fields on our main blog site https://github.com/mozilla/foundation.mozilla.org/issues/9464 so they should also be required fields for PNI Articles. Should this request be a new ticket or added to this one?

<img width="721" alt="image" src="https://user-images.githubusercontent.com/1068783/199579716-b78e7aa3-6fff-4b00-8577-dcb54e85bf09.png">

tbrlpld:
@kristinashu Though considering the questions this has raised before, maybe combine them here or have the "make required" ticket be blocking to this one here.

kristinashu:
Ok, I've added a note to the first comment in this ticket - _Note: This work is related to related ticket https://github.com/mozilla/foundation.mozilla.org/issues/9606_

[...]

PR #9836, opened 2022-12-12
Fix bug with featured_article_url

mmmavis:
# Description

This is to fix a [bug](https://github.com/mozilla/foundation.mozilla.org/issues/9603#issuecomment-1346963720) I introduced in my previous PR where I had `endwith` too soon in the file and caused the variable `featured_article_url` to "die" too early.  

To test, go to `/privacynotincluded` and make sure the title of the featured article is linked to the article and not /privacynotincluded.

Link to sample test page: /privacynotincluded
Related PRs/issues: #9603


danielfmiranda (approved):
Thanks for the tag @mmmavis! Just tested this out and can confirm that it is fixed. Approved 👍

Issue #7862, opened 2021-11-15
[PNI QA] Mobile tab alignment


sabrinang:
- [ ] Tab alignment (starts a bit too far in on smaller viewports)
<img width="300" alt="image" src="https://user-images.githubusercontent.com/2492510/138957843-14df377d-a3e6-4ee3-a90a-d44e6b09afc0.png">

After a breakpoint it shifts right:
![z09uSt3QOY](https://user-images.githubusercontent.com/2492510/141844587-29868cbc-d5d4-4e13-9c3d-cff5e514fa28.gif)

mtdenton:
Surprisingly, still relevant a year later. I'm sending this up to the top of the backlog.

fessehaye:
@sabrinang  should be ready on https://foundation.mofostaging.net/en/privacynotincluded/

sabrinang:
@fessehaye 
- The spacing alignment looks good on desktop (tested Firefox108, Chrome108, Safari 16, DuckDuckGo 0.31 on MaxOS 12.6)

- On mobile, the spacing alignment is fine though I notice there is a very narrow gap below the tab on Chrome & DDG browsers but not Firefox
(tested on Mobile Browsers: firefox107, chrome 108 ,ddg 5 on Android 13)
![Screenshot_20221215-142305](https://user-images.githubusercontent.com/2492510/207986024-0340cf45-5190-4b22-817e-61faf12cb5bb.jpg)

fessehaye:
I can not repro this on my devices! I created a followup ticket for this issues https://github.com/mozilla/foundation.mozilla.org/issues/10000
Since this ticket has been already been added to staging/prod we will close this ticket and hopefully a dev with an iPhone can repeat this and figure out why this is happening

PR #9843, opened 2022-12-14
7862 - Mobile tab alignment

fessehaye:
# Description

Adjusting mobile tab alignment for the product review page on small break points (<576px)
See related issue for gif example of issue
Link to sample test page: en/privacynotincluded -> click on product -> scroll down to tab on mobile breakpoint

Related PRs/issues: #7862


danielfmiranda (approved):
Hi Simon! 

Thanks for the tag, after testing this out on my local machine, I can confirm that there is no more "jump" when going from the small to xsmall breakpoint.

Approved!

Issue #9757, opened 2022-11-23
[PNI] Translating content categories leads to server error


tbrlpld:
## Description

This is a follow up to #9563. 

When trying to translate a buyer's guide content category a server error occurs. 

To reproduce:

* Visit the admin
* Go to the "buyer's guide content category" snippet screen
* Create a new category
* Try to translate the newly created category
* See error

```

Traceback (most recent call last):
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/core/handlers/exception.py", line 47, in inner
    response = get_response(request)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/core/handlers/base.py", line 181, in _get_response
    response = wrapped_callback(request, *callback_args, **callback_kwargs)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func
    response = view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/urls/__init__.py", line 125, in wrapper
    return view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail/admin/auth.py", line 195, in decorated_view
    return view_func(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/views/generic/base.py", line 70, in view
    return self.dispatch(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/views/submit_translations.py", line 242, in dispatch
    return super().dispatch(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/django/views/generic/base.py", line 98, in dispatch
    return handler(request, *args, **kwargs)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/views/submit_translations.py", line 190, in post
    return self.form_valid(form)
  File "/usr/local/lib/python3.9/contextlib.py", line 79, in inner
    return func(*args, **kwds)
  File "/app/dockerpythonvenv/lib/python3.9/site-packages/wagtail_localize/views/submit_translations.py", line 216, in form_valid
    single_translated_object = self.object.get_translation(
[...]

Issue #9107, opened 2022-07-14
[PNI] - Editorial Content Index Page - Filter by Content category


fessehaye:
https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=2303%3A126762

Content Filter Types names are All, Articles, Advice, Petition, and In the Press (which is visually separated and will not show up in "All")

Dev Tasks:

* [ ] Style Filter Component
* [ ] Add JS to filter articles by tag

Acceptance Criteria:

* [ ] Should be able to filter PNI articles/campaigns by page type or tag

![Article Index Page - All](https://user-images.githubusercontent.com/74557207/187467012-0fbc62c9-3d68-4096-85b7-269806d02079.jpg)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-300)

tbrlpld:
@fessehaye Are you sure you want to do this in JS? We have very similar functionality already on the backend and the index page from which we are going to build the index page.

cdanfon:
@nancyt1 to update design, @fessehaye to update the description

tbrlpld:
Functionality is outlined in the dev notes: https://docs.google.com/document/d/12AmWzlOVll6hHBPL-lu3E4rAtOM9vX6wgO6U8uOxCxk/edit#heading=h.qedu6bgm180y

beccaklam:
Nancy has updated the design. @fessehaye as Tibor mentioned in the comment above we have notes on the filtering functionality in the dev notes (see under 'Content index page'). Copying part of the notes here:

> - "All articles" lists all non-categorized articles, all advice-category (or otherwise categorized) articles and all campaign pages. It should NOT show the news updates!
> - Filter index by content category, there already is similar functionality on the index page. Show a filtering label for each content category (in the beginning there will only be one) that filters the list of entries to only those that belong to the content category. 
> - Filter index to show only campaign pages (this can be done based on the page type)
> - Show list news updates (these were not shown before)

tbrlpld:
Removing from this sprint since it's p3.

PR #9854, opened 2022-12-15
#9757 content category translations

tbrlpld:
# Description (Outdated)

Before, when trying to translate a Buyer's Guide Content Category, you would get a server error. 

This is issue was caused by having a `unique` constraint on the `slug` field. The issue is that the copy has the same slug, and saving fails and then the translate screen has no copy to present. 

This PR removes the slug from the panels, so the editors can not see it, because the value is set automatically. Additionally, the slug is automatically generated during the `clean` step and now contains the locale language code for the locale of the category. If the same slug already exists a `ValidationError` is raised. Since the `slug` field is not visible to the editor, it does not make much sense to show an error for that field. Rather, we raise an error for the `title` field, because that is the only editor controlled field that influences the `slug`. 

The move of the slug setting and validating from the `save` to the `clean` method enables the raising of the `ValidationError` and have that properly handled in the Wagtail admin interface.

As a small extra, this PR also updated the content category fragment to only take the list of categories and not the surrounding page object anymore. This should make that fragment more reusable. Also, I have removed the links to the content index page, because we are currently no able to filter the content index page and this was a rather confusing user experience. The filtering of the content index page will be implemented in #9107.

# Description (Update 2022-12-21)

While working on implementing the review feedback I came across the issue that I filed in Wagtail regarding the issue with `unique_together` constraints (https://github.com/wagtail/wagtail/issues/8918). The original work around attempt with the unique slug was what was causing the translation issue. 

Looking at the newer feedback on the issue made me realize that we might be able to actually achieve the intended functionality with a small fix to the validation logic. As it turns out, that hunch was correct. 
The original issue in Wagtail is cause by the `locale` field being set to `editable = False`. This leads to Django excluding the field when performing field validation including the uniqueness checks (see also https://docs.djangoproject.com/en/4.1/ref/models/fields/#editable). 
The small fix is to just remove the `locale` field from the `exclude` list.

We can now rely on the normal Wagtail form check to handle slugs that already exist per locale, while allowing the same slug to be used on a different locale (which is needed when first instantiating the copy for translation).

Link to sample test page: http://localhost:8000/cms/snippets/wagtailpages/buyersguidecontentcategory/?locale=en
Related PRs/issues: #9757 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented? Through tests and comments.


network-api/networkapi/wagtailpages/pagemodels/buyersguide/taxonomies.py
```diff
@@ -9,14 +11,42 @@ class BuyersGuideContentCategory(wagtail_models.TranslatableMixin, models.Model)
...
         verbose_name_plural = "Buyers Guide Content Categories"
 
     def __str__(self) -> str:
         return self.title
 
-    def save(self, *args, **kwargs) -> None:
-        if not self.slug:
-            self.slug = text_utils.slugify(self.title)
-        super().save(*args, **kwargs)
+    def clean_slug(self):
```
danielfmiranda, line 23: **Kudos** Just wanted to leave kudos for a really well documented function 👍 I also appreciate that this is being done it its own `clean_slug` function instead of the `clean` method itself! 


**Question**

While testing this method out in the CMS, I can confirm that translation seems to work. However, I do notice that I get the following warning after publishing. Was wondering if this is to be expected?

image

network-api/networkapi/templates/fragments/buyersguide/content_category_links.html
```diff
@@ -1,10 +1,12 @@
...
-        <span class="{{ classes }} {{ extra_classes }}">/</span>
-      {% endif %}
-    {% endwith %}
-  {% endfor %}
-{% endwith %}
+{% if content_categories %}
+  <div class="tw-h6-heading tw-mr-2 tw-mb-0 {{ extra_classes }}">
+    {% for content_category in content_categories %}
+        <span>
+          {{ content_category.title }}
```
danielfmiranda, line 5:
```suggestion
          {{ content_category.localized.title }}
```

Hey Tibor! 
I am thinking that we may need to update the template to render something such as `category.localized.title`.

The reason I bring this up, is because even after translating a category and then visiting a localized article page, it appears that the English category is still being rendered.

For example, I translated the category "Agency"'s French title to "Test Agency". 

image

However, when visiting a French article page, it is still rendering the English title. Even after running sync translations, etc:

Screenshot 2022-12-19 at 5 16 05 PM

danielfmiranda (commented):
Hey Tibor! 

Thanks for taking this work on, I gave the code a look through and tested this out on my local machine. 

Left some questions and suggestions, but besides that every things looking good to me!

network-api/networkapi/wagtailpages/tests/buyersguide/test_taxonomies.py
```diff
@@ -1,71 +1,69 @@
...
-        category_fr_locale.save()
+        category_default_locale.clean_slug()
+        category_fr_locale.clean_slug()
 
-        self.assertEqual(category_fr_locale.slug, category_default_locale.slug)
+        self.assertEqual(category_fr_locale.title, category_default_locale.title)
         self.assertNotEqual(category_fr_locale.locale, category_default_locale.locale)
+        self.assertNotEqual(category_fr_locale.slug, category_default_locale.slug)
+
+    def test_can_clean_and_save_copy_for_translation(self):
```
kevinhowbrook, line 61: Look like this test is just testing clean and save together but with no expected assertions. I'd suggest it's worth throwing some expectation in there, like asserting the saved objects slug is what you'd expect to have been stored.

kevinhowbrook (approved):
Thanks @tbrlpld, this all looks great to me. Good tests too.

I would suggest that the `clean_slug()` method is actually pulled out into it's own method for other models to use, just because it sounds like this may pop up again.

network-api/manage.py
```diff
@@ -4,7 +4,7 @@
 
 # Django runs twice to support live-reloading, so check Django's internal settings to determine whether or not
 # to start the debugger.
-if (os.environ.get("RUN_MAIN") or os.environ.get("WERKZEUG_RUN_MAIN")) and os.environ.get("VSCODE_DEBUGGER"):
+if (os.environ.get("RUN_MAIN") or os.environ.get("WERKZEUG_RUN_MAIN")) and os.environ.get("VSCODE_DEBUGGER") == "True":
```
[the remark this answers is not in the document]
tbrlpld: It was not, in the sense that there was no way to disable the debugger setup. Environment variables are strings, so setting `VSCODE_DEBUGGER=False` in `.env` would still pass this check, because the string `"False"` is still a "truthy" value.

So now, you will need to make sure to set `VSCODE_DEBUGGER=True` in `.env` to activate the debugger setup (which seems like it was the intended functionality of this check. 

I need to be able to disable the debugger because I don't use VSCode and therefore need a way to connect a different debugger. With the debugger server running though, the breakpoints would be swallowed by that and my debugger would never start.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/taxonomies.py
tbrlpld: Ah good catch! Thanks @danielfmiranda I had missed this. I believe this is because the translatable fields are not specified explicitly. I will add that.

network-api/networkapi/wagtailpages/tests/buyersguide/test_taxonomies.py
tbrlpld: Good call.

network-api/networkapi/templates/fragments/buyersguide/content_category_links.html
tbrlpld: Thanks for catching this. I must admit, I did not really test the frontend at all 🤦‍♂️ 

In general though, I would highly recommend against the use of `.localize` in templates. This is because that accessing that property will likely trigger an additional database query. As of now, this property is also not cached, so every access triggers another database query: https://github.com/wagtail/wagtail/blob/cbe9dc729b7cc92229b535c9bedd719b1a12c9e8/wagtail/core/models/__init__.py#L782-L796
This, I believe, is probably one of the reasons we end up with 80 queries on basically empty pages. 

As an alternative strategy, I would recommend using the `get_context` or other retrieval method on the page that ensures that the data used in the template is using the correct locale. In the case of this model, the `get_content_categories` seems like a good location to me.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/taxonomies.py
```diff
@@ -9,14 +11,42 @@ class BuyersGuideContentCategory(wagtail_models.TranslatableMixin, models.Model)
...
+        """
+        We would really want the slugs to be only unique per locale.
+        If we implement that with a UniqueContraint or unique_together,
```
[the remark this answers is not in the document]
tbrlpld: Thanks for the catch.

network-api/networkapi/templates/fragments/buyersguide/content_category_links.html
tbrlpld: This is addressed in 22a8e613535711c288c570066e05a99f9b26e354

danielfmiranda (approved):
Thanks for the updates Tibor! Looks good to me 👍

tbrlpld:
@danielfmiranda @kevinhowbrook Sorry for requesting re-review, but I just found a way to significantly simplify this fix and actually get the original intended unique-slug-per-locale validation in place.

danielfmiranda (approved):
Great find Tibor! 

Approved 👍

Issue #9720, opened 2022-11-18
[PNI] Implement sticky nav


beccaklam:
Implementation ticket for design ticket #9589

### Goal Description:
Implement sticky nav functionality for PNI's category nav

### Links
[Figma File](https://www.figma.com/file/GwdoUiY4fzVyeLfXsAJGPq/*Privacy-Not-Included-Website?node-id=5171%3A93008&t=qIBGqUXhA8SnRb7n-1)
[Redpen](https://redpen.io/p/wy75eeaf54e8ebcc04)

### Acceptance Criteria:
- [ ] As a user I can see the full PNI nav on page load (see image 1)
- [ ] As a user when I start scrolling down the page on PNI, and the main nav is longer visible, I see that the category nav becomes sticky whereas the upper nav stays out of sight (see image 2)
- [ ] As a user when I start scrolling back up, before I am at the top of the page, the full nav appears again and is sticky (see image 3)
- [ ] As a user I see this behaviour on mobile, tablet and desktop screens

**Image 1**
![Menu on page load.jpg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/9fd0d6cc-9cc0-4b3a-80de-25dee0cbe415)

**Image 2**
![Sticky Nav Scroll Down.jpg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/da61f73e-9974-4dad-8c84-b2a99c23ec8f)

**Image 3**
![Sticky Nav Scroll Up.jpg](https://images.zenhubusercontent.com/57acd65ce40e5714b16d4215/b4273e95-2994-49cd-a080-ca5defb4d964)

ArshErgon:
Hey @beccaklam  I would like to work on this, can you please assign it to me?

fessehaye:
@beccaklam can you check this on staging?

beccaklam:
@fessehaye Hey Simon, this looks to be working well on desktop and tablet screen sizes, but on mobile or at the mobile screen size I can't seem to access the drop down menu?

https://user-images.githubusercontent.com/20977376/212381844-c4169f1c-7855-4012-82cb-0bb1352da622.mov

fessehaye:
@beccaklam addressed feedback on staging

beccaklam:
@fessehaye This appears to be fixed! :) Looks ready to go to me!

PR #9855, opened 2022-12-15
9720 - Implement sticky nav

fessehaye:
# Description

Need to use position fixed instead of sticky since removing the top half of the nav with position sticky conflicts with the scroll event listener and makes the navigation shake and change the scrolling context. Since we are using position fixed, we are adding some padding top on the body to fill the space where the navigation would usually take

Implement sticky navigation for PNI pages when scrolling down the top half of the nav is hidden and is visible again if you scroll up.

Link to sample test page:  All PNI related pages
Related PRs/issues: #9720 

image
image
image

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_mobile_nav.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_mobile_nav.html
@@ -1,7 +1,7 @@
 {% load bg_nav_tags localization i18n static wagtailroutablepage_tags %}
 {% get_bg_home_page as home_page %}
 
-<div class="d-md-none mt-0 mb-0" id="pni-nav-mobile">
+<div class="d-md-none mt-0 mb-0 tw-bg-white" id="pni-nav-mobile">
     <div class="container">
         <div class="row px-3 px-sm-0">
             <div class="col-12 tw-border-0">
--- a/network-api/networkapi/templates/pages/buyersguide/base.html
+++ b/network-api/networkapi/templates/pages/buyersguide/base.html
@@ -76,7 +76,7 @@
 
   {% get_bg_home_page as home_page %}
 
-  <div class="primary-nav-container-wrapper tw-border-b tw-border-blue-10 tw-fixed tw-z-[999] tw-top-0 tw-w-full tw-transition-all tw-overflow-hidden">
+  <div class="primary-nav-container-wrapper tw-border-b tw-border-blue-10 tw-fixed tw-z-[999] tw-top-0 tw-w-full tw-transition-all">
     {% include "fragments/buyersguide/primary_nav.html" %}
     {% include "fragments/buyersguide/pni_mobile_nav.html" with pagetype=pagetype categories=categories current_category=current_category %}
     {% block category_nav %}
--- a/network-api/networkapi/templates/pages/buyersguide/base.html
+++ b/network-api/networkapi/templates/pages/buyersguide/base.html
@@ -62,7 +62,7 @@
 {% endblock %}
 
 
-{% block bodyclass %}pni{% endblock %}
+{% block bodyclass %}pni tw-pt-[150px]{% endblock %}
 
 
 {% block script_bundle %}
@@ -76,7 +76,7 @@
 
   {% get_bg_home_page as home_page %}
 
-  <div class="primary-nav-container-wrapper tw-border-b tw-border-blue-10">
+  <div class="primary-nav-container-wrapper tw-border-b tw-border-blue-10 tw-fixed tw-z-[999] tw-top-0 tw-w-full tw-transition-all tw-overflow-hidden">
     {% include "fragments/buyersguide/primary_nav.html" %}
     {% include "fragments/buyersguide/pni_mobile_nav.html" with pagetype=pagetype categories=categories current_category=current_category %}
     {% block category_nav %}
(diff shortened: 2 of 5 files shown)


network-api/networkapi/templates/fragments/buyersguide/pni_mobile_nav.html
```diff
@@ -1,7 +1,7 @@
 {% load bg_nav_tags localization i18n static wagtailroutablepage_tags %}
 {% get_bg_home_page as home_page %}
 
-<div class="d-md-none mt-0 mb-0" id="pni-nav-mobile">
+<div class="d-md-none mt-0 mb-0 tw-bg-white" id="pni-nav-mobile">
```
fessehaye, line 4: Need this else the dropdown is transparent background

network-api/networkapi/templates/pages/buyersguide/base.html
```diff
@@ -62,7 +62,7 @@
 {% endblock %}
 
 
-{% block bodyclass %}pni{% endblock %}
+{% block bodyclass %}pni tw-pt-[150px]{% endblock %}
```
mmmavis, line 65: Could you add a comment explaining why we need a 150px padding here?

network-api/networkapi/templates/pages/buyersguide/catalog.html
```diff
@@ -2,7 +2,7 @@
 
 {% load static env i18n static wagtailimages_tags cache bg_nav_tags localization %}
 
-{% block bodyclass %}pni catalog{% endblock %}
+{% block bodyclass %}pni catalog tw-pt-[150px]{% endblock %}
```
mmmavis, line 5: Same. Could you add a comment explaining why we need a 150px padding here?

source/js/buyers-guide/pni-sticky-nav.js
```diff
@@ -0,0 +1,21 @@
+const PNIStickyNav = {
+  init: () => {
+    const navContainer = document.querySelector(".primary-nav-container");
+
+    if (!navContainer) return;
+
+    let lastScroll = 0;
+    window.addEventListener("scroll", () => {
+      let currentScroll = window.pageYOffset;
+      if (currentScroll - lastScroll > 0 && currentScroll >= 300) {
```
mmmavis, line 10: I'm also curious why the number `300` is chosen. Is it possible that we use JS to find out the value instead of hardcode it to 300?

mmmavis (changes requested):
Hey @fessehaye , 

This PR seems to break the "More Categories" dropdown (dropdown becomes hidden). Could you look into that, please? I haven't looked at the code closely as I assume it's gonna change after you fix the dropdown issue.

mmmavis:
@fessehaye 

I don't know if this breaks other things but moving the category nav code from the wrapper seems to do the trick (category nav already has sticky property in CSS, see `header.scss` line 12)

In `base.hml` around line 78, change it to
```html
  <div class="primary-nav-container-wrapper">
    {% include "fragments/buyersguide/primary_nav.html" %}
    {% include "fragments/buyersguide/pni_mobile_nav.html" with pagetype=pagetype categories=categories current_category=current_category %}
  </div>
  {% block category_nav %}
    {% include "fragments/buyersguide/pni_category_nav.html" with show_all_reviews_as_active_category=False categories=categories current_category=current_category %}
  {% endblock %}
```

And apply the border classes `tw-border-b tw-border-blue-10` to `<nav id="multipage-nav" ...>`

source/js/buyers-guide/pni-sticky-nav.js
fessehaye: Honestly it was just a value I eyeballed and showed Rebecca since we don't want the nav to transform too much at the first bit of scrolling since it was a little distracting. Rebecca was good with the current setting. Should I just move it to a variable and comment it?

network-api/networkapi/templates/pages/buyersguide/base.html
fessehaye: Mentioned it in the description but since we are using position:fixed for the nav, everything on the body shifted up since the nav is around 150px so I thought adding that height as padding should be enough to make the rest of content look normal. This was the solution I thought about that didn't include using JS to calc the padding just in case people disable JS.

network-api/networkapi/templates/pages/buyersguide/catalog.html
fessehaye: Answered in comment above

fessehaye:
@mmmavis  Good callout on the category dropdown! Found removing one of the `overflow:hiddens` actually solved the problem!

mmmavis (commented):
@fessehaye thanks for the comments. Did you get a chance to check out the approach I put in this [comment](https://github.com/mozilla/foundation.mozilla.org/pull/9855#issuecomment-1355745284)? I'm wondering if it introduces any side bugs. Scroll event is especially costly so we should avoid using it and opt for a non-JS solution unless there's no other alternative. Arbitrary numbers in JS and CSS also make the codebase hard to maintain so let's not use them.

If design has any concern (e.g., "don't want the nav to transform too much at the first bit of scrolling"), let's chat as a group. I strongly vote for a non-JS solution for this ticket and will try to push back on design refinement request.

mmmavis:
Hey @fessehaye , I opened up a [draft PR](https://github.com/mozilla/foundation.mozilla.org/pull/9876) to demo what I meant with the JS-free approach. Navs are sticky for me. Could you give it a try and let me know if it for some reason still not working for you?

fessehaye:
We are going to work off https://github.com/mozilla/foundation.mozilla.org/pull/9876 instead

Issue #9548, opened 2022-10-26
Fix CI deprecation warnings


tbrlpld:
We currently get a bunch of deprecation warnings for every GitHub action run. I think we need to fix this to make sure CI is not going to break. 

![image.png](https://images.zenhubusercontent.com/61e1e4f0d1fded7471402d12/827a01f9-fa5f-4ce1-8ac7-5c49929a25fe)

mmmavis:
Fixed!

mmmavis:
Fixed!

PR #9952, opened 2023-01-13
Fix CI deprecation warnings

mmmavis:
Related PRs/issues: #9548 

To address the CI warnings below. End result with this fix can be found [here](https://github.com/mozilla/foundation.mozilla.org/actions/runs/3917214430).


.github/workflows/ci-image-diff-sync.yml
```diff
@@ -50,10 +50,10 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-python@v2
       with:
-        python-version: 3.9.9
+        python-version: '3.9.9'
```
mmmavis, line 53: Even though we have node version specified we are still getting the warning. I'm guessing it's because we are missing quotes? 

https://github.com/actions/setup-python#basic-usage

.github/workflows/ci-image-diff-sync.yml
```diff
@@ -50,10 +50,10 @@ jobs:
...
     - uses: actions/setup-node@v2
       with:
-        node-version: 18
+        node-version: '18'
```
mmmavis, line 56: Even though we have node version specified we are still getting the warning. I'm guessing it's because we are missing quotes? 

https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#example-only-run-job-for-specific-repository

.github/workflows/ci-image-diff-sync.yml
```diff
@@ -53,7 +53,7 @@ jobs:
         python-version: 3.9.9
     - uses: actions/setup-node@v2
       with:
-        node-version: 18
+        node-version: 16
```
mmmavis, line 56: Even though we have node version specified, we are still getting the warning. I'm wondering if it's because we need to get it to 16 as per [suggestion](https://github.blog/changelog/2022-09-22-github-actions-all-actions-will-begin-running-on-node16-instead-of-node12/)?

.github/workflows/ci-image-diff-sync.yml
```diff
@@ -47,13 +47,13 @@ jobs:
       XSS_PROTECTION: True
 
     steps:
-    - uses: actions/checkout@v2
+    - uses: actions/checkout@v3
```
mmmavis, line 50: because https://github.com/actions/checkout/blob/main/CHANGELOG.md#v300

.github/workflows/ci-image-diff-sync.yml
```diff
@@ -47,13 +47,13 @@ jobs:
...
     - uses: actions/setup-python@v2
       with:
         python-version: 3.9.9
-    - uses: actions/setup-node@v2
+    - uses: actions/setup-node@v3
```
mmmavis, line 54: because https://github.com/actions/setup-node/releases/tag/v3.0.0

.github/workflows/ci-image-diff-sync.yml
```diff
@@ -47,13 +47,13 @@ jobs:
...
       with:
-        node-version: 18
+        node-version: 16
```
mmmavis, line 56: Not sure if we wanna stick to `16` for consistency with `actions/checkout@v3`'s node version (16) or 18

https://github.com/actions/setup-node#usage

mmmavis:
ping @tbrlpld in case you missed the PR review request notification :D

.github/workflows/ci-image-diff-sync.yml
tbrlpld: I think 16 is the new minimum version, I don't think they mean we should go backwards from 18 to 16.

.github/workflows/ci-image-diff-sync.yml
tbrlpld: I think we should stick with '18' to make sure we use the same version locally, in CI and on staging and prod.

tbrlpld (commented):
@mmmavis Thanks for taking this on. 

I think we should stick with node version `'18'` to make sure we use the same version locally, in CI and on staging and prod.

mmmavis:
@tbrlpld fine with me with sticking to node 18 too. I've updated the PR. CI seems happy too.

tbrlpld (approved):

Issue #9968, opened 2023-01-17
Order snippets alphabetically


tbrlpld:
## Description

Without a proper order for the snippets, staff end up creating duplicate items on accident.

We should prevent this by at least ordering all snippets alphabetically if no other natural / required order exists for them. 

See also:
* https://mozilla.slack.com/archives/C98GW3P7X/p1673970641783389?thread_ts=1673970641.565049&cid=C98GW3P7X

## Acceptance criteria 

- [x] Snippets are ordered alphabetically (if no other order is required)

tbrlpld:
@kristinashu You had opened a related issue, would you mind giving this a quick check on staging? 

* https://foundation.mofostaging.net/cms/snippets/
* https://foundation.mofostaging.net/cms/snippets/wagtailpages/profile/

tbrlpld:
@kristinashu All snippets that did not specify a different ordering before are sorted alphabetically now.

* Blog page topic
* Buyer's guide content category
* Buyer's guide product update
* Callpower
* Donation CTA
* Area of focus
* Profile
* Pulse filter
* Research region
* Research topic
* Signup
* Tito event

The menu position should not have changed. This is a most curious thing 😵‍💫

tbrlpld:
I can reproduce the jumping menu position 😳 Though I have no hunch as to where this is coming from.

kristinashu:
Nice, all those snippets are in alphabetical order!

I thought the menu moved depending on the page but it actually sometimes moves even if I refresh the same page!?

tbrlpld:
Yea, i get the same havior on staging, but not locally... very weird. Tried FF and Safari and get the same result. Sometimes it's at the top sometimes at the bottom... 

Maybe we give it a little bit of time to out rule caching... cause right now I have no other idea where this is coming from.

PR #9969, opened 2023-01-17
Order snippets alphabetically

tbrlpld:
# Description

This PR adds alphabetically ordering to all snippets that don't already have an ordering defined. 

This is to help staff avoid accidental creation of duplicates.

Link to sample test page: http://localhost:8000/cms/snippets
Related PRs/issues: #9968

# Checklist

**Tests**
- ~~[ ] Is the code I'm adding covered by tests?~~ No, just a quick config fix.

**Changes in Models:**

**Documentation:**


danielfmiranda (approved):
Hi @tbrlpld!

Thanks for the tag, just tested this out and can confirm that everything is working as expected and that the code looks good to me as well.

Approved! 👍 

However, I do have one small comment about the `Highlight` snippet model that I wanted to bring up, it seems that these are being ordered by having the newest instances at the end of the list, meaning that I have to skip through 9 pages to get to my newly created highlight.

I know that this is what the code already had defined, but I feel like it might be worth double-checking if this is something that the team might also want to be updated in this ticket.

tbrlpld:
Thanks for the view @danielfmiranda. 

Regarding the `Highlights`: I think you bring up a good point that we might want to review this and check with other staff. Looking at the code though, the `Highlight` model has a dedicated "order" field that is used for the default ordering. Considering that we would need to find out the purpose for that and if a change would be desirable, I think this is outside of the quick fix scope of this PR and possibly better dealt with in a separate issue.

Issue #9821, opened 2022-12-08
[PNI] Using backspace in mobile search is very delayed


tbrlpld:
### Describe the bug

In the mobile search bar on the PNI homepage, when using backspace it takes a long time for the text to actually disappear. The problem seems to get worse the more often the key is used. 

It seems like this has something to do with the debounce.

### To Reproduce

1. Go to https://foundation.mozilla.org/en/privacynotincluded/ with an iPhone.
2. Click on the search icon.
3. Type something, e.g. "Germin".
4. Try to correct your typo by backspacing.
5. Notice that it takes a long time for the text to disappear.
6. Once the text does disappear, try typing new text, e.g. "Garmin"
7. This will also not appear immediately character-by-character, but take some time and then appear all at once.

### Expected behavior

When hitting the backspace the text should disappear immediately.

### Screenshots

Here is a screen recording. In times where it seems like nothing is happening, I am typing or trying to delete... That is the issue.

https://user-images.githubusercontent.com/24797493/206480747-c198e9f6-cb34-431d-bedf-eb16c304be4b.MP4

### Smartphone (please complete the following information):
 - Device: iPhone 14 Pro
 - OS: iOS 
 - Browser: Safari (but does not matter)
 - Version: 16

mmmavis:
I'm curious if this is also happening on Android/Windows phones.

tbrlpld:
😱 Are Windows phones a thing again?

mmmavis:
hahah I knew someone is gonna say something about it. I was trying to be inclusive-ish lol

fessehaye:
@beccaklam can you test staging on your mobile and see if this issue has gotten better or worse?

tbrlpld:
The tying delay has gotten better, but I am experiencing some weirdness with the visibility of the search bar now. Seems like the search bar is sometimes hidden or blocked by the dropdown when scrolling back up. Also, it sometimes takes quite some time for the grid items to show again. 


https://user-images.githubusercontent.com/24797493/213828314-353b449b-d1f8-4012-ae6e-e01d8fe9a479.MP4

beccaklam:
@fessehaye Hi! Sorry just testing this now. The backspace is working better but I'm also experiencing all the bugs that Tibor mentioned above?

tbrlpld:
@beccaklam Do you remember is these issues where an issue before the backspace fix? I don't remember, but it might have been because I was distracted with the backspace issue. 

[...]

PR #10002, opened 2023-01-20
9821 - Potential Safari Delay fix for Search

fessehaye:
# Description

What I came up after using safaris profile tools. (when it wasn't crashing)

* minor fix for `toggleCtaForCategory`
* Making it so only `toggleScrollAnimation()` is triggered only on load on non-category pages instead of triggered every time we search products. Reasoning for this, was noticing that a lot of the stalling on safari might be coming from resetting the ScrollTrigger libraries "triggers" whenever the products change.  `toggleCategoryAnimation()` will be used instead since the animation is way less performance reliant then the scrollTrigger() one. Doing this change made my safari browser had less issues with using the searchbox.
* There is still a chance that there might be other root causes of this issue but, I still think these changes are important to include since I think maintaining some of the scrollanimation stuff in this situation.

Link to sample test page: en/privacynotincluded
Related PRs/issues: #9821

### DIFF
--- a/source/js/buyers-guide/search/search-filter.js
+++ b/source/js/buyers-guide/search/search-filter.js
@@ -62,7 +62,7 @@ export class SearchFilter {
         )
     ).then(() => {
       if (this.categoryTitle.value === "None") {
-        Utils.toggleScrollAnimation(true);
+        Utils.toggleScrollAnimation();
       } else {
         Utils.toggleCategoryAnimation(true);
       }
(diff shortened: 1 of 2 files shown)


danielfmiranda (approved):
Just tested this out by running this branch on my localhost and then connecting through my Iphone. 

When following the workflow shown in the video in the original issue ticket, the delay seems to be resolved. I am not experiencing any delays.


Approved! 👍

Issue #9902, opened 2023-01-05
Unable to use anchor links for H2 text in Image Teaser Blocks on MozFest Website


marcwalsh:
### Describe the bug
When creating Image Teaser Blocks they have heading and body text. In other blocks we're able to link to those headings as an anchor link. However these headings are missing fragment ID that we would normally use for anchor links. 

### To Reproduce
1. Go to '[https://www.mozillafestival.org/en/issues-and-spaces](https://www.mozillafestival.org/en/issues-and-spaces)'
2. Click on MoZone (which links to https://www.mozillafestival.org/en/issues-and-spaces/transparency#mozone)
3. Page does not scroll to MoZone heading as part of Image Teaser Block

### Expected behaviour
I'm able to create anchor links using the headings added to Image Teaser Blocks.

mtdenton:
Hi March, also looking to get this in during Sprint 1.

Grabbing for discussion today.

cdanfon:
Hey @mmmavis can you update this ticket please? Thanks

tbrlpld:
@marcwalsh Can you provide a link to a page that contains blocks with linkable headings? 

Also, I am not sure what should happen when you click on "MoZone". The element seems to be just a simple paragraph and an image, so clicking usually does not have an effect.

tbrlpld:
@marcwalsh Was the first link meant to go to: https://www.mozillafestival.org/en/issues-and-spaces/transparency/ instead of https://www.mozillafestival.org/en/issues-and-spaces/?

tbrlpld:
## Prior art

`wagtailcustom_tags.py`

#7348
#6679
#5259
#2892

#2788


The styleguide does contain linkable headings, but these are all manually created in the template.

marcwalsh:
Hi @tbrlpld - I forgot we removed the links from the images on the [Issues and Spaces page](https://www.mozillafestival.org/en/issues-and-spaces).

Under [MozFest 2023 Spaces on the Issues and Spaces page](https://www.mozillafestival.org/en/issues-and-spaces/#mozfest-2023-spaces), we would link from MoZone (as an example) to the MoZone section on the [Transparency page](https://www.mozillafestival.org/en/issues-and-spaces/transparency#mozone) - I've added a screenshot of where the anchor for MoZone should go to:

[...]

PR #10037, opened 2023-01-21
#9902 Add anchor links to image teaser block headings

tbrlpld:
# Description

This PR adds anchor tags to the headings of the image teaser block. 
The anchor are set to link to the section id, which is a slug generated from the heading. 

The anchor does not have any content and thus will have no visual representation. 
While this is not ideal, this is a pattern that staff is already used to based on how anchor elements are generated for h2 headings in richtext (see https://github.com/mozilla/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailcustomization/templatetags/wagtailcustom_tags.py#L33). 

The difference to the richtext headings is that the anchor has an `href` attribute in stead of an `id` and that the `id` is set on the surrounding `section`. This was chosen to make sure that the whole section scrolls into view properly. 

This is best tested with production data (`inv copy-prod-db`) on this page: http://mozfest.localhost:8000/en/issues-and-spaces/transparency/#mozone
When on `main` the page will load scrolled to the top.
With this branch it will scroll the MoZone section into view. 

Link to sample test page: http://mozfest.localhost:8000/en/issues-and-spaces/transparency/#mozone
Related PRs/issues: #9902 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/customblocks/image_teaser_block.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/customblocks/image_teaser_block.py
@@ -1,7 +1,15 @@
 from django.forms.utils import ErrorList
+from django.utils import functional as func_utils
+from django.utils import text as text_utils
 from wagtail.core import blocks
 from wagtail.core.blocks.struct_block import StructBlockValidationError
 from wagtail.images.blocks import ImageChooserBlock
+
+
+class ImageTeaserValue(blocks.StructValue):
+    @func_utils.cached_property
+    def slug(self):
+        return text_utils.slugify(self.get("title", ""))
 
 
 class ImageTeaserBlock(blocks.StructBlock):
@@ -60,3 +68,4 @@ class ImageTeaserBlock(blocks.StructBlock):
         label = "Image teaser"
         icon = "doc-full"
         template = "wagtailpages/blocks/image_teaser_block.html"
+        value_class = ImageTeaserValue
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/image_teaser_block.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/image_teaser_block.html
@@ -3,9 +3,9 @@
 
 {% block block_content %}
     {% with card=value %}
-        <section class="tw-grid tw-grid-cols-1 large:tw-grid-cols-2 tw-gap-5 large:tw-gap-[60px] tw-mb-7 {{ divider_styles }}">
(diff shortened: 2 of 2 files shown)


network-api/networkapi/wagtailpages/pagemodels/customblocks/image_teaser_block.py
```diff
@@ -1,9 +1,17 @@
...
 from wagtail.core.blocks.struct_block import StructBlockValidationError
 from wagtail.images.blocks import ImageChooserBlock
 
 
+class ImageTeaserValue(blocks.StructValue):
+    @func_utils.cached_property
+    def slug(self):
+        return text_utils.slugify(self.get("title", ""))
+
+
```
kevinhowbrook, line 14: This is neat - I would have slugged the title and saved it as a slud value, but I prefer this a lot more!

kevinhowbrook (approved):

mmmavis (commented):
Hi @tbrlpld , do you know if the MozFest team would want the heading itself to appear as hyperlink? One concern with having an `<a>` without content like `<a href="#mozone"></a>` is not good with a11y.

network-api/networkapi/wagtailpages/pagemodels/customblocks/image_teaser_block.py
tbrlpld: Yea, I thought it's unlikely for us to retrieve the blocks by slug, so a prop felt simpler.

tbrlpld:
> Hi @tbrlpld , do you know if the MozFest team would want the heading itself to appear as hyperlink? One concern with having an `<a>` without content (e.g., `<a href="#mozone"></a>`) is that it's not good with a11y.

@mmmavis That's a fair concern. I think we should definitely revise how we do this. It might also make sense to make these links easier to discover, rather than using the dev tools. 

I have opted for this approach because there is prior art in that this is how headings are made linkable in richtext as well. 

I have had a look at https://amberwilson.co.uk/blog/are-your-anchor-links-accessible/ but decided to stick with something closer to the existing pattern. 


Are you ok with moving ahead with this an creating a follow up issue to deal with accessibility concern in all use cases and possibly also address the discoverability?

mmmavis:
@tbrlpld thanks for sharing how you decided with this approach. I'm good with visiting the a11y issue in a follow-up PR. Do you mind creating a ticket for that?

mmmavis (approved):

tbrlpld:
@mmmavis Created #10052 as follow up

PR #10049, opened 2023-01-24
Localize URLs in article_listing_compact.html

### LINKED ISSUE #10047: Localize URLs in article_listing_compact.html
Links in the Related section are hardcoded with `/en/`

Example: https://foundation.mozilla.org/sw/privacynotincluded/netatmo-smart-video-doorbell/

Capture d’écran 2023-01-24 à 11 44 55

TheoChevalier:
Fixes #10047

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/article_listing_compact.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/article_listing_compact.html
@@ -1,4 +1,4 @@
-{% load i18n static wagtailcore_tags wagtailimages_tags %}
+{% load i18n localization static wagtailcore_tags wagtailimages_tags %}
 
 {% with heading_level=heading_level|default:2 heading_classes=heading_classes|default:"tw-h3-heading tw-mt-2"  %}
     <div>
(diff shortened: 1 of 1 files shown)


tbrlpld:
@TheoChevalier Can you add some context why `relocalized_url` is necessary? Also, I would like to caution against the use of `.localize` on object. We should try to basically never use that in the templates because this is likely to create multiple database queries. We should make sure on a view/page level that data of the correct locale is added to the template context instead of requesting these in the template itself. 

There is a whole epic on removing as much of `.localize` from the templates as possible. 

This case is a particularly important one, because the `.localize` access happens in a loop, creating the database queries for every element in the loop (typical 1 + n situation).

I have just recently created a whole epic on this topic: https://app.zenhub.com/workspaces/mofo-engagement-585335eab729771d0736378d/issues/gh/mozilla/foundation.mozilla.org/9882 

You should be able to find some more context in there. 

Also, please let me know if you have any questions.

TheoChevalier:
Hi @tbrlpld, thanks, I wasn’t aware of the performance issues, I’ve just replicated what we have in `article_card_horizontal.html` to handle this (also the only way I know how to fix this)

As far as I remember, we can’t get Wagtail to use localized URLs with `pageurl` so we created `relocalized_url` for this use case.

I’ll have a look at the issue you linked, thanks!

tbrlpld:
@TheoChevalier You are right, this was the predominant way in which this was fixed in the past. I am not sure we were aware (I wasn't) that this is an uncached property that triggers 2 queries each time it's used. 

We should try to avoid introducing more of them and the linked epic is defined to remove the existing usages.

jhonatan-lopes:
Duplicate of #10248

Issue #10051, opened 2023-01-24
[PNI] Search bar becomes hidden on mobile when scrolling down then up


kristinashu:
Issue discovered in - [[PNI] Using backspace in mobile search is very delayed #9821 ](https://github.com/mozilla/foundation.mozilla.org/issues/9821#issuecomment-1399096152)

Via @tbrlpld 
> I am experiencing some weirdness with the visibility of the search bar now. Seems like the search bar is sometimes hidden or blocked by the dropdown when scrolling back up. Also, it sometimes takes quite some time for the grid items to show again. 

This bug has been confirmed by me and @beccaklam

### AC
- [ ] (on mobile) As a user, if I click on the magnifying glass icon to activate search mode, I can see the search bar becomes sticky as I scroll the page.
- [ ] (on mobile) As a user, if search mode is not on, I can see the category bar becomes sticky as I scroll the page.
- [ ] (on mobile) As a user, when search mode is activated, the top right magnifying glass icon becomes a 'x' button which can be used to close the search bar

### Describe the bug
On mobile, the search bar is sometimes hidden or blocked by the dropdown when scrolling down and then back up. 

### To Reproduce
1. Go to https://foundation.mozilla.org/en/privacynotincluded/ with an iPhone
2. Click on the search icon
3. Type something, e.g. "Garmin"
4. Scroll down 
5. Scroll back up
6. See error

### Expected behavior
The active search is still visible when I scroll back up.

### Screenshots
https://user-images.githubusercontent.com/24797493/213828314-353b449b-d1f8-4012-ae6e-e01d8fe9a479.MP4

### Smartphone
 - Device: iPhone12 mini
 - OS: iOS 16.1.2
 - Browser: Safari

tbrlpld:
Thanks for creating this @kristinashu 🙏

mmmavis:
@kristinashu @beccaklam  I'm wondering which component we want to make sticky as user scroll down the page? Right now the behaviour seems a bit unpredictable (buggy)? From the video you attached, the first time you scrolled down, the category nav wasn't sticky and it later became sticky after you scrolled up and down again. Since the purple search box takes over where the category nav is after user clicks on search icon, I'm wondering if we want to change the sticky pattern to

- if purple search box is visible, search box should become sticky (and not the category nav) as users scroll down the page
- if purple search box is not visible (search not toggled on), category nav should be sticky

mmmavis:
[...]

PR #10053, opened 2023-01-25
Fix PNI mobile search bar hidden bug

mmmavis:
# Description

PR review will be easier if you check commit by commit.

Major changes are below
- Moved `<div id="pni-mobile-container" ...">` code to its own file `pni_mobile_search.html` and renamed the div to `pni-mobile-search-container`
- Created a wrapper div and wrapped mobile category nav and mobile search bar in it. Conditionally toggle the visibility of category nav / search bar depending on the search mode state. If search mode is on, show search bar only and make it sticky. Otherwise, show sticky category nav.
- Some CSS/Tailwind class adjustments.

Link to sample test page: `/privacynotinclude`
Related PRs/issues: #10051


fessehaye (approved):
Reading the ticket, I guess we are making the search bar sticky? Everything looks good when tested

network-api/networkapi/templates/fragments/buyersguide/primary_nav.html
```diff
@@ -7,25 +7,7 @@
...
-                        type="text"
-                        class="tw-w-full tw-bg-blue-05 tw-outline-none tw-py-4 tw-text-base tw-text-blue-60 placeholder:tw-text-blue-60"
-                        placeholder="{% trans "Search All Products" %}"
-                    >
-                    <label for="pni-mobile-bar" class="clear-icon">&nbsp;</label>
-                </div>
-            </div>
-        </div>
-    </div>
+    {% include 'fragments/buyersguide/pni_mobile_search_container.html' %}
```
tbrlpld: What do you think about calling this `pni_movile_search.html`, so drop the `container` because the fragment contains the container with the rest of the content, the fact that there is a container in it should not matter to the including tempalte.

network-api/networkapi/templates/fragments/buyersguide/pni_mobile_nav.html
```diff
@@ -1,7 +1,7 @@
 {% load bg_nav_tags localization i18n static wagtailroutablepage_tags %}
 {% get_bg_home_page as home_page %}
 
-<div class="d-md-none mt-0 mb-0 tw-bg-white tw-border-b tw-border-blue-10" id="pni-nav-mobile">
+<div class="d-md-none mt-0 mb-0 tw-bg-white" id="pni-nav-mobile">
```
tbrlpld, line 4: If I am following this correctly, this element only contains the category navigation not the very main navigation. We do you think about updating the `id` to reflect that?

source/js/buyers-guide/template-js-handler/mobile-search-bar.js
```diff
@@ -25,8 +25,8 @@ export default () => {
   ) {
     setupSearchBar(
       "#pni-mobile-bar",
-      "#pni-mobile-container .clear-icon",
-      `#pni-mobile-container`
+      "#pni-mobile-search-container .clear-icon",
+      `#pni-mobile-search-container`
```
tbrlpld, line 31: Can you add some context why the we are using the backticks ``` for this string? It does not seem to do any [string templating](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals).

tbrlpld (commented):
Thank @mmmavis This is working nicely now. I like the restructuring too. 

I left some minor questions.

source/js/buyers-guide/template-js-handler/mobile-search-bar.js
mmmavis: Ah i've been sticking with backticks for all use cases. I remember that was how our lining rules set to be too. Do we have a different preference now?

network-api/networkapi/templates/fragments/buyersguide/primary_nav.html
mmmavis: Sure. Will update.

network-api/networkapi/templates/fragments/buyersguide/pni_mobile_nav.html
mmmavis: Agree the id could be improved. I need to check the codebase to see if there will be any collision though. There are quite a bit of relatively similar id names so I want to be extra careful.

mmmavis: @tbrlpld I'm gonna create a new ticket for this. There are too many occurrences of `pni-nav-mobile` in the codebase. (35 results in 9 files). I think this deserve its own ticket.

mmmavis: Filed https://github.com/mozilla/foundation.mozilla.org/issues/10088

tbrlpld: Wow. Thanks for that. I think that is going to be useful.

source/js/buyers-guide/template-js-handler/mobile-search-bar.js
tbrlpld: Intersting. I am ok either way. The linting rule does not seem to be active though, because all the other strings are not using it.

tbrlpld (approved):

source/js/buyers-guide/template-js-handler/mobile-search-bar.js
mmmavis: Just quickly checked the codebase. I see mixed uses of backticks and quotes everywhere. Looks like I remember it wrong. It seems it's always been fine either way.

tbrlpld: No worries. I think I could be a good idea for the future to focus on one style and use that consistent through the code base. Would be good to also make linting check for that.

Issue #9859, opened 2022-12-15
[PNI QA] Donate button is missing from nav on mobile


kristinashu:
The Donate button is missing at narrow browser widths. It should appear in the open hamburger menu.

@beccaklam @nancyt1 could you please add a screen cap of how it should look?

Dev AC
- [ ] As a user, I see the donate button when I open the main menu on my phone
- [ ] As a user, there is enough space between the nav items for my fat fingers (least 1.25rem between the links)

kristinashu:
Thank you Nancy! Yes, I agree the spacing is tight so I'll add that to the ACs.

mmmavis:
@kristinashu is this ready for review on staging https://foundation.mofostaging.net/privacynotincluded/

kristinashu:
Looks good Mavis, thank you!

PR #10061, opened 2023-01-27
Add donate link to PNI mobile nav

mmmavis:
# Description

- added donate link to PNI mobile nav.
- increased mobile primary nav link's y-margin

To test, go to `/privacynotincluded` and verify donate link is showing up on all screen sizes.

Related ticket: #9859


danielfmiranda (commented):
Hi @mmmavis! This is looking good to me so far, one thing I did notice however was that there was a request in the original issue to have the links be spaced out by 1.25rem (20px). 

After taking a look at the inspector, I can see that each nav item does have a margin of 10px top and bottom, but it looks like they are overlapping:

Screenshot 2023-02-01 at 2 21 14 PM


Would it be possible to fix this before merging?

Thanks!

source/sass/components/primary-nav.scss
```diff
@@ -268,7 +268,7 @@
           font-weight: 500;
           transition: transform 0.15s;
           display: inline-block;
-          margin: 10px 0;
+          margin: 1.25rem 0;
```
mmmavis, line 271: I thought about using a Tailwind class for this instead. However, there are multiple places outside of PNI affected by the stylings we define here. If we want to replace this line with Tailwind class, it means we have to update multiple template files to include the Tailwind class and that seems out of the scope for this ticket.

danielfmiranda (approved):
Looks good to me! Thanks @mmmavis 👍 approved

Issue #9790, opened 2022-12-01
MozFest: L10N for Tito widget


marcwalsh:
### Describe the feature request
The Tito widget allows for localisation by adding a parameter in the widget code. Adding this will allow us to have the ticketing process in Spanish for MozFest 2023, where we are looking to fully support Spanish across all platforms.

Currently the widget code looks something like this:
`<tito-button event="Mozilla/mozilla-festival-2023" releases="test-ticket"</tito-button>`

Adding the locales that are supported from Tito like this:
`<tito-button event="Mozilla/mozilla-festival-2023" releases="test-ticket" locale="es"</tito-button>`

### Screenshots / mock-ups
UI localisation already exists for multiple languages, for MozFest 2023 we're supporting `en` and `es` across all platforms. ~If the locale isn't supported it falls back to English.~ (The widget only falls back to English, if the language is support but no translation is provided)

<img width="950" alt="Screenshot 2022-12-01 at 17 50 03" src="https://user-images.githubusercontent.com/30226684/205124612-e0825fb8-06a2-473e-a7bb-d6cc2f6c7acf.png">
<img width="851" alt="Screenshot 2022-12-01 at 17 49 54" src="https://user-images.githubusercontent.com/30226684/205124620-dad4e614-013d-4dbf-83bb-c87af92800c4.png">

### Additional context
I briefly talked to @TheoChevalier about this and how it could be implemented, he mentioned that he would prefer if the local was handled in the code rather than us having this set within the snippets on wagtail. However it could introduce complexities where how we handle language identifiers doesn't match Tito's.

I'm happy with either route (and ideally it would be whichever would be able to be implements quickest) - since the use of ticketing is fairly limited at the moment.

cdanfon:
Hey @mtdenton Tom is happy to pick this ticket

tomusher:
I've added support for this with #9818

cdanfon:
@danielfmiranda @tbrlpld @fessehaye could one of you review this for Tom please?

cdanfon:
Hey @marcwalsh this should be ready to be tested on staging

marcwalsh:
Thanks @cdanfon - this is great - it's working for me on staging. Can we hold off pushing this to production as we're awaiting the human translations - currently it's only auto generated translations being used to test.

cdanfon:
Hey Marc, thanks for confirming. 

[...]

PR #10062, opened 2023-01-27
tito widget now checking if language code is in supported list

danielfmiranda:
# Description

This PR adds a new method to the Tito Widget Block titled `get_widget_language_code`, which checks the whether or not the user's requested language is supported in Tito.js. If so, let the Tito widget use that language. If not, default to EN, as sending an unsupported language will crash the widget.

Link to sample test page:
Related PRs/issues: #9790

### DIFF
--- a/network-api/networkapi/events/tests.py
+++ b/network-api/networkapi/events/tests.py
@@ -3,10 +3,12 @@ from unittest import mock
 
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
+from wagtail.core.models import Page, Site
 
-from .factory import TitoEventFactory
-from .utils import sign_tito_request
-from .views import tito_ticket_completed
+from networkapi.events.factory import TitoEventFactory
+from networkapi.events.utils import sign_tito_request
+from networkapi.events.views import tito_ticket_completed
+from networkapi.mozfest.factory import MozfestHomepageFactory
 
 
 class TitoTicketCompletedTest(TestCase):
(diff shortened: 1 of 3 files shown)


tbrlpld (commented):
Thanks @danielfmiranda I think this is looking good in general. 

Considering how much issues this has caused so far, what do you think about adding some tests for this?

network-api/networkapi/events/tests.py
```diff
@@ -3,10 +3,12 @@
...
 from django.test import RequestFactory, TestCase
 from django.urls import reverse
+from wagtail.core.models import Page, Site
 
-from .factory import TitoEventFactory
-from .utils import sign_tito_request
-from .views import tito_ticket_completed
+from networkapi.events.factory import TitoEventFactory
+from networkapi.events.utils import sign_tito_request
+from networkapi.events.views import tito_ticket_completed
```
danielfmiranda, line 10: Just updated these 3 lines to be the full path instead of the relative path.

network-api/networkapi/wagtailpages/pagemodels/customblocks/tito_widget.py
```diff
@@ -20,3 +23,22 @@ class TitoWidgetBlock(blocks.StructBlock):
...
+        context = super().get_context(request, parent_context=parent_context)
+        request_language_code = get_language_from_request(context["request"])
+        context["tito_widget_lang_code"] = self.get_widget_language_code(request_language_code)
+
+        return context
+
+    # Checking if the user's requested language is currently supported by Tito.
+    # If not, default to English, to prevent the Tito widget from crashing due to an unsupported language.
+    # For more info see: https://github.com/mozilla/foundation.mozilla.org/issues/9790
+    def get_widget_language_code(self, request_language_code):
```
tbrlpld, line 37: What do you think about turning the comment into a docstring for the function?

tbrlpld (approved):
Thank @danielfmiranda this looks great. 

I left one optional question. Merge at your own discretion.

network-api/networkapi/wagtailpages/pagemodels/customblocks/tito_widget.py
danielfmiranda: Sounds good to me!

Issue #10081, opened 2023-01-30
[Research Hub] Library shows protected pages


tbrlpld:
### Describe the bug

When visiting the research hub library page (https://foundation.mofostaging.net/en/research/library/) and there a non-public / protected page exists (https://foundation.mofostaging.net/en/research/library/alternative-data-governance-playbook/) then that page is listed. 

### To Reproduce

1. Go to the research hub library page: https://foundation.mofostaging.net/en/research/library/
2. In the admin, make one child page of the library page protected (e.g. password protected)
3. Refresh the library page
4. See see that the protected page is listed.

### Expected behavior

Needs discussion: *We should only show protected pages to logged-in users.*

kevinhowbrook:
> We should only show protected pages to logged-in users.

I've dealt with something similar recently, and really we deemed that 'teasers' were ok to be 'non-protected' as they advertise the content and the user is prompted to login.

A fix for this won't be as simple as modifying a queryset to add `.public()`, if we are definitely sure of the expected behaviour then we will need to offer different querysets depending on the `reuest.user` (one filtered by public() and the other not)

tbrlpld:
Good points @kevinhowbrook I agree, we need to define what the desired behavior should be.

I wonder how much we want to show protected pages in public. Not sure who to ask, but I would be confused as a user visiting the research hub and then find a page I can not access. Most users/visitors won't be editors as well. 

Outside of the check for the request user, what other considerations do you have in mind?

kevinhowbrook:
Other considerations, only that any public API was being filtered too... e.g search results. The two querysets would be fine, just raising it so we don't create further edge cases or create unreachable content

tbrlpld:
Staff are reporting this on production: https://mozilla.slack.com/archives/C98GW3P7X/p1675101252839289

tbrlpld:
@kristinashu I agree, I think the most expected would be to hide protected pages in the public library. 

If we later discover we want to show it to logged in users, we can to that then. To keep it simple, we could hide all non-public pages from the library for now. 

Can you explain what "in the thumbnails" means?

kristinashu:
Oh sorry, I mean the thumbnails on the library page. 

Yes, hiding all non-public pages from the library and search results sounds good like the right scope for this ticket.

tbrlpld:
@kristinashu Would you mind testing this on staging? https://foundation.mofostaging.net/en/research/library/

kristinashu:
[...]

PR #10087, opened 2023-02-01
#10081 Don't show protected pages in research hub library

tbrlpld:
# Description

Previously, when adding a research detail page which is set to "private" to the research library it would be listed in the library. 

This is not desirable because it leaks private information in the listing (e.g. page title and intro text). 
Also, it is a confusing user experience to see a page listed that might have interesting information and then be blocked from accessing this information. 

Staff are using the privacy setting for pages that are not ready for public consumption but are worked on internally. 
We should therefore avoid showing the pages publically if they are marked as private.

To test this, you should visit the library page and set any of it's child pages (research detail page) to private (e.g. via password protection). 

Link to sample test page: http://localhost:8000/en/research/library/
Related PRs/issues: #10081

# Checklist

**Tests**

**Changes in Models:**
- ~~[ ] Did I squash my migration?~~ No migrations added.

**Documentation:**


kevinhowbrook (approved):
LGTM, thanks @tbrlpld - from reading the issue it looks like we are ok with never having private pages listed so all good :)

danielfmiranda (approved):
Hi @tbrlpld just tested this out on my local machine and can confirm that its working for me. Great job! Approved 👍

tbrlpld:
Thanks for reviewing @danielfmiranda @kevinhowbrook 

> LGTM, thanks @tbrlpld - from reading the issue it looks like we are ok with never having private pages listed so all good :)


Yea. If this pops us as a requirement in the future, we can deal with it then I think.

Issue #10056, opened 2023-01-26
[PNI] Consider hiding speech bubble and creep-o-meter if products are sorted alphabetically


mmmavis:
Speech bubble and creep-o-meter change their content based on the creepiness score of the products visible on screen. This works great when products are sorted by creepiness. However, when products are sorted in alphabetically order, creepiness score can jump up and down product by product. This can make speech bubble and creep-o-meter change quite abruptly as users scroll. They don't always show the correct indicator either. (Think about a scenario where one super creepy product is on the same row of not creepy products. What should the speech bubble say?) We should consider hiding them if products are sorted alphabetically so they don't look wonky.

### Production/Desktop Screen Recording
https://user-images.githubusercontent.com/2896608/214915588-94a42f7d-47bb-4417-ae46-5bf8f8557b45.mov

### Mobile Screen Recording
**Note: Speech bubble and creep-o-meter currently don't update on smaller screens (bug #9942). This recording is taken locally from my machine with a potential fix to that.**

https://user-images.githubusercontent.com/2896608/214917727-cc2911ea-6791-4927-a859-d2423d498ee6.mov

simont-cr:
Closed per 2024 ticket purge.

Issue #9942, opened 2023-01-12
PNI - Speech bubble and creep-o-meter emoji don't update on small screens


mmmavis:
### Describe the bug
PNI homepage - Speech bubble and creep-o-meter emoji don't update on small screens (width < 768px)

### To Reproduce
1. Go to https://foundation.mozilla.org/privacynotincluded/
2. See if speech bubble and emoji are updating as you scroll up and down the page
3. Repeat Step 1-2 on a narrower screen

### Expected behavior
Speech bubble and emoji should update as you scroll.

### Screenshots (taken with window width 767px)

https://user-images.githubusercontent.com/2896608/212439612-4b1e890d-3841-4096-922b-1fb9c15f3b1d.mov

fessehaye:
@beccaklam can you check this on staging

beccaklam:
Yes this appears to be resolved!

PR #10089, opened 2023-02-01
Fixed PNI mobile speech bubble bug by removing old SCSS

mmmavis:
# Description

BUG: PNI homepage - Speech bubble and creep-o-meter emoji don't update on small screens (width < 768px)

The root cause of this bug is old SCSS adding "negative" 1px left/right margins to `product-box-list` which makes `product-box-list` wider than mobile viewport by 1px + 1px = 2px. This means `product-box`(children) in `product-box-list`(parent) can grow outisde of the viewport. We use [`isElementInViewport` function](https://github.com/mozilla/foundation.mozilla.org/blob/main/source/js/buyers-guide/homepage-c-slider.js#L30-L46) to check if an element is fully visibly in viewport so the speech bubble knows whether or not to update its content. An element that's bigger than viewport will always fail the check and that's the case for `product-box`. By removing old SCSS code, we make sure the parent container `product-box-list` is contained within the viewport size.

I also removed a bunch of SCSS code that isn't in use anymore. 

(For PR reviewer: It's easier to review code changes in this PR without whitespace differences.)

Link to sample test page: `/privacynotincluded` 
Related PRs/issues: #9942

### QA Steps
  - view PNI on mobile screen size
  - scroll up and down and verify speech bubble face/content get updated
  - note that speech bubble can act a little glitchy if products are sorted alphabetically instead of by creepiness. This is not a bug but more of a design issue. I filed ticket #10056 to address that.

### DIFF
--- a/source/sass/buyers-guide/views/catalog.scss
+++ b/source/sass/buyers-guide/views/catalog.scss
@@ -66,115 +66,93 @@ body {
   }
 
   .product-box-list-wrapper {
-    // Mixin for setting product box styling in product box list
-    @mixin product-box-size($box-margin: 1px, $items-per-row: 2) {
-      $num-of-gutter: $items-per-row;
-      $gutter-width: $box-margin * 2;
-      $row-total-gutter-width: $gutter-width * $num-of-gutter;
+    figure.product-box {
+      $padding-x: 12px;
+      $padding-y: 12px;
+      $privacy-ding-width: 30px;
+      $privacy-ding-height: 25px;
+      $recommendation-width: 17px;
+      $recommendation-height: 15.8px;
 
-      .product-box-list {
-        margin-right: -$box-margin;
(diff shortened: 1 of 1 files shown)


source/sass/buyers-guide/views/catalog.scss
```diff
@@ -66,115 +66,93 @@ body {
   }
 
   .product-box-list-wrapper {
-    // Mixin for setting product box styling in product box list
```
mmmavis, line 69: Since we don't want line 75-78 (negative margin) anymore and none of the variables `$num-of-gutter`, `$gutter-width`, `$row-total-gutter-width` get used anywhere, we don't need to have a mixin.

fessehaye (approved):
Thanks for removing the mixins, it will make future refactors to the scss way easier. Approved!

Issue #9881, opened 2022-12-21
PNI - homepage speech bubble content should be localized


mmmavis:
### Describe the bug
On PNI homepage, speech bubble content should be localized

### To Reproduce
1. Go to PNI homepage foundation.mozilla.org/privacynotincluded/
2. Scroll down to page bottom to choose a non-English language
3. Wait for page to reload
4. Scroll down and pay attention to speed bubble content (`creepy`, `not creepy`, etc)

### Expected behavior
Speech bubble content should be localized

### Screenshots
https://user-images.githubusercontent.com/2896608/208976643-8ba0bbce-5781-4f8f-9443-857d2debf06b.mov

tbrlpld:
This needs to wait as it is likely to clash with #10059

tbrlpld:
I assumed when estimating on this that this would be as simple as adding a `trans` block somewhere in the template. 

Turns out, these strings are defined in JS. 
`source/js/buyers-guide/homepage-c-slider.js` get strings from `source/js/buyers-guide/components/creepiness-labels.js`

I don't know how we deal with string translation in JS. When searching for `locale` in the JS files I am finding hardcoded strings and custom switcher logic.
 E.g.: 

- `source/js/components/petition/locales.js`  switches `source/js/components/petition/locale-strings.jsx`
- `source/js/components/tabbed-profile-directory/locales.js` switches `source/js/components/tabbed-profile-directory/locale-strings.jsx`

This seems to be a third way to do translations in the code base. Is this something we should be doing? 

@TheoChevalier is this the desired way to do this? 

How ever we are going to solve this, I have basically used the planned time for researching this with no result. I guess this needs more discussion.

tbrlpld:
cc @mtdenton @cdanfon for visibility

tbrlpld:
Also, not sure how we would be able to QA this because the strings are not in the translation database / dataset yet.

tbrlpld:
@fessehaye Mentioned that some locations use data attributes for this

TheoChevalier:
Hey there,

So yes, there are two different cases here: strings in simple .js files, and strings in React components.

The `locale-strings.jsx` file is a very hacky way of localizing strings in React components, we need to implement a proper i18n solution for this.

I’ve done some exploratory work a while back (and ended up spending way too many hours on this):
[...]

PR #10115, opened 2023-02-06
Localize creepiness labels in context

TheoChevalier:
Follow-up for #9881 by localizing the exclamation mark concatenated in the slider.

Interpolating the string with a placeable here allows to fully translate the string. This may seem over-engineered, but for example, Spanish will want to add `¡` before the label, French will want to add a non-breaking space before the exclamation mark.


mmmavis (approved):
Thanks for sharing your expertise and submitting this PR @TheoChevalier . Code looks good to me. I just have one question regarding if we should include a hardcoded punctuation `!` on line 123. Would love to hear what you think.

I'm going to merge this PR as is now because we are entering code freeze tmr for Wagtail version upgrade to take place. If you think this PR needs any further polish, please submit a PR and assign me as reviewer. Thank you!

TheoChevalier:
Thanks for the review, @mmmavis! We do have a hardcoded `!` in `"%(label)s!"`, the gettext variable in just %(label)s. This is also because both are so close and it’s not easy to tell whether the ! is part of the variable or not that I’ve mentioned it in the comment for translators ;)

mmmavis:
@TheoChevalier ah sorry I formed my sentence a bit weirdly.

>  I just have one question regarding if we should include a hardcoded punctuation ! on line 123. Would love to hear what you think.

I meant I was curious if including a hardcoded `!` was an ideal approach. I don't know much about localization and am curious if all languages are using `!` as the exclamation mark like what we use it for in English.

TheoChevalier:
Ah, sorry! I get it now, let me start again.

Exclamation marks should always be part of a string (so, translatable) because there are indeed changes that will happen in some locales. Spanish for instance will want to add `¡` at the beginning of the sentence (e.g `¡Una frase!`), French will want to add a non-breaking space before the exclamation mark (e.g `Une phrase !`). These are the most obvious changes I’m aware, but translators could also decide to replace the exclamation mark with a period if they need to “tune down” the tone a bit so it feels more appropriate in their language, depending on the context.

mmmavis:
@TheoChevalier thanks for the detailed explanation! 

> translators could also decide to replace the exclamation mark with a period if they need to “tune down” the tone a bit so it feels more appropriate in their language, depending on the context.

With `"%(label)s!"`, I'm curious how translators will be able to replace the hardcoded `!` with other punctuation marks. I assume they can only decide what string they want to show for the named interpolation "label"? I'm trying to wrap my head around our localization setup with Django. Thanks for having the patience to walk me through this.

tbrlpld:
Is there an argument against moving the `!` into the strings? It seems like that would be more flexible and easier to reason about. 
 https://github.com/mozilla/foundation.mozilla.org/blob/1b0298d1a23c78f9696fc5e315627cb90f041c35/source/js/buyers-guide/components/creepiness-labels.js#L6-L12

TheoChevalier:
> @TheoChevalier thanks for the detailed explanation!
> 
> > translators could also decide to replace the exclamation mark with a period if they need to “tune down” the tone a bit so it feels more appropriate in their language, depending on the context.
> 
> With `"%(label)s!"`, I'm curious how translators will be able to replace the hardcoded `!` with other punctuation marks. I assume they can only decide what string they want to show for the named interpolation "label"? I'm trying to wrap my head around our localization setup with Django. Thanks for having the patience to walk me through this.

They have access to everything between the quotes, here’s how it looks on the translation workbench where I was able to add a non-breaking space in the French translation:
Capture d’écran 2023-02-16 à 17 39 41

> Is there an argument against moving the ! into the strings? It seems like that would be more flexible and easier to reason about.

If if was the only context using these strings, no (it would be even preferable) but the strings are used in other places: https://github.com/mozilla/foundation.mozilla.org/search?q=CREEPINESS_LABELS

Issue #10102, opened 2023-02-02
Make Accordion block on Article pages the same width as the Content block


kristinashu:
We have a request from Eeva to make Accordions narrower on Article pages so they are the same width as the Content block so that she can use headings above the accordions without it looking broken. 

Screen cap showing the issue on Article and Publication pages:

<img width="1066" alt="image" src="https://user-images.githubusercontent.com/1068783/216437744-6678414c-0051-49a1-a146-9a0531f60cb4.png">

Screen cap showing how Accordions are the same width as Paragraph block on other page types:
<img width="922" alt="image" src="https://user-images.githubusercontent.com/1068783/216437188-6f30d040-95ed-4965-8e51-69ffe55f1442.png">

## AC
- [x] As a visitor, the Accordion block is the same narrower width as the Content block on Article pages

@mtdenton @cdanfon does this need further grooming?

cdanfon:
Hey @kristinashu are we meant to decrease the with of the _Accordion_ or increase the width of the _Content block_?

kristinashu:
Ah good point Claudia! I've update the copy to make it clear that the Accordion width should be decreased the match the Content block. Ty!

mmmavis:
Hi @kristinashu are we making this change for Article Pages only? It looks like accordion block is also available on other page types like PNI article page, blog page, campaign page, primary page etc)

kristinashu:
I believe it is only necessary on the Articles pages because the other Accordion block instances are all already narrow (same width as the paragraph text). But please let me know if you notice any that are wider.

mmmavis:
Hi @kristinashu , this is ready for QA on staging. I've included some links for testing.

### Article Page (accordion should be narrower on staging)

- PR: https://foundation.mofostaging.net/en/research/library/is-that-even-legal/builders-guide/
- Production: [https://foundation.mozilla.org/en/research/library/is-that-even-legal/builders-guide/](https://foundation.mozilla.org/en/research/library/is-that-even-legal/builders-guide/)

### Default Page (both should look the same)

- PR: https://foundation.mofostaging.net/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion
- Production: [https://foundation.mozilla.org/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion](https://foundation.mozilla.org/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion)

### Banner Page (both should look the same)

[...]

PR #10153, opened 2023-02-14
10102 - change width of Accordion block on Article pages

fessehaye:
# Description

The request was to change the width of accordions on article pages. Since article pages have a few noticeable adjustments, I made a tailwind variant to address future article page work. Also moved custom variants to its own file but if we do not add many in the future, it might be better to move it back to the config.

Link to sample test page:
Related PRs/issues: #10102

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?


tailwind.config.js
```diff
@@ -48,14 +49,13 @@ module.exports = {
           verticalAlign: "middle",
         },
       };
-      addVariant("summary-open", ["details[open] > summary > &"]);
-      addVariant("details-open", ["details[open] > &"]);
```
tbrlpld, line 52: Was this dropped on purpose?

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/accordion_block.html
```diff
@@ -5,7 +5,7 @@
 
     <div class="tw-row">
         {% for block in self.accordion_items %}
-            <details class="tw-w-full tw-font-sans tw-font-light tw-mx-8">
+            <details class="article-content:tw-w-4/5 article-content:tw-mx-auto tw-w-full tw-font-sans tw-font-light tw-mx-8">
```
tbrlpld, line 8: What do you think about moving these things out into the article page? I think it usually makes it harder to understand and maintain components if they need to know where they are being used. Rather, components should be like pure functions that don't have any side effects or know anything about their surrounding. 

In this case it seems like on article page you want to limit the block to a given width and center it. What do you thin about doing this via a wrapper on the article page? We would not need any of the Tailwind changes.

tbrlpld (changes requested):

tailwind.config.js
fessehaye: It was not being used so, it is being removed

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/accordion_block.html
fessehaye: But the accordions are streamfields. If the accordions were manually added to article pages in a predictable spot on the article pages, I would agree about adding a wrapper, but they will be mixed up with other CMS content. Still, I would be adding to either `network-api/networkapi/templates/pages/buyersguide/article_page.html#L49` and adding an if statement about accordions on there but you need to apply the width changes after the streamfield-content class since that class has different flex/width.  

Or does that mean we are modifying `network-api\networkapi\wagtailpages\templates\wagtailpages\blocks\base_streamfield_block.html`  and adding checks to accordions there? Should I just make the CSS changes on `source/sass/wagtail/blocks/article-blocks.scss` ? The whole purpose of adding these variants is to make stylings from different parent element changes easier to follow.  I definitely need more specific instructions moving forward with this idea.

tbrlpld: The accordions should be as wide as the rest of the blocks on the article page. Why are they wider now? Why are other blocks not wider? The width limitation should come from the article page and not from the blocks to make the blocks reusable. So yea, the change should probably be in the article page template. 

Components in general should not really know about their surrounding. Rather everything that is variable about a component should be passed in directly as flags. It makes components otherwise really hard to maintain because they include all these checks regarding where they are used. It's better to make these flags/checks semantic about the effect (in this case it should express something about the width) instead of the environment in which the flag should be set. A good analogy for components are pure functions, same input should lead to same output. Also, there should be no side effects should happen and they should not "pull" data, all data should be passed in as arguments. 

Also, I think that "variants" is the bit misleading place for this one off use case. Variants are widely available and should have something to do with styling. "Article page" is not really a concept in CSS. If we wanted to do this width adjustment in the component, which I still think is not the right way to do it, there should just be a flag that is passed into the component template which then applies that class or not. The flag should also not be `article_page=True/False` but rather `width='4_5'` or something of that nature. Or you even allow passing of extra classes to the component. 

If you need to handle a particluar block in a streamfield differently based on it's type, you can check the blocks type with `block.block_type`. https://docs.wagtail.org/en/stable/topics/streamfield.html#template-rendering

tbrlpld (dismissed):
See my comment: https://github.com/mozilla/foundation.mozilla.org/pull/10153#discussion_r1112087460

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/accordion_block.html
fessehaye: Adding a wrapper on the article page does not work because `base_streamfield_block.html` has container classes so the parent having 4/5 width would not achieve the same effect.

redid the solution by applying a width='4_5' value.

mmmavis:
@tbrlpld I will be wrapping up this PR. Will add you as reviewer again once this is ready for review!

mmmavis:
Hi @tbrlpld , this PR is ready for review!

I'm including some links for testing purposes. Once you are on the page, you can quickly jump to accordion block by entering this line on dev tool console: `document.querySelector(".accordion-block").scrollIntoView();`

### Article Page (accordion should be narrower on PR)

- PR: [http://localhost:8000/en/research/library/is-that-even-legal/builders-guide/](http://localhost:8000/en/research/library/is-that-even-legal/builders-guide/)
- Production: [https://foundation.mozilla.org/en/research/library/is-that-even-legal/builders-guide/](https://foundation.mozilla.org/en/research/library/is-that-even-legal/builders-guide/)

### Default Page (both should look the same)

- PR: [http://localhost:8000/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion](http://localhost:8000/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion)
- Production: [https://foundation.mozilla.org/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion](https://foundation.mozilla.org/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion)

### Banner Page (both should look the same)

- PR: [http://localhost:8000/en/youtube/regretsreporter-dataset/](http://localhost:8000/en/youtube/regretsreporter-dataset/) (Password: find on [edit page](http://test.example.com:8000/cms/pages/21685/edit))
- Production: [https://foundation.mozilla.org/en/youtube/regretsreporter-dataset/](https://foundation.mozilla.org/en/youtube/regretsreporter-dataset/) (same password from above)

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/accordion_block.html
```diff
@@ -2,19 +2,16 @@
 {% load wagtailcore_tags wagtailimages_tags static %}
 
 {% block block_content %}
-
```
mmmavis, line 5: A lot of these are whitespace changes.

https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10153/files?diff=unified&w=1

mmmavis:
PR description updated!

tbrlpld (approved):
Thank @mmmavis 👍 Clean and easy change. 


I left one optional question.

network-api/networkapi/templates/pages/article_page.html
```diff
@@ -54,7 +54,11 @@
 
     <div class="cms article-page-content article-blocks">
         {% for block in page.body %}
-            {% include_block block with parent_page=page page_type="article" %}
+            {% if block.block_type == "accordion" %}
+                {% include_block block with parent_page=page page_type="article" width="4/5" %}
```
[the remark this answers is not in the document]
mmmavis: Thanks for pointing this out! I did a find-all search in the html files in the codebase. Looks like `article_page.html` isn't the only file that passes `parent_page` and `page_type` to blocks. I think I'm gonna leave the PR as is for now for consistency purpose. Will file a ticket to do a proper audit.

image


image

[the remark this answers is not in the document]
tbrlpld: That's fine. 

However, I would like to point out that this exact pattern is somewhat of an anti-pattern to me, because it leads conditionals inside the fragments that check for the environment. These kinds of conditionals make reuse of fragments pretty hard, because you might want to use the same fragment in a different context in which that check makes no sense. It would be better to control the fragment through arguments that say something about their effect.

Issue #10155, opened 2023-02-15
Accordion shows detail triangles on Safari


tbrlpld:
### Description

The default toggle indicator triangles of the `details` element are showing on Safari. This is redundant with the +/- icon on the right and does not happen in other browsers.

On iOS the triangles are even smushed. 

### To Reproduce

1. Go to https://foundation.mozilla.org/en/research/library/is-that-even-legal/builders-guide/
2. Scroll down to the accordion section

### Expected behavior

The triangles are not shown. 

### Screenshots

<img width="300" alt="Screenshot 2023-02-15 at 07 49 37" src="https://user-images.githubusercontent.com/24797493/219080930-b0b63fae-2edc-4d7b-93fd-8a0832b81601.png">

<img width="300" src="https://user-images.githubusercontent.com/24797493/219080948-3f090a43-bddf-4756-bcf8-c5c85ec87dc0.jpeg">

### Desktop

 - OS: macOS
 - Browser: Safari
 - Version: 16.3

### Smartphone

 - Device: iPhone 14 Pro
 - OS: iOS 16.3.1
 - Browser: Safari

danielfmiranda:
Hi @tbrlpld this is now on staging and ready to be QA'd whenever you get a free moment. Thanks!

danielfmiranda:
Hi @mmmavis! 

Thanks for volunteering to do QA for this ticket. 


Here are the steps to test: 

- Using Safari, Go to https://foundation.mofostaging.net/en/research/library/is-that-even-legal/builders-guide/
- Scroll down to the accordion section
- Check if triangles are still showing on the left side of items, shown in the screenshot above.
- If the triangles are no longer there, testing is complete!

tbrlpld:
Thanks @danielfmiranda I can confirm this is fixed now 👍

mmmavis:
Oops, sorry just saw this. This is working great on prod as well. Closing!

PR #10179, opened 2023-02-17
added vendor prefix to accordion summary elements

danielfmiranda:
# Description
Link to sample test page: http://localhost:8000/en/docs/how-do-i-wagtail/streamfield-blocks/#accordion
Related PRs/issues: #10155 

Even though we are removing the triangle image from accordion elements using the class of `tw-list-none` which generates the following CSS:
```css
.list-none {
     list-style-type: none;
}
```

it appears that Safari is still requiring the use of a vendor prefix `::-webkit-details-marker` to remove it like so:

```css
summary::-webkit-details-marker {
  display: none;
}
```

This PR adds a new arbitrary tailwind class `[&::-webkit-details-marker]:tw-hidden]` in order to generate the required css above and remove the element on safari mobile and web browsers as expected

# Screenshots

Before:
image

After:

Screenshot 2023-02-17 at 3 31 21 PM


fessehaye (approved):
Part of me thinks we should update tw-list-none to include this but since this unique problem with accordions+ safari. I think this fix works!

Issue #10194, opened 2023-02-21
Dropdown button styles broken in Wagtail Dashboard


mtdenton:
### Describe the bug
A button has some bad styles in the Wagtail dashboard after the Wagtail 3.0 Update (affecting all browsers):

![bustedbutton](https://user-images.githubusercontent.com/41487159/220427002-42f8e6b4-4803-40f1-8a3f-249456ca3f93.JPG)

### To Reproduce
Go to https://foundation.mozilla.org/cms/snippets/wagtailpages/update/ as an example link. The button in the top right has the broken styles

mmmavis:
This bug comes from Wagtail again. I will add styling overrides on our end to fix this.

mmmavis:
@mtdenton this is ready for QA on staging: https://foundation.mofostaging.net/cms/snippets/wagtailpages/update/

mtdenton:
@mmmavis Good to go!

mmmavis:
Done!

Issue #10192, opened 2023-02-21
Wagtail CMS Dashboard icon is oversized


mtdenton:
### Describe the bug
The "right arrow" icon being used in the menus on the Wagtail dashboard seems to have inadvertently changed.

I'm seeing several tailwind classes added to this, so maybe that's the reason?

![menutailwind](https://user-images.githubusercontent.com/41487159/220406367-671521d0-7317-48da-8c84-5d0e3f7d080f.JPG)

### To Reproduce
1. In the CMS dashboard, go to Pages > Homepage > Blog, the arrow is seen at the bottom of the screen

### Expected behavior
The icon should be reduced in size so that the link to go to the blog view (https://foundation.mozilla.org/cms/pages/789/) isn't clipping out of the bottom of the page.

mmmavis:
@cdanfon do you mind sharing a screenshot of another Wagtail 3.0 site? I'm wondering how the arrow is supposed to look like.

mmmavis:
@cdanfon do you mind sharing a screenshot of another Wagtail 3.0 site? I'm wondering how the arrow is supposed to look like.

mmmavis:
I set up a fresh Wagtail site locally that runs on Wagtail v3.0 and confirmed that the issue came from Wagtail.

Should I use my best judgement to come up with CSS override to fix this bug?

mtdenton:
Ha, interesting. @cdanfon @tbrlpld I doubt it's a ticket-worthy bug for y'all but FYI.

+1 on the best judgement route @mmmavis . Thanks for looking into this!

tbrlpld:
If this is fixed in Wagtail 4 and this is not preventing editors from any important work, I would just wait. You can always opt to navigate the page tree with the list view (main page content) instead of the sidebar.

mmmavis:
I've filed a bug ticket on Wagtail's repo as I was able to reproduce this from Wagtail v3 to v4.2.
I will have a PR on our foundation site repo to address this issue.

mmmavis:
@mtdenton this is ready for QA on staging!

mtdenton:
@mmmavis looks good from here!

mmmavis:
Done!

PR #10198, opened 2023-02-22
Fix styling issues from Wagtail 3.0

mmmavis:
This PR addresses the following issues:

### #10192 oversized arrow icon on page explorer
- Using Page Explore to navigate to "Blog" or any directory that has more than 200 pages.
- Make sure arrow icon next to "See all" is sized properly.

### #10194 text in locale dropdown is not visible
- Go to http://test.example.com:8000/cms/snippets/wagtailpages/update/
- Verify text of the locale dropdown is visible.


network-api/networkapi/wagtailcustomization/static/wagtailadmin/css/custom-fix.css
```diff
@@ -0,0 +1,42 @@
...
+*/
+
+body#wagtail
+  main#main
+  form
+  ul
+  li
+  .field
+  .c-dropdown.t-inverted.c-dropdown--large
+  a {
```
mmmavis, line 27: I'm using very specific selector so there's less chance of these styling rules affecting elements that we are unaware of.

network-api/networkapi/wagtailcustomization/static/wagtailadmin/css/custom-fix.css
```diff
@@ -0,0 +1,42 @@
...
+body#wagtail
+  .sidebar-panel
+  .c-page-explorer__see-more
+  svg.icon.icon-arrow-right {
+  width: 20px;
+  height: 20px;
+  position: relative;
+  top: 5px;
+}
+
```
danielfmiranda, line 14: Hi @mmmavis!

This is looking great to me so far, after testing, I can confirm that your code changes fix both the issues mentioned in the ticket.

One thing that I did note and wanted to bring up, was that on screen sizes larger than M, it appears that the "see more" box is taller than it needs to be. According to the inspector, it is set to 50px in height (I am assuming to make room for the larger than usual SVG): 


Screenshot 2023-02-24 at 11 46 47 AM


I was wondering what you thought about adding something like this in order to reduce the unnecessary spacing around the text?:

```suggestion

body#wagtail .sidebar-panel .c-page-explorer__see-more {
  height: unset;
}

```

 Here is what it looks like with the suggested CSS above:


Screenshot 2023-02-24 at 11 43 16 AM


I have also tested it on smaller screensizes and can confirm that everything is looking as expected. 

Thanks!

mmmavis:
@danielfmiranda thank you for your awesome suggestion! I've updated the PR with a minor adjustment - I wrapped the styling rules in media query to make sure we are overriding the exact styling rule from Wagtail.

danielfmiranda (approved):
Thanks for the update Mavis! Approved 👍

Issue #10162, opened 2023-02-16
[PNI Refactor] Create a function for updating active state of a subcategory link


mmmavis:
Search the codebase for `a.subcategories`. Notice that there are a few implementations dealing with updating active state of the subcategory links.

e.g., 

```js
if (document.querySelector(`a.subcategories.active`)) {
      document
        .querySelector(`a.subcategories.active`)
        .classList.add(...defaultClasses);
      document
        .querySelector(`a.subcategories.active`)
        .classList.remove(...activeClasses);
    }
```

```js
document
      .querySelector(`a.subcategories[data-name="${categoryName}"]`)
      .classList.add(...activeClasses);

    document
      .querySelector(`a.subcategories[data-name="${categoryName}"]`)
      .classList.remove(...defaultClasses);
```
... etc

Let's extract the code and create a new function!

mmmavis:
Verified on staging!

mmmavis:
Done!

Issue #10161, opened 2023-02-16
[PNI Refactor] Create a function for scrolling subcategory link into view


mmmavis:
As of now, there are two copies of the same code in `source/js/buyers-guide/search/history.js`. Let's extract the code and turn it into a function.

I can't tell why this code is needed but since we are focusing just on improving code readability for now, let's revisit whether or not to keep this code at a later stage.

```js
document
      .querySelector(`a.subcategories[data-name="${history.state?.category}"]`)
      .scrollIntoView({
        behavior: "smooth",
        block: "nearest",
        inline: "start",
      });
```

mmmavis:
Done!

PR #10207, opened 2023-02-24
PNI code clean up

mmmavis:
For this PR I'm just focusing on extracting dupe of code into functions. I add function to `utils.js` only if that function is being called by different files. We'll figure out how to better group functions into modules at a later stage.

### Related PRs/issues:
- #10161 (see 1st commit in this PR)
- #10162 (see 2nd commit in this PR)


fessehaye (approved):
This looks good Mavis, thanks for refactoring this!

Issue #10193, opened 2023-02-21
Double scroll bar on Wagtail Dashboard


mtdenton:
### Describe the bug
On Firefox (issue doesn't seem to affect other browsers), a double scroll bar is showing up on certain dashboard pages. This is also causing an infinite scrolling issue on the page.

![doublescroll](https://user-images.githubusercontent.com/41487159/220408055-9ee5026b-21ef-4af4-af92-0879a28fff32.JPG)

### To Reproduce
1. In Firefox, go to https://foundation.mozilla.org/cms/pages/47/edit/. 

### Expected behavior
Scrolling behavior should be consistent across browsers, Chrome seems to have the correct behavior.

mmmavis:
I tested on Chrome/Firefox/Safari on Mac. I confirmed that the double scrollbar and infinite scroll issues are only on Firefox.

mmmavis:
I tested on Chrome/Firefox/Safari on Mac. I confirmed that the double scrollbar and infinite scroll issues are only on Firefox.

mmmavis:
@mtdenton do you mind QAing the fix on staging?

PR #10235, opened 2023-02-28
10193 - fix double scroll bar and infinite scrolling issue on FF on Wagtail 3.0

mmmavis:
# Description

Bug only appears on Firefox. There are some complex CSS rules going on in Wagtail's source code. Looks like they are to make both the sticky top bar and the toggleable sticky right sidebar to work. I didn't continue diving into their code as it was taking too much effort to reverse engineer everything.

The root cause seems to be the sidebar (toggleable by clicking on the "i" icon at top right) is set to be 100vh which is taller than the space that can be available for it without it overflowing. (See screenshot below).

A huge bottom margin on `<main>` seems to also contribute to the double scroll bar issue, so I reset it to `0` in this PR. 

I did a quick check on Chrome/Firefox/Safari and did not find any side effects with my fix. Hopefully I didn't miss anything.

---

Link to sample test page: `/cms` and find a page to edit
Related PRs/issues: #10193

image


danielfmiranda (approved):
Looks good to me after testing, thanks Mavis! 🙌

Issue #10060, opened 2023-01-26
Redirect deprecated category links into topic links


mtdenton:
We'd like to automatically send users who enter through these old `/category/` links to their `/topic/` equivalent. We also need to maintain l10n functionality for routing users to the correct language.

For instance:
`https://foundation.mozilla.org/blog/category/moz-news-beat/`

should route to

`https://foundation.mozilla.org/en/blog/topic/moz-news-beat/`

if the user should go to an `/en/` view

## Dev notes

@tbrlpld identified that the work would need to happen in these places:
https://github.com/mozilla/foundation.mozilla.org/blob/e03f22a51ae3bdc8faa75c7c9140030d25f429cf/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L91
https://github.com/mozilla/foundation.mozilla.org/blob/e03f22a51ae3bdc8faa75c7c9140030d25f429cf/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L348-L361

## Acceptance Criteria

- [x] As a site visitor, when I open a URL to filter for blog "categories"  I should be redirected to the equivalent topic filter URL:
  - From: `https://foundation.mozilla.org/en/blog/category/moz-news-beat/`   
  - To `https://foundation.mozilla.org/en/blog/topic/moz-news-beat/`.
- [x] As a site visitor, when I open a blog category URL without a language code in it, I am redirected to the blog topic URL on the default locale: 
  - From `https://foundation.mozilla.org/blog/category/moz-news-beat/` 
  - To `https://foundation.mozilla.org/en/blog/topic/moz-news-beat/`
- [x] As a site visitor, when I have visited the foundation site and changed my locale to something other than the default (e.g. `de`), redirects from URLs without language code will redirect to URLs for my previously selected locale. 
  - From `https://foundation.mozilla.org/blog/category/moz-news-beat/` 
  - To `https://foundation.mozilla.org/de/blog/topic/moz-news-beat/`
- [x] If the origin URL contains a language code, the redirect target should keep the same locale as the origin. 
- [x] This should be a permanent redirect.

tbrlpld:
@mtdenton Was it intentional that the origin URL does not contain the language code? 

`https://foundation.mozilla.org/blog/category/moz-news-beat/`
instead of 
`https://foundation.mozilla.org/en/blog/category/moz-news-beat/`

cdanfon:
[...]

PR #10237, opened 2023-03-01
added redirect from /category/ to /topic/

danielfmiranda:
# Description

Link to sample test page:
Related PRs/issues: #10060

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

### DIFF
--- a/network-api/networkapi/redirects.py
+++ b/network-api/networkapi/redirects.py
@@ -11,6 +11,12 @@ def foundation_redirects():
             r"^opportunity/(?P<subpath>.*)",
             lambda req, subpath: localized_redirect(req, subpath, "initiatives"),
         ),
+        # redirect /blog/category path to /blog/topic
+        # see https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10060 for context
+        re_path(
+            r"^blog/category/(?P<subpath>.*)",
+            lambda req, subpath: localized_redirect(req, subpath, "blog/topic", permanent=True),
+        ),
         # redirect /about Wagtail pages to /who-we-are
         re_path(
             r"^about/(?P<subpath>.*)",
--- a/network-api/networkapi/wagtailpages/views.py
+++ b/network-api/networkapi/wagtailpages/views.py
@@ -26,7 +26,7 @@ def custom404_view(request, exception):
         return HttpResponseNotFound(html.content)
 
 
-def localized_redirect(request, subpath, destination_path):
+def localized_redirect(request, subpath, destination_path, permanent=False):
     lang = request.LANGUAGE_CODE
     translation.activate(lang)
     request.session[translation.LANGUAGE_SESSION_KEY] = lang
@@ -35,4 +35,4 @@ def localized_redirect(request, subpath, destination_path):
     if request.META["QUERY_STRING"]:
         query_string = f'?{request.META["QUERY_STRING"]}'
 
-    return redirect(f"/{request.LANGUAGE_CODE}/{destination_path}/{subpath}{query_string}")
+    return redirect(f"/{request.LANGUAGE_CODE}/{destination_path}/{subpath}{query_string}", permanent)


network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -341,6 +342,18 @@ def generate_topic_entries_set_html(self, request, topic, *args, **kwargs):
...
+    def category_to_topic_redirect_view(self, request, topic_slug):
+        """
+        If a user attempts to visit the "entries by topic" page using
+        the deprecated /category/ route, we want to redirect them to the
+        correct page using the /topic/ route.
+        """
+        blog_index_slug = self.slug
+        return localized_redirect(
+            request, destination_path=f"{blog_index_slug}/topic", subpath=topic_slug, is_permanent=True
+        )
```
tbrlpld, line 355: I am surprised by this utils function (I have noticed we have a couple of those around and am wondering about their use). Have you consider generating the redirect url using `self.reverse_subpage` to get the topic URL? The necessary information about the locale etc. should already be on the page itself. So I am not sure for the need of that custom utils function. 

See also: https://docs.wagtail.org/en/stable/reference/contrib/routablepage.html#wagtail.contrib.routable_page.models.RoutablePageMixin.reverse_subpage

network-api/networkapi/wagtailpages/tests/test_blog.py
```diff
@@ -220,6 +220,66 @@ def test_topic_route_success(self):
...
+        topic = blog_factories.BlogPageTopicFactory(name="Test topic")
+        self.synchronize_tree()
+        blog_index__url_fr = self.blog_index.get_translation(self.fr_locale).url
+        category_route_url_fr = f"{blog_index__url_fr}category/{topic.slug}/?test_param=123"
+        topic_route_url_fr = f"{blog_index__url_fr}topic/{topic.slug}/?test_param=123"
+
+        response = self.client.get(path=category_route_url_fr, follow=True)
+
+        self.assertRedirects(response, expected_url=topic_route_url_fr, status_code=301, target_status_code=200)
+
```
tbrlpld, line 282: Kudos for a nice test setup 👌

tbrlpld (commented):
Thanks for taking this on @danielfmiranda I have a question about the use of that custom utils function for the redirect. 


I am guess we need the query parameters for tracking?

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: Hi @tbrlpld thanks for the great feedback regarding the redirect.

After looking into it a little more, I tried to see if I can rework this function to instead use django's `redirect()` function instead of our `localized_redirect`, but found it quite difficult to get both the language code **and** query parameters to work.

For that reason, I am voting that we stick with this custom `localized_redirect` function.

However, I do like the idea of using `self.reverse_subpage`, as that means we wont have the path `/topic/` hardcoded. I have updated the function to use reverse subpage to grab the needed path:

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/4927fe1e045b6f79e9298e85f9fe256e47785f8e/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L352-L355

mmmavis (approved):
Thanks for the instructions! I've tested everything mentioned in the PR description plus one additional test for hash. Everything works as expected! Code looks good to me but I'll defer the final approval to @tbrlpld just to be sure.

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
tbrlpld: Sorry for being inquisitive about this. I have just been noticing these localization utils all over the place and am not sure why they exist. Understanding their purpose is especially hard since there are no comments and no tests for them. If find them quite confusing because their functionality seems to be duplicating things that come out-of-the-box or simpler with Wagtail.
 
Can you add some more context of what is going wrong when trying to combine queryparameter and language code? What do you have to do to make the language code work? Is is exactly the part that I meant should come from using `self.reverse_subpage`. The queryparameters you might have to tag on. 

If the point of using `localized_redirect` is to keep queryparameters, we might be better off with a util for that specific use case. And probably also better to make that util only add the queryparameters from a request to a URL string (instead of tying it to the redirect functionality) because that increased the use cases for it.

danielfmiranda: Hi @tbrlpld!

After taking a second look at the docs I believe I found my mistake. 🤦 

I was attempting to redirect to the path returned by `self.reverse_subpage` like so:

```python
topic_path = self.reverse_subpage("entries_by_topic", args=(topic_slug,))

...
topic_path += f'?{request.META["QUERY_STRING"]}'

return redirect(topic_path, permanent=True)
```

Doing so would create some sort of infinite loop and redirect me to a path such as:
`localhost:8000/blog/topic/mozilla-festival/topic/mozilla-festival/topic/mozilla-festival/topic/mozilla-festival/topic/mozilla-festival/...`



However, after taking a 2nd look at the docs I see that they combine both `self.url` and `self.reverse_subpage` like so:
```python
url = page.url + page.reverse_subpage('events_for_year', kwargs={'year': '2014'})
```


After updating my code to the following, everything is working as expected (language code, query params, etc):
```python
topic_path = self.url + self.reverse_subpage("entries_by_topic", args=(topic_slug,))

if request.META["QUERY_STRING"]:
  topic_path += f'?{request.META["QUERY_STRING"]}'

return redirect(topic_path, permanent=True)
```

Issue #10149, opened 2023-02-13
[PNI] Feature a campaign/petition on PNI homepage


kristinashu:
Currently, the PNI homepage can only feature Buyers Guide Article Pages but I would also like to be able to feature a Buyers Guide Campaign Page.

Screen cap of homepage feature section
<img width="1237" alt="image" src="https://user-images.githubusercontent.com/1068783/218533074-dda65ac6-3da6-4d2a-bfc2-bc25a114ff18.png">

#### Screen caps of process
I've created a petition page
<img width="1417" alt="image" src="https://user-images.githubusercontent.com/1068783/218532623-9e865f56-688a-443b-9a2a-cadf4e6c70cb.png">

I would like to feature it on the PNI homepage
<img width="1425" alt="image" src="https://user-images.githubusercontent.com/1068783/218532848-8e3b5314-dc96-490f-9006-80af89c77779.png">

But the Campaign pages don't show up as an option in the page picker
<img width="1417" alt="image" src="https://user-images.githubusercontent.com/1068783/218532668-236f6b5c-18e1-40a8-9142-6c7493c63c33.png">

## AC
- [x] As an editor, I can add the Buyers Guide Campaign Page type as a Main or Supporting Article feature (the same way I would add an Article)
- [x] As a visitor, I can see a petition page as featured at the top of the PNI homepage the same way I would see an Article 

cc @tbrlpld @cdanfon @mtdenton

kristinashu:
@mtdenton can this make it into this sprint since the pni petition will launch soon and I believe it is a small update?

danielfmiranda:
Hi @kristinashu! I have one quick question regarding the wording used in the CMS.

Since now we are able to choose from both Article Pages and Campaign Pages, I was wondering if I should update the wording in the CMS from "Articles" to "Pages", it would look like this:

![Screenshot 2023-02-23 at 16-27-23 Wagtail - Editing Buyers Guide Page Privacy not included](https://user-images.githubusercontent.com/18314510/221062185-b2d279a6-6484-4f3c-a6b9-86afdccde9d9.png)


thanks!

kristinashu:
Nice catch! Yes, page works for me. Thank you!

cdanfon:
This ticket was originally estimated as a 2, but after having implemented the solution this looks more like an 8.

danielfmiranda:
[...]

PR #10245, opened 2023-03-02
renaming of fields + migrations + updated tests and factory

danielfmiranda:
# Description
Related issues: #10149

After having the solution for issue #10149 reviewed in PR #10206, we needed to separate the backend and frontend changes into two separate PRs per our [backwards-compatibility workflow](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations)

This PR houses the **back-end** changes from PR #10206


network-api/networkapi/wagtailpages/migrations/0079_update_bg_supporting_articles_to_supporting_pages.py
```diff
@@ -0,0 +1,52 @@
+# Generated by Django 3.2.16 on 2023-02-23 22:59
```
tbrlpld, line 1: What do you think about recreating the migration so that we only have one?

network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
```diff
@@ -392,25 +396,26 @@ def get_editorial_content_index(self):
...
-        #        a queryset of the articles (rather than a list) and that we have an
-        #        efficient way of pulling all items for a given locale.
-        return [a.localized for a in articles]
+        # FIXME: This implementation returns the localized, subclassed version of each page.
+        #        But, it is inefficient. Both ".localized" and ".specific" are N+1 queries.
+        #        It would be better to pull all pages for the correct locale at once,
+        #        and applying ".specific()" at the end of the queryset.
+        #        This would require the above returns a queryset of the pages (rather than a list)
+        #        and that we have an efficient way of pulling all items for a given locale.
+        return [page.specific.localized for page in supporting_pages]
```
tbrlpld, line 418: What do you think about moving the `.specific()` call to the queryset, rather than the individual page? That should be more performant.

https://docs.wagtail.org/en/stable/reference/pages/queryset_reference.html#wagtail.query.PageQuerySet.specific

tbrlpld (commented):
Thanks for these updates. Looking good. Just 2 small questions.

Issue #10253, opened 2023-03-03
Replace all the `.form-control` occurrences to `.tw-form-control`


mmmavis:
Currently blocked by #10252

Once Bootstrap's .form-control stylings and our custom SCSS for that class have been moved to Tailwind, replace all the `.form-control` occurrences to `.tw-form-control`.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-136)

mmmavis:
I think this is mostly done. I will wrap up the remaining task and update this ticket by tomorrow. The work should only take less 30 mins to complete.

Issue #8704, opened 2022-05-17
Update source\js\components\join\language-select.jsx


fessehaye:
See [[EPIC] React Components Stylings](https://github.com/mozilla/foundation.mozilla.org/issues/7042) for more details.

## Dev Work

- [x] Make sure there is no bootstrap classes remaining on the JSX file
- [x] If there are CSS classes that are from a local SASS file, refactor them on the JSX file using tailwind classes
- [x] After refactoring the classes from SASS files, delete the previous SCSS file
- [x] Might be needed to create a plugin if we are replacing a reused bootstrap class

## Acceptance Criteria

- [x]  Component should be able to work without a dependency from bootstrap
- [x]  Component should be able to work without a dependency from existing SASS files

mmmavis:
@fessehaye can you verify on staging that there's no regression?

### To QA 
- Go to https://foundation.mofostaging.net/en/campaigns/unknown-influence/ and check
There are three forms on the page 1) Nav bar via "Newsletter" button 2) in page petition form 3) footer newsletter sign up
- Go to https://foundation.mofostaging.net/en/blog/ and check the big "Subscribe to our newsletter" form

Thanks!

mmmavis:
@kristinashu taking for QAing this. Basically we just need to make sure the forms mentioned above are still looking the same as what's on prod.

Issue #8335, opened 2022-02-26
Cleanup source\sass\components\select-dropdown.scss


fessehaye:
We already have `tw-form-control` and `tw-form-control-lg` so we just need to remove this file and replace the dropdowns with the new classes

mmmavis:
@fessehaye can you verify on staging that there's no regression on `select.tw-form-control`

### To QA 
- Go to https://foundation.mofostaging.net/en/campaigns/unknown-influence/ and check
There are three forms on the page 1) Nav bar via "Newsletter" button 2) in page petition form 3) footer newsletter sign up
- Go to https://foundation.mofostaging.net/en/blog/ and check the big "Subscribe to our newsletter" form

Thanks!

danielfmiranda:
Thanks for the fix @mmmavis!

I have taken a look and can confirm that there is no regression. Moving this to ready for prod.

mmmavis:
@kristinashu taking for QAing this. Basically we just need to make sure the forms mentioned above are still looking the same as what's on prod.

kristinashu:
Looks good!

PR #10257, opened 2023-03-03
[SCSS/Bootstrap to Tailwind] .form-control and related stylings

mmmavis:
### Description

- Created a new Tailwind plugin `form-control.js`
- `form-control.js` is based on stylings on [Bootstrap v4.6's _form.scss](https://github.com/twbs/bootstrap/blob/v4.6.0/scss/_forms.scss)(only the relevant bits) and our custom SCSS overrides.
- Replace some classes with the newly created Tailwind ones. There are still some `.form-control` in the codebase that need to be updated to `.tw-form-control`. I've created a follow-up ticket for that so this PR doesn't grow too big #10253 

### To QA
1. Go to any campaign page with a petition form OR if you have staging db locally, go to `/en/campaigns/unknown-influence/`. Test the three forms on the page. There should not be any visual changes compared to what's on prod.
    - "newsletter" on nav
    - petition form (note that border radius and border colour have been updated. Changes are okayed by the design team)
    - footer form
2. Go to `/en/blog/` and test the newsletter form in the page (near footer)

### Related PRs/issues: 
  - #10252
  - #8704
  - #8335

### DIFF
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_category_nav.html
@@ -19,13 +19,13 @@
                     </div>
                     {% include "fragments/buyersguide/category_dropdown.html" %}
 
-                    <div id="product-filter-search" tabindex="0" class="tw-hidden large:tw-inline-flex tw-px-4 tw-items-center tw-bg-blue-05 tw-border-4 tw-border-blue-05 tw-w-[214px] tw-mr-4">
+                    <div id="product-filter-search" tabindex="0" class="tw-hidden large:tw-inline-flex tw-px-4 tw-items-center tw-bg-blue-05 tw-border-4 tw-border-blue-05 tw-w-[250px] tw-mr-4">
                         <img
                             class="tw-w-8 tw-h-8 tw-mr-4"
                             src="{% static "_images/buyers-guide/pni-search.svg" %}"
                         >
                         <input type="text" role="searchbox" id="product-filter-search-input" placeholder="{% trans "Search all products" %}" value=""
-                            class="tw-bg-blue-05 tw-outline-none tw-py-4 tw-text-base tw-text-blue-60 tw-min-w-0 placeholder:tw-text-blue-60" />
+                            class="tw-bg-blue-05 tw-outline-none tw-py-4 tw-text-base tw-text-blue-60 tw-min-w-0 placeholder:tw-text-blue-60 placeholder:tw-text-[14px] tw-w-full" />
                         <label for="product-filter-search-input" class="clear-icon">&nbsp;</label>
                     </div>
                 </div>
--- a/network-api/networkapi/templates/fragments/buyersguide/pni_sort_dropdown.html
+++ b/network-api/networkapi/templates/fragments/buyersguide/pni_sort_dropdown.html
@@ -5,34 +5,34 @@
         tw-list-none
         tw-relative
         tw-cursor-pointer
-        tw-m-0 tw-p-0 tw-ml-4
+        tw-m-0 tw-p-0 medium:tw-ml-4
         tw-flex
         tw-items-center
-        tw-w-[250px]
+        tw-min-w-[250px]
     "
 >
     <li
         role="button"
         id="pni-creepiness__selected"
-        class="tw-overflow-hidden tw-whitespace-nowrap tw-w-full tw-text-ellipsis tw-flex tw-items-center tw-m-0 tw-p-6"
+        class="tw-w-full tw-flex tw-items-center tw-m-0 tw-pl-0 medium:tw-pl-6 tw-p-6"
         tabindex="0"
     >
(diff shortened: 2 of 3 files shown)


fessehaye (commented):
Does `form-control-feedback` have anything to do with the class `form-control` I do not see them using the same CSS inheritance. I do not even think this has to be apart of the plugin since it is used very little. I think we can replace some of those instances with just some utility classes and call it a day right?

fessehaye (changes requested):
tw-has-danger is do not show up on IntelliSense so it is probably not being complied

tailwind-plugins/form-control.js
```diff
@@ -0,0 +1,181 @@
...
+        },
+        "select.form-control-lg": {
+          ...selectFormLgControl(theme),
+        },
+        "input[type='date'].form-control, input[type='time'].form-control, input[type='datetime-local'].form-control, input[type='month'].form-control":
+          {
+            appearance: "none",
+          },
+        ".has-danger": {
+          ...hasDanger(theme),
```
fessehaye, line 176: You can address the `.has-danger` class in two different ways. You can make a danger variant and address every HTML instance of `form-control` and utility classes. Or you can address has-danger the same way we kinda address adding dark class for form-control in L117. I do not think the has-danger function work in this case.

fessehaye:
When we use the `addComponents` API, we want to be able to use one CSS class and hover over that class and get all of the information we need about what is being included. In your example, we do that for `.has-danger` it doesn't really make sense to make a component for it.

When we are converting SASS files we have to keep in mind that `@extend` feature is used quite a bit and if we want to transition out of SASS or move it to a tailwind plugin we usually have to flatten the SASS. There will be some situations where it feels too awkward to remove certain SCSS lines but if that is the case keeping some remaining SASS is totally fine as long as we try to remove SCSS rules. That is the most essential part about cleaning the SASS files since we can always further clean up the remaining CSS at a later time after the conversion to tailwind is finished.

tailwind-plugins/form-control.js
```diff
@@ -0,0 +1,181 @@
...
+  paddingX: "12px",
+  paddingRight: "48px", // paddingX * 2 + backgroundSize
+};
+
+/**
+ * Stylings for .form-control (converted from Bootstrap 4.6 with our custom SCSS)
+ * @param {*} theme Tailwind theme config
+ * @returns {object} Stylings
+ */
+function formControl(theme) {
```
fessehaye, line 17: Thank you for using the theme variables.

fessehaye:
This is almost there @mmmavis. Just a few corrections we need to address. If you need any more help let me know. Thanks for doing this

tailwind-plugins/form-control.js
mmmavis: Thanks for the suggestion. Can you elaborate on `I do not think the has-danger function work in this case.`? I checked again and verified the `.tw-has-danger .form-control` and `.tw-has-danger .form-control-feedback` DOM elements are having the stylings I defined here.

fessehaye: image
When you hover over `.tw-form-control` it reveals zero information about the has-danger classes. 

Hovering over `tw-has-danger` does nothing as well and should not be a component.

fessehaye:
> > Does `form-control-feedback` have anything to do with the class `form-control` I do not see them using the same CSS inheritance. I do not even think this has to be apart of the plugin since it is used very little. I think we can replace some of those instances with just some utility classes and call it a day right?
> 
> Thanks for pointing out that `.form-control-feedback` doesn't rely on `.form-control` stylings. I can remove it from `form-control.js` and use utility classes inline instead. However, I do have a question - with this approach, how do we ensure future "form feedback" elements are using the same stylings? Does it require a dev to dig into the codebase, find a form that has "feedback" element, and grab the same set of utility classes to use again?

Yeah, good question! If we can remove a CSS class name it is always the better move. A lot of them do not have consistent naming conventions or at least ones we have not been consistent about. Along the way you will probably find css-class names that could just be built with tailwind classes. The exception is special cases like CMS content or tricky CSS rules/conditions.

Yeah, the goto is just digging the codebase. Sometimes you will find cases were refactoring the HTML could save you time,but it might be out of scope for these smaller tickets.

tailwind-plugins/form-control.js
mmmavis: @fessehaye I think the above problem is actually from Tailwind CSS IntelliSense. I tried to comment out `...formControlPlugins,` in `tailwind.config.js` and uncomment it again, everything then got generated as expected.

image

mmmavis:
@fessehaye PR updated! See individual commits for details.

tailwind-plugins/form-control.js
fessehaye: interesting, I guess since we are moving away from this implementation of has-danger, it doesn't matter atm

fessehaye (approved):

Issue #10163, opened 2023-02-16
Create checkbox for tables and graphs to have different widths on Article pages


nancyt1:
We've been getting a few requests to have simple tables, advanced tables, and datawrapper graphs fit the width of the content. Currently they break out the bounds of the content and it would be great to have a checkbox like the one in the image block to display the graphs at wide or narrow widths. By default, it should align with the content. 

AC
- As an editor, I'm able to check a box to have my table wider than the content.
- As a reader, I'm able to see tables the same width as the content if "wide" box is unchecked.

tbrlpld:
Do we ever want them to overflow the content width?

nancyt1:
I don't think so. Can you think of any cases where that might happen? @tbrlpld

tbrlpld:
@nancyt1 I does happen in the example you are showing above. Sorry, maybe I should have been more specific about my question. The question is: does this need to be an option or should we make sure that it always has the same width as the content?

nancyt1:
@tbrlpld Yes, it should be an option. I was thinking we have a "wide" option like the screenshot above and by default it will have the same width as the content.

tbrlpld:
Sorry, that was the question: Why do we want it to be wide? (that is what I mean with overflow... which I realize now has been poor wording)

nancyt1:
I would like the keep the width that we have currently and just label it as "wide". We've just been getting requests to make the tables narrower to align with the content but the wide option is good for tables that have a lot of columns or graphs that need more room for details. Sorry does that make sense or would u want to hop on a huddle?

tbrlpld:
@nancyt1 Thanks that makes sense. I guess it would have been a lot easier if I would have asked the initial question more straight clearly 😅

tbrlpld:
@nancyt1 I assume the checkbox would be ok to have an effect on other pages where the block types exist too?

nancyt1:
Yeah, I think that would be fine!

tbrlpld:
@nancyt1 Considering that the current layout is the "wide" one, I assume the check box should default to "checked" (True).

nancyt1:
That works!

kristinashu:
@mtdenton @cdanfon any chance this could get into the upcoming sprint? There's another report launching around March 20 and it would be nice to have this feature.

danielfmiranda:
Hi Everyone! 

[...]

PR #10333, opened 2023-03-15
wide option for advanced table block

danielfmiranda:
## Description

**Link to test page:** https://foundation-s-10163-wide-fbv9zi.mofostaging.net/en/publication-page-with-child-article-pages/fixed-title-article-page/

**Username/Password:** admin2/admin2

**Related PRs/issues:** #10163

This PR adds a new checkbox/boolean field named "wide" to the article page's `AdvancedTableBlock`.

When this field is **checked**, the table extends beyond the other content of the article page. 
When this field is **unchecked**, the table is in line with the other content.

This field is set to `True` by default. 

## Screenshots:

**"wide" set to True**
Screenshot 2023-03-15 at 09-40-09 Opaque and Overstretched Part II

**"wide" set to False**
Screenshot 2023-03-15 at 09-40-35 Opaque and Overstretched Part II


network-api/networkapi/wagtailpages/migrations/0079_advance_table_block_wide_option.py
danielfmiranda, line 1: This is just the auto-generated migration file needed to add the "wide" field to the `AdvancedTableBlock` model.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
```diff
@@ -2,7 +2,7 @@
 {% load l10n wagtailcore_tags %}
 
 {% block block_content %}
-    <div class="article-table-block table-responsive">
+    <div class="article-table-block tw-overflow-x-auto tw-mx-auto {% if self.wide %} tw-w-full {% else %} tw-w-4/5 {% endif %}">
```
danielfmiranda, line 5: The bootstrap class `table-responsive` added the following CSS Rules:

```css
overflow-x: auto;
width: 100%;
```

I am adding the same rules, this time using tailwind classes.

If the table is not set to "wide" then the width is reduced to 80% to match the width of the other article content.

beccaklam:
@danielfmiranda Desktop and mobile look good but on tablet the "wide" looks a more narrow than the rest of the content, would we be able to make it so that it's the same width as the other streamfield block content? 
image

mmmavis:
@beccaklam when I tested it the result seems to be different than what you described. 🤔 

On medium size screen, wide version seems to be in the same width of rest of the content while the regular version is narrower than other content. 

image

mmmavis:
@danielfmiranda is there a mockup we can reference for tablet design?

danielfmiranda:
Thanks for taking a look @beccaklam! I have added a fix for the error you found.  

Hi @mmmavis! I actually do not have a mockup. I believe that is the expected behavior, but I will leave it to Rebecca to confirm 👍

beccaklam:
@mmmavis I think we had the same result? Wide set to false was resulting in a narrower width, while wide set to true was resulting in a width that matched the rest of the content. 

@danielfmiranda The fix looks good! Thank you for doing that :)

Yeah I'm not sure what the intended behaviour is for wide on tablet and mobile but I think at those screen widths having everything match in width is probably the best since there's not a lot of space already.

beccaklam (approved):

mmmavis (approved):
Nice work!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/advanced_table_block.html
mmmavis: Smart!

Issue #9033, opened 2022-06-30
Detail page - Add a page chooser option to Research Links


bheasman:
### Describe the feature request
As there will be a lot of internal content that is being surfaced in the research hub, it'd be good to add a page chooser option to the 'Research links' section of the research detail page. This will then give editors the option of adding a file, an external URL or choosing an existing page. The key benefit of this is that if the internal page moves within the page tree, the links will remain intact.

### Acceptance Criteria

- [x] As an editor, I can use a page chooser to select an internal page as a research link

tbrlpld:
@kristinashu @beccaklam  Would  either of you mind doing the QA on this one? It would be greatly appreciated 🙏

kristinashu:
The page chooser is working well!

PR #10360, opened 2023-03-20
#9033 Add page chooser to research links

tbrlpld:
# Description

This PR adds page chooser to the research detail link model. These models are used on the research hub detail page to link to the original resource (whereas the detail page is only a summary page). 

The detail link model already existed with fields for URL or a document. Many resources are other pages on the Foundation site itself. Adding a page chooser makes it easier for editors to link to internal pages (because they don't have to visit the page anymore to copy the URL). It also ensures that links are not breaking then the pages are moved on the site. 

The validation rules are checking that only one (but at least one) of the fields is set, to be able to determine the final link URL. 

Link to sample test page: http://localhost:8000/en/research/library/assume-fill-mr-rich-ready-test-company/
Related PRs/issues: #9033

# Checklist

**Tests**

**Changes in Models:**
- ~~Did I update or add new fake data?~~
- [x] Did I squash my migration? One one migration added.

**Documentation:**
- ~~Is my code documented?~~
- ~~Did I update the READMEs or wagtail documentation?~~


network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
```diff
@@ -180,21 +186,26 @@ def __str__(self):
...
-                {"url": error_message, "document": error_message},
+                {"url": error_message, "page": error_message, "document": error_message},
                 code="required",
             )
 
     def get_url(self):
         if self.url:
             return self.url
+        elif self.page:
+            return self.page.get_url()
```
kevinhowbrook, line 209: **Potential issue**

I'm sure how much of a priority this edge case is but due to the nature of streamfield data storing only references, we probably want to add a check for `and self.page.live` here too. Because if an unpublished page is still referenced it will 404 for the user. Along with this, we will need to add an `if not get_url` check in the template for this link. Otherwise the link will render as `None`
Deleting a referenced page looks fine.

kevinhowbrook (commented):
Thanks @tbrlpld just one point that might be worth considering.

network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
tbrlpld: Thanks for catching this!

tbrlpld: I wonder if we should prevent linking to non-live pages... ah... but a page can become not-live after it has been linked to! 

Handling a non-live page it is.

tbrlpld: Implemented the check and am not rendering the link button for non-live pages.

network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
```diff
@@ -163,38 +172,52 @@ class ResearchDetailLink(wagtail_models.TranslatableMixin, wagtail_models.Ordera
...
+            if not self.page.live:
+                logger.warning(
+                    f"Detail link to unpublished page defined: { self } -> { self.page }. "
+                    "This link will not be shown in the frontend."
+                )
+                return ""
+            return self.page.get_url()
         elif self.document:
             return self.document.url
+        raise ValueError("No URL defined for this detail link. This should not happen.")
```
tbrlpld, line 223: @kevinhowbrook Mypy had failed before because this branch in the logic was not covered. I have added a test for this now too.

kevinhowbrook (approved):

Issue #10361, opened 2023-03-21
Detail page - Limit child page types to article and publication


tbrlpld:
## Description

Currently, research detail pages can have any page as a child page. This does not make much sense and can make it hard for editors to select the correct type. 

We might want to limit this to a few sensible options. E.g. publication page and article page. 

[Epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10339) | [Designs](https://www.figma.com/file/dPGD3acNbT75dvEGzNaGUb/Responsible-Computing?t=zoMb9zxg25R0T6yt-6)

## Acceptance criteria

- [x] As an editor, when adding a child page to a research detail page, I can add an article page or a publication page. 
- [x] As an editor, when adding a child page to a research detail page, I don't see any other options except the one listed above.

tbrlpld:
@kristinashu Does this sound sensible to you?

danielfmiranda:
Hi @kristinashu! This is now on staging and ready for QA. I was wondering if you could please take a look whenever you find a free moment? Thanks in advance!


## Steps to test:

1. Visit any detail page, for example, "[Shifting power through data governance](https://foundation.mofostaging.net/cms/pages/18276/edit/)"
2. Click "Add Child Page"
3. There should now be 2 options, "Article Page" and "Publication Page"
![image.png](https://images.zenhubusercontent.com/6046b7f1bc36adc9ce95a16a/f535b954-08e5-42f1-a1da-dbe0ee2962cb)
4. If everything looks as expected, testing is complete!

kristinashu:
This is working well, thank you!

Issue #9266, opened 2022-08-25
Detail page - Update which fields are required


kristinashu:
### Describe the bug
I need to create a detail page before I can create a Publication under it but the Landing page requires a link the Publication in order to create it, which I can't create until I have the Landing page.

### To Reproduce
1. Go to add a new Child page new in the Research Library https://foundation.mofostaging.net/cms/pages/18267/
2. Fill out the info that you can (no link to the Publication because it doesn't exist yet)
3. Publish the page
4. Can't Publish 

### Expected behavior
I understand that these fields should be required but unfortunately they shouldn't be required if it isn't possible.

### Screen cap
<img width="1176" alt="Screen Shot 2022-08-24 at 3 10 01 PM" src="https://user-images.githubusercontent.com/1068783/186713141-7ea736a9-b7ec-40a7-ad20-8456c8b44f9b.png">

tbrlpld:
@kristinashu This is ready for testing on staging: https://foundation.mofostaging.net/cms/pages/add/wagtailpages/researchdetailpage/18267/

kristinashu:
Looks good, thank you!

PR #10362, opened 2023-03-21
#9266 Detail page - Link not required

tbrlpld:
# Description

Follow-up PR to #10360 

This PR removes the requirement for the detail pages to have a link defined. 

The requirement has become cumbersome, because the detail pages are often used as the parent page of the article or publication page that the link should go to. With the link being required, we create a logical loop where the detail page can not be created because we can not link to the child and we can not create the child until the detail page exists. 

Removing the requirement for a link breaks this logical loop. The editors can now create a detail page without a link, create the child page and then come back to the detail page to add the link. They do have to remember to come back to add the link though. 

We could consider updating this to automatically link to the first child if no link is defined. But this needs to be checked with stakeholders and design if that is desirable. 

While testing I also noticed that we should limit the types of child pages we can create on a detail page. #10361 will deal with that. 

Link to sample test page: http://localhost:8000/cms/pages/add/wagtailpages/researchdetailpage/179/
Related PRs/issues: #9266

# Checklist

**Tests**
- ~~Is the code I'm adding covered by tests?~~ I could not find a good way to test this. 

**Changes in Models:**
- ~~Did I update or add new fake data?~~
- ~~Did I squash my migration?~~

**Documentation:**
- ~~Is my code documented?~~
- ~~Did I update the READMEs or wagtail documentation?~~

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
@@ -64,7 +64,7 @@ class ResearchDetailPage(research_base.ResearchHubBasePage):
 
     content_panels = wagtail_models.Page.content_panels + [
         image_handlers.FieldPanel("cover_image"),
-        edit_handlers.InlinePanel("research_links", heading="Research links", min_num=1),
+        edit_handlers.InlinePanel("research_links", heading="Research links"),
         edit_handlers.FieldPanel("original_publication_date"),
         edit_handlers.FieldPanel("introduction"),
         edit_handlers.FieldPanel("overview"),


network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
```diff
@@ -64,7 +64,7 @@ class ResearchDetailPage(research_base.ResearchHubBasePage):
 
     content_panels = wagtail_models.Page.content_panels + [
         image_handlers.FieldPanel("cover_image"),
-        edit_handlers.InlinePanel("research_links", heading="Research links", min_num=1),
+        edit_handlers.InlinePanel("research_links", heading="Research links"),
```
kevinhowbrook, line 67: I did notice when testing that this didn't have a max (which i was going to raise) but I think this change in itself is fine :+1:

kevinhowbrook (approved):

Issue #10391, opened 2023-03-28
Library page - Speed up test execution


tbrlpld:
## Description

Currently, the ~30 tests for the library page take me locally about 80-120 seconds to run. This is way too slow. 
We should speed this up to facilitate faster development cycles. 

You can probably move setup code to `setUpTestData` and switch from using the test client to testing the getter method for the detail pages directly.

[Epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10339) | [Designs](https://www.figma.com/file/dPGD3acNbT75dvEGzNaGUb/Responsible-Computing?t=zoMb9zxg25R0T6yt-6)

## Acceptance criteria

- [ ] As a developer, I can confirm that running the tests for the library page is faster than the baseline execution time I get with the target branch. 

:warning: Depends on #10392, as that ticket will be making changes to the functions that we will be using when updating the tests.

tbrlpld:
@danielfmiranda Let me know if you have any questions here.

danielfmiranda:
Thanks @tbrlpld!

Interestingly, my usual time to run all the test is about 230 seconds instead of  80-120.

After updating some of these to use the getter method like you mentioned above, I was able to bring my total runtime down around ~40%. I have been tracking test runtimes in this [google sheets doc](https://docs.google.com/spreadsheets/d/1JKB4CaOlfRldns8O5YfF5sFSTxjAa9cOIRlQmcoQ6-s/edit#gid=0). 

However, I have been experiencing an issue when trying to work on the test `test_filter_multiple_profiles` and I cant seem to find the cause of the issue. I was wondering if you might be able to help.

To start, I am trying to return a queryset of all ResearchDetailPages. I have marked them in the pseudocode below with a comment that starts off with ` ***** #NOTE:`.


```python
    def test_filter_multiple_author_profiles(self):
        start_time = time.time()
        detail_page_1 = detail_page_factory.ResearchDetailPageFactory(
            parent=self.library_page,
        )
        profile_a = detail_page_1.research_authors.first().author_profile
        detail_page_2 = detail_page_factory.ResearchDetailPageFactory(
            parent=self.library_page,
        )
        profile_b = detail_page_2.research_authors.first().author_profile
        # Make author of first page also an author of the second page
        relations_factory.ResearchAuthorRelationFactory(
            research_detail_page=detail_page_2,
            author_profile=profile_a,
        )

       # ***** NOTE: This line will NOT a populated queryset *****
[...]

PR #10488, opened 2023-04-11
Improved test execution times

danielfmiranda:
# Description

Link to sample test page:
Related PRs/issues: #10391

This PR updates the tests in `test_library_page.py` to use the `ResearchLibraryPage` model's `_get_research_detail_pages()` method directly, instead of preforming a `get()` request against the test client.

Doing so results in a reduction of runtime from an average of 358 seconds to  218 seconds. **(a 48.6% decrease)**

### Before:
Screenshot 2023-05-09 at 6 43 52 PM  
  
 

### After:

Screenshot 2023-05-09 at 6 55 50 PM


danielfmiranda:
Hi @jhonatan-lopes! 

I wanted to call out that all of these tests have been update except for one, which is the one that I mention was giving me trouble in this PR's linked ticket #10391.

Since we have already reduced the runtime by about 48%, and updating this test would make a negligible change in runtime (maybe 1-2 seconds), I have created ticket #10597 to track the investigation work and to avoid this ticket blocking me on my other deliverables this sprint.

Thanks!

jhonatan-lopes (approved):
Thanks @danielfmiranda! LGTM :+1:

PR #10536, opened 2023-04-25
Implement pytest as a test runner

jhonatan-lopes:
# Description

- Set up `pytest` as a test runner.
- Make sure that Django checks are ran before testing and during CI checks.
- Run tests in parallel as default (performance improvement)
- Reuse db between tests (performance improvement)
- Add missing migration

Why should we use `pytest` instead of Django's manage.py test command?

- Answered here: https://pytest-django.readthedocs.io/en/latest/#why-would-i-use-this-instead-of-django-s-manage-py-test-command
- Boils down to: simpler to write tests, still works with `unittest` (Django) style tests, more configurable, better output (personal preference)

Running checks (`manage.py check` and `manage.py makemigrations --check --dry-run`) are also implemented here and are crucial for catching mistakes. Part of the PR is an uncaught migration that went unoticed in our CI. Although there is test that aims to catch missing migrations, it either crashes before (when fields are added/removed) or it doesn't catch it at all (as is the case with the field change in this PR).

Those checks are also implemented at `invoke` level, when running tests locally. They should hopefuly avoid having those mistakes reach the CI in the first place.

# Checklist

**Tests**
- ~~[ ] Is the code I'm adding covered by tests?~~ Infrastructure change

**Changes in Models:**

**Documentation:**
- [X] Is my code documented?
- [X] Did I update the READMEs or wagtail documentation?


jhonatan-lopes:
#10531 shows an example where a field was added to the `Profile` model but a migration file was not included. Before the fixes in this PR, the failure occured inside the testing, with an error indicating that the field could not be located (since there were no migrations for it). Now, it fails much sooner with a more appropriate error message.

mmmavis (approved):
Looks good to me! Did a few tests with `inv test-python` and they all returned me the expected results. Thanks for improving the dev exp and quality of our codebase @jhonatan-lopes 🎉

danielfmiranda (approved):
Thanks for the help while reviewing @jhonatan-lopes! 

After testing can confirm this is working as expected. Excited to start using this!

Approved 👍

Issue #10538, opened 2023-04-25
Newsletter Bug - Fix null value for the source value that feeds into Acoustic


mmmavis:
This is a request from Slack. Full thread can be found [here](https://mozilla.slack.com/archives/C98GW3P7X/p1682437114403409).

Currently the majority of our Newsletter forms have a null value for the source value that feeds into acoustic.

A working example of the source value being returned is on any of the sign up forms via the common voice webpage. The data being returned is:

```json
"newsletters": [
  {
    "name": "common-voice",
    "subscribed": true,
    "format": "H",
    "lang": "en",
    "source": "https://commonvoice.mozilla.org/en",
    "unsub_reason": null
  },
  {
    "name": "mozilla-technology",
    "subscribed": true,
    "format": "H",
    "lang": "en",
    "source": "https://commonvoice.mozilla.org/en",
    "unsub_reason": null
  }
]
```

An example of it not working is on the Foundation home page. Data being returned is:

```json
"newsletters": [
  {
    "name": "mozilla-foundation",
    "subscribed": true,
    "format": "H",
    "lang": "en",
    "source": null,
    "unsub_reason": null
  }
]
```

mmmavis:
Did some investigation and found out the code used to include "source url" somehow got lost when we were doing the Cinchy/SQS to Basket transition in April 2021 😬 . I'll submit a fix soon. I will also file a new ticket to revisits tests for newsletter signup. We'll need to set up tests if we haven't done so already. More details in the to-be-filed ticket.

mtdenton:
@danielfmiranda this is clear to be pushed without stakeholder QA, they will test tomorrow.

danielfmiranda:
Got the OK from Tyler to push this to prod in the engineering slack channel. Moving to ready for prod!

PR #10539, opened 2023-04-25
Update basket.subscribe payload to include source_url

mmmavis:
# Description

Update `basket.subscribe` payload to include `source_url`. Basket doc can be found [here](https://basket.readthedocs.io/newsletter_api.html#news-subscribe).

For email signup through MozFest Ti.to ticket purchase - 
I'm not sure what to update [this line](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/events/views.py#L36) to because the request is coming from Tito and doesn't seem to include source url info. This is [the Tito endpoint doc](https://help.tito.io/en/articles/2011381-webhooks) I referred to, I'm not certain if this is the right place to look for answer though.

Related PRs/issues: #10538


mtdenton (approved):
Looks good!

danielfmiranda (approved):
Thanks Mavis for the work and also for providing the docs to reference. After giving them and the code a look, everything looks good to me 👍 Approved!

Issue #10418, opened 2023-03-31
Detail page - Use `select_related` to retrieve author profiles


tbrlpld:
## Description

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/templates/wagtailpages/research_detail_page.html#L64-L68

The `ResearchDetailPage` model is related to the `Profile` model via the `ResearchAuthorRelation` model. 

In the `research_detail_page.html` template the author profiles are rendered in the "project leads" section.

This is currently using `for research_author in page.research_authors.all` to retrieve the through model instances and then calls `research_author.author_profile` to get to the actual profile model that contains the data. 

This is a typical 1+N query problem. 

We should be using `select_related` to also pull the related `Profile` (`.author_profile`) when querying the `ResearchAuthorRelation` model. This is probably best done by adding a helper method to the model and adding the `ResearchAuthorRelation` queryset to the template context. Or, even better, pull only the `Profile`s without the through model (`ResearchAuthorRelation`) instances. Also make sure to select the related `locale` in the profile to avoid another database hit.

[Epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10339) | [Designs](https://www.figma.com/file/dPGD3acNbT75dvEGzNaGUb/Responsible-Computing?t=zoMb9zxg25R0T6yt-6)

## Acceptance criteria

- [ ] As a developer, I can confirm we rendering multiple profiles in the "project leads" section with less than 1+N queries. Ideally, regardless of the number of profiles it should only take 1 query to retrieve all of them. 
- [ ] As a site visitor, I can see all profiles related to the page displayed in the "project leads" section
- [ ] As a site visitor, when visiting a detail page in a non-default locale, the translated profile data is used if it exists. If a profile has not been translated into the locale used to visit, it displays the original (default locale) data of the profile).

PR #10542, opened 2023-04-26
added get_research_authors method to ResearchDetailPage

danielfmiranda:
# Description

Related PRs/issues: #10418 

**This PR adds the following:**
- A new method for the ResearchDetailPage model called `get_research_authors`. This method returns a localized queryset of all the `Profiles`  for research authors that are related to  the page through a [`ResearchAuthorRelation`](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/research_hub/relations.py#L7). I then call this method to add the list of profiles to the page's template context for use.
- Update to the template `research_detail_page.html` to now use the profile information from the context, instead of `page.research_authors.all` and then `research_author.author_profile`
- Some tests for the method to make sure it works and returns localized profiles as expected


network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
```diff
@@ -134,8 +136,15 @@ def get_context(self, request):
         context = super().get_context(request)
         context["breadcrumbs"] = self.get_breadcrumbs()
         context["authors_index"] = authors_index.ResearchAuthorsIndexPage.objects.first()
+        context["research_authors"] = self.get_research_authors()
         return context
 
+    def get_research_authors(self):
+        research_author_profiles = localize_queryset(
+            Profile.objects.filter(authored_research__research_detail_page=self)
```
jhonatan-lopes, line 144: I think we need a `prefetch_related` here to avoid the N+1 problem, something like:

```python
Profile.objects.prefetch_related("authored_research").filter(authored_research__research_detail_page=self)
```
Otherwise, for each `authored_research` in a `Profile`, we will have additional queries.

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
```diff
@@ -33,6 +36,60 @@ def test_research_detail_page_breadcrumbs(self) -> None:
         self.assertEqual(len(breadcrumbs), 2)
         self.assertEqual(breadcrumbs, expected_breadcrumbs)
 
+    def test_get_research_authors(self) -> None:
```
jhonatan-lopes, line 39: Here, it would be interesting to add a `Profile` linking to another page, and check that it doesn't show up in our `research_authors`.

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
```diff
@@ -33,6 +36,60 @@ def test_research_detail_page_breadcrumbs(self) -> None:
...
+        for _ in range(3):
+            research_author_relation = relations_factory.ResearchAuthorRelationFactory(
+                research_detail_page=detail_page
+            )
+            author_profiles.append(research_author_relation.author_profile)
+
+        research_authors = detail_page.get_research_authors()
+
+        self.assertEqual(len(research_authors), 3)
+        self.assertCountEqual(author_profiles, research_authors)
```
jhonatan-lopes, line 55: I think it would be best to assert that the author profiles actually show up in `research_authors` instead of counts, something like:

```python
self.assertIn(author_profiles[0], research_authors)
self.assertIn(author_profiles[1], research_authors)
self.assertIn(author_profiles[2], research_authors)
```

And if we had a profile that was not supposed to be in `research_authors`:

```python
self.assertNotIn(profile_not_in_this_page, research_authors)
```

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
```diff
@@ -33,6 +36,60 @@ def test_research_detail_page_breadcrumbs(self) -> None:
...
+        author_profiles = []
+        detail_page = detail_page_factory.ResearchDetailPageFactory(parent=self.library_page, research_authors=[])
+
+        for _ in range(3):
+            research_author_relation = relations_factory.ResearchAuthorRelationFactory(
+                research_detail_page=detail_page
+            )
+            author_profiles.append(research_author_relation.author_profile)
+
+        research_authors = detail_page.get_research_authors()
```
jhonatan-lopes, line 52: Can we [assert the number of queries](https://docs.djangoproject.com/en/4.2/topics/testing/tools/#django.test.TransactionTestCase.assertNumQueries) here, please? We need to make sure that we are not running into a N+1 problem again.

jhonatan-lopes (changes requested):
Thanks for implementing this @danielfmiranda! Could you take a look at the comments here, please?

network-api/networkapi/wagtailpages/pagemodels/research_hub/detail_page.py
danielfmiranda: Hey Jhonatan, thanks for walking through this comment with me over zoom. I have updated the PR to now include the `prefetch_related` method!

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
danielfmiranda: Thanks for the suggestion Jhonatan!

[AssertCountEqual](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertCountEqual) actually does not check the count of items, but instead checks that two items contain the same elements, regardless of their order. 

However, after your suggestion I can see that the name is very misleading, so much so that there has been a [github issue opened over it](https://github.com/python/cpython/issues/71258)! Though in the ticket they decided to leave the name as-is.

I like your suggestion of checking for each item with their index instead, as its easier to read/understand and have added it to my PR 👍 

I also added the other suggestion of creating an unrelated page and author and double checking that they do not appear in the list.

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
danielfmiranda: Updated! See comment below

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
danielfmiranda: Hi @jhonatan-lopes! 

Thanks for working through this with me over zoom, instead of doing the check here, I have added a new function titled "[test_get_research_authors_avoids_n1_queries](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/16637e81be7af07e285942f16607e9e59ef428b2/network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py#L63)" as discussed.

It creates 10 related authors, and then checks that the number of queries are less than that.

One concern I do have however and one I was hoping you might be able to help with, is when taking a look at the queries that the assert statement catches (currently catches 2), the queries both seem to be locale related 🤔 


Screenshot 2023-04-26 at 5 39 27 PM

From my understanding, I believe there should be at least 4:
```python
research_author_profiles = localize_queryset(
   Profile.objects.prefetch_related("authored_research").filter(authored_research__research_detail_page=self)
)
```
- 1 for pulling the profiles
- 1 for the `prefetch` statement from the PR feedback
- 1 for `localize_queryset` to get the default locale
- 1 for `localize_queryset` to get the active locale 

I do remember us discussing that the number of queries can vary from time to time. 

Should I wait until we merge in the testing framework from #10536 so I can change the statement to something along the lines of `self.assertNumQueries(< 10)`?

danielfmiranda:
Hi @jhonatan-lopes, thanks for the review! I have updated this PR based on your feedback and this PR is now ready for re-review whenever you find a free moment

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
jhonatan-lopes: Using the debug toolbar on a detail page, I can see that there is a query on profile: 
image

We know the prefetch is working since we can change the N+1 test to run with 100 or 1000 objects and it takes (roughly) the same amount of time. I won't block the PR because of this issue. If you would like to investigate this further, it's up to you. Otherwise, we can remove this test and create a ticket to revisit this another time.

jhonatan-lopes (approved):
Thank you for the changes, Daniel! I've put a comment for the number of queries assertion but it's non-blocking. LGTM :+1:

network-api/networkapi/wagtailpages/tests/research_hub/test_detail_page.py
danielfmiranda: Thanks @jhonatan-lopes! 

I have created https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10554 to track that work 👍

Issue #10422, opened 2023-03-31
Taxonomies - Add abstract taxonomy base class


tbrlpld:
All taxonomies, regardless of their specific relationships and associations are very similar. They are usually a simple model with a single name field. We could consider adding an abstract base taxonomy and have all taxonomies be based on that abstract base. This could even be broader than just the library site section. The abstract taxonomy could be used site wide. The abstract functionality for taxonomies is fairly well known (a name field). What we could save by adding an abstract base is the duplicate field, panel and translation specification.

See also [this section](https://docs.google.com/document/d/1DtBZvmd3Dpx6ssXWnXd8jk0jPLwfU9FGjnC9rl_CTfg/edit#heading=h.gf42qxkn6q78) of the tech spec. 

I am not 100% sure if this is needed as this introduces again the option to add features to the base class that are not really needed across the instances... 

Having an abstract base to share functionality will make more sense when we want to share some more complex configuration across all child classes, i.e. #10442.

[Epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10339) | [Designs](https://www.figma.com/file/dPGD3acNbT75dvEGzNaGUb/Responsible-Computing?t=zoMb9zxg25R0T6yt-6)

Gbaldassano:
@jhonatan-lopes to confirm if this still needs grooming after RCC meeting 5/3

jhonatan-lopes:
Yep, still needed and good to go

kevinhowbrook:
This can be reviewed here https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10582
Not entirely sure if it's what we need but I think it's in the ball park...

danielfmiranda:
Since this is more of a back-end change, there is not much to test besides making sure that everything is working/looking the same as before.

I have given the site and CMS a test and everything is looking good to me. Ready for prod!

PR #10582, opened 2023-05-04
Adds new abstract taxonomy base class

kevinhowbrook:
# Description

Link to sample test page: N/A
Related PRs/issues: #10422

# Checklist

**Tests**
- [ ] ~Is the code I'm adding covered by tests?~ No, any tests and factories will likely be made by taxonomies inherting this class.

**Changes in Models:**
- [ ] ~Did I update or add new fake data?~ No this is an abstract class
- [ ] ~Did I squash my migration?~ No - there are no migrations
- [ ] ~Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~ No migrations so this side is fine.

**Documentation:**
- [ ] Is my code documented?
- [ ] Did I update the READMEs or Wagtail documentation?


network-api/networkapi/wagtailpages/pagemodels/taxonomy.py
```diff
@@ -0,0 +1,31 @@
...
+            editing instances of this model.
+
+    Methods:
+        str(): Returns a string representation of the name of the taxonomy.
+
+    Note:
+        This model should be subclassed to create specific taxonomy models with additional fields.
+    """
+
+    abstract = True
```
jhonatan-lopes, line 23: I think this should be

```python
class Meta:
        abstract = True
```
otherwise it won't be picked up as abstract.

network-api/networkapi/wagtailpages/pagemodels/taxonomy.py
```diff
@@ -0,0 +1,31 @@
+from django.db import models
+from wagtail.admin import FieldPanel
+
+
+class BaseTaxonomy(models.Model):
```
jhonatan-lopes, line 5: It would be good to have the `TranslatableMixin` here, as I think all (most?) taxonomies are translatable.

jhonatan-lopes (changes requested):
Hi Kev, I think we also need to replace the models that are defining taxonomies to inherit from this base class, such as `ResearchRegion` and `ResearchTopic` in [taxonomies.py](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/research_hub/taxonomies.py). I'm not sure if the ticket was clear on this. Thanks

network-api/networkapi/wagtailpages/pagemodels/taxonomy.py
kevinhowbrook: :+1:  Added in 4ec53609107b1a6924596cd4972bfb0ed38bf084

network-api/networkapi/wagtailpages/pagemodels/taxonomy.py
```diff
@@ -0,0 +1,24 @@
+from django.db import models
+from wagtail.admin.panels import FieldPanel
+from wagtail.models import TranslatableMixin
+from wagtail.search import index
+
+
+class BaseTaxonomy(TranslatableMixin):
+    name = models.CharField(max_length=50, null=False, blank=False)
```
kevinhowbrook, line 8: I've left this as 50 just as that's what we have in the other taxonomies

jhonatan-lopes (approved):
Thanks Kev, I think this is ready to go!

We still have taxonomies on [buyersguide](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/buyersguide/taxonomies.py) and [blog](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/blog/blog_topic.py) but I think they differ too much from this abstract class to inherit from it. Perhaps it is better to leave them as they are.

Issue #10514, opened 2023-04-21
Update `BuyersGuideCallToAction`'s  "link_url" field to CharField


danielfmiranda:
## Description

At the moment, the `BuyersGuideCallToAction`'s `link_target_url` field is set to use Django's [`URLfield`](https://docs.djangoproject.com/en/4.2/ref/models/fields/#urlfield). 

This is useful if we want the CTA to link to another URL, such as "donate.mozilla.org".

However, since we have now updated the foundation site to render the donate modal whenever the UTM param `?form=donate` is appended to the current URL, it would be nice to have the CTA be able to append it as well.

Unfortunately, setting the field to `?form=donate` does not pass the URL fields validation, as it expects a full URL with `http://` included.

We can achieve this however by updating the `URLField` to `CharField` instead, and preforming the validation within the `clean()` method using a regex expression.

**We would need to make sure the attempted value falls under one of these 4 categories:**

1. The string is a valid URL. Example: `http://google.com/`
2. The string is a valid URL with query parameters. Example: `http://google.com/?form=donate`
3. The string is a valid query parameter. Example: `?form=donate`
4. The string is a valid group of query parameters. Example: `?form=donate&campaign_year=2022`

A regex expression that I found from a quick google search is:
```
^(http|https):\/\/[a-z0-9\-\.]+\.[a-z]{2,}(\/\S*)?(\?.+)?$|^(\?.+)$|^(\?.+)(\&.+=.+)*$
```

and the validation seems to work as expected.

## Dev Tasks:

- [ ] Update `link_target_url` from a `URLField` to a `CharField`, we are going to want to make sure that any existing data for this field gets carried over! 
- [ ] Add validation for this fields value in the models `clean()` method. An example of this being done can be found [here](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/2cd60a5f17913fb058cf8759bc494f507df5de46/network-api/networkapi/wagtailpages/pagemodels/blog/blog.py#L325). (**Note:** the example is preforming the validation within the `clean()` method directly, however, I would create a helper function to do the validation, and then call that within the clean method instead.
- [ ] Add some tests for the validation 

## Acceptance Criteria

- [ ] As a CMS user, I can set the `link_target_url` value to be any of the following formats:
  - Valid URL. Example: `http://google.com/`
  - Valid URL with query parameters. Example: `http://google.com/?form=donate`
  - Valid query parameter. Example: `?form=donate`
[...]

PR #10586, opened 2023-05-05
updated cta link_target_url to accept relative URLs

danielfmiranda:
# Description
Related PRs/issues: #10514

This PR updates the `BuyersGuideCallToAction` model's `link_target_url` field by changing it from a `URLField` to a `Charfield` with an additional regex validator. 

This regex makes sure that the attempted value for `link_target_url` meets one of these 4 formats:

1. The string is a valid URL. Example: “http://google.com”
2. The string is a valid URL with query parameters. Example: “http://google.com/?form=donate”
3. The string is a valid query parameter. Example: “?form_donate”
4. The string is a valid group of query parameters. Example: “?form=donate&campaign_year=2022”

Anything outside of these formats, will get a validation error with the message:

`Please enter a valid URL (starting with http:// or https://), or a valid query string starting with ? (Ex: ?form=donate)`

This PR also includes some tests for the validation, and the migration file needed for the update.

# Checklist

**Tests**


jhonatan-lopes (changes requested):
Thanks @danielfmiranda, the code looks good.

I'm a bit concerned about the number of tests, though. With regexes, we must be sure to properly test edge cases. Can we add more tests for should pass and should fail. There are some good examples on [this blog post](https://mathiasbynens.be/demo/url-regex).

danielfmiranda:
Thanks @jhonatan-lopes!

Using the link that you sent me for reference, I have updated the tests with a couple of more edge-case URLs to test against and re-requested review 👍

jhonatan-lopes (approved):
Awesome! Thanks Daniel!

Issue #10558, opened 2023-04-28
[FA] Implement design


mmmavis:
## Dev Tasks

- [x]  Work with designer to style the form.

## Acceptance Criteria

- [x]  Make sure form looks good on the "narrow" layout and the regular layout
- [x]  Make sure form looks good on different screen sizes

mmmavis:
Closing. PR has been merged into feature branch.

Issue #10556, opened 2023-04-28
[FA] Localization


mmmavis:
Localize form and validation. There are some limitation with FormAssembly which means we won’t be able to fully localize everything.

- Error messages are controlled by FormAssembly script. We can show localized error messages by loading the corresponding FormAssembly JS file (think about it like loading `en.js` vs `fr.js` vs `pl.js`).
    - Out of all languages we support on fo.mo `Frisian` and `Swahili` are currently not supported by FormAssembly.

## Dev Tasks

- [x]  Localization - form and validation should be localized as much as we can

## Acceptance Criteria

- [x]  As s site visitor, I'm able to see localized form and error messages (except for `Frisian` and `Swahili` )

mmmavis:
Closing. PR has been merged into feature branch.

PR #10589, opened 2023-05-05
FormAssembly localization and design work

mmmavis:
# Description

Link to sample test page: https://foundation-s-10556-1055-mvweww.mofostaging.net/en/campaigns/but-ahead-miss/?show_formassembly=true

## #10556

To review, please refer to the checklist in #10556 

## #10558 

To review, please refer to the checklist in #10558


mmmavis:
@nancyt1 Could you provide design feedback on the FA form? (Note that the after submission flow is not tied up yet and will be worked on in another PR). Here's[ the test page on the review app](https://foundation-s-10556-1055-nrhfmh.mofostaging.net/en/campaigns/initial-test-bannered-campaign-with-fixed-title/?show_formassembly=true). If you wanna test on other campaign pages you can simply append query param `show_formassembly=true` to the URL. The query param basically toggle the visibility of the new petition form -- the form on top is the FA form and the one below if our current form. This is an interim solution so we can easily compare the two version of forms during the transition/QA period. 

Please let me know if you have any q or the review app stops working. Thanks!

source/sass/formassembly-override.scss
```diff
@@ -0,0 +1,147 @@
+@import "mofo-bootstrap/colors";
```
mmmavis, line 1: Using SCSS for overrides instead of applying Tailwind classes directly to DOM elements. This way we won't have to re-apply classes every time we get a new version of code snippets from FA.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
```diff
@@ -1,3 +1,27 @@
+{% load i18n %}
+
+{% comment %}
```
mtdenton, line 3: This is awesome, thanks for this

mtdenton (approved):
Looks good, thanks for the hard work here!

PR #10598, opened 2023-05-10
[FRU] base and landing page

mmmavis:
Note: Original PR (#10541) had some weird CI issues. Creating this new PR seems to solve the problem.

---

- Implemented donate landing page + extra shared code to `base.html`, `nav.html`, and `footer.html`
- Nancy has [given her okay on the design](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10541#issuecomment-1540482875).

Link to test locally: `http://donate.localhost:8000/`
Related tickets: #10266, #10330

PR #10635, opened 2023-05-17
Localiza FRU page footer

TheoChevalier:
Quick follow up from #10598

I noticed this string was left unlocalized, I have no idea if it was done on purpose because it’s going to change or will be taken care of in a future PR, so feel free to close the PR if it’s the case!

Thanks!


mmmavis (approved):
Hello @TheoChevalier , this string slipped through the cracks. Thank you so much for catching and fixing it!

PR #10621, opened 2023-05-11
Split integration tests into two files

mmmavis:
Changed directory structure for integrations a bit.

From 
```
tests
├── integration.spec.js
├── ...
├── ...
├── ...

```
to

```
tests
├── integration
│   ├── foundation-homepage.spec.js
│   └── pni.spec.js
├── ...
```

Related PRs/issues: #10618 

### To test

1. on `main` branch, run `docker-compose up`
2. on a different terminal tab, run `npm run playwright:integration`
3. observe results
4. switch to this PR, run `docker-compose up`
5. on a different terminal tab, run `npm run playwright:integration`
6. observe results to make sure it matches what you see on Step 3
7. If results are the same, PR is good to go!

PR #10641, opened 2023-05-17
Fix package.json with the new path to Playwright integration tests

mmmavis:
Follow-up for https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10621. 

In #10621 I forgot to also update the "playwright" task to point to the new directory we created. Filing this PR to fix that.

## To QA

- have `docker-compose up` running on one tab
- run `npm run playwright` on another tab
- check the result 
  - if it's `1 skipped 1 passed` then we are good 👍🏻 
  - if you see "0 tests found" then this PR doesn't have the right fix 👎🏻


danielfmiranda (approved):
Tested and approved! Thanks @mmmavis 👍

PR #10701, opened 2023-06-08
Revert "Commented out a FormAssembly test that's failing until Mavis gets back"

mmmavis:
Reverts MozillaFoundation/foundation.mozilla.org#10698

I ran all tests locally and on a different PR, all of them passed. I think the error @jhonatan-lopes was seeing was most likely due to intermittent slow response from FA. I believe we can safely revert the code changes from the previous PR.


mtdenton (approved):
LGTM!

PR #10744, opened 2023-06-16
Removed reference to old redirect static file

danielfmiranda:
# Description

Link to sample test page: https://foundation-s-redirect-p-g4zvyb.herokuapp.com/cms/redirects/
Review app admin credentials: Admin2/Admin2

This PR removes a reference to an old static file related to the wagtail redirect page that no longer exists as it is no longer needed.

The reference to a non-existent file produced a 500 error whenever trying to visit the redirect page in the CMS on either prod or staging.

## Steps to test:

1. Visit the review app link above, with the credentials "Admin2/Admin2".
2. The page should render as expected.
3. If everything seems to be working as it should, testing is complete!


mtdenton (approved):
LGTM!

PR #10065, opened 2023-01-27
Automatic redirects failing

kevinhowbrook:
# Description

Adds tests to prove redirects are working with pages moving.
And hopefully... a fix if not

- Checked page types that inherit routable mixin
- Checked for save method overrides on testing pages
- Checked the custom middleware at _new_get_redirect override
- Checked with staging data (error occurs)
- Write tests for redirect creation on page move: **done, passes**
- Write tests for redirect creation on page move with translated pages / extra locales: **done, passes**
- Write tests that include multisite config: **done, passes**

Link to sample test page:
Related PRs/issues: #9961

# Checklist

**Tests**
- [x] Is the code I'm adding covered by tests?

**Changes in Models:**
- [ ] Did I update or add new fake data?
- [ ] Did I squash my migration?
- [ ] Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [ ] Is my code documented?
- [ ] Did I update the READMEs or wagtail documentation?

PR #10745, opened 2023-06-16
Fix CSP in redirects

jhonatan-lopes:
# Description

After latest prod push, redirects page is not working, showing `Content Security Policy: The page's settings blocked the loading of a resource at inline ("script-src").`.

This was tracked to #10065, where a custom template for `wagtailredirects/index.html` was introduced, replacing the one defined by the `redirects` module in Wagtail. The issue is that this template was extracted from the 2.11 version, which is not the current version in the repo (4.1.5).

After updating the template, the presence of a nonce in the script in the template made the other scripts not execute (since they did not define a nonce value themselves). The adopted solution was to keep the Wagtail admin consistent with the current code base and remove the custom template -- removing the nonce and the customizations fixed the CSP issue. 

Link to sample test page:
Related PRs/issues: #10065 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?


danielfmiranda (approved):
Thanks @jhonatan-lopes! LGTM!

Issue #10736, opened 2023-06-12
Libraries author detail page not showing default profile picture


jhonatan-lopes:
### Describe the bug

When an author does not have a profile picture, a default profile picture is shown:

![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/04477d75-bdac-4317-a3ee-2f51ef3bde8b)

![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/65d7b4eb-65a7-4025-b896-c72bceb1a9d6)

When going to the detail page of the author, if it has a profile picture, it will be displayed:

![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/175d99e0-5cb1-4288-bf44-5640699c09fd)

However, profiles without pictures do not display the default picture on the detail page:

![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/69986b81-1283-409a-a7cf-c24de32d1195)

### To Reproduce
1. Run `inv new-db`
2. On the CMS, delete the profile picture for one of the profiles that have a Research Hub article associated with them
3. Go to `en/research/authors`
4. Click the profile that has a default profile picture
5. See error

### Expected behavior
RCC/Research Hub author detail page should show default profile picture when there is no picture associated with it.

PR #10772, opened 2023-06-23
Add default picture to author detail page

jhonatan-lopes:
# Description

Fixes bug where the default profile picture is not shown on authors detail page.

Link to sample test page:
Related PRs/issues: #10736


jhonatan-lopes:
Authors index page:
image

Author's detail page:
image

@beccaklam I think we need a higher resolution default picture version for the detail page, as the one we have right now is only 100x100 pixels, so it is quite poor quality for big screens.

beccaklam:
@jhonatan-lopes Agreed, I'll take a look and see if I can find the original file today.

beccaklam:
Hi @jhonatan-lopes I found the original file. I've exported the SVG at a larger size, let me know if this is big enough? (It looks pixelated below but perhaps it will work for the size on RCC).
author-fallback-image

jhonatan-lopes:
Thanks for sending this through @beccaklam! Here's how it looks now:
image

danielfmiranda (approved):
Looks great! LGTM!

PR #10802, opened 2023-06-28
Make sure 'formassembly_head.html' is also included on PNI campaign pages

mmmavis:
Reasons of why [PNI campaign pages](https://foundation.mozilla.org/en/privacynotincluded/articles/tell-tiktok-stop-misleading-android-users-and-provide-full-transparency-about-your-privacy-practices/?show_formassembly=true) aren't showing FA form correctly is because `formassembly_head.html` isn't included on PNI campaign pages. This PR fixes that.

boxborders

---

Test page on review app: https://foundation-s-fa-include-kbitxl.mofostaging.net/en/privacynotincluded/articles/test-pni-campaign-pages/


mtdenton (approved):
Good to go!

Issue #10472, opened 2023-04-07
[FA] Clean up existing petition code


mmmavis:
We've decided to go with the HTML embed option with FormAssembly (for context see thread in #10468). Form submission will be handled by FormAssembly's own JavaScript code. This means once we have verified the new form is working, we can then remove existing petition related code.

## Acceptance Criteria
- [x] remove obsolete JS/JSX code
- [x] remove obsolete template code
- [x] remove obsolete SCSS code
- [x] remove petition related backend code that is no longer in use
- [x] remove old test code (if any)
- [x] review README and update it accordingly 

## To-Do on Staging and Prod
- [x] staging: remove old env vars from Heroku
  - [x] `CRM_AWS_SQS_ACCESS_KEY_ID`
  - [x] `CRM_AWS_SQS_REGION`
  - [x] `CRM_AWS_SQS_SECRET_ACCESS_KEY`
  - [x] `CRM_PETITION_SQS_QUEUE_URL`
- [ ] prod: remove old env vars from Heroku
  - [ ] `CRM_AWS_SQS_ACCESS_KEY_ID`
  - [ ] `CRM_AWS_SQS_REGION`
  - [ ] `CRM_AWS_SQS_SECRET_ACCESS_KEY`
  - [ ] `CRM_PETITION_SQS_QUEUE_URL`

Gbaldassano:
TBD on sizing - discuss when @mmmavis returns

mmmavis:
These env vars have been removed from **staging**:

- `CRM_AWS_SQS_ACCESS_KEY_ID`
- `CRM_AWS_SQS_REGION`
- `CRM_AWS_SQS_SECRET_ACCESS_KEY`
- `CRM_PETITION_SQS_QUEUE_URL`

I'm moving this ticket to `ready for prod` state!

mmmavis:
These env vars have been removed from **prod**:

- `CRM_AWS_SQS_ACCESS_KEY_ID`
- `CRM_AWS_SQS_REGION`
- `CRM_AWS_SQS_SECRET_ACCESS_KEY`
- `CRM_PETITION_SQS_QUEUE_URL`

PR #10850, opened 2023-07-11
Remove old petition code

mmmavis:
Related PRs/issues: #10472

There are two commits - first is for removing frontend code & second one is for removing backend code.

No test page for this PR. This PR is to delete code that is no longer in use. The [Playwright test](https://github.com/MozillaFoundation/foundation.mozilla.org/actions/runs/5516161815/jobs/10057224593?pr=10850#step:11:1871) set up for the current petition flow passed on this PR so there's no worry this PR can break the current flow.

For more context:

- the new petition flow (with FormAssembly) does not rely on our backend to send data (both petition and newsletter signup)
- form markup and front end logic are copied from FormAssembly's code snippets.


danielfmiranda (approved):
Thank you for the explanation in the ticket description @mmmavis!! This looks good to me 👍 approved!

Issue #10874, opened 2023-07-13
FormAssembly doesn't include asterisks for required fields in their embed code


mtdenton:
Stakeholder: Christina Hug

Original request:

> Looking for a way to indicate that fields are required in Form Assembly. Right now there is no indication and you get errors if you missed one - want that indicator up front? And if not can we just add a text asterisk beside the field names?

Let's go ahead and add asterisks to the embed code in the petition forms along with an indicator of what the asterisks mean.

Mock-up:
![requiredfield](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/41487159/ce135864-97a0-4d74-bde2-4727a5226f7f)

The embed code for the form lives in [`network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html`](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/39c38dcf3f1477273dbd5d11fc0a14aa0428c40f/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html#L4)

### Dev Tasks
- [ ] Add `*` to the following fields: `First Name`, `Last Name`, `Email`, `I’m okay with Mozilla handling my info as explained in this Privacy Policy`
- [ ] Add `* indicates a required field` above all fields at the top of the form
- [ ] Style all text to match the current text within the form

### Acceptance Criteria
- [ ] As a user, I see asterisks on the labels of the required fields on page load for all languages of a campaign page with a petition CTA.
- [ ] As a user, I see `* indicates a required field` above the fields on page load for all languages of a campaign page with a petition CTA. The text is presented in the corresponding language.

kristinashu:
Note, talk around the design is happening in the PR https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10875

danielfmiranda:
Hi @mtdenton @Gbaldassano! 

The PR for this ticket has now been merged into main and is ready for QA on staging. 

A test page can be found here: https://foundation.mofostaging.net/en/campaigns/tell-slack-make-dms-safe/

Thanks!

mtdenton:
Checked on three languages, also saw `aria-required`. Looks good!

PR #10875, opened 2023-07-13
updated form assembly template to include asterisks on required fields

danielfmiranda:
# Description

Link to sample test page: https://foundation-s-10874-fa-r-kocj6k.mofostaging.net/en/campaigns/single-page/
Related PRs/issues: #10874

This PR updates `formassembly-override.scss` in order to remove the rule that hides the asterisk from fields marked with the class `reqMark`. 

Doing so allows a red asterisk to render to the right of the label for fields that we have marked as required.

This PR also adds an additional label at the top, with the string `* indicated a required field`.

# Screenshots

**Mockup:**

image

**Review App:**

Screenshot 2023-07-13 at 17-20-12 Tell Slack Make DMs Safe


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
```diff
@@ -308,7 +310,7 @@
...
+                <label id="tfa_492-L" class="label preField" aria-label="Privacy Agreement  required">Privacy Agreement</label><br><div class="inputWrapper"><span id="tfa_492" class="choices vertical required" aria-required="true"><span class="oneChoice"><input type="checkbox" value="tfa_493" class="calc-DefaultName" id="tfa_493" name="tfa_493" aria-labelledby="tfa_493-L" data-tfa-labelledby="tfa_492-L tfa_493-L"><label class="label reqMark postField" id="tfa_493-L" for="tfa_493"><span class="input-checkbox-faux"></span>{% blocktrans with link="https://www.mozilla.org/privacy/websites/" %}I’m okay with Mozilla handling my info as explained in this <a href="{{ link }}" target="_blank">Privacy Notice</a>.{% endblocktrans %}</label></span></span></div>
```
danielfmiranda, line 313: Moving the class `reqMark` to this element allows the asterisk to render to the right of the label.

danielfmiranda:
Hi Everyone! This PR is now ready for review and for testing on the review app.

However, I do have two small questions:


1. @nancyt1 @mtdenton By default, form assembly renders the "required field" marker as a red asterisk. However, I see in the mockup they are grey. Would we like to stick with the default, or should I override them with CSS to make them the same color as the text? (hex#444)
2. @mtdenton By default, formassembly hides the asterisk once the checkbox is checked. Is this OK? Or, would you like me to override the template to always have a asterisk next to it. 

Thanks in advance!

mtdenton:
>     2. @mtdenton By default, formassembly hides the asterisk once the checkbox is checked. Is this OK? Or, would you like me to override the template to always have a asterisk next to it.

Thanks @danielfmiranda! Honestly that feels fine IMO

kristinashu:
Curious what @nancyt1 things but when we talk about this at crit we felt "* indicates a required field" wasn't even really necessary - the asterisk is very common convention and the extra text adds extra busyness/distraction and pushed the form down lower.

kristinashu:
I like the approach of making it smaller! 

cc @mmmavis

mmmavis:
Including a red asterisk and adding `* indicates a required field` prior to the form fields sounds good. Thanks team!

danielfmiranda:
Thanks all! 

I have updated the styling of the label to use the `tw-body-small` and `tw-italic` classes. Here is a screenshot of what it looks like:
Screenshot 2023-07-17 at 12-53-12 Tell Slack Make DMs Safe

nancyt1 (approved):
I think this looks good! One note tho, on the tablet size (on my desktop browser view) the width of the form changes size slightly when you check and uncheck the required box due to the asterisk getting hidden. It's not super urgent but it would be better if we can keep things static.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
```diff
@@ -46,6 +50,8 @@
             wFORMS.behaviors.condition.isConditionalSubmitAllowed = true;
         </script></div>
         <form method="post" action="https://mozillafoundation.tfaforms.net/api_v2/workflow/processor" class=" labelsAbove" id="9" role="form">
+            <label class="label tw-body-small tw-italic tw-mt-2">{% blocktrans %} * indicates a required field {% endblocktrans %}</label>
```
mmmavis, line 53: Hi @danielfmiranda after looking at this PR closely, I'm thinking it might be better to move this line outside of this file (i.e., move to `petition.html` before the `{% include ...` on line 28) because

1. This line is a completely new element we are adding to the page and not something came with the code snippets from FormAssembly
2. Including this line here means every time we update the FA form we'll have to re-add the `* indicates a required field` note in this file which creates more code maintenance overhead.

I would also suggest that we switch to use `<p>` instead of `<label>` (we probably need to adjust the vertical margin a bit as well)

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
mmmavis: Could you add a new note (probably as Step 6) at the top of the file so we remember to make this change in future updates? Thanks!

mmmavis (changes requested):
@danielfmiranda Thanks for taking care of this PR. I have left some comments. FormAssembly related bits in this codebase are not straightforward to understand so please let me know if I can help explain anything further.

danielfmiranda:
Thanks for the feedback @mmmavis! I have updated this PR taking your feedback into account and also have left a couple questions.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
```diff
@@ -24,17 +24,25 @@
...
     - Localize "value" attribute for <input type="submit" ...>
     - Go to formassembly_head.html and make sure everything has been updated according to the instructions
 
-7. Update the revision note on the next line. Revision number can be found on https://mozillafoundation.tfaforms.net/versions/index/9
+8. Update the revision note on the next line. Revision number can be found on https://mozillafoundation.tfaforms.net/versions/index/9
 
 The code snippet below is based on FormAssembly form revision #32
 
+*** NOTE *** 
+We have added a custom label that states "* indicates a required field" to the template.
```
danielfmiranda, line 43: Hi @mmmavis I was wondering if you thought it would make more sense to leave this comment here, or in `petition.html` where the label actually is being defined.

My thinking is that it makes more sense here since we are telling the user that they should take the label into account when making changes to the form/template, but was curious to hear what you think is best.

Thanks!

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
danielfmiranda: Hi Mavis! I have added a new comment [here](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/7038fa4851cfb5b4b1beb852d80cfc0983d013c6/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html#L27-L29).

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
danielfmiranda: Thanks for the feedback Mavis! Moving it outside of the template makes sense to me so I have done that in the latest commit. 

I also updated the `<label>` to a `<p>`, here is a screenshot confirming that visually nothing has changed except the vertical spacing of the element (made a small change from `tw-mt-2` to `tw-my-4`.)

Screenshot 2023-07-17 at 16-19-35 Tell Slack Make DMs Safe

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html
mmmavis: Thanks for sharing your thoughts. Leaving the note here sounds good. For extra clarification, I'm wondering what you think about updating this line to

```suggestion
We have added a custom label that states "* indicates a required field" to the template (i.e., in `petition.html`). 
```

danielfmiranda: That sounds great to me! Updated 👍

mmmavis (approved):
🎉 awesome work!

Issue #10579, opened 2023-05-03
[FA] [Post-Launch]  campaign_id should be a required field


mmmavis:
Ticket will need to be groomed.

In [campaigns.py](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/campaigns.py), `campaign_id` isn't a required field on various classes. I'm not sure if somewhere in the old pipeline we had a default value set up for campaign id, but it seems like we'll need turn it into a required field. 

From Mary Helen,
>  The campaign id needs to be sent to Salesforce because without that, Salesforce won't have a way to match to an existing Campaign and then create the corresponding Campaign member record for the person filling out the petition. 

## Dev Task
- [ ] find out what petition/signup/callpower snippets have an empty `campaign_id` and decide what to do with them
- [ ] turn `campaign_id` into a required field

mtdenton:
@mmmavis Wow, super surprised that it isn't already a required field. Thanks for calling this out! I'd say let's try to make this change as the very last step before we launch if possible (if we have time, we may not have that luxury). Otherwise, we can launch without making this a required field and this needs to be the very next ticket that we work after launch and we'd need to get this change to prod ASAP or we risk losing signatures on any new petitions that might get created.

@Gbaldassano I don't know if there's any petitions planned for late May/early June, but we will want to send out a PSA about this if there's not time to fix this before we're forced to cut over to FA. Basically, in the week following the switch, stakeholders *must* fill out the Campaign ID field when they're making a petition in the Wagtail dashboard or they will lose signs on their petitions. If we can't squeeze this in, a fix will come ASAP after launch making it where the user will get an error if they try to save the petition without filling out the Campaign ID field.

Gbaldassano:
Thank you for flagging @mtdenton & @mmmavis: a few clarifying questions before we give stakeholders a heads up:

- Has not having the campaign ID field filled out not been an issue in the past? Is this a new requirement that FA is adding to the equation, or have we always needed to fill out the campaign ID field and just haven't? 
- How has SF been tracking Campaign ID for petitions in their current state?
- How will this switch over affect existing petitions that don't have the campaign ID filled out? Will we need to go back and ask the team to fill out this field for pre-existing petitions?

@maryhelenyj FYI

maryhelenyj:
[...]

Issue #10846, opened 2023-07-10
Make petition modal CTA URL editable


mtdenton:
We want to be able to update the URL for the purple "Yes, I'll Chip In" button in the CMS

![donatemodal](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/41487159/7a0433be-61fe-4465-a93b-6da9e092cec8)

Gbaldassano:
Based on discussion of #10905, we should move forward and merge this when ready but not release to users until we have more information about how best to set up.

Gbaldassano:
Based on discussion of #10905, we should move forward and merge this when ready but not release to users until we have more information about how best to set up.

PR #10892, opened 2023-07-18
Made petition campaign_id a required field

danielfmiranda:
# Description

Related PRs/issues: #10579

**This PR does the following:**
- Makes the petition snippet's `campaign_id` field required. If a user tries to create or edit a petition without filling out this field, they will recieve a validation error.
- Updates `petition-thank-you.jsx` to pass the newly required `campaign_id` as a prop. The campaign ID is then appended to the donate modal's "donate_url". (Ex: `?form=donate&c_id=<id>`). That way FRU can use this ID for reporting purposes.
- Updates `002-thank-you.spec.js` to account for newly appended URL parameter.

# Screenshots:

**URL when hovering over "Yes, Ill chip in" Button:**

Screenshot 2023-07-18 at 5 34 04 PM

**GTM FRU Donate Modal opened event:**

image

# Steps to test

## Testing that campaign_id is required

1. Copy the prod db on your local machine
2. Visit http://localhost:8000/cms/snippets/wagtailpages/petition/edit/306/
3. Delete what is set for `campaign_id` and click "Publish"
4. You should receive a validation error stating that Campaign ID is required

## Testing that the campaign_id is being sent to FRU/Salesforce
1. Check this branch out on your local machine
5. run `inv copy-prod-db`
6. Log into [google tag manager](https://tagmanager.google.com/#/container/accounts/6002697367/containers/36502666/workspaces/22) and click "preview"
7. In the URL box, enter http://localhost:8000/en/campaigns/tell-slack-make-dms-safe/?thank_you=true
8. On the new tab that appears, hover over the "Yes Ill chip in" button, note the `c_id` that is shown in the URL preview at the bottom left of the window.
9. Click the "Yes, I'll chip in" button to render the modal.
10. Go back to the GTM screen, and click the event " Event: FundraiseUp.checkoutOpen" on the left hand side of the screen, under "Summary".
11. Expand the details, the c_id should match the c_id that was shown on step five.

Screenshot 2023-07-19 at 11 04 00 AM

9. If everything is working as expected, testing is complete!


danielfmiranda:
Hi @mtdenton! 

In this PR I am appending the campaign ID parameter as `campaign_id` however after looking further into [FRU documentation ](https://fundraiseup.com/support/utm-parameters/)and [Salesforce Documentation](https://help.salesforce.com/s/articleView?id=sf.mc_gai_utm_parameters.htm&type=5), it appears that they are both referencing the parameter as `utm_campaign`. 

I was wondering if you knew who I could reach out to to confirm what the param should be appended as? Thanks!

source/js/components/petition/donation-modal.jsx
```diff
@@ -85,6 +86,8 @@ class DonationModal extends Component {
     if (!this.donateURL) {
```
danielfmiranda, line 86: There seems to be no reference in the code base regarding donateURL as a prop. So at the moment this link is always defaulting to `form=donate`. However, I am leaving this here as I can use this prop while implementing #10846

mtdenton:
Hey Daniel! The UTM campaign and the Salesforce campaign are two separate things to track and send along here. The expectation is that you would have a Salesforce campaign ID to send through to the FRU modal 100% of the time. You would also send `utm_campaign` along with any other accompanying `utm_xx` parameters that come through on the page's URL if they're present.

Just looking at the above screenshot as well, I'm seeing `c_id` and `campaign_id` and they both have some sort of Salesforce-style ID. Where is `c_id` coming from?

danielfmiranda:
Great point @mtdenton I did not catch that! 

After taking a look inside FRU, it looks like that is a parameter that is set through FRU itself for our campaign titled "Mofo Main":

Screenshot 2023-07-19 at 10-57-02 Fundraise Up

tests/integration/petition/002-thank-you.spec.js
```diff
@@ -44,7 +44,7 @@ test.describe("Donation modal", () => {
   test("Donation modal can trigger FRU widget", async ({ page }) => {
     // test if FRU iframe pops up after clicking the Yes button
     const yesDonateButton = page.locator(
-      `.modal-content .tw-btn-primary[href="?form=donate"]`
+      `.modal-content .tw-btn-primary[href*="?form=donate&c_id="]`
```
danielfmiranda, line 47: Making sure that the donate URL includes the `c_id` URL param.

mtdenton:
Aha, good find @danielfmiranda. Let's get some stakeholder input here to make sure that's the intent. I'll have an answer soon

danielfmiranda:
Hi @mtdenton

I have confirmed with Esteban that parameter `c_id` is in fact what we are looking for. Now submitting this PR for review!

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -171,7 +171,6 @@ class Petition(TranslatableMixin, CTA):
         max_length=20,
         help_text="Which campaign identifier should this petition be tied to?",
```
mmmavis, line 172: Since we are updating this field, could we also update the help text to have "Salesforce" for extra clarity? e.g., something like `"Which Salesforce campaign identifier should this petition be tied to?"`
```suggestion
        help_text="Which Salesforce campaign identifier should this petition be tied to?",
```

source/js/components/petition/donation-modal.jsx
mmmavis: Thanks for the note. Either in this PR or in #10846, could we add a method for generating `donateURL`? Right now the variable name `donateURL` could be a bit misleading as it's default to a query string with `?` and later appended with more query params. I think using [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams) would probably be a good/cleaner choice.

mmmavis (changes requested):
Thanks for the inline comments, @danielfmiranda .It makes PR review a lot easier! I left some comments but they are mostly for adding extra clarity. We'll also need to update `PETITION_TEST_CAMPAIGN_ID` to have a non-empty default value since we are changing the field to be required now :)

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
danielfmiranda: Thanks @mmmavis! This is a great suggestion and has been implemented now.

danielfmiranda:
Thanks for the feedback @mmmavis! I have updated the help text to include "Salesforce" in the sentence now, and have also updated the env.default file to include the sandbox Salesforce campaign ID that was discussed in Slack.

Regarding the feedback about the `donateURL` I would vote to skip it in this PR, and I will take care of it in #10846 as I am going to be making changes to the donate URL in that ticket.

I have now requested review, thanks again!

env.default
```diff
@@ -62,7 +62,7 @@ CSP_INCLUDE_NONCE_IN=script-src
 
 
 # TEST ENVIRONMENT VALUES
```
mmmavis, line 64:
```suggestion
# Petition test campaign id for Salesforce Sandbox
```

network-api/networkapi/wagtailpages/migrations/0094_make_petition_id_required.py
```diff
@@ -0,0 +1,22 @@
...
+    dependencies = [
+        ("wagtailpages", "0093_auto_20230726_2248"),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name="petition",
+            name="campaign_id",
+            field=models.CharField(
+                help_text="Which Salesforce campaign identifier should this petition be tied to?",
```
mmmavis, line 17:
```suggestion
                help_text="Which Salesforce Campaign ID should this petition be tied to?",
```

Per new request on dev Slack channel

network-api/networkapi/wagtailpages/pagemodels/campaigns.py
```diff
@@ -169,9 +169,8 @@ class Meta:
 class Petition(TranslatableMixin, CTA):
     campaign_id = models.CharField(
         max_length=20,
-        help_text="Which campaign identifier should this petition be tied to?",
+        help_text="Which Salesforce campaign identifier should this petition be tied to?",
```
mmmavis, line 172:
```suggestion
        help_text="Which Salesforce Campaign ID should this petition be tied to?",
```

mmmavis (approved):
PR looks good! There are a few quick changes we can make to add more clarity around petition flow. I've left code suggestions inline. Feel free to modify them or just commit my suggestions. Either way we should squash commits for merging so it doesn't create issues with migration file.

Thanks @danielfmiranda !

PR #10897, opened 2023-07-20
10846 make donate modal url editable

danielfmiranda:
# Description

**Note**: This PR depends on #10892, please merge first and then take a look at this one.

Link to sample test page:
Related PRs/issues: #10846

This PR updates the model for the donation modal which appears after petition submission with a new field `donate_url`. 

This `donate_url` field allows users to enter a custom URL (Ex: http:// or https://) **or** relative URL (Ex: ?form=donate) if desired. 

If left blank, the donation modal will link to the same default donation form as before ("?form=donate").

This PR also adds tests for the donation modal's URL validation via the file `test_donation_modal.py`

# Steps to test

1. Check this branch out
2. Run `inv copy-prod-db`
3. Visit http://localhost:8000/cms/snippets/wagtailpages/donationmodal/edit/1/
4. Scroll down to "Donate Url" , enter an invalid URL (Ex: "test"), click "Publish", and note that you get a validation error.
5. Enter a valid URL, and click "Publish".
7. Visit [this campaign page](http://localhost:8000/en/campaigns/tell-slack-make-dms-safe/?thank_you=true) and click the purple "Yes, I'll chip in" button
8. If you are linked to the expected URL, testing is complete! 

# Screenshots

### CMS
Screenshot 2023-08-02 at 15-49-43 Editing donation modal - privacy_donate_modal - Wagtail

### Button Link
Screenshot 2023-08-02 at 3 50 11 PM

### CMS
Screenshot 2023-08-02 at 15-52-17 Editing donation modal - privacy_donate_modal - Wagtail

### Button Link
Screenshot 2023-08-02 at 3 53 01 PM

### CMS
Screenshot 2023-08-02 at 15-53-22 Editing donation modal - privacy_donate_modal - Wagtail

### Button Link
Screenshot 2023-08-02 at 3 53 52 PM

## Invalid URL (Neither Full URL or Relative)
Screenshot 2023-08-02 at 16-00-12 Editing donation modal - privacy_donate_modal - Wagtail


network-api/networkapi/wagtailpages/constants.py
```diff
@@ -0,0 +1,2 @@
+# Validates whether a string is either a valid URL, a query string (?param=test), or both.
+url_or_query_regex = r"^(https?://[\w.-]+(/\S*)?)?(\?[\w-]+(=[\w-]*)?(&[\w-]+(=[\w-]*)?)*)?$"
```
danielfmiranda, line 2: I moved this to a constants file because we are also using the same regex expression for the same type of validation [here](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10514). I am also opening this issue so we can discuss if this is the best place to store the regex moving forward, as I am assuming more and more fields are going to want the option to link to relative URL's such as `?form=donate`!

network-api/networkapi/wagtailpages/donation_modal.py
```diff
@@ -35,6 +38,25 @@ class DonationModal(TranslatableMixin, models.Model):
...
+                    "Please enter a valid URL (starting with http:// or https://), "
+                    "or a valid query string starting with ? (Ex: ?form=donate)."
+                ),
+            ),
+        ],
+        help_text=(
+            "If you would like this modal to link to a custom URL, "
+            "please enter a valid URL (starting with http:// or https://), "
+            "or a valid query string starting with ? (Ex: ?form=donate)."
+        ),
```
danielfmiranda, line 57: This is the help text that is going to appear in the CMS

network-api/networkapi/wagtailpages/donation_modal.py
```diff
@@ -35,6 +38,25 @@ class DonationModal(TranslatableMixin, models.Model):
...
+    donate_url = models.CharField(
+        max_length=255,
+        blank=True,
+        validators=[
+            RegexValidator(
+                regex=url_or_query_regex,
+                message=(
+                    "Please enter a valid URL (starting with http:// or https://), "
+                    "or a valid query string starting with ? (Ex: ?form=donate)."
+                ),
```
danielfmiranda, line 50: This is the red validation message that will render only if the user enters a invalid value.

source/js/components/petition/donation-modal.jsx
```diff
@@ -137,6 +135,19 @@ class DonationModal extends Component {
     );
   }
 
+  generateDonateButtonUrl() {
```
danielfmiranda, line 138: Hi @mmmavis!

I updated this function based on the feedback that you left [here](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10892#discussion_r1280940102).

I was wondering if you could take a look and let me know what you think?

It is working well for me, the only question that I have is this:

When a user enters a relative URL (ex: `?form=test`), this method generates a full URL (ex: `http://<url_of_current_page>/?form=test`. This means that a user clicking the button will trigger a page refresh.

Is this OK? 

Thanks!

network-api/networkapi/wagtailpages/donation_modal.py
```diff
@@ -35,6 +38,25 @@ class DonationModal(TranslatableMixin, models.Model):
...
+            ),
+        ],
+        help_text=(
+            "If you would like this modal's donate button to link to a custom URL, "
```
danielfmiranda, line 54: Hi @mmmavis, here is the updated help text! It now mentions that the url is for the modal's donate button

network-api/networkapi/wagtailpages/donation_modal.py
```diff
@@ -35,6 +38,25 @@ class DonationModal(TranslatableMixin, models.Model):
         default="Yes, I'll chip in",
     )
 
+    donate_url = models.CharField(
+        max_length=255,
+        default="?form=donate",
```
danielfmiranda, line 43: Now making the donate_url field required, with the default of `?form=donate`.

Note: since this is a new field, no custom migration is required!

source/js/components/petition/donation-modal.jsx
mmmavis: Hey @danielfmiranda , thanks for calling this out. In that case I think we can modify the code so that before `return`, we massage the URL string again.

We can update `const url = new URL(this.props.donateUrl, window.location.href);` to `const url = new URL(this.props.donateUrl);`

And instead of returning `url.toString();`, we can do something like 

```js
const current = new URL(window.location.href);
const baseUrl = current.origin + current.pathname;

return url.toString().replace(baseUrl, "");
```

mmmavis: I'm also curious if Python/Django has built-in functions for generating URL like this?  If so, we should probably utilize that.

mmmavis (changes requested):
@danielfmiranda sorry for the wait! I left comments regarding your question for `generateDonateButtonUrl` function :)

danielfmiranda:
Hi @mmmavis thank you for the review and for the feedback! 

After taking a look at our options on the python/django end, in my opinion using JS still seems to be the easiest.

I have updated the code with your suggestions, and can confirm that it is working great 🙌 

The only difference is that I left the first line as `const url = new URL(this.props.donateUrl, window.location.href);`. 

If I remove the 2nd parameter like suggested, the JS then returns an error when trying to use relative URLs (ex: "?form=donate is not a valid URL").

However after implementing your other suggestions and testing full URLs, full URLs with parameters, and relative URLs, I can confirm that they are all working as expected!

mmmavis (commented):
Hi @danielfmiranda , I think I found the issue with page reload.

Say if the page is on `https://test.example.com/en/campaigns/single-page/?thank_you=true` (notice the trailing slash before `?`) and on CMS, user inputs the following as donate url: `https://test.example.com/en/campaigns/single-page?hello=world&form=donate` (notice the absence of the slash before `?`), then clicking on the button will trigger page reload.

I'm thinking we can leave this as is because the trailing slash could be intentionally entered by users. What do you think?

danielfmiranda:
Hi @mmmavis thanks for taking a look!  I now see why the page is not reloading on my end but it is on yours.

I see in your example, where the page's URL is `https://test.example.com/en/campaigns/single-page/?thank_you=true`, you are updating the button to append UTM parameters by setting the URL to a full URL `https://test.example.com/en/campaigns/single-page?hello=world&form=donate`.

However, this is not necessary as using a relative URL (ex: `?hello=world&form=donate`) to append query parameters will suffice. 

Adding a full URL (with http/https included) is only necessary when the user wants to link to either a new page on the site, or, a new page somewhere else.

mmmavis (approved):
We are ready to land this awesome work. Thanks @danielfmiranda !

Issue #10854, opened 2023-07-11
[Report] Add latest publication info to Page Types report


jhonatan-lopes:
Add information about the latest page published under each Page type to the report, changing it to confirm with the following structure:

| Page Type  | Number of Pages |Last published at | Last published page | Latest changes by |
| --- | --- | --- | --- | --- |
| Blog Detail Page  | 20  | January 1st, 1970 | `url.to.the.page.com` | J. Doe |
| RCC Detail Page  | 50  | January 1st, 1971 | `url.to.the.page.com` | J. Doe |

## Acceptance Criteria

- [ ] As an editor, I can see a **Pages** report with:
  - [ ] Wagtail page type
  - [ ] Number of pages under that type
  - [ ] Last page published under that page type
  - [ ] When the last page was published
  - [ ] Last person that published a page under that page type

Issue #10853, opened 2023-07-11
[Report] Page Types Report


jhonatan-lopes:
Implement a Wagtail report with the following structure:

| Page Type  | Number of Pages |
| --- | --- |
| Blog Detail Page  | 20  |
| RCC Detail Page  | 50  |

## Acceptance Criteria

- [ ] As an editor, I can see a **Pages** report with:
  - [ ] Wagtail page type
  - [ ] Number of pages under that type

PR #10928, opened 2023-07-25
Page types report

jhonatan-lopes:
# Description

Add a Page types report.

Covers tickets #10853 and #10854.

Link to sample test page: "cms/reports/page-types-report/"

To test:

1. Login to admin
2. Go to reports
3. Go to "Page Types"
4. See report

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**


network-api/networkapi/reports/views.py
```diff
@@ -0,0 +1,68 @@
...
+        return (
+            ContentType.objects.filter(model__in=page_models)
+            .annotate(
+                count=Count("pages"),
+                active_locale_count=Count("pages", filter=Q(pages__locale=self.active_locale)),
+                last_edited_page=Subquery(latest_edit_log.values("page")[:1]),
+                last_edited_by=Subquery(latest_edit_log.values("user")[:1]),
+            )
+            .order_by("-count")
+        )
```
jhonatan-lopes, line 61: I'm not particularly fond of these queries. The subqueries are quite expensive (they take up to more than a second on my machine with production data) and make the page load slowly.

However, my attempts to optimise the query have been in vain. In addition, this is the most readable form of the query and, I think, the form that takes the least amount of database queries.

We can use less expensive queries by prefetching the `pages` on `ContentType` and then doing another query to get the users from that through `PageLogEntry` in `decorate_paginated_queryset`, but that only trades the database bottleneck for a CPU bottleneck.

jhonatan-lopes: Thanks Tom, I think leaving it as it is seems like the best course of action now. I appreciate the input

jhonatan-lopes:
Thank you for the review @tomusher! For sure, we'll cross that optimisation brigde when (or if) needed!

PR #10957, opened 2023-08-02
Unlocalize strings from the Page types report

### LINKED ISSUE #10956: Unlocalize Page types report
PR #10928 landed a few strings for a page that isn’t public-facing, we should rather hardcode these strings since they can only be seen by CMS users.

TheoChevalier:
Fixes #10956.

PR #10928 landed a few strings for a page that isn’t public-facing, we should rather hardcode these strings since they can only be seen by CMS users.

I appreciate the “Let’s localize everything by default” attitude, though!

### DIFF
--- a/network-api/networkapi/templates/pages/reports/include/_list_page_types.html
+++ b/network-api/networkapi/templates/pages/reports/include/_list_page_types.html
@@ -1,4 +1,4 @@
-{% load i18n wagtailadmin_tags wagtailcore_tags %}
+{% load wagtailadmin_tags wagtailcore_tags %}
 
 <table class="listing {% block table_classname %}{% endblock table_classname %}">
     <col width="20%" />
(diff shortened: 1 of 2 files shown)


TheoChevalier:
@jhonatan-lopes Are you able to have a quick look at this?

network-api/networkapi/templates/pages/reports/include/_list_page_types.html
```diff
@@ -12,25 +12,25 @@
...
-                    {% trans "App Model" %}
+                    App Model
                 </th>
                 <th class="count">
-                    {% trans "Pages" %}
+                    Pages
                 </th>
                 <th class="default-locale-count">
-                    {% trans "Pages in" %} {{ active_locale_name }}
+                    Pages in {{ active_locale_name }}
```
jhonatan-lopes, line 24: In this case, we are showing the count of pages in the locale that the Wagtail admin interface of the editor is currently setup in, e.g.:

- If I'm an editor and have my language setup as "English" in the admin, I should see the count of pages that are currently published in English
- If I'm an editor and have my language setup as "Portuguese (Brazil)" in the admin, I should see the count of pages that are published in Portuguese

The idea here is to have a comparison between the total amount of pages published per page type and the amount of pages in a page type in a locale, e.g. we have 9 `HomePage`s but only 1 in each locale.

However, if this page won't be localized, I'm wondering if having `{{ active_locale_name }}` makes sense now. Maybe we should display the count in English for all editors regardless of their configured Wagtail admin language.

What do you think?

jhonatan-lopes (commented):
Hi @TheoChevalier, thanks for catching this. I thought all pages needed to be localised. I've included a comment on the behaviour of a localised column that I was hoping to get your feedback on.

network-api/networkapi/templates/pages/reports/include/_list_page_types.html
TheoChevalier: Heya!

"Pages in {locale}" may be a bit misleading indeed, as it includes alias pages, that, by definition, aren’t translated.
If we can exclude alias pages from the count, it could be useful data to have.

However I’m not sure it’s super intuitive from a user perspective that this feature is tied to the Wagtail UI language. A filter dropdown on the locale would be ideal, but probably means more work. Alternatively, reusing the filter on the target locale from the Translations report could be easy to add? https://foundation.mozilla.org/cms/localize/reports/translations/?content_type=&source_title=&source_locale=&target_locale=fr

jhonatan-lopes (approved):
LGTM :+1: thanks Theo!

PR #10978, opened 2023-08-08
Revert "Long author names now wrap instead of breaking out of container"

danielfmiranda:
Reverts MozillaFoundation/foundation.mozilla.org#10807


jhonatan-lopes (approved):
LGTM :+1: Thanks @danielfmiranda

Issue #10659, opened 2023-05-24
[FA] [Post-Launch] Rename three fields in Petition Snippet


mmmavis:
For context, first part of the work is tracked here: https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10628

Once FA work is fully launched, we can then update the three properties to
- [ ] requires_country_code to show_country_dropdown
- [ ] requires_postal_code to show_postal_code_field
- [ ] comment_requirements to show_comment_requirements

mtdenton:
Moving estimate from 4 to 1 for this sprint

PR #11003, opened 2023-08-14
Petition Snippet - Specify what panels we wanna show on admin

mmmavis:
Petition Snippet - Specify what panels we wanna show on admin.

---

Now that a new field `show_comment_field` has been added. Before we remove the old field `comment_requirements` from model, let's hide it from admin to avoid confusing.

Related PRs/issues: #10659


jhonatan-lopes (approved):
Tested locally and it hides the comment field well. Thanks @mmmavis! :+1:

PR #11007, opened 2023-08-15
Fix petition fields migration

jhonatan-lopes:
# Description

Rename the old boolean fields `requires_country_code` and `requires_postal_code` to `show_country_field` and `show_postal_code_field` to preserve the data.

Previous migration file was deleting the fields and then creating the new fields, so the data was lost.

Link to sample test page:
Related PRs/issues: #10961 

# Checklist

**Tests**
- [ ] Is the code I'm adding covered by tests?

**Changes in Models:**
- [ ] Did I update or add new fake data?
- [ ] Did I squash my migration?
- [ ] Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [ ] Is my code documented?
- [ ] Did I update the READMEs or wagtail documentation?

**Merge Method**
**💡❗Remember to use squash merge when merging non-feature branches into `main`**

PR #11028, opened 2023-08-22
Revert campaign changes

jhonatan-lopes:
# Description

There is currently a large backlog of changes waiting to be deployed to production but being held back by the changes to the campaign pages, which require sequential migrations to not disturb the production environment, since fields were renamed/deleted.

To avoid pausing production deployment further, the changes related to the campaign pages will be reverted and the work will be picked up again once the team is back from All-Hands.

Reverts:
- #10961 
- #11003 
- #11007 

Also fixes the numbering of the remaining migrations.


kevinhowbrook:
Thanks @jhonatan-lopes, as confirmed on slack this has been tested against the prod BD as good.

One thing to think about is this making it's way to the stage site, it's possible you will end up in a bit of a tricky situation once main is merged back into staging

mmmavis (approved):
Looks good to me. Thanks @jhonatan-lopes !

Issue #10920, opened 2023-07-24
Move petition CTA above the fold on campaign page mobile views


mtdenton:
via @Gbaldassano, split from #10896 
Stakeholder: ***needs grooming***

### Context
Optimizing the Petition forms to maximize conversions is a priority for the Engagement team. As a result, they would like us to move the petition form/CTA up so it appears above the fold on mobile (similar to what was done for donate forms)

![Screen Shot 2023-06-29 at 11.01.21 AM (2).png](https://images.zenhubusercontent.com/640f8c755e71df27c4397000/7c90c34b-d24e-427a-9acc-7ef057a1e704)

However, there are some considerations to take into account: 

- Screen readers would have a different experience on desktop sizes if paragraphs were kept to the left, which means it would start reading the petition side first. While not optimal to give these users a different experience, it seems like we don't have an option here without radically rearranging or redesigning this page.
- We will have to rearrange the DOM through the template and make it respond to different viewport sizes with `order` sort of like we did in [donate-wagtail #1755](https://github.com/MozillaFoundation/donate-wagtail/issues/1755). The `h1` page title will need to be independent of the current top paragraph `div` and always remain at the top of the page despite the order of the other elements.
- The compactness of the form doesn't improve on mobile sizes. We may want to consider any other ticket coming from #10896 addressing this as a requirement as well.

I've mocked up what this could look like but omitted needed input from the design team on arrangement, margin/padding, etc. (cc @kristinashu):

![mocked](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/41487159/2d5d3774-ad06-47c5-a91a-55ded665bc20)

This takes the liberty of removing a few elements that were causing spacing issues:
- The gray bar to be removed in #10919 (consider making this a blocking issues)
- The form footer `div`s, we'll need input on what to do with these as well:
![supportInfo](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/41487159/8e525027-55b7-4eee-9e88-18b14cb11f84)

mtdenton:
We might consider splitting this as well, maybe assigning to an overarching petition page improvements epic. Suggested ticket ideas:

- make `h1` page title independent of the `campaign-page-content` parent div
- set visibility of `wFormFooter` and `supportInfo` elements to hidden if there is no content to be displayed

kristinashu:
[...]

Issue #10919, opened 2023-07-24
Hide unnecessary gray bar between the main nav and the heading on campaign pages if there's nothing to display


mtdenton:
Stakeholder: @kristinashu

On mobile sizes for campaign pages, the `multipage-nav-mobile` container appears even if there isn't any subpages.

### To Reproduce
1. Go to https://foundation.mozilla.org/en/campaigns/facebook/
2. Use dev tools to go to a mobile size, gray bar is at the top

### Expected behavior
This container shouldn't appear if there's nothing to display

### Screenshots
![multipage](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/41487159/405b1c31-8871-4194-8bc1-1a7bd7690be2)

Present on all browser/os/version combinations

### Additional context
It's been this way since at least 2019 according to the internet archive.

PR #11044, opened 2023-08-29
Hide multipage nav (secondary nav) if there's no links to show

mmmavis:
- Hide multipage nav (secondary nav) if there's no links to show (#10919)
- Hide FA's .wFormFooter and .supportInfo (as mentioned in [#10920](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10920#issuecomment-1648085670)) Note that the remaining tasks in #10920 will be tackled in new PRs.

### Test pages
- test page 1 (no secondary nav): https://foundation-s-10920-peti-ejxqei.herokuapp.com/en/campaigns/single-page/
- test page 2 (with secondary nav): https://foundation-s-10920-peti-ejxqei.herokuapp.com/en/campaigns/multi-page/

---

Tagging you @kristinashu as I'm unsure which designer I should assign this PR review to 😅

### DIFF
--- a/source/js/multipage-nav.js
+++ b/source/js/multipage-nav.js
@@ -1,6 +1,5 @@
 /**
  * Inject mobile version of multipage navs (secondary navs)
- *
  */
 export default () => {
   const targetNode = document.querySelector(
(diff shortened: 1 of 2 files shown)


source/js/multipage-nav.js
```diff
@@ -1,54 +1,70 @@
 /**
  * Inject mobile version of multipage navs (secondary navs)
- *
```
mmmavis, line 3: A lot of changes in this file are just white space changes.

kristinashu (approved):
Much better, thank you Mavis!

danielfmiranda (commented):
Hi @mmmavis thanks for taking on this work! 

After testing this out, I can confirm that everything is working as expected. The grey bar is gone, and both divs have been hidden.

However, I do have one question:

Regarding the .formFooter and .supportInfo divs that are now being set to `display:none;`, will these divs **always** be empty? Or, do users have the option of adding content within these divs through FormAssembly?

If so, I am concerned that permanently adding the property `display:none;` to these divs through SCSS will hide elements that are meant to be seen.

Thanks!

mmmavis:
> Regarding the .formFooter and .supportInfo divs that are now being set to display:none;, will these divs always be empty? Or, do users have the option of adding content within these divs through FormAssembly?

@danielfmiranda thanks for the q. Both sections will always be empty!

danielfmiranda (approved):
Thanks for the clarification Mavis! Approved 🙌

Issue #8797, opened 2022-05-24
[Blog] - [Author] - Author Details Header


danielfmiranda:
<img width="533" alt="image" src="https://user-images.githubusercontent.com/18314510/171496813-c66acb43-1579-4cb7-890e-aaaa317a9c91.png">

The goal of this ticket is to create a new fragment template for the header section of the author detail page. 

Please note that the social buttons are to be implemented at a later date.

Dev Steps:
---

- [ ] Create new HTML template for component
- [ ] Include the Breadcrumbs Component #8790 
- [ ] Include the Search Bar Component #8778 
- [ ] Render and style the authors name, image, and bio.
- [ ] Include this component in the Author Detail Page template #8796 

Acceptance Criteria:
--- 
- [ ] Bread Crumbs
- [ ] Author name
- [ ] Author Bio
- [ ] Author image
- [ ] Search Bar

danielfmiranda:
Hi @nancyt1! I was wondering if you could please help with a question regarding this ticket? 

In this ticket’s mock ups, I see that there are social media buttons. At the moment, author profiles do not have any fields for social media platforms.

We can add these, but should they be required? Or should we only render buttons for social media handles that they save.

Also, I see that the “copy to clipboard” button is in the mockups. Do we want to include that?

Thank you in advance!

danielfmiranda:
Leaving a note that after discussing with design, we have decided to  omit the social media buttons until a later date!

danielfmiranda:
Hi @nancyt1! 

This feature has now been merged into staging and is ready for QA, can you please take a look here https://foundation.mofostaging.net/en/blog/authors/daniel-miranda-646/ and confirm if everything looks OK? 

Thanks in advance!

kristinashu:
Yay thank you @danielfmiranda! Do you know if there's a follow up ticket to make the names on /blog and /blog/[post] clickable? I think they were clickable but then it was removed because they didn't go anywhere but I guess we'll be ready to make them clickable again!

danielfmiranda:
Hi @kristinashu! 

I have just taken a look at our repo's issues and see no tickets related to making the author names clickable. I can definitely create one for us, however, I am wondering whether the ticket should be marked as blocked until the Author Detail Page is complete, or, if its good to go now?

kristinashu:
Yes thank you Daniel, that makes sense!

nancyt1:
[...]

PR #11080, opened 2023-09-07
Author Detail Page Header

danielfmiranda:
# Description

Link to sample test page:
Related PRs/issues: #8797

This PR creates a new template for the "Author Detail Header" and adds it to the "Author Detail Page".

This PR also introduces a new helper function in `blog_index.py`, that takes a count of all the topics used by an author's blog posts, and then returns the top 3 in descending order. This function is used to return the topics found in the "Frequent Topics" section. With the time given this sprint, I was not able to add any tests for this function, but I will be happy to when I return from PTO 👍 

Since the design no longer asks for the Profiles tagline, I also updated the test `test_author_detail` to remove the two statements looking for them, as this was causing the tests to fail. 

# Screenshots

## Desktop
Screenshot 2023-09-07 at 18-15-23 Blog - Mozilla Foundation


danielfmiranda:
Hi @nancyt1!

I have pretty much wrapped up the changes needed for this PR, minus a couple of things on the back end that I want to update. However, the front end is ready for testing!

I was wondering if you could please take a look at these pages and let me know if everything looks OK? 

The way the top 3 topics section is set up is that the page renders them in descending order of times used. If a profile only has used two topics, then two will render. If a profile has used 0 topics, the section wont render at all.

Author Detail Page with profile picture set: https://foundation-s-8797-blog--5pdwru.herokuapp.com/en/blog/authors/xavier-harding-12/
Author Detail page without profile picture set: https://foundation-s-8797-blog--5pdwru.herokuapp.com/en/blog/authors/samuel-washington-7/
Author Detail page without any used topics: https://foundation-s-8797-blog--5pdwru.herokuapp.com/en/blog/authors/daniel-miranda-13/


Thanks in advance!

network-api/networkapi/wagtailpages/tests/test_blog.py
```diff
@@ -835,10 +835,8 @@ def test_authors_detail(self):
         response = self.client.get(path=blog_author_url)
         self.assertTemplateUsed(response, "wagtailpages/blog_author_detail_page.html")
         self.assertContains(response, self.profile_1.name)
-        self.assertContains(response, self.profile_1.tagline)
```
danielfmiranda, line 838: It was requested to not include Profile taglines in the author detail page, so I removed these from the tests!

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
```diff
@@ -407,15 +409,34 @@ def blog_author_detail(self, request: "HttpRequest", profile_slug: str, *args, *
...
+        BlogPage = apps.get_model("wagtailpages.BlogPage")
+
+        # Retrieve all BlogPages authored by this profile.
+        authored_blog_pages = localize_queryset(BlogPage.objects.filter(authors__author=author_profile))
+
+        # From the previous QS, get the related BlogPageTopic objects.
+        frequent_topics = BlogPageTopic.objects.filter(blogpage__in=authored_blog_pages)
+
+        # Calculate the count of each topic using Counter.
+        topic_counts = Counter(frequent_topics)
```
jhonatan-lopes, line 433: I feel like we can defer all of this to the db in one or two queries (excluding the localisation part). Maybe:

```python
top_topics = BlogPageTopic.objects.filter(blogpage__authors__author=author_profile).annotate(count=Count("name")).order_by("count")[:3]

top_topics = localize_queryset(top_topics, order_by="name")
```
I've tested it with a local copy of prod's db and it seems to work, but I haven't done extensive testing.

In any case, it would be good to have backend tests for this function before we merge.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
```diff
@@ -0,0 +1,49 @@
+{% load l10n i18n localization static blog_tags wagtailimages_tags wagtailroutablepage_tags wagtailcore_tags %}
+
+{% get_root_or_page as blog_index_page %}
+
+<div class="tw-grid tw-grid-cols-12 large:tw-gap-40">
+    <div class="tw-col-span-12 large:tw-col-span-8">
+        <div class="tw-text-sm tw-text-gray-40">
+            <a class="tw-text-gray-40" href="{% pageurl blog_index_page %}">Blog</a>
```
jhonatan-lopes, line 8: Here, I don't think we need `blog_index_page`. We can use `page` directly, as it will get the url for the `BlogIndexPage`:

```html
 href="{% pageurl page %}"
```

If we needed any of the routes defined in the `BlogIndexPage`, we could use the `wagtailroutablepage_tags`, e.g.:

```html
{% load wagtailroutablepage_tags %}

 href="{% routablepageurl page "authors_index" %}"
```

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
```diff
@@ -0,0 +1,49 @@
...
+                        {% else %}
+                            <img src="{% static '_images/author-placeholder.jpg' %}" width="58px" height="58px" alt="{{ author.name }}" class="{{ profile_image_classes }}" />
+                        {% endif %}
+                    {% endwith %}
+                </div>
+            </div>
+            <h1 class="tw-mb-8">{{ author.name }}</h1>
+        </div>
+        <p class="tw-text-xl tw-pb-16">
+            {{ author.localized.introduction }}
```
jhonatan-lopes, line 28: We should avoid using `localized` in the templates, as it generates extra queries. I think `author` should already be localized, though I'm not sure. This would be a good scenario to add tests.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
```diff
@@ -0,0 +1,49 @@
...
+        </p>
+        {% if frequent_topics %}
+            <p class="tw-text-lg tw-text-gray-40 tw-uppercase">{% trans "Frequent Topics" %}</p>
+            <div class="tw-inline-flex tw-gap-6 tw-flex-wrap">
+                {% with link_classes="tw-text-black tw-px-8 tw-border-black tw-border tw-rounded-full tw-text-lg tw-font-normal tw-whitespace-nowrap tw-block tw-transition-colors" link_hover_classes="hover:tw-bg-black hover:tw-text-white hover:tw-no-underline hover:tw-border-white" %}
+                    {% for topic in frequent_topics %}
+                        {% localized_version topic as localized_topic %}
```
jhonatan-lopes, line 35: Do we need this line? The topics will be localized at the `get_authors_frequent_topics` method, using `localize_queryset`...

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
```diff
@@ -0,0 +1,49 @@
...
+                        <a href="{% localizedroutablepageurl blog_index_page "entries_by_topic" topic.slug %}" class="{{ link_classes }} {{ link_hover_classes }}">
```
jhonatan-lopes, line 36: Nice routing :+1:

jhonatan-lopes (changes requested):
Hey @danielfmiranda! Thanks for this work! It looks great, I just have some changes to make sure that it works better

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: Great suggestion Jhonatan! I have updated the code to use your suggested query, I think it looks a lot cleaner and is easier to understand. 

Working on adding tests now 👍

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
danielfmiranda: Updated! Working on adding tests now 👍

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
danielfmiranda: Great point! Updated

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_author_detail_header.html
danielfmiranda: Hey @jhonatan-lopes, after taking a further look at this, I agree that the `.localized` is no longer needed here.

That is because this page is using profile slugs to grab/render the data needed from the DB, and each translation of a profile has its own unique slug (ex: EN would be `daniel-miranda-1` and FR would be `daniel-miranda-4`).

That means if a user visits the blog author index page under their preferred locale and selects my name, they will be redirected to my author detail page with the appropriate slug if applicable.( if not, it will default to the EN slug). 

One concern that I have, is that this means that if a user visits `/fr/blog/authors/daniel-miranda-1` (the English slug), that the profile details will still render in English. 

However, I know that we have implemented profile slugs like this [by design](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10578), so I think that this is OK for now, and we can discuss this issue further at a later time.

I have added it to the agenda on the next sprintly engineering meeting, and will add some simple tests making sure that the pages of translated profiles render the appropriate language. 

Thanks!

danielfmiranda:
Hi @jhonatan-lopes! Thanks for the great review and feedback. I was able to successfully use the blog breadcrumb component that you merged in while I was gone, added some tests for the get_frequent_topics method [here](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/a433c733ae2c91d7bc01b394fef2630792e4179c/network-api/networkapi/wagtailpages/tests/test_blog.py#L850-L935), and re-requested review. 👍

network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py
danielfmiranda: Hi Jhonatan! Wanted to let you know that your query is working great, the only difference is that I had to update it with `order_by("-count")` and remove the `order_by="name"` from localize queryset 🙌 

Thanks!!

jhonatan-lopes (approved):
Thanks @danielfmiranda! Great job on the tests!

nancyt1 (approved):
Looks amazing, thanks Daniel!

Issue #8790, opened 2022-05-24
[Blog] - Global Breadcrumbs


danielfmiranda:
<img width="580" alt="image" src="https://user-images.githubusercontent.com/18314510/169957239-db5fdf98-2f08-4ab4-8820-b856eb5f2a66.png">

According to the designs, topic detail and author pages should include breadcrumbs above the pages title. 

Since they are styled the same in both pages, we can probably make this its own component, similar to how we did in the [research hub](https://github.com/mozilla/foundation.mozilla.org/pull/8628), and render it on both pages.

Acceptance Criteria:
---

- [ ] A new template is created that will accurately display and link back to "parent" pages when used.

PR #11092, opened 2023-09-12
Blog global breadcrumbs

jhonatan-lopes:
# Description

Add breadcrumbs to blog pages.

The author detail page breadcrumbs will be implemented as part of the "Author details header" ticket (#8797) and wasn't included in this work.

Link to sample test page:
Related PRs/issues: #8790 
Jira: https://mozilla-hub.atlassian.net/browse/TP-9

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

**Merge Method**


network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_breadcrumbs.html
```diff
@@ -0,0 +1,6 @@
+{% load wagtailcore_tags %}
+
+<div class="tw-text-sm tw-text-gray-40">
+    <a class="tw-text-gray-40" href="{% pageurl page %}">{{ page.title }}</a>
+    <span> › </span>
```
mmmavis, line 5: Let's change this so it's more a11y friendly. 

Using HTML entity `&gt;` for the symbol > makes sure the separator is properly encoded and can be read by screen readers etc. The surrounding spacing would be better if it's done via CSS than the actual space character.

One thing I'm unsure because the lack of attached mockup is that using Tailwind classes to style as well as `&gt;` does give a different look (screen shot no.1) than what's currently implemented in this PR (screen shot no.2). We might need to change the font styling/type so the separator look the way we wanted.

```suggestion
    <span aria-hidden="true" class="tw-mx-1">&gt;</span>
```

Screencap no.1

image

Screencap no.2

image

mmmavis (changes requested):
Hi @jhonatan-lopes , PR looks good. 😃 There's just one minor change we can do to make the breadcrumb more a11y friendly.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/blog_breadcrumbs.html
jhonatan-lopes: Thanks @mmmavis, those are awesome suggestions! Accesibility is definitely not my strongest suit, I appreciate the input.

One thing to note is that the char is not the "greater than" quotation mark (>) but rather the "single right-pointing angle quotation mark" (`›`) -- see https://www.compart.com/en/unicode/U+203A.

The HTML entity seems to be `&rsaquo;`.

mmmavis (approved):
🍞 👍🏻

Issue #8788, opened 2022-05-24
[Blog] - [Blog Page] - CTA Block Component


danielfmiranda:
<img width="832" alt="image" src="https://user-images.githubusercontent.com/18314510/171538373-547d521c-7553-4238-ae2b-ff5e3b3a7360.png">

The goal of this ticket is to create a new block for blog posts, called a "CTA block".

This new component should be able to be left aligned, right aligned, and full width. Similar to the blocks in publications.

It can be styled one of 3 ways:

Pop:

<img width="565" alt="image" src="https://user-images.githubusercontent.com/18314510/171538861-434d5f5a-6523-40c1-a378-9908a070c2cf.png">

Outline:
<img width="565" alt="image" src="https://user-images.githubusercontent.com/18314510/171540744-ef22c2da-6894-4b7c-89ce-bb5db7d6f961.png">

Filled:
<img width="565" alt="image" src="https://user-images.githubusercontent.com/18314510/171540795-f333be04-1267-4742-8a4d-48bd0f7f8b7b.png">

Full width:
![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/50ff44b2-a343-429c-bc7f-2beede005fd3)

Optional button or image:
![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/a577711f-74ec-459d-b346-4e91baf25386)

It should have the following fields:

- [ ] Image - optional
- [ ] Body (rich text) - required
- [ ] Link Button (choice of primary or secondary) - Optional
- [ ] Link URL - Optional

Dev Steps:
---

 Note: We already have images with the same "left/right" with text wrapping around it styling on PublicationPages, so we can probably use that for reference on this ticket.

- [ ] Create a new streamfield block for use on the BlogPage, with the following fields above
- [ ] Make sure that this new block works as expected, with the other body text wrapping around it.

mtdenton:
Hey @nancyt1 ! When you get a sec, could you take a look here? Thanks!

PR #11127, opened 2023-09-18
[TP-13] [Blog] Blog page CTA block component

jhonatan-lopes:
# Description

Implement a CTA Block component.

Some changes were made from the original issue. I had to break this down into two components:

- A "CTA card" block, where we can display the required info
- A "CTA card with text" block, where we link a card to a "paragraph" so that the card can be placed either to the left or the right.

The "CTA card" block will display the "Full width" option. There is no option to add a "full width" card on the "CTA card with text" block, as this is equivalent to a regular "CTA card" block followed by a previously-available "Paragraph" block.

The `image`/`button` are optional, as well as a `title` for the card, which is styled differently from the rest of the text.

The `body` is a rich text field, allowing for:

- Bold
- Italics
- Links
- Unordered lists
- Ordered lists
- H4 and H5
- Horizontal rules

Link to sample test page: https://foundation-s-feature-87-h6m1io.mofostaging.net/en/blog/initial-test-blog-post-with-fixed-title/
Related PRs/issues: #8788 

# Checklist

**Tests**
- [X] Is the code I'm adding covered by tests? (PS.: only new custom blocks are covered, as blog page has no tests setup and its factory relies on the [old method that is going to be replaced](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11067))

**Changes in Models:**
- [ ] Did I update or add new fake data? -- Nope, see above. I've added factories, but not new fake data.

**Documentation:**

**Merge Method**


jhonatan-lopes:
Hey @nancyt1, the review app is ready now. I've added some cards to the CMS so that you can see the design

nancyt1 (changes requested):
Could we increase the padding on all sides by .5rem? 

Can we decrease the size of the top border to 4px?
SCR-20230922-nxzq

Also decrease shadow of this box to 4px?
SCR-20230922-nzae

On the mockups, I originally wanted them to break out of the body gird, is this still possible?
Screenshot 2023-10-02 at 1 37 31 PM

jhonatan-lopes:
Hi @nancyt1, the review app is ready for you again

jhonatan-lopes:
Hi @nancyt1, I've addressed your comments, can you please take a look at this again?

More specifically:
* Increased the padding inside the card by 0.5 rem
* Decreased the shadows to 4px
* Implemented the break outside the grid card (finally able to get this one)

image


To make the card break outside the grid, I'm using a negative margin of 12rem, please let me know if you would like to adjust this value.

I've only put this on cards that are aligned to the right. Cards aligned to left are then aligned to the text column.

I've also put a media query to make sure that this out of the grid effect only activates for large screens.

nancyt1 (changes requested):
It's looking great! I'm just trying to identify edge cases at this point. Let me know if there's anything I should elaborate on!

- Can we constrain images to fit the width of the card? Currently it makes the CTA huge if the image is big.
- Is it possible to also break out the left align one in a similar way as the right aligned one? Or maybe we can just remove left align option? Alternative is making left align smaller width so that it doesn’t take up more than 1/3 of the body columns.
- Do you have suggestions on how to prevent gaps like this if the paragraph text is shorter than the CTA text? Would adding character count be enough (~600-700)? 
SCR-20231006-phne



**For Mobile/Tablet:** 
- Can the cards be full width for mobile? Or at least be center aligned?
- For smaller screens like tablet, can the width of the card be ~20rem or take up no more than half of the body columns?
- Can we also reorder it so that the card comes after the paragraph text? Maybe we can also reorder in wagtail too.

**For Wagtail UI:** 
- Move the style and alignment options to the top
- Image and button should be optional, can we collapse them and let staff add it themselves?
- Can the title styling be H3? And can we add H5 to the styling options?

jhonatan-lopes:
Hi @nancyt1, thanks for the feedback!

I have altered the cards to also break from the grid when left aligned:
image

I've put a validation that the paragraph content cannot be shorter than the card's content:
image


Full-width cards for mobile:
image


For tablet, I've set it up to take 20rem max. I've also disabled the grid offset on tablet size:
image

Cards should come after the paragraph on mobile screens now.

For Wagtail/UI:

- I've moved the alignment option to the top, but the style can only be at the top of the "card" section (so not at the top of the whole block, but at the top of the "card" sub-block)
- Image and button should be collapsed by default now
- I've changed the card title to be H3 and put options for H4 and H5 on the card body

The only thing left now is to address the image situation for full-width cards, as we discussed on slack.

jhonatan-lopes:
It was decided that the images for "full-width" cards should have their heights capped at `7rem`.

Here is the new review app:
https://foundation-s-feature-87-h6m1io.mofostaging.net/en/blog/initial-test-blog-post-with-fixed-title/

network-api/networkapi/wagtailpages/tests/customblocks/test_blog_cta_card_block.py
```diff
@@ -0,0 +1,34 @@
+from django.test import TestCase
+
+from networkapi.wagtailpages.factory import customblocks as customblock_factories
+
+
+class TestBlogCTACardBlock(TestCase):
+    def test_blog_cta_card_all_optionals(self):
+        full_card = customblock_factories.BlogCTACardBlockFactory()
+        print(full_card)
```
danielfmiranda, line 9: Hey @jhonatan-lopes just checking if we want to print `full_card` during this test or if maybe this was left here by accident

jhonatan-lopes: Great catch, thanks! I have a tendency to leave those pesky prints behind :see_no_evil:

danielfmiranda (approved):
Thanks @jhonatan-lopes! I think this PR sets a great example of creating factories for streamfield blocks.

Everything looks good to me. Approved!

nancyt1 (changes requested):
Sorry I only have 3 more notes!

1. Can we move the Alignment field back next to Styling? Sorry, it's more confusing now that it's by itself. 

2. Can we increase margin-right for the left-align block to 1.75rem? On tablet, can we also remove the margin-left indentation similar to the right-align block? SCR-20231017-mhng

3. Can we update the full width CTA to also be full width for mobile? Currently they are slightly off from each other.
Screenshot 2023-10-17 at 5 56 45 PM

jhonatan-lopes:
Hi @nancyt1, no problem! I've implemented the changes in your notes now

nancyt1 (approved):
Looks good, thank you!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/blog_cta_card_with_text_block.html
```diff
@@ -0,0 +1,14 @@
+{% extends "./base_streamfield_block.html" %}
+{% load l10n wagtailcore_tags wagtailimages_tags %}
+
+{% block block_content %}
+    <div class="blog-cta-card-with-text">
+        <div class="blog-cta-card-with-text__card-container blog-cta-card-with-text__card-container--{{ value.alignment }}">
```
mmmavis, line 6: Hi @jhonatan-lopes , we don't use BEM naming convention in this codebase so I would appreciate if we don't  introduce a new naming pattern. To be honest, there's no one consistent naming convention that we have been sticking to in this codebase 😓 , however, we try to use shorter names possible.

I see that we are also adding new SCSS code in this PR. As we are trying to move away from SCSS, could we convert the rules into Tailwind classes instead? If the rules will be used more than just this block, we can even create custom component on Tailwind config. What do you think? If this sounds good then that also mitigate the naming convention headache here.

mmmavis (changes requested):
Hey @jhonatan-lopes sorry for getting to this PR late and thanks for the great work! I just have some questions regarding SCSS/Tailwind and class naming conventions. Please see my inline comments.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/blog_cta_card_with_text_block.html
jhonatan-lopes: Hi @mmmavis, thanks for raising these points! It was definitely inertia from working with BEM before :smile: I've migrated the styles to only use Tailwind now.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/blog_cta_card_block_regular.html
```diff
@@ -0,0 +1,32 @@
+{% load l10n wagtailcore_tags wagtailimages_tags %}
+
+<div class="[&_.rich-text_a]:tw-underline [&_.rich-text_a]:tw-text-black
```
jhonatan-lopes, line 3: I've tried to create a [variant plugin](https://tailwindcss.com/docs/plugins#static-variants) for this selector as we might need to use it often but it wouldn't work.

After adding

```js
    plugin(function({ addVariant }) {
      addVariant("rich-text-links", ["&_.rich-text_a"]);
    })
```
to our `tailwind.config.js` and changing the calls to `rich-text-links:tw-underline rich-text-links:tw-text-black`, the styles were no longer applied to links inside the cards.

This is a weird issue, since the syntax for the plugin seems to be correct, as we can see in this [Tailwind play environment](https://play.tailwindcss.com/eZmFo3HlUs).

jhonatan-lopes: It seems like this is caused by a bug with Tailwind and prefixes: https://github.com/tailwindlabs/tailwindcss/issues/10125#issuecomment-1382724160

There is a workaround, which I have now implemented here :)

tailwind.config.js
```diff
@@ -244,6 +245,10 @@ module.exports = {
         purple: "#353552",
       },
     },
+    boxShadow: {
```
mmmavis, line 248: @jhonatan-lopes Do you know if these "pop" and "outline" styles are also used elsewhere on the site? Or if they are defined in the design system? I'm a little worried that we add these styles as "boxShadow/pop/outline" to Tailwind configs because they are very specifically to this use case, and there are many other existing elements on the site have some sort of box shadow that we eventually need to make them adapt to Tailwind classes. If we want to keep these in the config, maybe rename these classnames to something less general could be an option?

mmmavis (changes requested):
Left a comment regarding Tailwind config. Everything else looks good!

tailwind.config.js
jhonatan-lopes: I see your point @mmmavis , thanks! I got the names from the design team, so I think these would be reused on the blog as it is defined with their design system. In the end, I thought this was "cleaner" than one time styles, but maybe I could rename them to be on a blog scope, e.g. "blog-pop" and "blog-outline"? What do you think?

jhonatan-lopes: Here it is on the [design system](https://www.figma.com/file/528nyaJFzvCXAiw2gglwzh/Blog-Improvements?node-id=627%3A22439&mode=dev):
image

mmmavis (approved):
Chatted with designers. It's good that we add these styles to Tailwind config.

Issue #11134, opened 2023-09-18
[PNI Privacy Index] Implement By The Numbers section


mmmavis:
## Dev Tasks

- [x] Build "By the Numbers" Section according to the [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1812%3A18398&mode=dev) and [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1812%3A18399&mode=dev) mockups
- [x] Static illustrations are in place. (animation / effects will be tackled in a separate ticket)

mmmavis:
Closing. Work has been merged into the feature branch. 🎉

Issue #11133, opened 2023-09-18
[PNI Privacy Index] Implement Best & Worst Products section


mmmavis:
## Dev Tasks

- [x] Build "Best & Worst Products" Section according to the [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1812%3A17154&mode=dev) and [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1812%3A17404&mode=dev) mockups
- [x] Static illustrations are in place. (animation / effects will be tackled in a separate ticket)

mmmavis:
Closing. Work has been merged into the feature branch. 🎉

Issue #11132, opened 2023-09-18
[PNI Privacy Index] Implement Big Picture Trends section


mmmavis:
## Dev Tasks

- [x] Build "Big Picture Trends" Section according to the [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1743%3A28059&mode=dev) and [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1743%3A28309&mode=dev) mockups
- [x] Static illustrations are in place. (animation / effects will be tackled in a separate ticket)

mmmavis:
Closing. Work has been merged into the feature branch. 🎉

Issue #11131, opened 2023-09-18
[PNI Privacy Index] Implement Sub Nav section


mmmavis:
## Dev Tasks

- [x] Build "Sub Nav" Section according to the [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1677%3A15830&mode=dev) and [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1677%3A16082&mode=dev) mockups
- [x] Static illustrations are in place. (animation / effects will be tackled in a separate ticket)
- [x] Link buttons should have hover state

mmmavis:
Closing. Work has been merged into the feature branch. 🎉

PR #11156, opened 2023-09-22
PNI CC Page - Sub Nav + 3 middle sections

mmmavis:
Note for code reviewer @danielfmiranda :

This PR is similar to the previous one you reviewed, except it has a lot more content. Because this is a highly customized one-off page that won't be updated frequently (or at all), I didn't end up creating custom Tailwind theme configs for these new styles. The long class names are making the template files really hard to review. Apologies! If you have any suggestions on how to make the code organization better, please advice here or on this [ticket](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11154)! Since the code changes in this PR are all HTML and assets changes, I think it's safe for you to just do a quick sanity check (vs in depth line by line code review). 
(Note that this PR almost passes design review. I'm just one or two tweaks away to earn that R+ 😄 )

---

Review app link: https://foundation-s-11133-pni--vrmku2.mofostaging.net/en/privacynotincluded/articles/annual-consumer-creep-o-meter/

This PR implements a few sections. Many of these sections are decorated with multiple illustrations and are a bit cumbersome to make everything look awesome on all screen sizes while sticking to using sematic HTML and good a11y practice. I've gone through rounds of tweaking myself and thought I should just move on lol. I would appreciate if we could focus on finding "bugs" for this PR review and leaving the fine-tuning till later.

## Subnav 
- #11131

## Big Picture Trends 
- #11132

## Best & Worst Products
- #11133

## By the Numbers
- #11134
- This is the trickiest section to get illustrations look balanced with the text. For implementation I had to use a 12-column grid and that's why the width of each card doesn't look exactly the same as the mockup (mockup wasn't using grid based layout). There are some cases where text breaks at a less optimal place. There are ways to improve this but all of them require structuring the code in an overly complicated way (which I'd like to avoid). Maybe we could consider reducing font sizes or increasing the space we reserve for text?


tessheinricks (changes requested):
**Subnav**
- For tablet, can we reduce the button sizing (perhaps mobile sizing but still in a horizontal line) so that they can all remain one line tall (please let us know if this should be prioritized later of if we can design a mock for this state)

**Big Picture Trends**
- Example title and body copy too large on mobile. It is currently 20 point size (should be the same size as the other body copy within the container) As shown in figma both body copy fonts should be 18 

I've noted that this is different than the desktop treatment so apologies if this complicates things, if you would like a universal treatment, please revise all desktop, tablet, and mobile body copy sizing for this section to the same font size (18/27)

**By the numbers**
- Title is missing s, should be "BY THE NUMBER**S**"

sabrinang (changes requested):
**By the numbers**
Great work on this so far! Here are some feedback that might help refine this section _(specifically for the 991<1200 breakpoint)_:
- In the mockup, the confetti cones face outward so that can be swapped
- In the areas where the numbers and illustrations overlap, the illustrations perhaps could shrink to have some white space in between (the confetti cones can be a bit larger to also balance these out)
- the 50% stat text could be shifted lower to balance with the box beside it like in the mockup
image

mmmavis:
@tessheinricks @sabrinang thanks you for your valuable feedback. I've updated the PR accordingly. Could you review the page again to see if I have captured everything?

mmmavis:
@tessheinricks np! I've adjusted the sizing a bit. Please have a look.

network-api/networkapi/templates/buyersguide/pages/consumer_creepometer_page_2023.html
```diff
@@ -40,7 +40,7 @@ <h1 class="tw-font-normal tw-font-zilla tw-text-[48px] tw-leading-[44px] medium:
   </div>
 
   {% comment %} Quiz Section {% endcomment %}
```
danielfmiranda, line 42: I appreciate you separating these sections using `{% comment %}`. 

In previous PR's of mine, I have tried to show separation of different sections by using `ID` or classnames, however this seems like a better idea to me as it shows the separation of sections in a cleaner, more readable manner. Plus, adding ID's to an element when we were not using the ID for anything else such as JS, felt strange to me.

Definitely going to keep this in mind moving forward!

network-api/networkapi/templates/buyersguide/fragments/consumer_creepometer_page_2023/big_picture_trends.html
```diff
@@ -18,13 +18,13 @@ <h3 class="{{ h2_class }} tw-mb-9">Products are getting more secure, but also a
         </div>
         <div class="tw-col-span-2 medium:tw-col-span-1 tw-flex tw-flex-col tw-justify-between">
           <p class="{{ body_large_class }} tw-font-light">The mental health app BetterHelp shares your data with advertisers, social media platforms, and sister companies.</p>
-          <a class="{{ logo_class }} tw-mb-9 medium:tw-mb-0" href="https://foundation.mozilla.org/privacynotincluded/betterhelp/" target="_blank" >
+          <a class="{{ logo_class }} tw-mb-9 medium:tw-mb-0" href="https://foundation.mozilla.org/privacynotincluded/betterhelp/" target="_blank" aria-label="read more about BetterHelp">
```
danielfmiranda, line 21: **Kudos** for better a11y :D

danielfmiranda (approved):
**Kudos** @mmmavis amazing job! 🙌 taking a look through the commits, I can see that this was a lot of work. However after taking a look at the review app linked in the description, I think everything is looking great! Very excited to see this live with the eye functionality.

Approved 👍

network-api/networkapi/templates/buyersguide/pages/consumer_creepometer_page_2023.html
mmmavis: Thanks @danielfmiranda ! I still need to read upon best practices around identifying sections/blocks without creating new id/class solely for that purpose. This wasn't an issue before as we had to create id/class for SCSS to kick in. Now that we are using Tailwind, I do find debugging on dev tool very challenging because what we have now is a series of Tailwind utility classes 😱

tessheinricks (approved):
Amazing work @mmmavis. Thank you!

sabrinang (approved):
Looking fantastic so far Mavis! 👍

mmmavis:
Merging this into the feature branch now. Thanks all!

Issue #11172, opened 2023-09-26
Create DonateHelpPage model and template


danielfmiranda:
## Description 

The goal of this ticket is to create the model needed for the new "Donate Help Page". 

This page should have one field titled `body` and should share streamfield/block choices with the CampaignPage model. This can be done by importing the [base_fields](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/customblocks/base_fields.py) list for use. An example of this can be found [here](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/modular.py#L31).

For the time being, this page's template should be blank, only rendering the pages title.

## Dev Tasks

- [ ] Create new page model titled `DonateHelpPage`, that can only be created under a `DonateLandingPage`
- [ ] Add new `body` field to the page model, which is a streamblock that uses [base_fields](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/customblocks/base_fields.py)).
- [ ] Create a new template for this page, that at the moment only renders  a div with the class of `tw-container` and 4 vertically stacked placeholder elements
    - `<h2> {{page.title}} </h2>`
    - `<p> Intro </p>`
    - `<p> Form </p>`
    - `<p> Body </p>`
- [ ] Add some tests for this new model (ex: make sure the right template renders, that it can only be created under DonateLandingPage, etc)

## Acceptance Criteria
- [ ] As a CMS user, I can create a new page of the type `DonateHelpPage` as a child of a `DonateLandingPage`.
- [ ] As a CMS user, the CMS allows me to edit the page's `body` field, which has the same selection of streamblocks as `CampaignPage`

PR #11182, opened 2023-09-27
DonateHelpPage model, factory, and tests

danielfmiranda:
Related Ticket: #11172

Link to review page: https://donate-foundation-s-11177-dona-5stlkm.mofostaging.net/en/help/

# Description

Originally, there were only two page models related to donations on the foundation stack. These were:
- `BaseDonationPage` - Which serves as the base page model, in case we need to add any shared fields/functionality across Donation related models. (currently, there are none so its pretty much blank.)
- `DonationLandingPage` - Which serves as the landing page for the donation related pages.

These both lived in a file `models.py`, which goes against what we do for other models in our codebase. 

If you take a look at the folder `wagtailpages` for example, you will see that `models.py` imports all page models, which all live under their respective folders/files. (for example `blog/blog_page.py`)

Since we are so early into the development of the donate side of the foundation site, I thought that this PR would be the best time to do some refactoring so the donate models can fall under the same convention. 

That is the reason that alongside the `DonateHelpPage` model changes, you will also see the following additional file changes:

- `pagemodels/__init__.py`
- `pagemodels/base_page.py`
- `pagemodels/landing_page.py`

The code related to these page models has already been approved in [this PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10401), so I do not believe there is a need to look too into these.

**That being said, this PR introduces the following DonateHelpPage related updates:**

- A new model `DonateHelpPage` which at the moment only has one  `StreamField` field `body`. The streamblocks available for use are the same as `CampaignPage` per request of design. (They both use our pre-defined list called `base_fields`)
- A template for the new model, which at the moment only renders the page's title, as well as placeholders for "Intro", "Form", and "Body"
- A new factory for the model, and an update to existing factories so new db's will include a help page titled "Donate Help"
- A basic test suite for the new model, as well as one small update to the `DonateLandingPage` tests, confirming that it can create a `DonateHelpPage`.

# Screenshots
Screenshot 2023-09-27 at 15-40-58 Donate Help

# Steps to test

1. Checkout this branch
2. Run `inv new-db`
3. Log into the CMS
4. Alongside "Homepage" and "Mozilla Festival", there should be a page titled "Donate Now" with a subpage "Donate Help".
5. Click into "Donate Help" and take a look around, it should only have two fields "title" and "body".
6. Visit the page on the front end, it should look like the screenshot above.
7. If everything is working as expected, testing is complete!


mmmavis (approved):
Awesome! Thanks for the clear review instructions as well as breaking the existing code into separate files.

network-api/networkapi/donate/pagemodels/help_page.py
```diff
@@ -0,0 +1,34 @@
...
+
+    translatable_fields = [
+        # Promote tab fields
+        SynchronizedField("slug"),
+        TranslatableField("seo_title"),
+        SynchronizedField("show_in_menus"),
+        TranslatableField("search_description"),
+        SynchronizedField("search_image"),
+        # Content tab fields
+        SynchronizedField("body"),
```
jhonatan-lopes, line 33: Is this correct? Do we want to show the same `body` for all translated pages? Shouldn't this be a `TranslatableField`?

network-api/networkapi/donate/factory.py
```diff
@@ -18,6 +18,13 @@ class Meta:
     featured_image = SubFactory(ImageFactory)
 
 
+class DonateHelpPageFactory(PageFactory):
+    class Meta:
+        model = DonateHelpPage
+
+    title = Faker("text", max_nb_chars=140)
```
jhonatan-lopes, line 25: Just FYI, I'm usually using `Faker("sentence")`, since it's usually better for a phrase.

jhonatan-lopes (commented):
Looks good, I just have one question for a translated field...

network-api/networkapi/donate/pagemodels/help_page.py
danielfmiranda: Hey @jhonatan-lopes! Great point, after taking a look at the rest of our codebase it appears that I put this here by mistake. All of the other "body" fields are set to be a `TranslatableField` like you mentioned. I have updated this file to do the same! 

Thanks for taking a look 👍

danielfmiranda:
Hey @jhonatan-lopes thanks for taking a look and the great feedback! I have updated the code accordingly and re-requested review for whenever you find a free moment 🙌

jhonatan-lopes (approved):
LGTM :+1:

Issue #11095, opened 2023-09-13
[PNI Privacy Index] Implement Quiz section (Part 1)


mmmavis:
This is part 1 of Quiz section work which focuses on implementing the intro text and quiz screens (product selections + result page). Newsletter signup related bits and share buttons work will be tackled in [part 2](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11194) and [part 3](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11213) tickets.

## Mockups

- [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1090%3A13156&mode=dev)
- [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1635%3A8065&mode=dev)

## Copy and Other Quiz Details

See the very bottom section on this [doc](https://docs.google.com/document/d/1xY3-NacjCfaMKoYM1EHhd6KNhPRowkdKnGwd5PBd5TA/edit).

## Dev Tasks

- [x] Build intro text section
- [x] Build the "product selection screen"
- [x] Build the "result screen"
- [x] Optimize image files

## Acceptance Criteria

- [x] Quiz results are generated based on the calculation listed in the copy doc's [Quiz Details section](https://docs.google.com/document/d/1xY3-NacjCfaMKoYM1EHhd6KNhPRowkdKnGwd5PBd5TA/edit)
- [x] Implementation matches Figma design

mmmavis:
Hi @kristinashu, this is the ticket that I need your help with regarding Shareprogress. Thanks!

kristinashu:
Hi @mmmavis, do you need just one Share Progress link or one for each results screen? For example, when I click share after voting on a product creep-o-meter it gives a different message depending on what I voted.

<img width="719" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/5cad2b68-6159-4e19-a785-b02f341fd91a">
<img width="581" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/5a7006b6-e6cc-475c-a728-6d7347b61c4c">

mmmavis:
[...]

PR #11211, opened 2023-10-03
PNI CC Page - Quiz section (part 1) 

mmmavis:
To code reviewer @danielfmiranda :
Hey Daniel, there are 30ish files changed in this PR but most of them are from image optimization. I would say the file that needs a closer look at would be [the `.jsx` file that renders the quiz component](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11211/files#diff-d8536ec8cd7f2d79fed291dcb46a71dd0da0c1b21570dcb6cf8a8c80860dcebd). I didn't add a ton of inline comments in that file but please let me know if any parts need further clarification. Thanks!

---

- Related ticket: #11095 
- Review app link: https://foundation-s-new-11095--nuyv94.mofostaging.net/en/privacynotincluded/articles/annual-consumer-creep-o-meter/
- Quiz details / logics can be found in this doc's [Quiz Details section](https://docs.google.com/document/d/1xY3-NacjCfaMKoYM1EHhd6KNhPRowkdKnGwd5PBd5TA/edit)

Note for reviewers: 
- This only cover's part 1 of Quiz section work (details see #11095)
- "Retake Quiz" button doesn't have the icon yet. Could we add this as a todo to the "after MVP" ticket? (Update: Tess has added it. Thanks, @tessheinricks )


tessheinricks (changes requested):
Hi @mmmavis amazing work! 

As you mentioned in this ticket, I have added the button icon for "Retake Quiz to the after MVP ticket

**Universal**
- [ ] The Apple Watch image has a white background instead of transparent and/or grey. Please let me know if you need any additional artwork exported.
Screenshot 2023-10-04 at 3 01 58 PM

**For Desktop**
- [x] Are we able to add an ease to the product image hover (so that it optically behaves like the Winners & Losers hover or more similarly to the figma prototype). However, I think this request can be moved to the post MVP ticket - please let me know if that works for you!

**For tablet**
- [x] The product list unfortunately becomes quite long, are to conform the tablet version to the mobile solution for this (a contained, scrollable box)? 

**For Mobile:**
- [x] When reviewing on my phone, the top copy does not appear (unsure if this is a device specific problem as it shows when inspecting for mobile on my laptop)
- [ ] On my mobile device, the scroll bar seems to be operating behind the content
- [ ] Are we able to conform the vertical spacing between items on mobile to M_Vertical-Gap_Quaternary (8)
Mobile_Testing

mmmavis:
Hi @tessheinricks , thanks for catching those issues especially the bug on Safari! There are always some gotchas when implementing sticky elements 😅  

I've updated the PR and it's for your review again!

mmmavis:
@tessheinricks @danielfmiranda I just realized my latest commit ruined the spacing in the product selection grid a bit (on larger screens sizes). I will submit a commit to fix that soon.

tessheinricks (changes requested):
Hi @mmmavis thank you for such swift work!

I've reviewed and it's looking great! My final comments are regarding the product viewport heights for mobile and tablet, I'm recommending that we reduce their height so that the quiz module does not extend beyond the screen size. However, everything else looks and is functioning amazingly! So if this is something you would like to save for the final cleanup ticket, I would be comfortable with that as well. 

I'm suggesting that we reduce the mobile products viewport height to showcase 4 and a half at products maximum (more similar to what is shown within the figma).
Mobile_Recommendation

For tablet I am recommending the same, with a reduction to 5 and a half products maximum.
Tablet_Recommendation

Again, please do not hesitate to shift these adjustments to post MVP if they will create launch concerns. The quiz section still operates beautifully and this is more so an optimizing concern. 

Thanks again for all your hard work on this!

mmmavis:
@tessheinricks thanks for the recommendation. I've reduced the height from 60% of the screen viewport height to 45%. I want to avoid setting it to a fixed pixel value or by the number of product items as these approaches might not work well responsively or when users have custom font settings on their browser (e.g., extra large font type by default).

tessheinricks (approved):
Looking amazing @mmmavis! Thanks for all your hard work!

danielfmiranda (approved):
@mmmavis Amazing work as always! After taking a look through the code and testing, everything is looking good to me. Approved 👍

mmmavis:
Lovely! Thank you both for the reviews!

Issue #10159, opened 2023-02-15
[EPIC] PNI Refactoring Stage 1


mmmavis:
As discussed in the [spike ticket](https://github.com/mozilla/foundation.mozilla.org/issues/9948), let's start the refactoring work with low-risk tasks. I've broken them down into three categories and will be filing tickets for individual tasks.

1. Extract code dupes and turn them into functions. This will reduce the code by hundreds+ lines. **[Tickets filed under this Stage 1 Epic]**
2. Use JSDoc style to document every functions and modules. We have quite a bit of functions with long implementation. Many functions look like they have gone through rounds of quick patches and now their function names don’t represent what they do anymore. Ideally, we would want to break the long implementation into smaller functions to make sure each function has one sole purpose. To do so, we will first need to know what the current functions do, which we can achieve by adding documentations to existing functions and modules. **[Tickets filed under this Stage 1 Epic]**
3. Update function names. There are some functions with very similar names but have quite different purposes. We should update them to make them more distinguishable. **[Tickets to be filed after Step 1 and 2 tickets are done]**
4. Audit to see if JS files need to be moved to a diff directory (for better file organization)
5. Change [utils.js](https://github.com/mozilla/foundation.mozilla.org/blob/1b0298d1a23c78f9696fc5e315627cb90f041c35/source/js/buyers-guide/search/utils.js#L15) so functions are not wrapped in a class. **[Mavis to file ticket]**

tbrlpld:
@mmmavis I have also notices in the JS that some util modules group all their functions on a single object the is imported. If this is a style we want to keep, can we add some description as to what is supposed to be useful about this? I don't quite understand it. If we want it for namespacing, I feel like `import "module-name"` or `import * as newName from "module-name"` provides the same benefits.

mmmavis:
@tbrlpld Thanks for raising this issue. I haven't paid much attention to that front and I'm unable to give opinions until the codebase is more readable and digestable lol 😅 Do you think we can revisit this at a later stage? I can make a note in the epic so it doesn't get lost.

tbrlpld:
@mmmavis Sure, we can revisit this. I was mostly wondering if this is something that has an obvious reason to a more seasoned JS dev that I am missing. If this does not sound like the natural thing to do to you too, yea, adding it to / noting it in the epic would be great.

mmmavis:
Note added to epic description.

[...]

Issue #9813, opened 2022-12-06
[PNI] Product grid performance - Reduce queries to get products data


tbrlpld:
## Description

This is a follow-up ticket to #9545 and #9812. 

The investigation in #9545 has shown that the `get_context` method generates 209 database queries to render 50 items. That means the method generates `4 * N` queries. Consider that it is generally deemed bad performance when `1 + N` queries are required to retrieve information about `N` items. 

I presume (though this is not confirmed) that many of these queries have to do with fetching related data like translations, images, categories etc. The use of `select_related` and `prefetch_related` should help in many cases.

Related issues: #9948

## Acceptance criteria

(These criteria assume a sufficiently large number `N` of products, e.g. `N = 50`)

- [ ] Soft target: As a developer, I can confirm from the unittests that the `get_context` method (or other dedicated method) requires **no more than `N`** number of database queries to retrieve `N` products.
- [ ] Real target: As a developer, I can confirm from the unittests that the `get_context` method (or other dedicated method) requires **significantly less than `N`** number of database queries to retrieve `N` products.

Issue #9812, opened 2022-12-06
[PNI] Product grid performance - Retrieve data in view


tbrlpld:
## Description

This is a follow up ticket to #9545. 

The investigation in #9545 has shown that a large number of queries is generated in the template when rendering the product grid on the buyer's guide homepage. 

Rendering 50 products generated 955 database queries, while executing the `get_context` method resulted in 209 database queries. 

The goal of this ticket would be to reduce the amount of extra data fetching that happens because of the template. All the necessary data should be retrieved with the queries that are already generated by the view. 

While this is not confirmed, I assume that many of these queries have to do with repeated fetching or additional related data, like categories, translations and the product images. 

Instead of retrieving this information in the template, make sure that the view / `get_context` method passes the necessary information to the template and uses caching where appropriate to avoid duplicate retrieval. Also, the use of `select_related` and `prefetch_related` should help in many cases.

Related issues: #9948

## Acceptance criteria

- [ ] As a developer, I can confirm from the unit tests that it does not take significantly more database queries to render the full page than it takes to retrieve the products in the `get_context` method.
- [ ] As a developer, I can confirm from the unit tests that the number of database queries to render the full buyer's guide homepage has been reduced significantly below 955.
- [ ] As a developer, I can confirm from the unit tests that the number of database queries generated in the `get_context` method of the buyer's guide homepage has not been significantly increased above 209.

PR #11214, opened 2023-10-05
[PNI] New PNI vote tallying model

jhonatan-lopes:
# Description

This PR replaces the current voting model by a vote model that registers each vote as a row in the database. This should allow for a more performant fetch of voting data.

A `ProductVote` model was introduced to hold the value of a single vote. The collection of votes is then tied to a `ProductPageEvaluation` instance, with a ForeignKey to a `ProductPage`. We can then use a `SynchronizedField` property to sync evaluations from translated instances. This removes the need to fetch an "original_product", reducing the number of queries.

In addition, holding the votes in a separate table makes it easier to do aggregations, such as getting the number of votes per bin or the average creepiness. It also separates data storage from data display. If a need arises to display a different number of bins than 5, for instance, the code can easily be changed to do that without changing the data inside the database. 

Since the current voting data does not register the value of each vote (only the cumulative creepiness), it has to be converted to an equivalent amount of votes that results in the same cumulative creepiness. This is handled by a data migration.

:warning: This deployment should require some downtime. Please disable Heroku's pre-boot before deploying this feature :warning:

This is the first stage in the PNI refactoring (#10159) and should address some of the concerns in #9813, #9812.

Link to sample test page:
Related PRs/issues: #10159

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

**Merge Method**


danielfmiranda (approved):
Thanks @jhonatan-lopes for your great work! Everything looks good to me, I also set a reminder for the following prod deploy day to disable pre-boot 👍 approved!

PR #11215, opened 2023-10-05
Container bash utility

jhonatan-lopes:
# Description

Add utility to run bash in a local container. Defaults to the "backend" one. Side note: we should rename this to `web`/`webapp` as it now has frontend and backend stuff in here.


mtdenton (approved):
LGTM!

PR #11216, opened 2023-10-05
Migrate node install script

### LINKED ISSUE #11130: Migrate node install script
We're getting a notice during environment setup about the script we're using to install Node in our Dockerfile being deprecated:

migrate

Let's migrate. Instructions are [here](https://github.com/nodesource/distributions/wiki/How-to-migrate-to-the-new-repository)

jhonatan-lopes:
# Description

Migrate deprecated node install script for new version. Use `nvm` to install node based on `.nvmrc`. This is simpler than the proposed migration script.

Fixes #11130


mtdenton (approved):
Yup! Thanks for taking this on @jhonatan-lopes

PR #11238, opened 2023-10-09
Update nodejs install script

jhonatan-lopes:
# Description

The install script updated in #11216 is giving an error when the server starts:

```sh
foundationmozillaorg-backend-1   | 14:20:45 system           | frontend-watch.1 started (pid=9)
foundationmozillaorg-backend-1   | 14:20:45 frontend-watch.1 | /bin/sh: 1: npm: not found
foundationmozillaorg-backend-1   | 14:20:45 system           | frontend-watch.1 stopped (rc=127)
```
In fact, after installing, it seems that the `~/.profile` is not picked up and neither `nvm` nor `npm` are recognised when starting a shell.

Attempts to fix this have been in vain. Instead, this PR migrates to the [official node installation script](https://github.com/nodesource/distributions/wiki/How-to-migrate-to-the-new-repository) without `nvm`.

Link to sample test page:
Related PRs/issues: #

# Testing

1. Run `inv setup`
2. See that command finishes without errors
3. Once it finishes, run `inv start` and check that both processes in `backend` container start without errors
4. Run `inv sh` and then run `node --version`. This should give a version `18`.

**Merge Method**

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -26,7 +26,8 @@ RUN npm run build
 #
 # Note: This stage builds the base image for production. Presently we are not
 # using this on the production site, but only use it as base for the dev build.
-FROM python:3.9.9-slim as base
+# Pin "bullseye" as it matches Ubuntu 20.04 -- Heroku 20 stack currently used in production.
+FROM python:3.9-slim-bullseye as base
 
 # Install dependencies in a virtualenv
 ENV VIRTUAL_ENV=/app/dockerpythonvenv
@@ -68,10 +69,12 @@ EXPOSE 8000
 
 # Install operating system dependencies.
 RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
+    ca-certificates \
     build-essential \
     libpq-dev \
     curl \
     git \
+    gnupg \
     gettext \
     && apt-get autoremove && rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -26,7 +26,8 @@ RUN npm run build
 #
 # Note: This stage builds the base image for production. Presently we are not
 # using this on the production site, but only use it as base for the dev build.
-FROM python:3.9.9-slim as base
+# Pin "bullseye" as it matches Ubuntu 20.04 -- Heroku 20 stack currently used in production.
+FROM python:3.9.9-slim-bullseye as base
```
jhonatan-lopes, line 30: This is a related change but a good one to make. We should pin the Debian version to avoid surprises in the future.

Dockerfile
```diff
@@ -68,10 +69,12 @@ EXPOSE 8000
 
 # Install operating system dependencies.
 RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
+    ca-certificates \
     build-essential \
     libpq-dev \
     curl \
     git \
+    gnupg \
```
jhonatan-lopes, line 77: Those are necessary packages to use this installer.

tomusher: question: If you only need these in the dev container, would it make sense to move them in to the `apt-get` run in the dev stage (around line 140)?

Dockerfile
```diff
@@ -120,6 +123,18 @@ FROM base as dev
 # Swap user, so the following tasks can be run as root
 USER root
 
+# Install node (Keep the version in sync with the node container above)
+# Download and import the Nodesource GPG key
+# This step is necessary as the previous method of using setup_18.x is now deprecated.
```
tomusher, line 128: suggestion: Remove this last comment line as it requires knowledge of previous configurations to make sense.

Dockerfile
tomusher: (although just realised you'd need to run `apt-get` twice anyway, but probably still worth limiting dependencies to where they are required)

Dockerfile
jhonatan-lopes: Done

Dockerfile
jhonatan-lopes: Done

Dockerfile
```diff
@@ -123,9 +121,15 @@ FROM base as dev
 # Swap user, so the following tasks can be run as root
 USER root
 
+# Install `psql`, useful for `manage.py dbshell`, and dependencies for installing nodejs
+RUN apt-get update --yes --quiet && apt-get install --yes --quiet --no-install-recommends \
+    postgresql-client \
+    ca-certificates \
+    gnupg \
+    && apt-get autoremove && rm -rf /var/lib/apt/lists/*
```
tomusher, line 129: suggestion (non-blocking): It doesn't matter that much for the dev image, but this last line will slow down the next `apt-get update`, and because you're not doing the same thing after the `nodejs` update, the final image will be a bit bigger.

You could both `apt-get` calls, and the commands to add the `nodesource` repo in to a single `RUN` command to keep this layer quite small, but up to you whether you think that's worthwhile.

jhonatan-lopes: That makes total sense, thanks! I've moved that line to after the node installation

tomusher:
Looks good to me!

mtdenton:
Thanks for this @jhonatan-lopes and thank you @tomusher for all of the awesome input!

Issue #11213, opened 2023-10-04
[PNI Privacy Index] Implement Quiz section (Part 3) 


mmmavis:
This is part 3 of Quiz section work which focuses on adding share buttons to the results screen.

## Mockups

- [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1090%3A13156&mode=dev)
- [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1635%3A8065&mode=dev)

## Dev Tasks

- [x] Share buttons on the results screen are linked with Shareprogress

kristinashu:
I will provide the three share progress links!

kristinashu:
Sorry @mmmavis, now I'm confused. Do the creep-o-meter share buttons use Share Progress https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11095#issuecomment-1745691528? I'm trying to find them in the Share Progress app to use as a reference but am having trouble.

mmmavis:
Hey @kristinashu , you are totally right about it! Petitions are using Shareprogress buttons but PNI doesn't 😓

kristinashu:
Thank you Mavis for looking into this! Which option is easier for you to implement?

danielfmiranda:
Hi @kristinashu! I have been tasked with this ticket, which is asking that I implement the share buttons for the results page of the quiz.

I was wondering, what should these button's share messages be?

For example, on the product page, when a user clicks a share button the message is pre-populated with: 

<img width="581" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/5a7006b6-e6cc-475c-a728-6d7347b61c4c">


Also, if you would like to send me the share progress links/ids, I may be able to implement them as well if needed

mtdenton:
I just made the buttons in Shareprogress:

Head Script
`<script src="//c.shpg.org/352/sp.js"></script>`

Email
`<div class='sp_227930 sp_em_large' ></div>`

Facebook
`<div class='sp_228230 sp_fb_large' ></div>`

Twitter
`<div class='sp_228231 sp_tw_large' ></div>`

[...]

PR #11239, opened 2023-10-09
11213 quiz share buttons

danielfmiranda:
# Description

Link to sample test page: https://foundation-s-11213-quiz-ldr5ma.mofostaging.net/en/privacynotincluded/articles/test/
Related PRs/issues: #11213

This PR adds the social share buttons to the bottom of the results panel found after taking the PNI product quiz.

Through the use of javascript the buttons are responsive, and are of the "circle mini" variation on mobile and tablet, and then turn into the "rectangle" variation on desktop and larger.

When the user clicks on either the Facebook, Twitter, or Email buttons, a new window opens and they are prompted to share through the use of shareprogress.

When the user clicks on the copy button, the URL is copied to their clipboard.

# Screenshots

**Desktop:**

Screenshot 2023-10-09 at 14-35-00 Privacy Not Included A Buyer’s Guide for Connected Products

**Mobile:**


source/js/buyers-guide/components/social-share/product-quiz-share-buttons.jsx
```diff
@@ -0,0 +1,120 @@
...
+        shareProgressButton.click();
+      }
+    } else {
+      const linkButton = event.target;
+
+      linkButton.classList.add("copied");
+      linkButton.title = window.gettext("Copied");
+      const spanElements = linkButton.querySelectorAll("span");
+
+      // Iterate through each <span> element and update its inner text
```
danielfmiranda, line 40: Since the copy button features two spans (one that is "Screenreader only" for mobile and one that renders the word "copy", we are updating both to "Copied", that way if a user clicks copy on mobile and resizes the screen, the desktop button will still say "Copied" as well

danielfmiranda:
Wanted to leave a comment that It was a little tricky to get our share buttons to be responsive (circular on mobile/tablet and rectangular on desktop) but I was able to do so through the use of JS. However, if the code reviewers find the JS to be a little too "hacky",  I think leaving the buttons as rectangular for all screen sizes would simplify this PR quite a bit! 

Screenshot 2023-10-09 at 15-41-26 Privacy Not Included A Buyer’s Guide for Connected Products

nancyt1 (approved):
All looks good to me, thank you!

mmmavis (changes requested):
@danielfmiranda Thanks for taking on this ticket and for creating a new file to host the share button code!

~~Generally we want to avoid using `document` related functions to manipulate the DOM in React code. If we have to, we should switch to use `ref` instead (see [React doc](https://react.dev/learn/manipulating-the-dom-with-refs)).~~ I re-reviewed the code again and realized this might not be possible for this use case.

I would also recommend replacing `resize` handler by simply applying Tailwind classes like `tw-hidden medium:tw-block` to toggle visibility of the two different sets of the buttons (full size vs mini circle) according to the screen sizes. This way we can avoid the more costly operations happening in the `resize` handler and also allow us to turn `ProductQuizShareButtons` to a stateless/functional component.

Let me know what you think or if you have any question! :)

danielfmiranda:
Thanks Mavis! The two set of button approach sounds great to me 👍 I have updated my PR to use that approach and re-requested review

source/js/buyers-guide/components/social-share/product-quiz-share-buttons.jsx
```diff
@@ -0,0 +1,108 @@
...
+  }
+
+  render() {
+    const shareButtonClasses =
+      "tw-btn tw-bg-white tw-h-22 tw-w-22 tw-border-black hover:tw-bg-black hover:tw-text-white btn-share after:tw-hidden";
+
+    const mobileShareButtons = (
+      <div className="circle share-button-group tw-justify-center tw-flex medium:tw-hidden">
+        <div className="subgroup">
+          <button
```
mmmavis, line 36: Since the `<button>` on mobile and desktop version are almost identical except their inner element, what do you think about creating a function to render them? Something like

```jsx
renderFacebookButton(child) {
    return (
      <button
        className={`${this.shareButtonClasses} facebook-share`}
        onClick={(e) => this.shareButtonClicked(e, "share-progress-fb")}
      >
        {child}
      </button>
    );
  }
```

And you should be able to build the button group by calling `this.renderFacebookButton(<span className="sr-only">Share on Facebook</span>)`, for example.

mmmavis (changes requested):
Thanks for updating the PR, @danielfmiranda ! I just have one more suggestion on making the code "DRYer". 😁

source/js/buyers-guide/components/social-share/product-quiz-share-buttons.jsx
danielfmiranda: Great idea Mavis! I have updated this PR to do the following and re-requested review 👍

mmmavis (approved):
Perfect! Thank you @danielfmiranda ! 🎉

mmmavis:
I'm merging this awesome work!

PR #11242, opened 2023-10-10
PNI CC Page - fixes and improvements

mmmavis:
https://foundation-s-pni-cc-qui-n071am.herokuapp.com/en/privacynotincluded/articles/annual-consumer-creep-o-meter/

Hi @danielfmiranda , this is a design tweak based PR and Tess has already review everything in the list below. Could you give code changes a quick look as well as double check the signup form in Quiz section works on the review app link? I forgot to fix the basket url in review app config and that was why [it wasn't working for Tess when she tested out the form](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11242#pullrequestreview-1672073822). Thank you!

---

### Quiz section

- updated some text sizes
- updated some spacing
- gave each screen in the quiz flow a minimum height of 530px
- removed dupe text on the results screen (description was repeating the header)
- hyperlinked `privacynotincluded.org` on the results screen
- added rotate icon to Retake Quiz button on the results screen

### Other

- optimized image files
- replaced asterisk face with the darker `:(` version
- lazy load some images
- replaced `face-5.svg` with the pre image optimization version (image optimization stripped off the texture from this particular image)


tessheinricks (changes requested):
Hi @mmmavis looking great!

All devices:
Unfortunately, it seems like the email subscription within the quiz is not working (when all the information is filled, the "Sign Up" button becomes active but does not do anything when pressed)

Tablet:
It seems like the icons and copy within the share buttons are a bit misaligned. 
image


Thanks in advance!

mmmavis:
@tessheinricks 

>  All devices:
Unfortunately, it seems like the email subscription within the quiz is not working (when all the information is filled, the "Sign Up" button becomes active but does not do anything when pressed)

Ah right, this is actually something to do on the review app config which I forgot to set it up before sharing the test link. I'll update the config.

> Tablet:
It seems like the icons and copy within the share buttons are a bit misaligned.

Will look into this.

mmmavis:
@tessheinricks The misaligning issue on share buttons cannot be easily fixed due to [the current styling rules](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/f8002a347c7c985a985cec978fe607777fe45453/source/sass/components/share-button-group.scss#L55) applied to ALL the share buttons on the site. I tried to look for why we forced them to have that `align-items: baseline !important;` rule but there was no clear info associated with the code commit so at this point, I'm hesitant to update the rule as it will affect to all share buttons on the site.

Tess, since you are OOO. I'm gonna dismiss your review and ask @danielfmiranda to help with testing on the newsletter signup form. I will merge this PR after I get R+.

danielfmiranda (approved):
Hi @mmmavis! Just gave the quiz and newsletter a test and can confirm its working. Also, after taking a look at the code, everything looks great. Approved!

mmmavis:
🎉

Issue #11135, opened 2023-09-18
[PNI Privacy Index] Implement Stay In Control section


mmmavis:
<img width="1724" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/34ddb2af-e952-4331-8a70-4785d5dc6630">

## Dev Tasks

- [ ] Build "Stay in Control" Section according to the [desktop](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1090%3A13156&mode=dev) and [mobile](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?node-id=1090%3A13156&mode=dev) mockups

**We might need to come up with a non-slider version if we run short on development time.**

mmmavis:
Closing. PR has been merged into the feature branch. 🎉

PR #11245, opened 2023-10-10
PNI CC Page - MVP version of 'Stay in Control' section

mmmavis:
https://foundation-s-11135-pni--ku8eqo.herokuapp.com/en/privacynotincluded/articles/annual-consumer-creep-o-meter/

Related PRs/issues: #11135

Things we might want to leave as post-MVP:

1. Shared gradient background on the three blocks: 
There is no straightforward to implement this. We can probably achieve the effect by applying individual gradient segment to each block to make them look like they are sharing the same gradient.

2. Image and header positions on smaller screens.
Because the images are decorative, I implemented them as background images (instead of ``) for better a11y. This, however, means that we lose the ability to auto wrap the header text when screen gets resized. We can finetune the positions with a series of styling rules but will be challenging to make them look perfect on every screen sizes.


tessheinricks (changes requested):
Hi @mmmavis, amazing work! It's looking great and I agree that we can shift some of those final details to post launch (will add 1 & 2 to the clean up ticket).

1. Yes, I understand this is tricky. Your solution proposal is actually how we achieved this effect in figma (with a gradient applied to each block). I've attached images if it helps at all to reference. 

2. This is definitely something we can find tune, but it seems to perform well at current!

3. The only outstanding change is the body copy on desktop is too large and should be the [Desktop/Body] style of 18/27.  
Screenshot 2023-10-11 at 10 13 45 AM

Screenshot 2023-10-11 at 10 15 23 AM

Thank in advance and let me know if you have any questions!

mmmavis:
@tessheinricks thanks for the feedback. When I inspected it on Figma, it showed me (see below) so I thought the effect was designed by covering all of the three block in one shared gradient. I guess it's time for me to get back to my Figma course 🙈 

image

> The only outstanding change is the body copy on desktop is too large and should be the [Desktop/Body] style of 18/27.

On it. Will update PR soon!

tessheinricks (approved):
Incredible work @mmmavis!! 🥇

danielfmiranda (approved):
Looks great!! Thanks @mmmavis

Issue #11185, opened 2023-09-28
[PNI Privacy Index] Cleanup & Final Edits


tessheinricks:
### This is a space to capture any final tweaking of the PNI Creep-O-Meter development

**Hero**
- [ ] **[P2]** Creep-O-Meter Emoji brought up to remain visible - top of speech bubble stem is brought more in line with top of emoji head
- [x] Speech bubble on mobile: Can be less wide for stem to be closer to left side of emoji face like mockup 
- [ ] Lead copy font weight: Appears to renders heavier than figma file (although inspector says it is using ExtraLight?)
- [ ] Asterisk illustration location between 768<1200 breakpoints: It gets obscured so perhaps placing it like mobile in upper left but larger
- [x] Swap current asterisk illustration with spot illo 1, the dark purple sad face one

**Quiz**
- [x] Add "Retake Quiz" icon in button
- [x]  **[P1]** Make scroll bar permanently visible (so users always know there is additional products to select)
- [x] **[P1]** Twitter share button links to Facebook
- [ ] **[P2]** Improve quiz result images load speed to be instant the first time (they have a bit of a delay on first load but once they have appeared they load instantly)
- [x] **[P2]** Share button icon disappears after clicking on the button
<img width="600" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/89d7a6f7-8f30-47ff-98db-c755a47489c1">

- [ ] Revise [See Results button module](https://www.figma.com/file/idw7tpYbaRXCRqzk01poEX/Privacy-Index-Page?type=design&node-id=2872%3A10272&mode=design&t=J3ZG0pJGyLNmsclS-1):
Adding a button for users who don't own any of the products or don't select any products. The see results button should be inactive until a user selects products, and if a user selects the "I don't own any" button they should continue through the flow.

![Quiz Addition - _I don't own any_](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/141766274/0ba29519-5108-4f5c-88a4-2b18f1c259f7)

**Stay in Control** 
- [ ] Refine shared colour gradient 
- [ ] Refine header & image relationship for mobile & smaller screen sizes
- [x] Within "Wield Your Pocketbook" Add [link](https://foundation.mozilla.org/en/blog/applause-for-apples-idfa-decision/) to "Apple" - Ensure page opens in a new tab
[...]

PR #11273, opened 2023-10-17
PNI CC Page - quick fixes and tweaks

mmmavis:
https://foundation-s-pni-new-ta-7h7alv.mofostaging.net/en/privacynotincluded/articles/annual-consumer-creep-o-meter/

I created this PR mainly to address `make privacynotincluded.org link in Quiz Result screen to open on a new tab`.

As bonus 😅 - I adjusted a few other things mentioned in #11185 
 
- Speech bubble on mobile: Can be less wide for stem to be closer to left side of emoji face like mockup
- [P2] Share button icon disappears after clicking on the button

and added a quick code change that that might improve

> - [P2] Improve quiz result images load speed to be instant the first time (they have a bit of a delay on first load but once they have appeared they load instantly)


source/js/buyers-guide/components/product-quiz/product-quiz.jsx
```diff
@@ -304,7 +304,13 @@ class ProductQuiz extends Component {
       <div className="tw-flex tw-flex-col medium:tw-flex-col-reverse tw-dark">
         <div className="tw-text-center">
           <div className="tw-mx-auto tw-w-[80px] tw-h-[80px] medium:tw-w-[180px] medium:tw-h-[180px]">
-            <img src={resultType.imgSrc} alt="" width="100%" height="100%" />
+            <img
+              src={resultType.imgSrc}
+              alt=""
+              width="100%"
+              height="100%"
+              loading="eager"
```
mmmavis, line 312: > Improve quiz result images load speed to be instant the first time (they have a bit of a delay on first load but once they have appeared they load instantly)

Adding `loading="eager"` here which may or may not help with the above task because the image is rendered dynamically by a state update anyway. But it doesn't harm leaving this attribute here...

kristinashu (approved):
Looks good, thank you Mavis the bonus work! I think the quiz image is loading faster now.

source/js/buyers-guide/components/product-quiz/product-quiz.jsx
danielfmiranda: Interesting! I never knew about this attribute 📝

danielfmiranda (approved):
LGTM! 👍

mtdenton:
Thanks for putting in the work Mavis!

I think you're correct in thinking that `loading="eager"` doesn't make the images load faster in this context since they're already in the viewport and would trigger the load anyhow. Having a smaller source image size would have been the best route here and/or putting them into a more optimized file format is how we would need to handle this to get it to load faster, but there was no time for this before launch. Either way, seems to work just fine even on a throttled connection. We can optimize later if this is really necessary

Issue #11173, opened 2023-09-26
DonateHelpPage FA form


danielfmiranda:
## Description
 ⚠️ **[Note]** - Code changes should only start once the dev has finished styling the FA form, and received approval from the design team. 

![Screenshot 2023-09-25 at 23-36-53 Donate Help](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/18314510/7e86e358-f9ed-40ef-a4f9-9fbe68505650)

The goal of this ticket is to style and implement the donate help pages [FormAssembly form](https://mozillafoundation.tfaforms.net/12), keeping in mind these guidelines:
- The form should be styled to fit MoFo Brand Guidelines as much as possible
- The form should hide the captcha box when the following options are selected from the menu: "Technical Support for Firefox or Thunderbird" and "To Request a feature for Firefox".
- The form should pull language/locale code from main page URL and add it to the form into a hidden field (see [epic ticket](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10840#issue-1796903917)
for more context)

## Dev Tasks
- [ ] Style the FormAssembly form as close as possible to fit Mofo Brand guidelines, and get approval from design.
- [ ] Implement recaptcha box and locale code functionality. According to [this comment](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10840#issuecomment-1635032472) this functionality has been implemented by the Thunderbird team, and can be used for reference upon request.
- [ ] Copy and paste the formassembly code into the donate pages template, underneath the intro section. For reference, we can take a look at tickets in [this previous formassembly epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10532)

## Acceptance Criteria

- [ ] As a CMS user, I can visit a `DonateHelpPage` and see that the design-approved FormAssembly form is rendering as expected. 
- [ ] As a page visitor, form should hide the captcha box when the following options are selected from the menu: "Technical Support for Firefox or Thunderbird" and "To Request a feature for Firefox".

danielfmiranda:
All FormAssembly form work has been complete and merged into the feature branch. Closing this ticket!

PR #11306, opened 2023-10-24
Donate Help Page FA Form basic rendering

danielfmiranda:
# Description

Link to sample test page: https://donate-foundation-s-11173-dona-jtqs8w.mofostaging.net/en/help/
Related PRs/issues: #11173

This PR handles the basic rendering of the unstyled FA form.

# Screenshots

Screenshot 2023-10-25 at 20-13-25 Donate Help


mmmavis (changes requested):
Hey @danielfmiranda , 

At a glance it's not possible to tell what code is copied from FormAssembly vs our own custom code so it's tricky to do code review. Could you mimic what we have for petition form ([formassembly_body.html](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_body.html#L3-L48) and [formassembly_head.html](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/formassembly_head.html#L3-L26)) and add notes to the top of the two FA related files? Having notes/instructions will also help future code edits from breaking the form.

We'll also need to edit default CSP values on various files for recaptcha to work!

Thanks!

network-api/networkapi/templates/donate/fragments/formassembly_body.html
```diff
@@ -0,0 +1,168 @@
...
+                    'es-MX': 'tfa_231',
+                    'es-AR': 'tfa_231',
+                    'es-ES': 'tfa_231',
+                    'es-CL': 'tfa_231'
+                }
+
+    // Assign the locale, or default to 'Other' if the locale isn't supported.
+                varLocale.value = localeMap[locale] || 'tfa_227';
+            });
+
```
danielfmiranda, line 73: Hi @mmmavis just calling out this custom localization JS that I found from Mia and the thunderbird team. I am probably going to have to update this in my localization PR as they are referencing locales we do not use on the foundation site (ex: en-US, es-MX, etc.).

danielfmiranda:
Hi @mmmavis! Thanks for the review. 

I have updated this PR with what we discussed on slack which includes:
- Updated env.default to include the URLs needed to render the form / recaptcha
- Some comments at the top of the files, which will be updated as I make more changes to the form such as styling/localization


This pr is now ready for re-review whenever you find a free moment. Thanks!

mmmavis (changes requested):
Looking great! If we can update CSP configs in `app.json` and `continous-integration.yml` as well, that'd be rad 💯

danielfmiranda:
Thanks for the review @mmmavis! Just updated both `app.json` and `continuous-integration.yml` 👍

mmmavis (approved):
💯

PR #11331, opened 2023-10-27
[PNI] Reduce product vote/creepiness memory consumption

jhonatan-lopes:
# Description

Reduces the memory consumption on PNI's homepage by pre-calculating all product's creepiness on the database via an annotation. This way, we don't need to prefetch all vote objects (a significant number), which was causing an increase in memory usage.

Link to sample test page:
Related PRs/issues:

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

**Merge Method**

### DIFF
--- a/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
@@ -5,6 +5,7 @@ from django.apps import apps
 from django.conf import settings
 from django.core.cache import cache
 from django.db import models
+from django.db.models.functions import Coalesce
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils.text import slugify
 from django.utils.translation import gettext
@@ -24,7 +25,6 @@ from wagtail_localize.fields import SynchronizedField, TranslatableField
 
 from networkapi.utility import orderables
 from networkapi.wagtailpages.pagemodels.base import BasePage
-from networkapi.wagtailpages.pagemodels.buyersguide.utils import sort_average
 from networkapi.wagtailpages.templatetags.localization import relocalize_url
 from networkapi.wagtailpages.utils import (
     get_language_from_request,
@@ -524,11 +524,11 @@ def get_product_subset(cutoff_date, authenticated, key, products, language_code=
         products = products.live()
 
     products = products.prefetch_related(
-        "evaluation__votes",
         "image__renditions",
         "product_categories__category",
-    )
+    ).annotate(
+        _average_creepiness=Coalesce(models.Avg("evaluation__votes__value"), float(0))
+    ).order_by("_average_creepiness")
 
-    products = sort_average(products)
     cache.get_or_set(key, products, 24 * 60 * 60)  # Set cache for 24h
     return products
--- a/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
@@ -523,12 +523,14 @@ def get_product_subset(cutoff_date, authenticated, key, products, language_code=
     if not authenticated:
         products = products.live()
 
-    products = products.prefetch_related(
-        "image__renditions",
-        "product_categories__category",
-    ).annotate(
-        _average_creepiness=Coalesce(models.Avg("evaluation__votes__value"), float(0))
-    ).order_by("_average_creepiness")
+    products = (
+        products.prefetch_related(
+            "image__renditions",
+            "product_categories__category",
+        )
+        .annotate(_average_creepiness=Coalesce(models.Avg("evaluation__votes__value"), float(0)))
+        .order_by("_average_creepiness")
+    )
 
     cache.get_or_set(key, products, 24 * 60 * 60)  # Set cache for 24h
     return products
--- a/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
@@ -5,7 +5,6 @@ from django.apps import apps
 from django.conf import settings
 from django.core.cache import cache
 from django.db import models
-from django.db.models.functions import Coalesce
 from django.shortcuts import get_object_or_404, redirect, render
 from django.utils.text import slugify
 from django.utils.translation import gettext
@@ -528,7 +527,7 @@ def get_product_subset(cutoff_date, authenticated, key, products, language_code=
             "image__renditions",
             "product_categories__category",
         )
-        .annotate(_average_creepiness=Coalesce(models.Avg("evaluation__votes__value"), float(0)))
+        .with_average_creepiness()
         .order_by("_average_creepiness")
     )
 
--- a/network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
@@ -327,12 +327,12 @@ class ProductPageEvaluation(models.Model):
         """
         try:
             return self._total_creepiness
-        except AttributeError:
-            raise ValueError(
+        except AttributeError as e:
+            raise AttributeError(
                 "Can't calculate total creepiness without `_total_creepiness` annotation. "
                 "Make sure to annotate the evaluation queryset by calling `.with_total_creepiness()` "
                 "method before accessing this property."
-            )
+            ) from e
 
     @property
     def average_creepiness(self):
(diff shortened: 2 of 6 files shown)


network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
```diff
@@ -523,12 +523,14 @@ def get_product_subset(cutoff_date, authenticated, key, products, language_code=
...
-    products = products.prefetch_related(
-        "evaluation__votes",
-        "image__renditions",
-        "product_categories__category",
+    products = (
+        products.prefetch_related(
+            "image__renditions",
+            "product_categories__category",
+        )
+        .annotate(_average_creepiness=Coalesce(models.Avg("evaluation__votes__value"), float(0)))
```
jhonatan-lopes, line 531: We need to deal with cases where there aren't any votes yet, in which case we want the creepiness to be 0. I tried using the `default` parameter on `models.Avg` but it would return `None` instead. Hence, the `Coalesce`.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -292,26 +293,44 @@ class ProductPageEvaluation(models.Model):
...
+        except AttributeError:
+            return self.votes.count()
 
     @property
     def total_creepiness(self):
-        return sum([vote.value for vote in self.votes.all()])
+        try:
+            return self._total_creepiness
+        except AttributeError:
+            return sum([vote.value for vote in self.votes.all()])
```
tomusher, line 306: **suggestion:** Should this throw an exception rather than falling back to a manual calculation? This would prevent someone from accidentally loading all those votes in to memory without realising it.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -292,26 +293,44 @@ class ProductPageEvaluation(models.Model):
 
     @property
     def total_votes(self):
-        return self.votes.count()
```
tomusher, line 295: **suggestion:** Add a comment indicating which QuerySet method should be used to populate `_total_votes`

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -292,26 +293,44 @@ class ProductPageEvaluation(models.Model):
...
+        try:
+            return self._total_votes
+        except AttributeError:
+            return self.votes.count()
 
     @property
     def total_creepiness(self):
```
tomusher, line 302: **suggestion:** Add a comment indicating which QuerySet method should be used to populate `_total_creepiness`

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -292,26 +293,44 @@ class ProductPageEvaluation(models.Model):
...
-        return sum([vote.value for vote in self.votes.all()])
+        try:
+            return self._total_creepiness
+        except AttributeError:
+            return sum([vote.value for vote in self.votes.all()])
 
     @property
     def average_creepiness(self):
```
tomusher, line 309: **suggestion:** Add a comment indicating which QuerySet method should be used to populate `_average_creepiness`

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -663,7 +682,16 @@ def total_vote_count(self):
...
+        try:
+            # Try an annotation made above the ProductPage level
+            return self._average_creepiness
+        except AttributeError:
+            try:
+                # It failed, let's try an annotation made at the ProductPageEvaluation level
+                return self.evaluation._average_creepiness
+            except AttributeError:
+                # We don't have an annotation, let's calculate it ourselves
+                return self.evaluation.average_creepiness
```
tomusher, line 694: **question:** Is there ever a case where it makes sense to have this fallback? With the DB optimisation available, perhaps this is another case where raising an Exception would help avoid future developers avoid potentially expensive calls?

network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
tomusher, line 531: **Suggestion:** Perhaps avoid the duplication of this calculation by using a [Prefetch](https://docs.djangoproject.com/en/4.2/ref/models/querysets/#django.db.models.Prefetch) object to retrieve the associated evaluation from an Evaluation queryset using the `with_average_creepiness` method?

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -663,7 +682,16 @@ def total_vote_count(self):
 
     @property
     def creepiness(self):
-        return self.evaluation.average_creepiness
+        try:
+            # Try an annotation made above the ProductPage level
+            return self._average_creepiness
```
tomusher, line 687: **Suggestion:** This logic might be simpler if the `Prefetch` annotation was used as above - you'd only need to check for the attribute on the Evaluation then.

tomusher (commented):
This is looking good! Just a few comments on duplication and future safety.

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
jhonatan-lopes: That's a great point, thanks! I suppose we can shoot ourselves in the foot now without noticing it! I'll remove this and throw an exception

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
jhonatan-lopes: You're right, I'll throw an exception here as well

network-api/networkapi/wagtailpages/pagemodels/buyersguide/homepage.py
jhonatan-lopes: That's a great suggestion, thanks! I was wondering how I could remove this duplication

jhonatan-lopes: After trying it, I'm afraid the duplication is inevitable. If we use a `Prefetch` object, it will do another query to prefetch the evaluation objects and then annotate the `_average_creepiness` property onto it. The issue now is that we can't order the original query anymore, i.e.

```python
products = (
        products.prefetch_related(
            "image__renditions",
            "product_categories__category",
            Prefetch("evaluation", ProductPageEvaluation.objects.with_average_creepiness()),
        )
        .order_by("evaluation___average_creepiness")
    )
```
doesn't work since `order_by` won't be able to solve the property in a single query.

Having to reorder the query in Python would be a performance hit, I think, so it justifies leaving the code duplication.

tomusher: Ah got you - the duplication could be warranted here then, although maybe it still makes sense to move it to a method on the Product QuerySet for future convenience?

jhonatan-lopes: True, a `ProductQuerySet` method would be more convenient and better documented. I'll do that!

network-api/networkapi/wagtailpages/pagemodels/buyersguide/products.py
```diff
@@ -293,26 +293,71 @@ class ProductPageEvaluation(models.Model):
...
+        """
         try:
             return self._total_creepiness
         except AttributeError:
-            return sum([vote.value for vote in self.votes.all()])
+            raise ValueError(
+                "Can't calculate total creepiness without `_total_creepiness` annotation. "
+                "Make sure to annotate the evaluation queryset by calling `.with_total_creepiness()` "
+                "method before accessing this property."
+            )
```
jhonatan-lopes, line 335: I'm not sure if `ValueError` is the best type of exception here as it is more of a "Don't do this thing or we'll get in trouble" :thinking:

tomusher: I think perhaps re-raising `AttributeError` might make sense - you're accessing an attribute and it doesn't exist, just for a special reason that you're explaining well in the exception.

Perhaps also worth using exception chaining here to keep the trace from the original exception:

```python
...
except AttributeError as e:
   raise AttributeError("...") from e
```

jhonatan-lopes: Done :+1:

tomusher (approved):
Looks great to me!

Issue #11448, opened 2023-11-20
Localize recaptcha on donate help page


danielfmiranda:
Stakeholders have found that the recaptcha box on the donate help page is not localized when visiting the page using a language other than the default locale (EN)

![Screenshot 2023-11-20 at 15-05-21 Donate Help](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/18314510/0914d692-120d-4f85-915d-8e19222b5463)

Taking a look at [recaptchas supported language codes](https://developers.google.com/recaptcha/docs/language), we should be able to fix this by using the `LANGUAGE_CODE` variable set at the top of the file, and updating the line [here](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/0fcd9c5d2d97362b9c143bf0f3254c83e550ce8c/network-api/networkapi/templates/donate/fragments/formassembly_head.html#L159) to include `hl={{LANGUAGE_CODE}}`. At the moment it is hardcoded to en_US.

PR #11450, opened 2023-11-20
localized DonateHelpPage recaptcha box

danielfmiranda:
# Description

Link to sample test page:
Related PRs/issues: #11448

This PR updates the `src` of the script `<script src='https://www.google.com/recaptcha/enterprise.js?onload=gCaptchaReadyCallback&render=explicit&hl=en_US' nonce="{{ request.csp_nonce }}" async defer></script>` to include `hl={{ LANGUAGE_CODE }}` instead of being hardcoded to `en_US`. That way the recaptcha box renders with the users active locale.

# Screenshots

EN (English):
Screenshot 2023-11-20 at 3 38 41 PM

ES (Spanish):
Screenshot 2023-11-20 at 3 38 53 PM

DE (German):
Screenshot 2023-11-20 at 3 39 10 PM

FR (French):
Screenshot 2023-11-20 at 3 39 22 PM

fy-NL (Frisian):
Screenshot 2023-11-20 at 3 39 40 PM

NL (Dutch):
Screenshot 2023-11-20 at 3 39 54 PM

PL (Polish):
Screenshot 2023-11-20 at 3 40 08 PM

pt-BR (Portuguese (Brazil)):
Screenshot 2023-11-20 at 3 40 24 PM

SW (Swahili):
Screenshot 2023-11-20 at 3 40 38 PM

### DIFF
--- a/network-api/networkapi/templates/donate/fragments/formassembly_head.html
+++ b/network-api/networkapi/templates/donate/fragments/formassembly_head.html
@@ -18,6 +18,9 @@ Please follow the steps below every time after you paste a new version of FormAs
     - Update script file name https://mozillafoundation.tfaforms.net/wForms/3.11/js/localization-en_US.js?v=...
         with {% fa_locale_code %}
         e.g., https://mozillafoundation.tfaforms.net/wForms/3.11/js/localization-{% fa_locale_code %}.js?v=...
+    - Update script file name https://www.google.com/recaptcha/enterprise.js?onload=gCaptchaReadyCallback&render=explicit&hl=en_US...
+        with {{ LANGUAGE_CODE }}
+        e.g., https://www.google.com/recaptcha/enterprise.js?onload=gCaptchaReadyCallback&render=explicit&hl={{ LANGUAGE_CODE }}...
     - Go to formassembly_body.html and make sure everything has been updated according to the instructions
 
 5. Update the revision note on the next line. Revision number can be found on https://mozillafoundation.tfaforms.net/versions/index/12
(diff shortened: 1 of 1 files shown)


mmmavis (changes requested):
Hey @danielfmiranda , 

I'm wondering if FA supports all languages our site supports? If not, we will need to fallback to `en` so that the script file can be properly loaded (e.g., using our custom `fa_locale_code` tag)

danielfmiranda:
Hi @mmmavis!! Thanks for the review 👍 

The external script that this PR updates is actually not directly related to form assembly, but instead a google script that is used the render the recaptcha box. 

Taking a look at the recaptcha docs [here](https://developers.google.com/recaptcha/docs/language), it appears that they support all of the languages that we support, and after testing I can confirm they do. For example, the screenshots above show a localized recaptcha box for every language in our [supported languages list](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/0fcd9c5d2d97362b9c143bf0f3254c83e550ce8c/network-api/networkapi/settings.py#L463-L471).

Taking a look at the code for `fa_locale_code`, it appears that we would lose support for fy-NL and SW if we were to use that tag.

However, I do see the benefit in sharing one tag.

Do you think we should stick with `LANGUAGE_CODE` as recaptcha supports all languages that the foundation site does?
Or, should I switch it out for `fa_locale_tag` to reduce complication, with the catch that the recaptcha box would lose support for fy-NL and SW.

Thanks!

mmmavis:
@danielfmiranda ah apologies! From my quick review yesterday I didn't realize we were talking about a script from Google. Thank you for your detailed explanation! In that case I would say sticking to your current implementation (using `LANGUAGE_CODE`) 😃

mmmavis (approved):
Thanks @danielfmiranda !

PR #11538, opened 2023-12-13
[PNI] Handle 100 average creepiness

jhonatan-lopes:
# Description

Fixes bug where products with 100 average creepiness would have a `KeyError` when trying to get the `average_bin` property.

Link to sample test page:
Related PRs/issues: #

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

**Merge Method**


network-api/networkapi/wagtailpages/tests/buyersguide/test_product_page_evaluation.py
```diff
@@ -286,8 +286,8 @@ def test_avg_bin_with_avg_vote_between_60_and_80(self):
         self.assertEqual(evaluation.average_creepiness, vote_value)
         self.assertDictEqual(evaluation.average_bin, {"label": "Very creepy", "localized": gettext("Very creepy")})
 
-    def test_avg_bin_with_avg_vote_between_80_and_100(self):
-        vote_value = self.fake.random_int(min=80, max=100)
+    def test_avg_bin_with_avg_vote_between_80_and_99(self):
```
mmmavis, line 289: Hi @jhonatan-lopes , thanks for catching this!

What do you think about renaming this `def` and the ones above a bit so at a glance, it's easy to tell what creepiness range is covered in each test?

e.g., 

```py
def test_avg_bin_with_avg_vote_between_1_and_20(self):
        vote_value = self.fake.random_int(min=1, max=19)
```

I'm wondering if changing them to the following so it's more explicit that the two numbers (`1`, and `19` in this case) are both inclusive.

```py
def test_avg_bin_with_avg_vote_from_1_to_19(self):
        vote_value = self.fake.random_int(min=1, max=19)
```

mmmavis (changes requested):

network-api/networkapi/wagtailpages/tests/buyersguide/test_product_page_evaluation.py
jhonatan-lopes: I think the ranges are correct. Take the `20 to 40`, for instance. For that one, the ideal scenario would be to have floats that range from 20 until 40 not including 40 itself (40 itself will belong in the next bucket, since `int( 40 // BUCKET_SIZE) = 2` which is the next bin). So on that bin we could have floats for average creepiness like `20`, `20.001`, `35`, `39`, `39.99`, `39.99999`, etc.

However, I'm only doing the test with one vote and each vote can only hold an integer value. Thus, the average creepiness and average vote will also hold integer values. So to make sure that I don't get the ends of the interval, I'm doing the `N - 1` here. I thought that would be simpler than mocking the `average_creepiness` call.

If that's confusing, I can add a comment with some explanation to the tests.

One alternative would be to do:

```
# Get a random float for the vote between 0 and 20 (not including ends) and convert that to an integer
vote_value = int(self.fake.pyfloat(min_value=0.01, max_value=19.99, right_digits=2))
```

For this case I would have to be careful with how the integer conversion works to make sure we're not getting the ends of the interval...

mmmavis: Got it! I'm coming from the perspective of this `def` name is using a different pattern than the rest.

From the code above:

```py
def test_avg_bin_with_avg_vote_between_1_and_20(self):
        vote_value = self.fake.random_int(min=1, max=19)
...
...

def test_avg_bin_with_avg_vote_between_20_and_40(self):
        vote_value = self.fake.random_int(min=20, max=39)
...
...

def test_avg_bin_with_avg_vote_between_40_and_60(self):
        vote_value = self.fake.random_int(min=40, max=59)
...
...

def test_avg_bin_with_avg_vote_between_60_and_80(self):
        vote_value = self.fake.random_int(min=60, max=79)
```
Notice the `max` argument is always one integer smaller than the upper bound number in the `def` name. Maybe for consistency's sake, we can update `def test_avg_bin_with_avg_vote_between_80_and_99(self):` to `def test_avg_bin_with_avg_vote_between_80_and_100(self):`?

I'm going to approve the PR since I'm just nitpicking on this. Please go with what you think is more appropriate. Thanks, @jhonatan-lopes !

mmmavis (approved):
Left a comment (nitpick). Feel free to merge this PR.

network-api/networkapi/wagtailpages/tests/buyersguide/test_product_page_evaluation.py
jhonatan-lopes: I see. I agree with you, it's just that the interval was messed up to being with. It should have ended on 99. That's why I have the 100 on a special case.

PR #11539, opened 2023-12-13
6650 - MozFest Dark Quote

chris-lawton:
# Description

- Changes the mozfest-base template to have one column
- Adds the `mozfest_dark_quote.html` template
- BE work to integrate this is to come

Desktop/Tablet/Mobile screens 

<details><summary>Details</summary>
<p>

Screenshot 2023-12-13 at 14 03 10

Screenshot 2023-12-13 at 14 03 19

Screenshot 2023-12-13 at 14 03 51

</p>
</details> 

Link to sample test page:
Related PRs/issues: #

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

**Merge Method**

### DIFF
--- a/network-api/networkapi/mozfest/templates/fragments/blocks/mozfest_dark_quote.html
+++ b/network-api/networkapi/mozfest/templates/fragments/blocks/mozfest_dark_quote.html
@@ -4,12 +4,12 @@
     <div class="tw-py-16 tw-bg-black tw-dark">
         <div class="tw-container">
             <div class="tw-row tw-justify-center">
-                <blockquote class="feature-quote">
-                    <div class="tw-quote tw-px-2 [&_p]:tw-h2-heading">
+                <div class="feature-quote">
+                    <div class="quote-content [&_p]:tw-text-white">
                         {{ self.quote|richtext }}
                     </div>
-                    <cite class="tw-text-gray-40 tw-h6-heading tw-max-w-4xl">{{ self.cite }}</cite>
-                </blockquote>
+                    <p class="tw-text-gray-40 tw-h6-heading tw-text-lg tw-max-w-4xl">{{ self.cite }}</p>
+                </div>
             </div>
         </div>
     </div>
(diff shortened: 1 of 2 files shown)


network-api/networkapi/mozfest/templates/fragments/blocks/mozfest_dark_quote.html
tbrlpld, line 1: Can you add some context why this is building a totally new block template over reusing the existing quote block that is mentioned on the ticket? It looks like we are reusing the CSS but not the HTML. Could this be handled as a variant in mostly CSS?

tbrlpld (commented):
@chris-lawton Can you add some context about the reuse of existing code and new code? It looks like we are using one existing CSS class but none of the markup. I am also a bit concerned about the mix of Tailwind and component CSS classes. This feels like it will be complex to maintain. Is there anything we can do to make the two components more alike / consistent to maintain?

network-api/networkapi/mozfest/templates/fragments/blocks/mozfest_dark_quote.html
chris-lawton: We've held some conversations since the ticket was created and it was decided that the dark background was enough to warrant it being a new block - i'll update the ticket to make this clearer. 

The dark quote block is heavily based of the existing `single_quote_block.html` template (https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/single_quote_block.html) which also contains a mix of tailwind and css classes. Unfortunately we can't use this template directly as it extends 
 `"./base_streamfield_block.html"` which wraps everything in a container (https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/base_streamfield_block.html#L5) meaning we couldn't achieve the full width black background seen in the design. 

The main difference between the dark quote block in the design and the existing one is the font size of the quote so i'll check in standup which one we should be following.

chris-lawton: Discussed in standup and we're going to follow the styles of the current quote block. Updated here: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11539/commits/ad110bc9f8425b71b0a918947c4da513f6abfea3

tbrlpld (approved):

PR #11545, opened 2023-12-14
9300 MozFest - List block updates

chris-lawton:
# Description

- Updates the listing block to be 12 cols wide when used on the MozFest pages. As this block is used on the rest of the site i've done this using the `mozfest` body class modifier. 
-  Move spacing classes to the heading tag to remove the need for an extra div

As we've re-used the listing block for these styles there may be some discrepancies between the design and the build. 

Desktop/Tablet/Mobile:

<details><summary>Details</summary>
<p>

Screenshot 2023-12-14 at 14 10 29

Screenshot 2023-12-14 at 14 10 37

Screenshot 2023-12-14 at 14 10 45

</p>
</details> 

Link to sample test page:
Related PRs/issues: #

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

**Merge Method**


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/listing_block.html
tbrlpld, line 1: I had called out the metadata section when Kevin added it. 
```
                                        {% if card.highlighted_metadata %}
                                            <div class="tw-px-0"><p class="tw-h6-heading tw-text-blue-80 tw-mr-2 tw-py-2 tw-mb-0">{{ card.highlighted_metadata }}</p></div>
                                        {% endif %}
                                        {% if card.metadata %}
                                            <div class="tw-px-2"><p class="tw-h6-heading tw-text-gray-40 tw-py-2 tw-mb-0">{{ card.metadata }}</p></div>
                                        {% endif %}
```

I think we can remove the wrapper `div`s and change from padding to flex box with gap to make this more resilient. E.g. I think the currently implementation can have layout issues due to the applied padding if the first meta data section is not used but the second is.

source/sass/wagtail/blocks/listing-block.scss
```diff
@@ -0,0 +1,12 @@
+.listing-block {
+  .mozfest & {
+    .streamfield-content {
```
tbrlpld, line 3: I seem to be missing it. Can you add some context where this class is applied to this block?

source/sass/wagtail/blocks/listing-block.scss
tbrlpld, line 1: Since the listing block has already been completely converted from SCSS to Tailwind (which has been, and I guess still is, a big effort in the project): is there a way we can apply these changes without adding SCSS for this block? Maybe something with an arbitrary variant like `[body.mozfest &]:...`

tbrlpld (commented):
@chris-lawton I left some question about a fixup of the metadata styling and the use of SCSS

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/listing_block.html
chris-lawton: Thanks, updated in https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11545/commits/3c3a4a64fbc3c46b2438d5a5690c0351b17dc844

source/sass/wagtail/blocks/listing-block.scss
chris-lawton: It gets added in [base_streamfield_block.html](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/integration/mozfest-2024/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/base_streamfield_block.html#L7) which [listing_block.html](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/integration/mozfest-2024/network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/listing_block.html#L1) extends.

source/sass/wagtail/blocks/listing-block.scss
chris-lawton: Do you mean by adding something like this? https://github.com/MozillaFoundation/foundation.mozilla.org/blob/integration/mozfest-2024/tailwind.config.js#L52

I'm not that familiar with variants so if you are able to provide a code sample it would be appreciated. 

In looking into this I wonder if the styles would be better off living in https://github.com/MozillaFoundation/foundation.mozilla.org/blob/integration/mozfest-2024/source/sass/mozfest.scss as that seems to be the place overrides are being handled.

tbrlpld: The move to the `mozfest.scss` file makes sense. 

However, given that the projects wants to transition to Tailwind, a Tailwind solution may be more desirable. 

The config allows you to define custom variants. This is would work, but I find overkill for one-time uses. ["Arbitrary variants"](https://tailwindcss.com/docs/hover-focus-and-other-states#using-arbitrary-variants) allow you to build one-off variants/selectors. You basically write the selector in the brackets. Similar to SCSS, you use `&` to refer to the current element. 

E.g. this would set a gray background on the heading if the body also has the class `.mozfest`.

```html
<body class="mozfest">
  <h1 class="[body.mozfest_&]:bg-gray-500">Gray on Mozfest</h1>
</body>
```

https://play.tailwindcss.com/AR1KYZ6EIX

An arbitrary variant is also used in `list-block.html` to remove the last body child's bottom margin:

```html
<div class="[&_p,li,ul]:last:tw-mb-0">
```

chris-lawton: Thanks for the example, that's super helpful.

It works for tailwind classes but I can't seem to override the grid system classes. I need to extend `col-lg-12` which I added here: https://github.com/MozillaFoundation/foundation.mozilla.org/blob/feature/9300-list-block-updates/source/sass/mozfest.scss#L358

In `listing_block.html` i've added the following but it doesn't apply the class

```
{% block streamfield_content_classes %}[body.mozfest_&]:col-lg-12{% endblock %}
```

Does it only work for `tw-` prefixed classes?

tbrlpld: Yea, that is only going to work for Tailwind classes I think. Or at least classes that are defined in a layer known to Tailwind: https://stackoverflow.com/questions/71561324/tailwindcss-how-to-use-custom-classes-with-responsive-variants

In this case though, I think the `col` classes are outdated. I could not even find where they are defined 🤔. Are these old bootstrap columns? I think we should rather use [Tailwind equivalents](https://foundation.mofostaging.net/en/style-guide/#columns) for the columns. For `col-lg-12` (which I think means "full width at large screens") this should be `lg:tw-w-full` (combined with the Mozfest variant, this would be `[body.mozfest_&]:lg:tw-w-full`). 

Other columns an be made with fraction definitions, e.g. `tw-w-11/12`.

tbrlpld (approved):

PR #11575, opened 2023-12-18
Feature/5643 new signup snippet block be

kevinhowbrook:
# Description

This PR adds a new signup snippet block with a different style:

image

Link to sample test page: http://mozfest.localhost:8000/en/
Related PRs/issues: https://torchbox.monday.com/boards/1334760528/pulses/1336915643

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

**Merge Method**


network-api/networkapi/mozfest/models.py
```diff
@@ -59,6 +59,61 @@ def clean(self):
             raise exceptions.ValidationError("Please provide both link URL and link text, or neither")
 
 
+@register_snippet
+class NewsletterSignupWithBackground(TranslatableMixin):
```
tbrlpld, line 63: This looks very similar to the `CTA` classes from the `campaigns` module. 

Most of the fields and even the core markup seem to be shared with the general `Signup` and `NewsletterSignupBlock` and the same JS seems to be responsible for the actual functionality. 

Would it make sense to reuse some of those definitions?

tbrlpld (approved):

network-api/networkapi/mozfest/models.py
kevinhowbrook: I'm reluctant too, but yes I'll inherit the CTA. I'm of the opinion there is _far_ too much inheritance in this codebase, but I take your point given the are both technically sign ups :)

tbrlpld: I tend to agree with that assessment of too much inheritance. In this case though, the behavior is coupled and needs to match so that they can be handled by the same JS. 

Maybe the split of classes is at a weird place. E.g. instead of having a bunch of different CTA snippet classes (that are each probably only used once), why not make the styling and option on the instance 🤷‍♂️

If you see a good reason to not use inheritance here, then please feel free stick with the separate definition.

network-api/networkapi/mozfest/models.py
```diff
@@ -59,6 +60,30 @@ def clean(self):
...
+    translatable_fields = campaign_models.CTA.translatable_fields + [
+        SynchronizedField("background_image"),
+    ]
+
+    class Meta(TranslatableMixin.Meta):
+        ordering = ["name"]
+        verbose_name = "Newsletter Signup With Background"
+
+    def __str__(self):
+        return self.heading
```
kevinhowbrook, line 84: woops - I fixed this on a follow pr that's gone in :face_with_head_bandage:

Issue #9484, opened 2022-10-12
Refactor invalid `disabled` attribute on `a` elements to class


tbrlpld:
# Description

The [`disabled` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled) is only valid on form elements such as `<button>` and `<input>`. [We are using it on `<a>` elements](https://foundation.mofostaging.net/en/style-guide/#buttons), on which it is not officially supported , for styling purposes. Mainly, we are using it to display button style links as grayed out.  

We should refactor this to valid HTML by using a `disabled` class for styling of `<a>` elements. 

[`pointer-events: none;`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events) is probably going to be useful here. 

References:
* https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/disabled
* [`pointer-events: none;`](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events)

# Dev tasks

- [ ] Create a class that applies the styles for disabled buttons but with a class instead of the `disabled` attribute.
- [ ] Replace all use cases of the `disabled` attribute with the class.
- [ ] Ensure typical link events (like hover and click) have not effect on disabled links.

# Acceptance criteria

- [ ] As a user, I can see no difference between a disabled button element and a link that is disabled with a class.
- [ ] As a user, when hovering a disabled link, the cursor does not turn into a pointer (the little hand).
- [ ] As a user, when clicking on a disabled link, I am **not** navigate to a new page.

mtdenton:
Linked to #9875

PR #11579, opened 2023-12-19
9484 Use `aria-disabled` instead of `disabled` attribute on links

tbrlpld:
# Description

Since `disabled` is not a valid HTML attribute on `a` elements, we needed a different solution to style links as "disabled". The only current use case is in the pagination. 

At first I assumed we should go with a class. It actually turns out that `tw-disabled` could already be used on `tw-btn-...` elements to style them as disabled. But, the class approach would not benefit screenreader users as the class only makes stylistic changes. 

So, since we already need to disable the links for screenreader users with `aria-disabled='true'`, I decided to use that instead of the styling.

Link to sample test page:
Related PRs/issues: #9484

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [-] Is my code documented?

**Merge Method**


mmmavis (approved):
Hey @tbrlpld , this is great! Thanks!

I noticed that this PR is compared against `main`. Since we are technically on a code freeze now, we will need to keep this PR open until after the break. Alternatively, we can have it merged into the `2023-eoy` branch (which will be merged after the break).

tbrlpld:
> Hey @tbrlpld , this is great! Thanks!
> 
> I noticed that this PR is compared against `main`. Since we are technically on a code freeze now, we will need to keep this PR open until after the break. Alternatively, we can have it merged into the `2023-eoy` branch (which will be merged after the break).

Thanks @mmmavis Happy to update and merge to the `2023-eoy` branch 👍

Issue #11600, opened 2023-12-21
Refactor `generic_card` to be more generic


tbrlpld:
When creating new cards, it would be helpful to have a truly generic card template which can be reused in many ways. 

Currently, the "generic card" only works for internal links because it always requires a page to be passed. 

To make this more generic, we should enable the option to directly pass a URL as well and only use the page URL when not URL was passed. 

Similarly, the image should be possible to be passed and also only fall back to the page meta image if nothing else was provided. 

Also, the title should be passable. Basically everything that currently is pulled from the page should be possible to pass directly and only fallback to using the pages info. Passing a page should also be optional.

Something like this could achieve the above:
```django+html
{% load wagtailcore_tags wagtailimages_tags %}
{% if page %}
    {% if not url %}
        {% pageurl page as url %}
    {% endif %}
{% endif %}
{% with title=title|default:page.localized.title image=image|default:page.localized.specific.get_meta_image %}
<div
    class="
           tw-flex
           tw-flex-col
           tw-gap-6
           tw-w-full
           {% if horizontal %}
               medium:tw-flex-row
               medium:tw-gap-16
           {% endif %}
          "
>
    <div
        {% if horizontal %}
            class="medium:tw-max-w-xs large:tw-max-w-sm medium:tw-shrink-0"
        {% endif %}
    >
        <a href="{{ url }}">
            {% image image fill-1200x628 %}
        </a>
    </div>

    <div>
        <div class="tw-flex tw-flex-wrap">
            {% block tags %}{% endblock %}
            {% block published_date %}{% endblock %}
        </div>

        <a href="{{ url }}" class="tw-group tw-block hover:tw-no-underline">
            <p class="tw-h4-heading d-inline-block mt-1 mb-2 group-hover:tw-underline">{{ title }}</p>
            <div class="tw-body-small my-0">
                {% block description %}{% endblock %}
            </div>
        </a>
        {% block byline %}{% endblock %}
    </div>
</div>
{% endwith %}
```

This idea came up in: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11585#discussion_r1434385666

mmmavis:
@danielfmiranda do you mind QAing this on staging since you have more context?

danielfmiranda:
No problem @mmmavis! I have just taken a look and can confirm everything is looking as expected 👍

mmmavis:
Thanks @danielfmiranda @tbrlpld !

PR #11601, opened 2023-12-21
#11600 Make generic card more generic

tbrlpld:
# Description

This PR updates the generic card template to be more generic. 

Previously it was only possible to pass a page. While this works often it fails in cases where we might want to link to external pages. See #11585. 

This PR adds the option to pass url, image, and title directly to the card template making it more re-usable. It also adds extra blocks for overrides.

Currently, the generic card is used as basis for the blog card (on the blog index) and for the related links in each blog page. 

Link to sample test page: http://localhost:8000/en/blog/, http://localhost:8000/en/blog/initial-test-blog-post-with-fixed-title/
Related PRs/issues: #11600

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [-] Is my code documented?

**Merge Method**

### DIFF
--- a/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic_card.html
+++ b/network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/generic_card.html
@@ -5,47 +5,47 @@
     {% endif %}
 {% endif %}
 {% with title=title|default:page.localized.title image=image|default:page.localized.specific.get_meta_image %}
-<div
-    class="
-           tw-flex
-           tw-flex-col
-           tw-gap-6
-           tw-w-full
-           {% if horizontal %}
-               medium:tw-flex-row
-               medium:tw-gap-16
-           {% endif %}
-          "
->
     <div
-        {% if horizontal %}
-            class="medium:tw-max-w-xs large:tw-max-w-sm medium:tw-shrink-0"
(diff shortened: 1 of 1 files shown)


danielfmiranda (changes requested):
Hi @tbrlpld thanks for taking this work on! the changes to `generic_card.html` look good to me. 

However, it appears that there may be some [unintended changes/commits in the history](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11601/commits/67365e10fea5dcc0b6b46cd3c2e10170a6cd23dd). 

Just wanted to double check if this change was intended and if not, would it be possible to remove it so we can merge in the intended changes?

Thank you!

tbrlpld:
> Hi @tbrlpld thanks for taking this work on! the changes to `generic_card.html` look good to me.
> 
> However, it appears that there may be some [unintended changes/commits in the history](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11601/commits/67365e10fea5dcc0b6b46cd3c2e10170a6cd23dd).
> 
> Just wanted to double check if this change was intended and if not, would it be possible to remove it so we can merge in the intended changes?
> 
> Thank you!

Ah good catch! Thanks @danielfmiranda. I have rebased and dropped the commit in question.

Issue #11644, opened 2024-01-04
Fix HTML validation errors on banner page


tbrlpld:
Banner page (`/en/what-we-do/`, `/en/what-we-fund/`, `/en/what-we-fund/fellowships/`)

Most of the issue come down to duplicate IDs that occur because we have duplicate markup for mobile and desktop (see `campaign_page.html` which is the basis for `bannered_campaign_page.html`). We probably want to find a more elegant solution here that makes use of flexbox to reorder the elements for the different page conditions.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-226)

Morsey187:
I fixed some errors for this page in this [PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11874), however, some painpoints came up trying to resolve the issue around duplicate IDs (`html_content_desktop` and`html_content_mobile`).

>We probably want to find a more elegant solution here that makes use of flexbox to reorder the elements for the different page conditions.

Unfortunately this isn't possible without switching to duplicating the CTAs instead of the page's main body content, hiding and showing the CTAs on mobile and desktop depending on whether in the sidebar or not. 
This is almost more complicated due to the different states and values the CTAs can be in depending on the type of campagin page, however this would be possible if CTAs could be at the bottom of the page at mobile instead of at the top.

Then theres the issue of the duplicate ids themselves, they are being applied via the `with_heading_ids` tag, however I can't see how they're used on this page, are they being used elsewhere? and are they accessible to screen-readers?

```
<h3 data-block-key="2dsou">
    <a class="fragment-id" id="what-we-fund"></a>
    What we fund
</h3>
```

I also spotted another issue where theres additional HTML at the bottom of the campagin templates, I can’t tell why this is needed as theres no content or height to the elements, however I was hesitant to remove them.
```
                    <div class="
                                {% if page.cta and page.specific.narrowed_page_content == False %}
                                    tw-w-7/12 tw-px-8
                                {% elif page.signup and page.specific.narrowed_page_content == False %}
                                    tw-w-7/12 tw-px-8
                                {% else %}
                                    tw-w-full tw-px-8
                                {% endif %}"
                    >
                    </div>
```

tbrlpld:
I am not sure anymore if this is even the right page type I am thinking about, but I am wondering if the IDs are maybe needed when the page has multiple tabs.

mmmavis:
[...]

PR #11653, opened 2024-01-05
11644 Only add the form assemly head if the body is added too

tbrlpld:
# Description

I am running out of time and there are still a several validation errors left. Most them are the duplicated IDs that result from the duplicate markup that #11147 introduced. 

The PR is excluding the FormAssembly header code when no CTA is present on the page. The FormAssembly code contains a bunch of validation issues. We probably won't be able to fix those, but we can exclude the code when we don't need it. 

Note: This PR does not solve all the issues on the page type. The ticket should remain open. After the PR is merged.

Link to sample test page: http://localhost:8000/en/what-we-do/
Related PRs/issues: #11644 

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [-] Is my code documented?

**Merge Method**


mmmavis (approved):
Thanks for catching this, @tbrlpld !

Issue #11175, opened 2023-09-26
DonateHelpPage integration tests


danielfmiranda:

danielfmiranda:
An file with integrations tests has been merged into `main` via #11658. This ticket can now be closed!

PR #11658, opened 2024-01-08
Donate Help Form Integration tests

danielfmiranda:
# Description

Related PRs/issues: #11175

Using https://github.com/MozillaFoundation/foundation.mozilla.org/pull/10646 for reference, this PR adds integration tests for the Donate Help Page's FormAssembly form, including tests for:
- **Form and Container Loading**: The tests verify that the form and its container load in as expected.
- **Prepopulated, Hidden Fields**: The tests check that prepopulated, hidden fields such as `language` and `thank_you_url` are being set to their expected values.
- **Field Rendering**: The tests verify that fields render when expected. For example:
  - Fields such as `name`, `email`, and `other_details` should only render once the user has selected an option from the "I need..." dropdown menu.
  - Only some dropdown options are set to render the optional `screenshot` field.
- **Form Submission Without Required Fields**: The tests simulate form submission without filling out the required fields and verify that the correct validation errors are displayed.
- **Post-Submission Redirects:** After simulating form submission, the tests verify that post submission redirects are working (users are redirected to the appropriate `thank_you_url` with any existing query parameters, and a "thank you message" replaces the form on the template).


network-api/networkapi/templates/donate/pages/help_page.html
```diff
@@ -24,7 +24,7 @@ <h1 class="tw-h1-heading"> {{ page.title }} </h1>
             <div class="tw-border-t tw-border-b tw-py-10 tw-mt-28 tw-mb-10 tw-border-gray-20">
                 <p id="contact-us" class="tw-h3-heading">{% trans "Contact Us" %}</p>
                 {% if show_formassembly_thank_you %}
-                    <div class="tw-py-22">
+                    <div id="thank-you" class="tw-py-22">
```
danielfmiranda, line 27: Added this ID so I can easily locate this div in [001-help-form-spec.js](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/e1fb131ca01dd43b200c7c02c1b87a929320a3c0/tests/integration/donate/help/001-help-form.spec.js#L156)

tests/integration/donate/help/001-help-form.spec.js
```diff
@@ -0,0 +1,161 @@
...
+          expect(await screenshotInput.isVisible()).toBe(false);
+        }
+      }
+
+      // Test if submitting the form without filling out the required fields creates validation errors.
+      await submitButton.click();
+      expect(await page.locator(".errFld").count()).toBe(2);
+      expect(await page.locator(".errMsg").count()).toBe(2);
+
+      // Navigate to the form's "thank you" URL without submitting, to avoid flooding FormAssembly with test submissions.
```
mmmavis, line 145: Thanking for taking test submission flood into consideration! My concern with this is we won't be able to test out the submission functionalities. There could be scenarios when the code/form on FA site and the one we copied over to the codebase became different (e.g., maybe a field's `name` attribute was changed or removed). Without having a test to trigger the entire submission flow, it's tricky for us to know when something breaks. Is there any possibility that we can at least submit one test submission through the integration test (e.g., don't do it for all possible locale)? Maybe we can work with the fundraising team to come up with a test email address to use which they can (hopefully) easily filter out when the fake help requests get to their end.

tests/integration/donate/help/001-help-form.spec.js
```diff
@@ -0,0 +1,161 @@
...
+      expect(await screenshotInput.count()).toBe(1);
+      expect(await screenshotInput.getAttribute("type")).toBe("file");
+      expect(await screenshotInput.isVisible()).toBe(false);
+
+      // Test that the "Lang" input exists, is hidden, and is being appropriately mapped the correct FA language code.
+      const langInput = wFormContainer.locator(utility.FA_HIDDEN_FIELDS.lang);
+      expect(await langInput.count()).toBe(1);
+      expect(await langInput.isHidden()).toBe(true);
+      expect(await langInput.inputValue()).toBe(
+        localeMap[localeToTest] || localeMap["other"]
```
mmmavis, line 99: If I understand the requirement correctly 😅 , I think we'll need to break this into two cases

- if current page locale is one of the FA supported locales, test to see if `langInput.inputValue()` equals to ` localeMap[localeToTest]`
- if current page locale is * not * one of the FA supported locales, test to see if `langInput.inputValue()` equals to ` localeMap["other"]`

mmmavis (changes requested):
Yay more test coverage! Thanks, @danielfmiranda . I left two comments - one should be easy to address and the other is more involved. I think having a test to cover the entire submission flow is essential with anything linked with FA. We can create a follow-up ticket for that work to avoid it becoming a blocker for this PR, but we should ensure that ticket gets addressed soon and not be buried in our backlog. 😁

tests/integration/donate/help/001-help-form.spec.js
danielfmiranda: Thanks Mavis! I agree that actually submitting the form would be a great test to include here. 

I have created issue https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11699 to house the work/discussion, and also to keep this ticket moving.

tests/integration/donate/help/001-help-form.spec.js
danielfmiranda: Sounds good to me! I have implemented both test cases using this if/else statement:
https://github.com/MozillaFoundation/foundation.mozilla.org/blob/7d875b08dbffaa0352920e21522d552b8d239ea7/tests/integration/donate/help/001-help-form.spec.js#L95-L105

danielfmiranda:
Thanks for the feedback @mmmavis!

I have created a new [issue](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11699) to handle the discussion and implementation of actual form submissions for testing purposes, and have also [updated the language check ](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11658#discussion_r1445563271)to include an if/else statement like you suggested 👍 

This PR is now ready for re-review whenever you find a free moment. 

Thanks again!

Issue #11584, opened 2023-12-19
Refactor "2-column" version of newsletter signup form


mmmavis:
This is part of the refactoring work for newsletter signup form (epic #11225)

For the "2-column" version of newsletter signup form, extract the existing code from `join.jsx`, and create a similar component based on what we did for [the blog body streamfield block implementation](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/source/js/components/newsletter-signup/organisms/blog-body-signup.jsx).

For detailed audit report, see [this sheet](https://docs.google.com/spreadsheets/d/1EeWshn5N7f6eEZ8eBdMS_IEF1RtFuaXFeUrwLEX96js/edit#gid=0). Look for the "Blog Pages (callout box near footer)" column

Example: https://foundation.mozilla.org/en/blog/mozillas-responsible-computing-challenge-holds-its-first-student-unhackathon/

<img width="1146" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/ce2431df-52a6-40f0-bf27-ddb19a3bfbaf">

danielfmiranda:
QA'd on staging. Ready for prod!

mmmavis:
Live on prod! Closing! 🎉

PR #11698, opened 2024-01-08
Blog near-footer signup form is now rendered using the new React components

mmmavis:
# Description

Note for the first commit:

- I decided to add a new data attribute `data-form-style` and use it to determine what style in `form-specific-style.js` the signup module should be using. Reasons being
  1.  `data-form-style` is more self-explanatory for this purpose
  2.  Using this new attribute, forms from multiple locations (e.g., different `data-form-location` value) can "re-use" existing styles by defining `data-form-style`.
- I've changed all `data-module-type` value to be `default` so it corresponds with the file name `default-layout-signup.jsx` that's used to inject the React component. I'm thinking to rename that file to `default-signup.jsx` but that is outside of the scope for this ticket.

Note for the second commit:

- I renamed `designSystemStyle` to `fieldStyle` because the possible value "outlined" and "filled" are applicable on the field level and not the form level.

---

Link to sample test page: https://foundation-s-11584-blog-wbgmww.mofostaging.net/en/blog/offer-near-raise-they-poor-research/
Related PRs/issues: #11584


danielfmiranda (approved):
Looks great! Thanks @mmmavis 🙌 approved

Issue #11651, opened 2024-01-05
Add Playwright tests for mobile primary navs


mmmavis:
Add Playwright tests for mobile primary navs.

For main site, PNI, and MozFest, make sure primary nav can be toggled on/off on mobile screen size.

mmmavis:
Forgot to close this ticket. Work has been done and merged!

PR #11706, opened 2024-01-09
Add Playwright tests for primary nav on mobile

mmmavis:
# Description

Added Playwright tests to make sure primary nav on mobile can be toggled on and off.

Related PRs/issues: #11651

### DIFF
--- a/tests/integration/navigation/001-site-nav.spec.js
+++ b/tests/integration/navigation/001-site-nav.spec.js
@@ -1,0 +1,102 @@
+const { test, expect } = require("@playwright/test");
+const waitForImagesToLoad = require("../../wait-for-images.js");
+
+const MOBILE_VIEWPORT = { width: 360, height: 800 };
+
+test.describe("Main site primary nav", () => {
+  test(`Mobile nav`, async ({ page }) => {
+    const PAGE_URL = `http://localhost:8000/en/`;
+
+    await page.setViewportSize(MOBILE_VIEWPORT);
+    await page.goto(PAGE_URL);
+    await page.locator(`body.react-loaded`);
+    await waitForImagesToLoad(page);
+
+    const primaryNavContainer = page.locator(".primary-nav-container");
+    expect(await primaryNavContainer.count()).toBe(1);
+
+    // check if mobile nav is not visible by default
+    const nav = primaryNavContainer.locator(".narrow-screen-menu-container");
+    expect(await nav.count()).toBe(1);
+    await expect(nav).not.toBeInViewport();
+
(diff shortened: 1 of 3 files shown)


tests/integration/navigation/001-site-nav.spec.js
```diff
@@ -0,0 +1,37 @@
+const { test, expect } = require("@playwright/test");
+const waitForImagesToLoad = require("../../wait-for-images.js");
+
+const MOBILE_VIEWPORT = { width: 360, height: 800 };
+
+test.describe("Main site primary nav", () => {
+  test(`Primary nav (mobile)`, async ({ page }) => {
+    const PAGE_URL = `http://localhost:8000/en/`;
```
jhonatan-lopes, line 8: This `localhost:8000/en` value seems to be repeated throughout our integration tests. It would be nice if it could be defined in a single place, like a `constants` file or something. Same thing for `mozfest.localhost:8000`... Not a block though, just an observation

jhonatan-lopes (commented):
Hey Mavis, from my perspective it seems like the three tests are essentially the same, except for the URL that they target. Would it be better if we had a single test parameterised with the URLs that we are targeting? This way, we could easily introduce new sites (like the donation once the migration is over). What do you think?

mmmavis:
> Hey Mavis, from my perspective it seems like the three tests are essentially the same, except for the URL that they target. Would it be better if we had a single test parameterised with the URLs that we are targeting? This way, we could easily introduce new sites (like the donation once the migration is over). What do you think?

Hey @jhonatan-lopes , nice suggestion! I thought about doing that but didn't end up going with that route for two reasons: 
1. The PNI nav test is actually slightly different than the main site and MozFest site
2. Going forward the three test files are going to differ more as the new nav work goes. So I thought keeping them separate could be slightly more future proofing. What do you think?

tests/integration/navigation/001-site-nav.spec.js
mmmavis: I like this idea! I'll create a new ticket for this as there are multiple places in the existing code that can benefit from this.

jhonatan-lopes:
Nah, that's totally fair, you have way more context into these tests than me so your judgement is for sure better than mine. They look great, thanks!

jhonatan-lopes (approved):

tests/integration/navigation/001-site-nav.spec.js
jhonatan-lopes: Awesome, thanks!

Issue #11671, opened 2024-01-08
Refactor "PNI Consumer Page" version of newsletter signup form


mmmavis:
This is part of the refactoring work for newsletter signup form (epic #11225)

For the "PNI Consumer Page" version of newsletter signup form, extract the existing code from `join.jsx`, and create a similar component (or re-use one) based on what we did for [the blog body streamfield block implementation](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/source/js/components/newsletter-signup/organisms/blog-body-signup.jsx).

For detailed audit report, see [this sheet](https://docs.google.com/spreadsheets/d/1EeWshn5N7f6eEZ8eBdMS_IEF1RtFuaXFeUrwLEX96js/edit#gid=0). Look for the **"PNI (2023 Consumer Creep-o-meter Page)" column**

mmmavis:
Hi @danielfmiranda , apology for another QA request from me. Could you help QA the signup form in the quiz section: https://foundation.mofostaging.net/en/privacynotincluded/articles/annual-creep-o-meter/

mmmavis:
Hi @danielfmiranda , I realized the merged PR actually missed a part. I've filed a fix to that here (https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11788). Let's wait until this new fix gets merged before you QA on staging. Thanks & sorry for the confusion.

mmmavis:
Closing. Live on prod now. 🎉

PR #11729, opened 2024-01-18
Updated newsletter module on PNI consumer page's quiz section

mmmavis:
Updated newsletter module on PNI consumer page's quiz section with the new implementation approach.

Related PRs/issues: #11671 

---

- Review app: https://foundation-s-11671-news-tsuohl.herokuapp.com/en/privacynotincluded/articles/annual-consumer-creep-o-meter/
- Prod: https://foundation.mozilla.org/en/privacynotincluded/articles/annual-creep-o-meter/

# To QA

1. Go to the page
2. Scroll to the quiz section and select "See Results"
3. Sign up button should be disabled by default and whenever "email" field is empty

Note that on prod, there's no gap in between the email field and "Sign up" button but on review app there is. This is expected behaviour and not a regression.

image

image


danielfmiranda (approved):
Looks good to me! also thanks for the note at the end of the description regarding the gap in between the form field and submit button 👍 approved!

Issue #11657, opened 2024-01-06
Refactor "newsletter_signup_sidebar" version of newsletter signup form


mmmavis:
This is part of the refactoring work for newsletter signup form (epic #11225)

For "newsletter_signup_sidebar" version of newsletter signup form, extract the existing code from `join.jsx`, and create a similar component (or re-use one) based on what we did for [the blog body streamfield block implementation](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/source/js/components/newsletter-signup/organisms/blog-body-signup.jsx).

For detailed audit report, see [this sheet](https://docs.google.com/spreadsheets/d/1EeWshn5N7f6eEZ8eBdMS_IEF1RtFuaXFeUrwLEX96js/edit#gid=0). Look for the "MozFest (page body)" column

Example: https://www.mozillafestival.org/en/

<img width="379" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/be283a6b-fdd3-4df3-8f5f-38d901218836">

PR #11733, opened 2024-01-18
Remove MozFest's partials/signup.html because it's no longer in use

mmmavis:
# Description

Looks like `network-api/networkapi/mozfest/templates/partials/signup.html` was replaced as [part of the EOY 2023 MozFest update work](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11527/files#diff-9fb27af5fe66f957ba6d13679154a6b390547bb6199bdb5a9864117958ff16d6L63). I looked through the codebase and didn't see it was being used elsewhere. Removing this file as the result. Please help double check in case I overlooked something. Thanks!

Related PRs/issues: #11657


jhonatan-lopes (approved):
Great catch!

Issue #9373, opened 2022-09-16
PNI - Fix PNI Playwright tests


danielfmiranda:
At the moment, the PNI search test for playwright is failing due to the animations that load in the products. This is causing a random number to be returned in the assert statements, we should figure out a way to get a consistent number to test against.

mmmavis:
PR has been merged into `main`. Closing!

PR #11765, opened 2024-01-24
Fix PNI Playwright tests

mmmavis:
Related #9373

This PR enables existing Playwright tests for PNI search related functionalities by fixing broken code and breaking code into individual tests using test hooks.

File is relocated under a new folder `pni` and renamed into `search.spec.js`

### To test locally

- Run `docker-compose up` on one tab
- Approach 1: 
  - On `package.json`, update line 52 to `"playwright:integration": "wait-on -i 3000 -t 120000 -v http://127.0.0.1:8000/cms && playwright test ./tests/integration/pni"` so you can skip all other tests unrelated to PNI search
  - Open a new terminal tab, and run `npm run playwright:integration`
- Approach 2: Alternatively, you can run the PNI tests directly using VSCode's Test Explorer.


tests/integration/pni/search.spec.js
```diff
@@ -0,0 +1,160 @@
+const { test, expect } = require("@playwright/test");
+const waitForImagesToLoad = require("../../wait-for-images.js");
+
+/**
+ * Perform several PNI tests related to searching/filtering
+ *
+ * NOTE: this requires a `new-db` run with the seed value set
+ *       through RANDOM_SEED=530910203 in your .env file
```
jhonatan-lopes, line 8: I'm thinking that this makes the integration tests quite fragile, right? Is this the reason why we have these tiny changes on Percy visual regression?

This is, of course, unrelated to this PR. I'm just thinking out loud on how we (and if) we could mitigate this issue to prevent integration tests being broken due to dev forgetting to set the `RANDOM_SEED`. I totally did not waste half an hour investigating a broken test on this today...

jhonatan-lopes (approved):
Hey @mmmavis, thanks for this! It's working well LGTM :rocket:

jhonatan-lopes:
Another thing that I noticed here is that `playwright` is not installed inside the container. I think it would be better to install it alongside the dependencies of the dev Dockerfile and make sure that it runs inside the container (and with the right seed) so that we don't have tiny differences between the `playwright` env and the Docker env.

jhonatan-lopes:
Ticket to clean up `package.json` scripts: #11767

tests/integration/pni/search.spec.js
mmmavis: Hey @jhonatan-lopes , for this I think we'll need to rely on `RANDOM_SEED` set to a fixed value. An alternative would be changing how factory work - so instead of generating a random number of products per category, we decide how many and use that value in our Playwright tests. Any thoughts on this?

mmmavis: Also it looks like most of the Percy flags are due to date difference. I haven't taken a look at the code yet but I'm guessing if we instead use a fixed dates on those, then the false positive flags should go away.

mmmavis:
@jhonatan-lopes thanks for the suggestions and filing follow-up ticket to make our dev setup more concrete!

tests/integration/pni/search.spec.js
jhonatan-lopes: >  An alternative would be changing how factory work - so instead of generating a random number of products per category, we decide how many and use that value in our Playwright tests.

While I appreciate the randomness of it, maybe it would be best to have some controlled/reproducible situations, for instance a predictable category with 3 products that we can use in testing, similar to how we do with the "Percy blog post with a fixed title". Just an idea

jhonatan-lopes: > Also it looks like most of the Percy flags are due to date difference. I haven't taken a look at the code yet but I'm guessing if we instead use fixed dates on those (or how to generate dates), then the false positive flags should go away.

That makes sense

PR #11771, opened 2024-01-25
[WIP] Fixed how to factory generate dates

mmmavis:
WIP


network-api/networkapi/highlights/factory.py
```diff
@@ -18,17 +18,12 @@ class Meta:
             "footer_sentence",
         )
 
-    class Params:
```
mmmavis, line 21: These details seem a bit unnecessary to me. I'm deleting it to avoid complication.

network-api/networkapi/news/factory.py
```diff
@@ -32,34 +32,15 @@ class Meta:
         exclude = ("headline_sentence",)
 
     class Params:
-        unpublished = Trait(
```
mmmavis, line 35: Similarly, these details seem a bit unnecessary to me. I'm deleting them to avoid complication.

network-api/networkapi/news/factory.py
```diff
@@ -32,34 +32,15 @@ class Meta:
...
-            expires=Faker("date_time", tzinfo=timezone.utc)
-            if RANDOM_SEED and not TESTING
-            else Faker("past_datetime", start_date="-30d", tzinfo=timezone.utc)
-        )
         video = Trait(is_video=True)
 
     headline = LazyAttribute(lambda o: o.headline_sentence.rstrip("."))
     outlet = Faker("company")
-    date = Faker("date") if RANDOM_SEED and not TESTING else Faker("past_date", start_date="-30d")
+    date = Faker("past_date", start_date=datetime(2020, 1, 1), tzinfo=timezone.utc)
```
mmmavis, line 39: I'm unsure why we needed to have two ways to generate values for `date` here. Any insights?

network-api/networkapi/news/factory.py
```diff
@@ -32,34 +32,15 @@ class Meta:
...
     link = Faker("url")
     excerpt = Faker("paragraph", nb_sentences=3, variable_nb_sentences=True)
     author = Faker("name")
-    publish_after = (
-        Faker("date_time", tzinfo=timezone.utc)
-        if RANDOM_SEED and not TESTING
-        else Faker("past_datetime", start_date="-30d", tzinfo=timezone.utc)
-    )
+    publish_after = Faker("past_datetime", start_date=datetime(2020, 1, 1), tzinfo=timezone.utc)
```
mmmavis, line 43: Same question here. I'm curious why RANDOM_SEED and TESTING matter when it comes to generating value for `publish_after`

network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -88,10 +88,7 @@ class Meta:
 
     title = LazyAttribute(lambda o: o.title_text.rstrip("."))
     body = Faker("streamfield", fields=blog_body_streamfield_fields)
-    first_published_at = (
-        Faker("date_time", tzinfo=timezone.utc)
-        if RANDOM_SEED and not TESTING
-        else Faker("past_datetime", start_date="-30d", tzinfo=timezone.utc)
```
mmmavis, line 94: Same q as above.

network-api/networkapi/wagtailpages/factory/blog.py
```diff
@@ -88,10 +88,7 @@ class Meta:
...
+    first_published_at = Faker("past_datetime", start_date=datetime(2020, 1, 1), tzinfo=timezone.utc)
```
mmmavis, line 91: Same q as above.

jhonatan-lopes (commented):
Hey @mmmavis, thanks for these changes. I'm a bit concerned that these won't be enough to solve the issue though. The dates randomly changing on RCC/Research Hub does not seem to be caused by the factory date generation that you changed here, for instance.

I think it relates more to controlling the randomness. If the randomness is indeed reproducible, the dates shouldn't be changing like that from one Percy test to another.

network-api/networkapi/highlights/factory.py
mmmavis: Cool I'll add them back!

network-api/networkapi/news/factory.py
mmmavis: Will add them back!

mmmavis:
Hi @jhonatan-lopes , thanks for the feedback! I'll add the `Trait` stuff back.

Regarding the date generation issue - from my observation in the past, the date difference seemed always to be "x days" from when the `main` branch was last updated. For example, if the last Percy build for `main` was on January 1, 2023 and an article's published date was generated to be June 1, 2020, if I ran a Percy build again on the next day (Jan 2, 2023) then the same article's published date would become June 2, 2020. I suspected the issue was because we used "today" as the baseline to generate the date (e.g., `Faker("past_datetime", start_date="-30d", tzinfo=timezone.utc)`) and that became a new variable used in faker's algorithm which ultimately caused the "x days" difference if we ran the Percy build x days apart. I'm thinking by having a fixed date as baseline will always give us an expected value given that the same SEED value is provided, just like how `get_random_option` works. That said, as a proof of concept I ran another Percy build today on this PR and compared the results with yesterday's build. There are still some leftover issues I need to address. I'm gonna revise my code and have this PR open for another few days to keep my proof of concept going before re-requesting a new review from you. 😁

jhonatan-lopes:
That sounds great, thanks @mmmavis! I tried some stuff locally here but couldn't get to the bottom of the problem either but your comment definitely does make sense. I'll try some other stuff with that in mind (trying to "fudge today's date" and re-generating data) to see if I can repeat the problem

mmmavis:
Very outdated PR. Closing.

Issue #11769, opened 2024-01-25
Blog topic description not showing


kristinashu:
Blog landing page https://foundation.mozilla.org/en/blog/ has generic intro text "Mozilla’s blog features guides to making your online life better, stories from the movement, and critical analysis of issues around internet health."
<img width="1241" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/e48baa90-3be8-4156-9edf-724d8d9c65c8">

Blog topic pages have a field to enter custom topic intro text. Example https://foundation.mozilla.org/cms/snippets/wagtailpages/blogpagetopic/edit/5/
<img width="1119" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/ae0aa8f6-d6af-40d5-8aa6-453ad988b6ad">

[Bug] However the blog topic pages are showing the generic text
<img width="1231" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/1068783/ede5725c-14ed-4250-a2d2-7619cf09ac6a">

[Expected behavior] It should show the custom topic text.

danielfmiranda:
Hi @kristinashu! 

The fix for this ticket has been merged and is now live on staging. I was wondering if you could please give https://foundation.mofostaging.net/en/blog/topic/advocacy/ a look whenever you find a free moment, and leave a comment if everything looks as expected? 

Thank you!

kristinashu:
Perfect, thank you Daniel!

PR #11772, opened 2024-01-26
Updated blog topic page to show topic intro/description

danielfmiranda:
# Description

Link to sample Blog page: https://foundation-s-feature-11-kietq3.mofostaging.net/en/blog/
Link to sample Advocacy blog page: https://foundation-s-feature-11-kietq3.mofostaging.net/en/blog/topic/advocacy/
Link to Localized Advocacy blog page: https://foundation-s-feature-11-kietq3.mofostaging.net/fr/blog/topic/advocacy/

Related PRs/issues: #11769

This PR updates the `blog_index.html` template with a `if/else` statement that checks if an updated `"index_intro"` is set in the context. 

If so, such as when visiting a blog topic page (ex: blog/topic/advocacy), the template's intro section should update to render the topic's `intro` field.

If not, the template renders the page's `intro`, as it was before.

# Screenshots:

**Blog Page:**

Screenshot 2024-01-25 at 16-21-38 Blog

**Advocacy Topic Page:**

Screenshot 2024-01-25 at 16-22-47 Advocacy

**Localized Advocacy Topic Page:**

Screenshot 2024-01-25 at 16-23-18 Advocacy


kristinashu (approved):
Looks great, thank you Daniel! Also thank you for including a localized example!

network-api/networkapi/wagtailpages/tests/blog/test_blog_index.py
```diff
@@ -305,6 +306,41 @@ def test_topic_route_shows_only_entries_of_topic(self):
...
+        # Expecting the name of the topic -> the name of the index page
+        expected_index_title = "Test Topic Name Blog"
+        topic = blog_factories.BlogPageTopicFactory(name=topic_name, title="")
+
+        url = self.get_topic_route(topic=topic.slug)
+
+        response = self.client.get(path=url)
+
+        self.assertEqual(response.status_code, HTTPStatus.OK)
+        self.assertEqual(response.context["index_title"], expected_index_title)
```
mmmavis, line 342: Hi @danielfmiranda , I'm not familiar with how these tests are usually written. Do you mind explaining where `expected_index_title` is getting set? I see it was defined on line 334 but can't' figure out how the value is used. Is it because the factory always generates a blog index page with title "Blog"? If so, I'm wondering why `expected_index_title` wouldn't just be a single word "Blog".

danielfmiranda: Hi @mmmavis! No problem. 

Blog Topic detail pages such as https://foundation.mozilla.org/en/blog/topic/advocacy/ are technically still the `BlogIndexPage`, just filtered for only blog posts that are linked to that following topic.

In the blog index page's [`filter_entries_for_topic` function,](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/f9a4c7ca64801437e40413dcb480b653f4910861/network-api/networkapi/wagtailpages/pagemodels/blog/blog_index.py#L177C1-L177C58), we have the following lines:


```python
...
        locale = get_locale_from_request(context["request"])
        try:
            localized_topic = topic.get_translation(locale)
        except ObjectDoesNotExist:
            localized_topic = topic

        context["index_intro"] = localized_topic.intro
        context["index_title"] = titlecase(f"{localized_topic.name} {self.title}")

        if localized_topic.title:
            context["index_title"] = localized_topic.title
...
```

What this code is pretty much saying, in regards to the index title, is "if the topic has a custom `title` set, update `index_title` accordingly."

If the topic does **not** have a title set, have `index_title` default to the value of `<topic_name> <index_page_title>`.  Which in my previous commit's case would be "Blog".

However, taking a second look at that approach, I can see that its quite confusing.

So instead, I have updated/renamed the function in my latest commit, as well as adding an additional comment in hopes it makes it easier to understand.

```python
    def test_index_title_default_works_with_no_title(self):
        topic = blog_factories.BlogPageTopicFactory(name="Test Topic", title="")

        # If a topic has no title set, the blog index page will set the context's
        # "index_title" field to the default value of "<topic_name> <blog_index_page_title>"
        expected_index_title_value = f"{topic.name} {self.blog_index.title}"

        url = self.get_topic_route(topic=topic.slug)

        response = self.client.get(path=url)

        self.assertEqual(response.status_code, HTTPStatus.OK)
        self.assertEqual(response.context["index_title"], expected_index_title_value)

```

Do you think you could please take a look whenever you find a free moment and let me know what you think?

Thanks!

mmmavis (approved):
Thanks for the explanation, @danielfmiranda . I love the updated approach. R+ 🎉

Issue #11647, opened 2024-01-04
Fix HTML validation errors on "participate page2"


tbrlpld:
Participate page2 (`/en/what-you-can-do/` )

danielfmiranda:
The PR for this ticket #11774 has been merged into main. However, since it was just updating the formatting of the template to avoid DJLint errors, there is nothing much for someone to QA on staging as DJlint checks in CI passed during the PR process. 

Moving to ready for prod!

PR #11774, opened 2024-01-26
fixed linting errors for participate_page2.html

danielfmiranda:
# Description

Related PRs/issues: #11647

This PR updates the template `participate_page2.html` to remove any existing linting errors.

According to the file `pyproject.toml`, this template only had one linting error being temporarily ignored which was:
`T003: Endblock should have name. Ex: {% endblock body %}.`

So adding the `{% block %}` name to the `{% endblock %}` tags was all that was needed in order to have the template pass all linting tests 👍 

# Steps to test

1. If this PR is passing all linting tests in CI, no further testing is required!


jhonatan-lopes (approved):
Looks great, thanks @danielfmiranda

PR #11786, opened 2024-01-29
Update MozFest newsletter signup modules

mmmavis:
Related to #11672 

- updated MozFest newsletter signup modules with the new implementation approach
- removed file that doesn't seem to be used
- removed obsolete code in `join.jsx`

I didn't set up a review app for this PR as some parts can only be tested locally (on a Mozfest landing page, for example). Please see inline comments for details.


network-api/networkapi/mozfest/templates/fragments/blocks/sign_up_block.html
```diff
@@ -1,22 +0,0 @@
-{% load i18n l10n static %}
```
mmmavis, line 1: This file was first added in [PR 7028](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11571/files#diff-569abab24729f8c302dea6a60364db3da2ec5738a91aef3e89c34502bb60f07d) as part of the MozFest EOY 2023 work. However, in the [main MozFest feature branch](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11527/files) I don't see it being used anywhere. From the code I suspect this file was replaced by a newer file `newsletter_signup_with_background_block.html` and the team somehow forgot to delete this one. I'm deleting this file now unless I overlooked something here.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/newsletter_signup_block.html
mmmavis, line 1: This file / block is currently not available on the CMS because of a bug (https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11731). I'm updating the code here in case it's something we wanna re-enable in the future. To QA this. Test QA locally, please go to `network-api/networkapi/mozfest/models.py` and give any of the `newsletter_signup` a different key so it can show up on the block selection on MozFest Primary Page for you to add and test.

danielfmiranda (approved):
Hi @mmmavis!

Thanks again for always leaving detailed descriptions, updates, and comments in your PRs. They help greatly, especially in PRs such as this where it could have been confusing to see two sets of newsletter blocks for mozfest.

I have tested this PR on my local machine, and was also able to get a "newsletter signup with background" on a review app [here](https://mozfest-foundation-s-mozfest-ne-ivbqtl.mofostaging.net/en/)

Everythings looking good to me! Approved 👍

Issue #11795, opened 2024-01-31
[Mozfest] White gap between hero and dark section


jhonatan-lopes:
### Describe the bug
After addressing the spacing issue on Mozfest (https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11756), there is now a white gap between the hero and dark sections that shouldn't exist.

> On the home and landing pages now it leaves a gap between the hero and the black section that should be able to sit directly linked.
https://www.mozillafestival.org/en/Test-landing/

![image](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/61277874/b75c6a9e-5512-45e0-85e8-d48265b1e04c)

### To Reproduce
1. Go to 'https://www.mozillafestival.org/en/Test-landing/'
2. Scroll down after "Heading Heading"
3. See error

### Expected behavior
White gap should not be visible (or should be dark).

mmmavis:
Looks like we only need to [revert these changes](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11756/files#diff-5632e11459c8dca75fd1fd48f3ce59e88e90b3815cbc90d364d886fc3227124f) to get rid of the white gap. @jhonatan-lopes do you know why the changes were introduced in that PR? I could file a PR to revert the code but would like to make sure I don't create unwanted effects with that.

jhonatan-lopes:
Hey @mmmavis, thanks for taking a look at this! It seems that the margin/padding between the blocks was not adequate as it can be seen on [this visual regression test](https://percy.io/Mozilla-Foundation/foundation.mozilla.org/builds/32400510/changed/1778962378?browser=firefox&browser_ids=48%2C49&subcategories=approved&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280). That PR was made to increase that gap.

If this feels like a simple task for you, I would appreciate you working on it, thanks

Please let me know if you need more details

robdivincenzo:
Hey @jhonatan-lopes and @mmmavis- just to clarify, do we want to remove only the gap between the blocks and keep the gap between the content and footer sections which existed before https://github.com/MozillaFoundation/foundation.mozilla.org/issues/11795?

![Screenshot 2024-02-08 at 11 50 40 AM](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/3083419/39ab5869-bf13-49b1-8353-b4deab07867c)

jhonatan-lopes:
[...]

PR #11862, opened 2024-02-13
Remove unwanted gaps between hero and footer when using dark sections

robdivincenzo:
# Description

The problem is that Issue #11756 introduced unwanted white gaps in between the dark hero and an immediately following dark section as seen by screenshots in Issue #11795 . Additionally, there is an unwanted white gap between the body section and the footer when the page ends with a dark section.

The solution is to conditionally add spacing above a paragraph block when it immediately follows a dark hero instead of after the dark hero unconditionally. The solution for the second unwanted white gap is to simply remove the margin class from the mozfest-content wrapper.

The implementation introduces a streamfield custom block which will be overridden to include a new streamfield_spacer.html template on mozfest_homepage.html and mozfest_landing_page.html templates. This new template extends the existing streamfield.html template to include a new body_spacer custom block containing a spacer div. The body_spacer block will then be rendered only if a paragraph is the first block in the body section. The body_spacer block is empty by default, meaning the spacer div is only rendered when the two following conditions are met:

1. Mozfest Homepage or Mozfest Landing Page
2. First block in body is a paragraph

There are further considerations. We could reduce coupling by adding margins to the paragraph block unconditionally to have adequate spacing in all cases regardless of placement. But the paragraph block is used generically and currently assumes spacing from blocks above it (see for example the various uses on just this one page: https://www.mozillafestival.org/en/highlights/). Modifying each dependency in order to introduce a generic paragraph margin seems out of scope for this issue, but perhaps could be a larger follow-up issue.

Link to sample test page: https://mozfest-foundation-s-issue-1179-lpznkg.mofostaging.net/en/test-landing/
Related PRs/issues: #11756 and #11795

# Steps to test

1. Review the dark section example page to confirm the white spaces are removed: https://mozfest-foundation-s-issue-1179-lpznkg.mofostaging.net/en/test-landing/
2. Review the landing page example page to confirm the margin / spacer div is present above the first paragraph block and below the hero: https://mozfest-foundation-s-issue-1179-lpznkg.mofostaging.net/en/landing-page-1/
3. Review another page to confirm the spacing is still accurate / untouched: https://mozfest-foundation-s-issue-1179-lpznkg.mofostaging.net/en/tickets/


network-api/networkapi/mozfest/templates/partials/streamfield.html
```diff
@@ -1,5 +1,9 @@
 {% load wagtailcore_tags %}
 
 {% for block in page.body %}
+    {# optionally render a spacer if the first block is a paragraph. Only for landing_page / homepage page types with the dark hero #}
+    {% if forloop.first and block.block_type == 'paragraph' %}
+        {% block body_spacer %}{% endblock body_spacer %}
+    {% endif %}
```
jhonatan-lopes, line 7: Hey @robdivincenzo, I think the paragraph block needs a spacer every time, not only when it's the first block:

image

jhonatan-lopes (commented):
Thanks for taking on this work, @robdivincenzo! I think it's almost there, we just need to adjust the paragraph spacing

network-api/networkapi/mozfest/templates/partials/streamfield.html
robdivincenzo: Thanks for reviewing @jhonatan-lopes ! I've made a couple commits with your feedback to include the paragraph spacer each time on these dark themed pages, and then also to include a post_paragraph spacer so that spacing is balanced.

Screenshot 2024-02-16 at 8 59 37 AM

jhonatan-lopes (approved):
LGTM :+1: thanks Rob!

Issue #11865, opened 2024-02-14
Newsletter signup module - remove obsolete code


mmmavis:
(This is the last ticket of the EPIC.)

Remove the now obsolete code including

- join.scss 
- join-us.js
- join.jsx
- and other relevant code

mmmavis:
Done!

PR #11876, opened 2024-02-16
Remove old newsletter code

mmmavis:
This is the final PR for the newsletter module refactor work. (Thanks for reviewing all my previous PRs for the epic @danielfmiranda ) This PR basically removes the code that is no longer in use. There should not be any functionality or UI changes.

Related PRs/issues: #11865


maintenance/maintenance.html
```diff
@@ -28,11 +28,6 @@
         <div class="takeover"></div>
         <div class="wrapper">
             <div class="sticky-top">
-                <div class="wrapper-join-us">
```
mmmavis, line 31: We never load our compiled JS file into this page which means the old React component for rendering newsletter sign form never worked.

network-api/networkapi/wagtailpages/templates/maintenance/maintenance.html
```diff
@@ -27,11 +27,6 @@
         <div class="takeover"></div>
         <div class="wrapper">
             <div class="sticky-top">
-                <div class="wrapper-join-us">
```
mmmavis, line 30: Same. We never load our compiled JS file into this page which means the old React component for rendering newsletter sign form never worked.

danielfmiranda (approved):
Looks good to me! Thanks again @mmmavis for taking on all this work, approved! 👍

Issue #11406, opened 2023-11-10
Add error handling to newsletter signup forms


mmmavis:
All of our existing newsletter signup forms **don't** have error handling set up for any server errors related to form submission (e.g., the server or Basket went down in the middle of the POST call etc). Currently the form doesn't change or show any user feedback when errors are encountered. 

# Current Form State

Form logic can be found in [`join.jsx`](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/source/js/components/join/join.jsx). 

I also wonder why we are only taking [http status `201`](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/source/js/components/join/join.jsx#L175-L177) and not all of the `2xx` as the "ok" response. (More on this [Slack thread](https://mozilla.slack.com/archives/C043Q45BBMF/p1699648412480669))

## Default State

<img width="1181" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/fae30a30-4843-41e3-adaa-b2f9bc36cffa">

<img width="1165" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/dd44ac7c-1a46-4863-9217-82ec0f788c65">

## Form Validation Errors

<img width="1162" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/0aa2d540-b66f-4f8b-91cb-fbff3894b44e">

## Form Submission Successfully Went Through

<img width="1163" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/f0254f55-0431-4132-90cd-5876d2b37fe3">

## Form Submission Did Not Go Through Due To Server/API Error

(Form stays the same. No feedback to users to indicate errors have occurred in our end.)
<img width="1187" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/24485751-f985-4aab-98fd-3833bd4aab05">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-203)

robdivincenzo:
The error messaged mocked up. Design gave the thumbs up via Slack.
![Screenshot 2024-03-04 at 4 13 04 PM](https://github.com/MozillaFoundation/foundation.mozilla.org/assets/3083419/9ef7ad92-b9d6-460d-82e2-f59a56393b12)

Issue #11825, opened 2024-02-05
makemessages command no longer works


TheoChevalier:
For some reason, since last week `inv makemessages` doesn’t update .po files anymore, only .pot files. This setup is required to reproduce: https://github.com/MozillaFoundation/foundation.mozilla.org?tab=readme-ov-file#translations

I’ve tried updating my toolchain, but it doesn’t seem to fix the issue. I wonder if the Django upgrade might have to do with this?

jhonatan-lopes:
Hey @TheoChevalier, thanks for flagging this up. I'll investigate

jhonatan-lopes:
From the release notes:
- The [makemessages](https://docs.djangoproject.com/en/5.0/ref/django-admin/#django-admin-makemessages) command no longer changes .po files when up to date. In older versions, POT-Creation-Date was always updated. (https://docs.djangoproject.com/en/5.0/releases/4.1/#miscellaneous)

Is that the issue? Or is it not updating files that have changes besides the date?

simont-cr:
@TheoChevalier Based on @jhonatan-lopes reply, is there any other details that you would wish to add as part of this ticket or this can be considered as resolved?

TheoChevalier:
Sorry, looks like my reply to @jhonatan-lopes  didn’t get posted…

I was saying that yes, it’s not updating .po files that have actual content changes pending, not just the date, and that our custom makemessages override is the curlprit in `network-api/networkapi/management/commands/makemessages.py`. Removing the file fixes the issue.

However, the override was added to disable fuzzy strings, so we need to figure out what’s causing the issue or an other way to disable fuzzy

simont-cr:
Thank you Theo! I have added a P1 status to this ticket so we can take a look on our upcoming sprints to see the best way to resolved based on your feedback.

PR #11910, opened 2024-02-26
Newsletter Signup Error Handling

robdivincenzo:
# Description

There currently is no error message returned to the user in the case that the 3rd party newsletter sign up service is down or does not return 201. This pull request will now display an API error to the user. These API Errors are different and separate from validation errors, and are treated as such.

Link to sample test page: https://foundation-s-11406-news-g7ddcn.mofostaging.net/en/test-the-invalid-newsletter-signup/
Related PRs/issues: #11406

Login: admin
Password: 4@IMcwJ!(GqoL9^Z

# To Test

An invalid signup snippet (non-existant newsletter) has been created and added to the sample test page. Navigate to that link and attempt to sign up. You should now receive an error message.

Repeat for any locale like /fr/ instead of /en/ and verify localization is occurring.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP-63)


source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
```diff
@@ -193,9 +193,19 @@ function withSubmissionLogic(WrappedComponent) {
         // [TODO] We will need to update this logic depending on what comes out of
```
mmmavis, line 193: This comment can finally be removed :D

mmmavis (changes requested):
PR is working well. Thanks @robdivincenzo ! 

Two little things - 

- There are some comments in `with-submission-logic.jsx` that can be removed.
- Regarding the error message, do you think we can change it to something more generic? `Signup was not created.` and  `Unable to resolve signup.` feel quite technical and not so user friendly in my opinion. Maybe we can ask in the team channel to brainstorm copy ideas? 👀

mmmavis (changes requested):
For localization, we are moving away from using `getText` as it's reading hardcoded strings in `source/js/components/petition/locale-strings.jsx` and not hooked up with our localization flow. I wonder if there's a way to expose the new strings (the two error messages we are adding in this PR) to Pontoon and have our backend to [send back the localized error message directly](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/campaign/views.py#L99-L102) ? @jhonatan-lopes @danielfmiranda

jhonatan-lopes:
Hey @mmmavis! Localisation on JS is handled by serving `.mo` files to the client (see https://docs.djangoproject.com/en/4.2/topics/i18n/translation/#internationalization-in-javascript-code).

We have that enabled on our codebase already. See:

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/bf8bd1d4116b5162e087742cd6d8cf85d7854db0/network-api/networkapi/urls.py#L120

and

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/bf8bd1d4116b5162e087742cd6d8cf85d7854db0/network-api/networkapi/templates/pages/base.html#L135

It seems that from the JS perspective, you just need to call `gettext`, which will then issue a request to the server to fetch the relevant `.mo` file.

We are also prepared to generate messages from JS files:

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/bf8bd1d4116b5162e087742cd6d8cf85d7854db0/tasks.py#L33-L45

I believe that covers all cases but if we have any further questions, I suggest getting in contact with Théo to confirm.

source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
robdivincenzo: You got it! 👍

robdivincenzo:
@mmmavis @jhonatan-lopes Thanks for all the extra context related to localization!

Interestingly, I do not see strings exporting from JS, only from django (i.e through `blocktrans`). Though `makemessages` is running the correct `-d djangojs` parameter.

I'm tempted to include a hardcode `getText` string for this issue, but spin off a separate issue related just to looking at JS localization. I could use the already translated "Something went wrong" string from `source/js/components/petition/locale-strings.jsx` which would also address the error message being too technical. Thoughts?

jhonatan-lopes:
Hey @robdivincenzo, are you running `inv makemessages`? If so, could this be broken due to #11825?

robdivincenzo:
Hey @jhonatan-lopes! Yes, that's the command I ran to check for detecting JS strings. I was also using the updated command from that related issue's [PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12019). I could verify the .po files updated with a new Django string but not a new JS string. Is there an additional step in the workflow?

jhonatan-lopes:
:thinking: that's really weird. We can check if it worked before on main but from a Django perspective it seems like everything is in order.

robdivincenzo:
Based on slack discussions, we're going to approach it the old, hard-coded way to unblock this issue. We'll spin off a separate issue to look at djangojs and a second ticket to refactor this component's localization. We are also going with "Something went wrong and your signup wasn't completed. Please try again later." for the less technical error message per this slack convo: https://mozilla.slack.com/archives/C04DYP6T6VD/p1710357586006909

robdivincenzo:
Hey @TheoChevalier, we have some hard-coded localized strings in this PR. Can you review the new translations located in `locale-strings.jsx`? Thank you!

TheoChevalier:
Hi @robdivincenzo, I can get these translated by our translation agency alongside another project (hoping to be able to submit it later today or tomorrow)

Regarding your earlier question about using `window.gettext("My string.")` in jsx files, we just need to make this small change to the makemessages command in tasks.py:

Capture d’écran 2024-03-20 à 18 15 35

I’ve tried locally and was able to extract your string. The only caveat is that gettext doesn’t support backticks.

source/js/components/petition/locale-strings.jsx
```diff
@@ -123,6 +125,8 @@ export default {
         und Ihre E-Mail-Einstellungen aktualisieren.
       </Fragment>
     ),
+    "Something went wrong and your signup wasn't completed. Please try again later.":
+      "Es ist ein Fehler aufgetreten und Ihre Anmeldung wurde nicht abgeschlossen. Bitte versuchen Sie es später noch einmal.",
```
TheoChevalier, line 129:
```suggestion
      "Etwas ist schiefgelaufen und Ihre Anmeldung konnte nicht durchgeführt werden. Bitte versuchen Sie es erneut.",
```

source/js/components/petition/locale-strings.jsx
```diff
@@ -184,6 +188,8 @@ export default {
         y actualizar tus preferencias de correo electrónico.
       </Fragment>
     ),
+    "Something went wrong and your signup wasn't completed. Please try again later.":
+      "Algo salió mal y tu registro no se completó. Por favor, inténtelo de nuevo más tarde.",
```
TheoChevalier, line 192:
```suggestion
      "Algo salió mal y tu registro no se ha completado. Vuelve a intentarlo más tarde.",
```

source/js/components/petition/locale-strings.jsx
```diff
@@ -242,6 +248,8 @@ export default {
         et modifier vos préférences de messagerie.
       </Fragment>
     ),
+    "Something went wrong and your signup wasn't completed. Please try again later.":
+      "Quelque chose s'est mal passé et votre inscription n'a pas été finalisée. Veuillez réessayer plus tard.",
```
TheoChevalier, line 252:
```suggestion
      "Quelque chose s’est mal passé et votre inscription n’a pas été finalisée. Veuillez réessayer plus tard.",
```

source/js/components/petition/locale-strings.jsx
```diff
@@ -299,6 +307,8 @@ export default {
         i zaktualizować preferencje.
       </Fragment>
     ),
+    "Something went wrong and your signup wasn't completed. Please try again later.":
+      "Coś poszło nie tak i Twoja rejestracja nie została ukończona. Spróbuj ponownie później.",
```
TheoChevalier, line 311:
```suggestion
      "Coś poszło nie tak i rejestracja nie została ukończona. Spróbuj ponownie później.",
```

source/js/components/petition/locale-strings.jsx
```diff
@@ -357,5 +367,7 @@ export default {
         e atualizar suas preferências de e-mail.
       </Fragment>
     ),
+    "Something went wrong and your signup wasn't completed. Please try again later.":
+      "Algo deu errado e sua inscrição não foi concluída. Por favor, tente novamente mais tarde.",
```
TheoChevalier, line 371:
```suggestion
      "Algo deu errado e sua assinatura não foi concluída. Tente novamente mais tarde.",
```

TheoChevalier (changes requested):
Hey @robdivincenzo added the updated translations

TheoChevalier (approved):
Thanks, looks good to me!

mmmavis (approved):
Yay! Thanks for the awesome work, @robdivincenzo !

Issue #12006, opened 2024-03-05
[Main Navigation Dropdown] NavLinkBlock


data-sync-user:
Implement a `NavLinkBlock` for a `navigation item`.

# Acceptance criteria
* [ ] A navigation item needs to link to either a CMS page, a relative URL or an external link
* [ ] External links on a navigation item must be properly identified by using different rendering than internal links
* [ ] External links always open on a new tab

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP-292)

PR #11954, opened 2024-02-29
Setup `LinkBlock` to handle multiple types of links

jhonatan-lopes:
# Description

:warning: Please disable **preboot** for the deployment of this feature! :warning: 

Setup `wagtail-link-block` with a custom `LinkBlock` to handle multiple types of links:
- Pages
- Documents
- External/relative URLs
- Anchor links
- Mail to
- Phone

Old `LinkButton` interface (doesn't allow `mailto:` links):

image

New interface:

image

Link to sample test page:
Related PRs/issues: #11875

# Checklist

**Tests**
- [x] Is the code I'm adding covered by tests?

**Changes in Models:**
- [x] Did I update or add new fake data?
- [x] Did I squash my migration?
- [ ] Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?

**Documentation:**
- [x] Is my code documented?
- [x] Did I update the READMEs or wagtail documentation?

**Merge Method**
**💡❗Remember to use squash merge when merging non-feature branches into `main`**

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP-240)

PR #12030, opened 2024-03-11
Custom base for `LinkBlock`

jhonatan-lopes:
# Description

#11954 changed `LinkBlock` to handle multiple types of links by introducing `wagtail-link-block` as a dependency to manage multiple link types. `wagtail-link-block` seemed like a good solution for standardising the way we handle links, since it was ready-made and was a good fit to our needs.

However, as I was defining a `NavLinkBlock` (#12006) that only allowed for pages, external urls and emails to use in our navigation menu, I ran into several limitations with `wagtail-link-block`. Creating a child of `wagtail-link-block` while removing fields somehow seemed like an anti-pattern and in any case its code did not handle that well.

The level of customisation necessary to remove the extra fields would make it almost a complete rewrite.

Hence, to have a more scalable approach that works for both the current `LinkBlock` and the future `NavLinkBlock`s (used in the main nav dropdowns), I decided to recreate `wagtail-link-block` supporting a dynamic number of fields.

## Implementation

This PR introduces a `BaseLinkBlock` with links to:
- Page
- External URL
- Email
- Anchor

The `BaseLinkBlock` is an abstract class, so it won't be recognised by Django/Wagtail on its own.

The only concrete implementation right now is on `LinkBlock`, which extends `BaseLinkBlock` to include also file and phone links.

This same `BaseLinkBlock` can be reused later for the `NavLinkBlock`.

## Production data

This work has also modified the migrations in #11954 (not deployed yet) to work with this new approach. The migrations will convert all current links in production to an "external URL" type of link.

## Testing

1. Log in to the CMS
2. Go to "Reports -> Block types" and find the `LinkBlock` column
3. Select a page type on in the `LinkBlock` column and verify that their links have been migrated to the new interface
4. Verify that you can add a new link block to the relevant field for that page type
5. Repeat steps 3 and 4 for all page types on the `LinkBlock` column

Link to sample test page:
Related PRs/issues: 
- #12006 
- #11954
- https://mozilla-hub.atlassian.net/browse/TP-292

# Checklist

**Tests**

**Changes in Models:**
- ~~[ ] Are my changes [backward-compatible](https://github.com/mozilla/foundation.mozilla.org/blob/main/docs/workflow.md#django-migrations-what-to-do-when-working-on-backward-incompatible-migrations). If not, did I schedule a deploy with the rest of the team?~~ :warning: This will result in downtime. preboot should be **disabled** :warning:

**Documentation:**
- [X] Is my code documented?

**Merge Method**

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP-316)


jhonatan-lopes:
I think playwright is going bonkers. Mozfest looks fine locally so I'm not sure why it's failing on CI

network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
```diff
@@ -0,0 +1,136 @@
...
+        return self.get("link_to")
+
+
+class BaseLinkBlock(blocks.StructBlock):
+    link_to = blocks.ChoiceBlock(
+        choices=[
+            ("page", "Page"),
+            ("external_url", "External URL"),
+            ("relative_url", "Relative URL"),
+            ("relative_url", "Relative URL"),
```
robdivincenzo, line 51: Appears to be a duplicate choice

jhonatan-lopes: Nice catch, thanks!

network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
```diff
@@ -0,0 +1,135 @@
...
+        if errors:
+            raise blocks.StreamBlockValidationError(block_errors=errors, non_block_errors=ErrorList([]))
+
+        return clean_values
+
+
+class BaseLinkBlockAdapter(StructBlockAdapter):
+    """Custom adapter to register JS for conditionally hiding/showing fields.
+
+    Can be used with any block that extends `BaseAbstractLinkBlock`.
```
robdivincenzo, line 121: Should this comment be for extending `BaseLinkBlock`?

jhonatan-lopes: Thanks Rob! That's fixed now

robdivincenzo (approved):

PR #12033, opened 2024-03-12
[TP-292][Main Nav] Nav link block

jhonatan-lopes:
# Description

- Add a `nav` app to Django
- Add a `NavLinkBlock`
- Make `BaseLinkBlock` more flexible and adjust it to `NavLinkBlock`

Link to sample test page:
Related PRs/issues:
- #12006
- https://mozilla-hub.atlassian.net/browse/TP-292

# Testing
- Covered by unit tests

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [X] Is my code documented?

**Merge Method**

### DIFF
--- a/network-api/networkapi/nav/blocks.py
+++ b/network-api/networkapi/nav/blocks.py
@@ -1,0 +1,27 @@
+from wagtail.telepath import register
+
+from networkapi.wagtailpages.pagemodels.customblocks.common.base_link_block import (
+    BaseLinkBlock,
+    BaseLinkBlockAdapter,
+    BaseLinkValue,
+)
+
+
+class NavLinkValue(BaseLinkValue):
+    @property
+    def open_in_new_window(self) -> bool:
+        link_to = self.get("link_to")
+        if link_to == "external_url":
+            return True
+        return False
+
+
+class NavLinkBlock(BaseLinkBlock):
+    class Meta:
+        value_class = NavLinkValue
+        label = "Navigation Link"
+        icon = "link"
+        template = "nav/blocks/nav_link_block.html"
+
+
+register(BaseLinkBlockAdapter(), NavLinkBlock)
--- a/network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
@@ -37,6 +37,8 @@ class BaseLinkValue(blocks.StructValue):
 
 
 class BaseLinkBlock(blocks.StructBlock):
+    label = blocks.CharBlock()
+
     link_to = blocks.ChoiceBlock(
         choices=[
             ("page", "Page"),
--- a/network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
+++ b/network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
@@ -4,7 +4,7 @@ from django.utils.functional import cached_property
 from wagtail import blocks
 from wagtail.blocks.struct_block import StructBlockAdapter
 
-from networkapi.wagtailpages.validators import AnchorLinkValidator, RelativeURLValidator
+from networkapi.wagtailpages.validators import RelativeURLValidator
 
 
 class BaseLinkValue(blocks.StructValue):
@@ -21,12 +21,6 @@ class BaseLinkValue(blocks.StructValue):
 
     def get_relative_url_link(self):
         return self.get("relative_url")
-
-    def get_email_link(self):
-        return f"mailto:{self.get('email')}"
-
-    def get_anchor_link(self):
-        return f"#{self.get('anchor')}"
 
     @property
     def url(self):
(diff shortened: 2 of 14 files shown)


network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
```diff
@@ -97,6 +81,9 @@ def clean(self, value):
         url_default_values = self.get_default_values()
         url_type = clean_values.get("link_to")
 
+        if not url_type:
+            errors["link_to"] = ErrorList(["Please select a link type"])
```
robdivincenzo, line 85: Could we instead use required=True in the model's field? Or do we need the additional flexibility?

network-api/networkapi/nav/blocks.py
```diff
@@ -0,0 +1,27 @@
...
+            return True
+        return False
+
+
+class NavLinkBlock(BaseLinkBlock):
+    class Meta:
+        value_class = NavLinkValue
+        label = "Navigation Link"
+        icon = "link"
+        template = "nav/blocks/nav_link_block.html"
```
robdivincenzo, line 24: I can't find the template being referenced. Is that the frontend piece Mavis is working on separately?

robdivincenzo (changes requested):
Thanks @jhonatan-lopes, looking great! Just have some minor questions to help clarify a few things for me.

network-api/networkapi/wagtailpages/pagemodels/customblocks/link_block.py
```diff
@@ -34,27 +39,23 @@ class LinkBlock(BaseLinkBlock):
...
-        super().__init__(local_blocks, **kwargs)
-        # Reoder child_blocks so that label is first
-        self.child_blocks = self.base_blocks.copy()
-        child_blocks = OrderedDict(
-            {
-                "label": self.child_blocks.pop("label"),
-            }
-        )
-        child_blocks.update({k: v for k, v in self.child_blocks.items()})
-        self.child_blocks = child_blocks
```
robdivincenzo, line 54: Double-checking that this __init__ method isn't needed. I noticed it wasn't moved with the label to `BaseLinkBlock`.

Doesn't seem necessary with how that's been organized now, but asking for clarification because in retrospect maybe it wasn't needed in the first iteration?

network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
jhonatan-lopes: This is a tricky one. All blocks have `required=True` by default but since `link_to` is a `ChoiceBlock`, it contains a blank choice (the empty string `""`). So, if editors do not select a link type, technically the block still has some data.

If we put `required=True` but remove that validation check, it will allow editors to save without selecting a "valid" type of link:
image

That's why we have to put that `if not url_type` check, which will trigger an error if editors have not made a selection (the choice is blank):
image

network-api/networkapi/nav/blocks.py
jhonatan-lopes: That's force of habit :see_no_evil: the templates will be worked on later but there's no harm on leaving this here

network-api/networkapi/wagtailpages/pagemodels/customblocks/link_block.py
jhonatan-lopes: Yep, now that the label is in the right place we can remove this `__init__` method :+1:

network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
robdivincenzo: Gotcha, thanks @jhonatan-lopes! I just tried that locally with the `BaseLinkBlock` and do see what you mean here.

I messed around with it a little bit further, and it looks like if we remove the required=False from the `LinkBlock` class, it will give us the red * and the required validation even in the case that the user selects the blank option and the additional `url_type` check is removed. Not a big deal, and it's up to you. But if we can, I feel like it'd be a consistent editor experience to have the * and out of the box validation.

Screenshot 2024-03-13 at 9 37 33 AM

network-api/networkapi/nav/blocks.py
robdivincenzo: No worries! 😆

network-api/networkapi/wagtailpages/pagemodels/customblocks/link_block.py
robdivincenzo: Awesome, thanks!

network-api/networkapi/wagtailpages/pagemodels/customblocks/common/base_link_block.py
jhonatan-lopes: I see what you mean. Yes, having the required * is indeed an advantage. I'll add that.

jhonatan-lopes: You were absolutely right, @robdivincenzo! Removing the `required=False` and doing some new migrations makes it so that the check is redundant. I've fixed that now. Thanks

robdivincenzo (approved):
LGTM! 🔥

Issue #12001, opened 2024-03-05
[Main Nav] Front-end (desktop)


data-sync-user:
Implement desktop version of main nav

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-41)

Issue #12000, opened 2024-03-05
[Main Nav] Front-end (mobile/tablet)


data-sync-user:
Implement mobile/tablet version of main nav

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-42)

PR #12146, opened 2024-04-05
[Main Nav][Frontend] NavItem template

jhonatan-lopes:
# Description

- Add HTML template for `NavItem`
- Add template tag to render SVGs from static dir
- Add SVGs for nav arrows

Link to sample test page: http://localhost:8000/pattern-library/render-pattern/fragments/blocks/nav/nav_item.html
Related PRs/issues:
- #12001
- #12000
- https://mozilla-hub.atlassian.net/browse/TP1-42
- https://mozilla-hub.atlassian.net/browse/TP1-41

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**

**Merge Method**

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-369)


jhonatan-lopes:
I just need to confirm whether we need `blue-03` colour with Nancy of if `blue-05` is fine

mmmavis (commented):
Hey @jhonatan-lopes , http://localhost:8000/en/pattern-library/render-pattern/fragments/blocks/nav/nav_item.html/ gives me a 404 😬 Should I be checking the template using a different url or method?

For the SVGs, I don't see the files being used in this PR. Depending on how we plan to use the icons, one way to do it is to include the SVG inline and use CSS/Tailwind classes to update the stroke colour (something like [this](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/11864/files#diff-6d4022516b1b49aab0d3819d14095a02da9dc5c09b2e30b2746c4c93bc9ba70fR3)). This way we don't have to include multiple SVGs of the same icon and can potentially transition between the states more smoothly. How I usually do it is to run SVGO on the SVG file, copy the source code into a html file, and add appropriate classes to it. This is just a thought. We can chat more about this in later PRs when the icons are actually used.

jhonatan-lopes:
Hey @mmmavis, thanks for taking a look at this!

Initially, I went with the approach of including the SVG from the static dir and then rendering it inside another `<svg>` tag using the `<use>` tag. However, this involved using a custom Django template tag (which I had to create) and seemed quite convoluted.

One thing I noticed is that the `cta-link` class adds an "arrow right" character automatically, using the `content` and `after` CSS:

https://github.com/MozillaFoundation/foundation.mozilla.org/blob/f3e228853b83fa28d8854d17f19068539191a87b/tailwind-plugins/type.js#L349-L355

So, I replaced that implementation with rendering the "arrow right up" (↗) directly and adding it via the `after` pseudo-selector. You can check how I did on this commit: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12146/commits/926e3831e9c37b13eac94d3429f42ec062b2a432

I think that made the code more readable and simpler but let me know what you think. Changing colors, size and all becomes trivial because it behaves like a font (it is a font glyph, after all).

This implementation indeed makes the SVGs that I've added obsolete, as we can do everything with the font glyph. I've kept them in the PR for reference but let me know if you'd prefer to delete them.

jhonatan-lopes:
For getting the 404s out of the pattern library you need to enable it. Please add `PATTERN_LIBRARY_ENABLED=True` to your .env :)

network-api/networkapi/templates/fragments/blocks/nav/item.html
```diff
@@ -0,0 +1,17 @@
+{% load static i18n wagtailadmin_tags %}
+
+{% fragment as font_styles %}tw-text-xl tw-font-semibold tw-capitalize tw-text-black{% endfragment %}
+{% fragment as hover_styles %}group-hover:tw-text-blue-80 group-hover:tw-underline{% endfragment %}
+{% fragment as after_arrow_styles %}after:tw-content-['_↗'] after:tw-font-['Nunito_Sans'] after:tw-text-gray-20 group-hover:after:tw-text-blue-80{% endfragment %}
```
mmmavis, line 5:
```suggestion
{% fragment as after_arrow_styles %}after:tw-content-['_↗'] after:tw-font-sans after:tw-text-gray-20 group-hover:after:tw-text-blue-80{% endfragment %}
```

mmmavis (changes requested):
Thanks for the clarification as well as a reminder to turn on `PATTERN_LIBRARY_ENABLED` 😂 . I left two comments regarding two Tailwind classes. For SVGs - yay let's go with font glyph instead!

network-api/networkapi/templates/fragments/blocks/nav/item.html
```diff
@@ -0,0 +1,17 @@
...
+
+<div class="tw-group tw-flex tw-flex-col tw-gap-2 tw-p-2 hover:tw-bg-blue-03">
+    <a class="{{ font_styles }} {% if value.is_external %} {{ after_arrow_styles }} {% endif %} {{ hover_styles }}"
+       href="{{ value.url }}" {% if value.is_external %}target="_blank"{% endif %}>
+        {{ value.label }}
+    </a>
+    {% if value.description %}
+        <p class="tw-font-light tw-text-sm tw-text-gray-80 tw-hidden medium:tw-inline">
```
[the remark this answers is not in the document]
jhonatan-lopes: For sure, done!

network-api/networkapi/templates/fragments/blocks/nav/item.html
jhonatan-lopes: The glyphs do not work properly with `tw-font-sans`. It seems that one of the other fonts takes precedence and then the arrow is displayed weirdly:

image

Forcing "Nunito Sans" (using `tw-font-['Nunito_Sans']`), we get the expected result:

image

I can add a comment if it helps...

mmmavis: Ah I see. Don't worry about this now. We can do a final check at a later phase prior to requesting review from designers.

mmmavis (approved):

data-sync-user:

PR has already been merged.

Issue #12141, opened 2024-04-04
Investigate about FRU Cookie Request


data-sync-user:
The FRU team has share a document with our team regarding a request to use cookies as currently our cookie policy means that they need to handle two separate versions of the same product, and our version also stops them from being able to track the user journey through the foundation process, meaning that they have less information available to track issues and resolve them on time.

You can read about the initial request [here](https://docs.google.com/document/d/1Of--c02sfPxURaFYLNE_FYfyAcj0YQ_8YLhIBn1VtU0/edit).

Coincidentally, we are aware that MZLA has experience the same type of issues but we have been asked to analyze the document and come up with some feedback, mainly around either agreeing or disagreeing with their proposed solution.

A few things that can aid on this investigation are:

* Review how the FRU app is set up right now from our side
* Evaluate how is the MZLA team using FRU app and how it is configured

Important Note: The CRM team is also taking a look at this currently, as Shing mentioned she expected this to be a joint responsibility between CRM/Web teams. Once he have feedback, we will be compiling it in a document that the CRM team also uses.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-364)

PR #12147, opened 2024-04-05
Update referrer policy to use strict-origin-when-cross-origin

robdivincenzo:
# Description

This PR updates the referrer policy to use `strict-origin-when-cross-origin` instead of `same-origin`. While `same-origin` is a Django default, it can be too restrictive when integrating 3rd party tools like the donation platform as extensively documented in [TP1-364](https://mozilla-hub.atlassian.net/jira/software/c/projects/TP1/boards/1918?selectedIssue=TP1-364) / #12141.

`strict-origin-when-cross-origin` is a modern, [balanced approach](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy) that still respects security & privacy and **only** includes the origin (i.e. https://foundation.mozilla.org) in cross-origin requests like 3rd party tools. It is the default of modern browsers and the current setting for mozilla.org and thunderbird.net (who uses the donation modal without issue).

Note: this is just the default .env variable / Django setting, and we will still need to change the config var in heroku.

Related PRs/issues: #12141 / [TP1-364](https://mozilla-hub.atlassian.net/jira/software/c/projects/TP1/boards/1918?selectedIssue=TP1-364)

# To Test
Open Browser console go to "Network" and view request headers to confirm ReferrerPolicy is set to `strict-origin-when-cross-origin`.

### DIFF
--- a/env.default
+++ b/env.default
@@ -63,6 +63,9 @@ CSP_INCLUDE_NONCE_IN=script-src
 # Security config
 SECURE_CROSS_ORIGIN_OPENER_POLICY=" 'same-origin-allow-popups' "
 
+# Referral policy
+REFERRER_HEADER_VALUE =" 'strict-origin-when-cross-origin "
+
 # Petition test campaign id for Salesforce Sandbox
 PETITION_TEST_CAMPAIGN_ID=7017i000000bIgTAAU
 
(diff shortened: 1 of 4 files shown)


jhonatan-lopes (changes requested):
Hey @robdivincenzo, awesome work! Totally agree with the change, it seems like going with the `strict-origin-when-cross-origin` is the way to go.

One thing I noticed, though, is that Django now provides a way to [specify the referrer policy directly through their security middleware](https://docs.djangoproject.com/en/5.0/ref/middleware/#referrer-policy).

I believe this is the perfect opportunity to remove our custom middleware implementation in favour of Django's builtin.

robdivincenzo:
Thanks for pointing that out @jhonatan-lopes! I've adjusted the PR to remove the custom middleware.

jhonatan-lopes (approved):
LGTM :+1: :fire: thanks Rob

data-sync-user:

This PR was merged on April 9th.

PR #12148, opened 2024-04-05
Match FRU CSP Directives

robdivincenzo:
# Description

This PR will make additional adjustments to CSP per FRU directives [noted here](https://fundraiseup.com/support/content-security-policy/). This code contains updates to local environments. The only changes in heroku / prod config env are:

1) Missing *.paypalobjects.com from img-src
2) Removing single quotes from SECURE_CROSS_ORIGIN_OPENER_POLICY `'same-origin-allow-popups'`. Double-checking [specifications](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cross-Origin-Opener-Policy#specifications), it says it should be a token.

Related PRs/issues: #12141 / TP1-364


.github/workflows/continous-integration.yml
```diff
@@ -145,10 +145,11 @@ jobs:
       XSS_PROTECTION: True
       CSP_CONNECT_SRC: "*"
       CSP_FONT_SRC: "'self' https://fonts.gstatic.com https://fonts.googleapis.com https://code.cdn.mozilla.net https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/fonts/ data: https://static.fundraiseup.com/common-fonts/"
-      CSP_FRAME_SRC: "'self' https://www.google.com/recaptcha/ https://*.stripe.com https://pay.google.com https://*.paypal.com"
+      CSP_IMG_SRC: "* data: blob: *.fundraiseup.com ucarecdn.com pay.google.com *.paypalobjects.com"
```
mmmavis, line 148: Let's be more specific and include the `https://` protocol for extra safety 😁 . Please also update other files as well.

.github/workflows/continous-integration.yml
```diff
@@ -145,10 +145,11 @@ jobs:
...
+      CSP_FRAME_SRC: "'self' https://www.google.com/recaptcha/ https://*.stripe.com https://pay.google.com https://*.paypal.com *.fundraiseup.com"      
       CSP_SCRIPT_SRC: "'self' 'unsafe-inline' https://www.google-analytics.com/analytics.js http://*.shpg.org/ https://comments.mozillafoundation.org/ https://airtable.com https://platform.twitter.com https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/gsap.min.js https://cdnjs.cloudflare.com/ajax/libs/gsap/3.8.0/ScrollTrigger.min.js https://*.googletagmanager.com https://*.fundraiseup.com https://mozillafoundation.tfaforms.net https://www.google.com/recaptcha/ https://www.gstatic.com/recaptcha/ 'unsafe-eval' https://*.stripe.com https://m.stripe.network https://*.paypal.com https://*.paypalobjects.com https://pay.google.com"
       CSP_STYLE_SRC: "'self' 'unsafe-inline' https://code.cdn.mozilla.net https://fonts.googleapis.com https://platform.twitter.com https://mozillafoundation.tfaforms.net https://stackpath.bootstrapcdn.com/font-awesome/4.7.0/css/font-awesome.min.css"
-      SECURE_CROSS_ORIGIN_OPENER_POLICY: "'same-origin-allow-popups'"
```
mmmavis, line 151: I know the special CSP source keywords like `self` or `unsafe-inline` need to be wrapped in quotation marks. I'm not sure about [SECURE_CROSS_ORIGIN_OPENER_POLICY](https://django.readthedocs.io/en/stable/ref/settings.html#secure-cross-origin-opener-policy). @jhonatan-lopes do you know?

network-api/networkapi/settings.py
```diff
@@ -628,7 +628,7 @@ class DatabasesDict(TypedDict):
 # Security
 SECURE_BROWSER_XSS_FILTER = env("XSS_PROTECTION")
 SECURE_CONTENT_TYPE_NOSNIFF = env("CONTENT_TYPE_NO_SNIFF")
-SECURE_CROSS_ORIGIN_OPENER_POLICY = env("SECURE_CROSS_ORIGIN_OPENER_POLICY", default="'same-origin'")
+SECURE_CROSS_ORIGIN_OPENER_POLICY = env("SECURE_CROSS_ORIGIN_OPENER_POLICY", default="same-origin-allow-popups")
```
mmmavis, line 631: Let's keep the default to the safest option (i.e., `same-origin`), similar to how we are setting `self` as the default value for the CSP.

mmmavis (changes requested):
@robdivincenzo thanks for organizing the loooong CSP lists. I left some comments and also pinged @jhonatan-lopes for a question regarding whether or not to wrap `same-origin-allow-popups` in quotation marks.

.github/workflows/continous-integration.yml
jhonatan-lopes: I think it doesn't because it's a [straight Django setting](https://docs.djangoproject.com/en/4.2/ref/middleware/#cross-origin-opener-policy) instead of being managed through the `django-csp` package

network-api/networkapi/settings.py
robdivincenzo: Thanks for pointing that out how we're handling the defaults @mmmavis !

mmmavis (approved):
Perfect! Thank you so much for updating these configs @robdivincenzo !

data-sync-user:

This PR was merged on April 9th.

PR #12100, opened 2024-03-25
Enable js and jsx for exporting via makemessages cmd

robdivincenzo:
# Description

This pull requests enables exporting from js and jsx strings for locallization via makemessages command.

To export a string: window.gettext("my text for export")

Related PRs/issues: #11910

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP-393)

PR #12249, opened 2024-04-23
Refactor newsletter petition localization

robdivincenzo:
# Description

This PR refactors the localization methods for the newsletter and petition components. Now that gettext is exporting localization strings properly via `makemessages`, we're revisiting these components to replace the hardcoded `getText` method.

This first PR is a confirmation that everything is exporting properly for Pontoon translation. Further work is needed to remove `locale-strings.jsx` and `getText` method across the codebase, and will be completed via follow-up issues.

Link to sample test page:
Related PRs/issues: #11910 #12100

# To Test

1) Ensure your localization string repo is set up
2) Export the strings via `inv makemessages`
3) Review for new changes in .po files to note that the gettext strings are exported for translation

### DIFF
--- a/source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
+++ b/source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
@@ -1,7 +1,6 @@
 import React from "react";
 import PropTypes from "prop-types";
 import { ReactGA } from "../../../common";
-import { getText } from "../../petition/locales";
 
 /**
  * Higher-order component that handles form submission logic and validation
@@ -27,20 +26,20 @@ function withSubmissionLogic(WrappedComponent) {
       this.validators = {
         email: (value) => {
           if (!value) {
-            return getText(`This is a required section.`);
+            return gettext("This is a required section.");
           }
 
           // Regex copied from join.jsx
           const emailRegex = new RegExp(/[^@]+@[^.@]+(\.[^.@]+)+$/);
           if (!emailRegex.test(value)) {
-            return getText(`Please enter a valid email address.`);
+            return gettext("Please enter a valid email address.");
           }
 
           return null;
         },
         privacy: (value) => {
           if (value !== "true") {
-            return getText(`Please check this box if you want to proceed.`);
+            return gettext("Please check this box if you want to proceed.");
           }
           return null;
         },
@@ -191,16 +190,16 @@ function withSubmissionLogic(WrappedComponent) {
 
         if (res.status !== 201) {
           this.setState({
-            apiError: getText(
-              `Something went wrong and your signup wasn't completed. Please try again later.`
+            apiError: gettext(
+              "Something went wrong and your signup wasn't completed. Please try again later."
             ),
           });
           throw new Error(res.statusText);
         }
       } catch (error) {
         this.setState({
-          apiError: getText(
-            `Something went wrong and your signup wasn't completed. Please try again later.`
+          apiError: gettext(
+            "Something went wrong and your signup wasn't completed. Please try again later."
           ),
         });
         throw error;
@@ -219,7 +218,7 @@ function withSubmissionLogic(WrappedComponent) {
       if (
         this.state.apiSubmissionStatus === this.API_SUBMISSION_STATUS.SUCCESS
       ) {
-        message = getText(`Thanks!`);
+        message = gettext("Thanks!");
       }
 
       return message;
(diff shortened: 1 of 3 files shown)


robdivincenzo:
Great catch @TheoChevalier ! I've updated the `gettext` calls to include the actual string rather than the aliases. From my [understanding of the docs](https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#internationalization-in-template-code), it's ok to include html tags as part of the translation. And they do seem to export fine. But I'm curious to hear if everything imports back from Pontoon ok. Any thoughts?

TheoChevalier (changes requested):
This doesn’t work unfortunately, I suspect we need to do interpolation for tags, I can have a look tomorrow

Capture d’écran 2024-04-23 à 18 53 17

TheoChevalier:
Note that the documentation linked is for templates only, here we can only use what is in the JavaScript section and it works slightly differently: https://docs.djangoproject.com/en/5.0/topics/i18n/translation/#internationalization-in-javascript-code

robdivincenzo:
Thanks Theo, I see it now 😅. Since that value is passed back to React, React is going to escape it by default. The html needs explicitly defined. How would you feel with the translated text fragmented out? Something like:

```
    const label = (<span>
      {pgettext("Fragment of: I'm okay with Mozilla handling my info as explained in this Privacy Notice","I'm okay with Mozilla handling my info as explained in this ")}
      <a target='_blank' href='https://www.mozilla.org/privacy/websites/'>{pgettext("Fragment of: I'm okay with Mozilla handling my info as explained in this Privacy Notice","Privacy Notice")}</a>
    </span>)

    return (
      <InputCheckboxWithLabel
        id={this.ids[name]}
        name={name}
        label={label}
        value={this.getFormFieldValue(name)}
        checked={this.getFormFieldValue(name) === "true"}
        onChange={(event) => this.handlePrivacyChange(event)}
        required={true}
        errorMessage={this.props.errors[name]}
      />
    );
  }

```
We could use pgettext to export a comment you'd think necessary for an accurate translation of the fragments. I made it "Fragment of: [alias]" so it could be identified as a fragment of whatever alias. We could also put the whole html string there if you wanted, but that might be too verbose.

TheoChevalier:
Thanks @robdivincenzo , let’s please use the whole HTML then, because it won’t work in some languages as translators often need to move the linked part (maybe not true for our current set of languages, but will surely be the case in other strings or when we support new languages)

robdivincenzo:
Hey @TheoChevalier, knowing that translators might need to move the link around, I've tweaked the idea slightly. The translation now has 3 pieces for `pre-link text`, `link text`, and `post-link text`. This will enable translators to edit any part of the privacy notice text, rendering the link wherever as necessary, and securely presenting the HTML.

source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
```diff
@@ -239,8 +238,16 @@ function withSubmissionLogic(WrappedComponent) {
...
-            <p>{getText(`confirm your email opt-in`)}</p>
-            <p>{getText(`manage your subscriptions`)}</p>
+            <p>
+              {gettext(
+                "If you haven’t previously confirmed your opt-in to a Mozilla-related email subscription you may have to do so now. <strong>Please check your inbox or spam filter for an email from us to click and confirm your subscription</strong>."
+              )}
+            </p>
+            <p>
+              {gettext(
+                "If you have already confirmed your opt-in to receive Mozilla-related emails, you can now <a href='https://www.mozilla.org/newsletter/recovery/' target='_blank'>manage your subscriptions</a> and update your email preferences."
```
TheoChevalier, line 248: Thanks @robdivincenzo , your approach works, let’s just apply it on these two strings as well

TheoChevalier (changes requested):

source/js/components/newsletter-signup/organisms/with-submission-logic.jsx
robdivincenzo: Thanks @TheoChevalier! Feel free to give it another go

TheoChevalier (changes requested):
Unfortunately, strings with back-ticks are not extracted anymore

robdivincenzo:
Thanks @TheoChevalier, please try once more. The docs suggested it was available in our version of gettext, but you're right that it doesn't export.

TheoChevalier (changes requested):
Hm, looks like the translator notes are not fully extracted now, I suspect the concatenation needs to be done slightly differently with pgettext in JS

Capture d’écran 2024-05-15 à 17 49 57

Issue #12233, opened 2024-04-18
[Main Nav] Setup dropdown animations


data-sync-user:
Setup animations for closing and opening dropdowns both on all viewports according to the specs on Figma.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-547)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/62049">MainNavReload.mkv</a>

PR #12260, opened 2024-04-26
[Mobile Nav] Added sliding effect and allowed multiple open accordions

mmmavis:
# Description

- Fixed a few spacing issues.
- Mobile Nav: 
    - added sliding effect
    - allow multiple open accordions

---
Link to review app: https://foundation-s-feature-ma-36qyhj.herokuapp.com/en/ (CMS cred: `admin2`/`admin2password`)
Related PRs/issues: #12233

### DIFF
--- a/source/js/components/accordion/accordion.js
+++ b/source/js/components/accordion/accordion.js
@@ -16,7 +16,7 @@ class Accordion {
   bindEvents() {
     this.title.addEventListener("click", (e) => {
       e.preventDefault();
-      let open = !this.content.classList.contains("tw-hidden");
+      let open = this.title.getAttribute("aria-expanded") === "true";
 
       if (open) {
         this.close();
@@ -37,14 +37,12 @@ class Accordion {
   }
 
   open() {
-    this.content.classList.remove("tw-hidden");
     this.chevron.classList.remove("tw-rotate-180");
     this.title.setAttribute("aria-expanded", "true");
     this.content.setAttribute("aria-hidden", "false");
   }
 
   close() {
-    this.content.classList.add("tw-hidden");
     this.chevron.classList.add("tw-rotate-180");
     this.title.setAttribute("aria-expanded", "false");
     this.content.setAttribute("aria-hidden", "true");
--- a/source/js/components/nav/mobile-dropdown.js
+++ b/source/js/components/nav/mobile-dropdown.js
@@ -27,12 +27,6 @@ class NavMobileDropdown extends Accordion {
     this.title.classList.add("tw-border-s-4");
   }
 
-  getOpenSibling() {
-    return document
-      .querySelector(`${NavMobileDropdown.selector()} [aria-expanded="true"]`)
-      ?.closest(NavMobileDropdown.selector());
-  }
-
   bindEvents() {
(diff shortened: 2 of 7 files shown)


sabrinang (changes requested):
**Desktop**
- When moving from one item to another, I think there is a quick fade/ease out that makes it look like there's a bit of background flashing through going on _(this could be simplified)_

**Tablet**
- Line below navigation disappears after opening item like “who we are” or after when other items snap to top
- Border reappears when closing but items do not reset so top item (so “who we are” gets hidden) appears fully hidden 
- The snap to top is working inconsistently for me, but mainly “Blog” moves up a bit more all the time while the others often stay in place which makes it feel inconsistent/unpredictable

**Mobile**
- When closing a section other than the first section, a strip is seen of “who we are” but does not fully reset back to the top

I’m thinking overall, having close/open animation without snap to top would provide a more predictable/consistent experience

nancyt1:
**Mobile**
- Would we be able to animate the initial open and close of the menu? 
- I also noticed that when a nav item is open but i haven't clicked on anything and I close the whole menu, after opening it again the original nav item is still opened. Should we reset once someone closes the menu? 
- +1 to sabrinas comment about not doing snap to top to be more predictable

sabrinang:
The tablet/mobile issues resolved with removing the snap to top combo. Wanted to check in on desktop, if it was possible to not have the background flashing through when transitioning from one primary nav item to another?

source/js/components/accordion/accordion.js
```diff
@@ -16,7 +16,7 @@ class Accordion {
   bindEvents() {
     this.title.addEventListener("click", (e) => {
       e.preventDefault();
-      let open = !this.content.classList.contains("tw-hidden");
+      let open = this.title.getAttribute("aria-expanded") === "true";
```
mmmavis, line 19: Checking `aria-expanded` instead as mobile nav doesn't use `tw-hidden` to toggle visibility.

source/js/components/nav/mobile-dropdown.js
```diff
@@ -63,30 +58,20 @@ class NavMobileDropdown extends Accordion {
   }
 
   open() {
-    super.open();
```
mmmavis, line 66: Summary of changes in this file:

- Adding sliding effect by animating section's height
- We've decided to allow multiple open accordions at the same item. There's no longer needs to reset sibling sections.

nancyt1 (approved):

sabrinang (approved):

source/sass/components/primary-nav.scss
```diff
@@ -124,7 +124,9 @@
 
     .wide-screen-menu {
       line-height: $logo-height; //Tied to logo height so that the "active" page border shows up at the correct alignment consistently
+      height: auto;
```
mmmavis, line 127: I can't wait to get rid of this file once we fully phase out the old nav 😂

robdivincenzo (approved):
LGTM 🚀. And very clear comments 📝, thanks Mavis!

data-sync-user:

PR has already been merged.

Issue #12289, opened 2024-04-30
[Main Nav] Styling Adjustments


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-605)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/62298">image-20240501-171923.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/62300">image-20240501-172145.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/62299">image-20240501-172550.png</a> | <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/62313">SCR-20240501-mufy.png</a>

PR #12296, opened 2024-05-02
Various styling adjustments for new nav

mmmavis:
@danielfmiranda This PR contains quite a big of styling fixes and can be tricky to do in-depth code review. However, I will say since these are all styling related fixes, as long as we get the okay from designers and the code changes don't look concerning to you, it should be safe to merge. 😁 

---

### Changes
- showed desktop styled nav on xlarge screens instead of large
- fixed various spacing issues
- changed 'title' to 'aria-label' to avoid tooltip

# Description

Link to sample test page: https://foundation-s-12289-nav--ns14nh.herokuapp.com/ (CMS creds: `admin2/admin2password`)
Related PRs/issues: #12289 

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-613)


tailwind-plugins/site-nav.js
```diff
@@ -0,0 +1,58 @@
+const plugin = require("tailwindcss/plugin");
```
mmmavis, line 1: I'm creating this `site-nav.js` Tailwind plugin in this PR so we can create new utility class for the nav work. Once we can fully sunset the old nav, my plan is to clean up the template code a bit by converting the loooong class lists into a new class defined in this file.

nancyt1 (commented):
Not sure if it's just the review app but when I open a section, I can't scroll all the way down to the newsletter button and the newsletter button doesn't work. But all other spacing looks good!

One mobile change:
SCR-20240506-oapb

mmmavis:
@nancyt1 Thanks for the feedback. It seems like the newsletter button functionality is also broken on staging/production. I will file a new ticket for it. 

> Not sure if it's just the review app but when I open a section, I can't scroll all the way down to the newsletter button 

This seems to be regression introduced in this PR. I'll take a look. 

> Spacing

SCR-20240506-oapb

Did you mean increasing all sections' bottom padding by 1rem?

nancyt1:
@mmmavis If there isn't padding already at the bottom of the sections then yes. But it seemed like all other sections with a button or had the featured section had padding except ones that end in just a nav column.

mmmavis:
@nancyt1 I adjusted the bottom padding as noted in your screencap to 40px/2.5rem. Let me know what you think!

For the two issues related to newsletter button - I realized they are both appearing on staging and not newly introduced in this PR. I have created a new ticket and the devs will tackle in a different PR.

nancyt1 (approved):
Lovely! Thank you!

mmmavis:
I managed to find a fix to make newsletter button show again. Gonna slide the fix into this PR.

To code reviewer / @danielfmiranda -
Here's the [commit](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12296/commits/0c1e28cbd79f7c6ca4d548d362804050efa7a1f8). Since primary_nav.html is shared in between the old nav and new nav (and MozFest site), I had to find a new way to target `.narrow-screen-menu` on the new nav. I've done so by adding a new class `new-nev-enabled` to `<body>` and move some styling over to SCSS so they can be specific for the new nav. This is not the cleanest solution but is the safest as we have too many code shared by different variation of the navs. I can wait for the day to tidy up the code!

source/sass/components/primary-nav.scss
```diff
@@ -245,11 +245,12 @@
...
       overflow-y: auto;
-      height: 100vh;
-      transition: all 0.5s;
+      transition:
+          opacity 0.2s,
+          height 0.5s;
 
       &.hidden {
-        height: 0;
+        opacity: 0;
```
mmmavis, line 253: adding this rule back as the old nav needs it

danielfmiranda (approved):
Looks great! Thanks @mmmavis 👍 approved!

data-sync-user:

PR has already been merged.

Issue #12348, opened 2024-05-16
Host Google Fonts internally


data-sync-user:
### Describe the feature request
We currently host Google Fonts via [http://fonts.googleapis.com/](http://fonts.googleapis.com/) which is not ideal per our current policy. We should be able to host the fonts ourselves in our project to resolve this situation.

### Additional context
Initial request per Slack message from Nneka Soyinka.
We also want to ensure that any other sites that fall under core stack such as Mozfest and PNI have this update applied as well.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-671)

PR #12351, opened 2024-05-17
Self Host Fonts

robdivincenzo:
# Description

This PR will introduce self-hosted fonts rather than relying on google's font delivery service.

Link to sample test page:
Related PRs/issues: TP1-671 #12348


mmmavis (approved):
Great work! Thanks @robdivincenzo ! I didn't realize this many files would be affected until I saw your PR 😅

As a follow-up discussion, I wonder if we want to move fonts related bits out of the SCSS/SASS ecosystem to improve our build time. It's very unlikely that we wanna modify those font files (we probably shouldn't anyway). If we can convert the font `.scss` files into `.css` and just include them as is, it will shorten the watch/build cycle during development. SCSS does take very looooong to compile and that is that one thing I dislike about SCSS. I will file a new ticket to discuss about this.

robdivincenzo:
Thanks @mmmavis, I agree with you and prefer it being a non-compiling CSS file. Didn't see any directory where we're doing that currently, but I don't see why we can't start! I've split this off into [TP1-685](https://mozilla-hub.atlassian.net/browse/TP1-685) / #12362 so that we can merge this in / unblock main now

data-sync-user:

PR has already been merged.

Issue #10801, opened 2023-06-28
Strange horizontal panning on Opportunity Pages on iOS Safari


mmmavis:
This is a follow-up for https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10232 as my original PR didn't seem to fix the issue on certain Opportunity Pages.

For example, 

* works fine: https://foundation.mozilla.org/initiatives/web-literacy/core-curriculum
* has panning issue: https://foundation.mozilla.org/initiatives/web-literacy/core-curriculum/participate/

There's some overflowing element on the page which causes the strange horizontal panning. See recording below.

I suspect the problem to be mismatched layout classes between Bootstrap and Tailwind. e.g., using `.container` from Bootstrap with `.tw-row` from Tailwind. Since the issues don't occur on all Opportunity Pages, I'm guessing the underlying problem may be due to specific type of streamblock.

https://github.com/MozillaFoundation/foundation.mozilla.org/assets/2896608/b535a612-03b6-4e8d-b0dc-6c0ed01f2033

## Dev Task
* [ ] Find out what's causing the overflow issue
* [ ] Update this ticket with findings
* [ ] If the underlying issue is coming from specific type of streamblock, ensure the fix doesn't create regression for other page types.

## Acceptance Criteria
* [ ] As an user, when I navigate [this page](https://foundation.mozilla.org/initiatives/web-literacy/core-curriculum/participate/) on iOS Safari, I don't get any strange panning issue.

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/TP1-168)

mtdenton:
@mmmavis to groom here

data-sync-user:
➤ Ramon Ramos commented:

The horizontal scrolling is caused by a margin mismatch between bootstrap and tailwind row container inside specific content blocks: `.tw-container.image_text-block .streamfield-content .tw-row` being the main offender.

Solution proposed involves targeting said elements and applying a calc() to trim  1px from right margin.

Seems to be working fine in iOS v17.5.1.

PR #12444, opened 2024-06-06
Fix horizontal scrolling on pages including Streamfield ImageText block

ramram-mf:
# Description

The `Image Text` block, specifically the `.image_text-block .streamfield-content .tw-row` elements, when rendered on mobile viewports, calculates differently horizontal margins adhering to tailwind unit scales, having `16px` to each side, providing the following box model:
image

This PR attempts to mitigate the resulting user experience behavior in mobile devices, which introduces horizontal body wide scrolling. This horizontal scrolling gets in the way of vertical scrolling and immediately disrupts user experience in affected pages.

The solution proposed uses a `calc()` css function to remove the resulting extra `1px` from the right of the elements so it doesn't overflow from body.

image

Fix will only be applied to small viewports.
image

Link to sample test page: https://foundation-s-tp1-168-st-l8zo7d.herokuapp.com/en/initiatives/web-literacy/core-curriculum/participate/
Related PRs/issues: #10801

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-768)


source/sass/wagtail/blocks.scss
```diff
@@ -97,3 +97,12 @@
...
 }
+
+// Fixes horizontal overflow for Image Text blocks
+.tw-container.image_text-block {
+  .streamfield-content {
+    .tw-row {
+      margin-right: calc(-1rem + 1px);
+    }
+  }
+}
```
robdivincenzo, line 108: There's an [ongoing epic](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/7045) to move our codebase away from Bootstrap and Sass. So when possible, we'll use tailwind classes instead of adding in more scss / css.

Think we can try the same rule via a tailwind class in the template file instead `/networkapi/wagtailpages/templates/wagtailpages/blocks/image_text.html`?

robdivincenzo (changes requested):
Great job @ramram-mf, the new rule addresses the issue and the horizontal scroll is no more! I left one small comment on adjusting the implementation. Please reach out if you have any questions!

source/sass/wagtail/blocks.scss
ramram-mf: Thanks Rob, I was trying to figure out a better place to implement it while discussing the fix approach. I think the block template is definitely the way to go.

robdivincenzo (approved):
LGTM, thanks @ramram-mf! 🚀

data-sync-user:

PR has been merged.

Issue #10858, opened 2023-07-11
[Design Updates] Secondary menu cuts off if there's too many nav items or long titles


nancyt1:
Currently on desktop sizes, it seems the secondary menu will decrease column size on smaller screens and this makes long titles stack a little weirdly ([example on what we fund page.](https://foundation.mozilla.org/en/what-we-fund/)) Then if there's too many items, it will get cut off. There's not a live example of this cut off but it's an edge case that I think we should address. 

My initial idea was to allow the secondary menu to grow and scroll horizontally but open to any suggestions on how to make this more accessible.

**Example page:**

<img width="812" alt="image" src="https://github.com/MozillaFoundation/foundation.mozilla.org/assets/74557207/1a6be36a-5bc9-4498-9abf-1af47598350e">

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-175)

mmmavis:
Somewhat related: [PNI - Product Review - Tabs not accessible for certain locales](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10101)

nancyt1:
Yes i think this is still needed!

simont-cr:
Thanks Nancy! I will leave this as a P2 for the time being, so we can focus on the new nav and we can then revisit this issue.

data-sync-user:
➤ Rob DiVincenzo commented:

Agree with Nancy’s suggestion! 

I think here’s a start for acceptance criteria:

The overflowing secondary menu should be handled by a horizontal scroll below the element. Ideally, add a slight fade to the text as it tapers off-screen. On browsers/systems where scrollbars are hidden until used, it would help indicate to the user there are more links and to scroll. The links should also be keyboard navigable for accessibility. And as you tab through the links, the links should become visible on the screen as if horizontally scrolling.

PR #12559, opened 2024-06-30
Add scroll shadow to `#multipage-nav` wrapper for overflowing screens

ramram-mf:
# Description

This PR attempts to mitigate the overflowing wrapping layout of secondary navigation for non mobile screens by adding a scroll shadow feature to the container, which displays a subtle shadow on the sides where overflowing elements are present.

Current state:
image

Fixed state:
image
image

Link to sample test page: https://foundation-s-tp1-175-se-pxywdd.herokuapp.com/en/what-we-fund/
Related PRs/issues: #10858

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-889)


robdivincenzo:
Thanks @ramram-mf, adding Nancy for a design review as well!

Nancy, what do you think about this treatment of a horizontal scroll in response to #10858?

robdivincenzo (approved):
Code approved @ramram-mf! Ideally our plan is to move away from using .scss, but we may have to for this element since it's still using it and a full refactor of the element seems out of scope for this issue.

Please hold off on merging until Nancy takes a look. Thanks!

nancyt1:
@robdivincenzo Can this be a white gradient instead? or a lighter grey if white doesn't look right?

ramram-mf:
This would be a lighter shadow variant:
image
image

nancyt1:
@ramram-mf This [page](https://foundation.mozilla.org/en/privacynotincluded/categories/toys-games/) on the category pill buttons has an example of what I mean with the white gradient. Is that solution feasible in this context?

robdivincenzo:
If I'm understanding this a bit better- it sounds like to do the white gradient, we will lose the disappear / reappear gradient effect. But since we prefer the design to be consistent with the rest of the site, so we should continue to mimic the white gradient example that @nancyt1 linked. 

My 2 cents- I'm ok with losing the additional effects to opt for consistency (even though they're really nice!). And I agree with @ramram-mf that we shouldn't pivot to JS.

nancyt1:
@ramram-mf Actually, talking about this further at crit, we were thinking the better solution for now is just stacking the elements into 2 rows like [mozilla festival](https://www.mozillafestival.org/en/highlights/mozfest-house-amsterdam-2023/) has it. This will allow all links to be visible at once and doesn't need anything fancy. Would that work?

ramram-mf:
Ok here is the revised version of the secondary navigation, this time wrapping around overflowing elements following UI style taken from Mozfest navigation.

image

nancyt1 (approved):

PR #12654, opened 2024-07-23
Add main navigation visual regression tests

ramram-mf:
# Description

This PR adds visual regression tests for the main navigation, with an open state.

Link to sample test page:
Related PRs/issues: #

### DIFF
--- a/tests/visual.spec.js
+++ b/tests/visual.spec.js
@@ -67,9 +67,26 @@ function testMozfestURL(path, locale = `en`) {
   return testURL(mozfestBaseUrl(locale), path);
 }
 
+async function expandDropdown(page, dropdownSelector) {
+  await page.hover(dropdownSelector);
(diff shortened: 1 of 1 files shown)


ramram-mf:
Currently, the following menu elements have hardcoded target values and will reset the users language to english by default, besides breaking the integration test since the link wont exist as it redirects the hardcoded link to the english language one:

1. `What you can do` \ `Learn` \ `Explore the resarch hub`
2. `What you can do` \ `Donate` \ `Make a donation`

I have tested updating the first one to a `Page` value which will maintain the site language when clicking it. The second one uses a query string (`?form=donate-menu`) that seems to be unreachable by the CMS capabilities at the moment, so I suggest to detect and skip that specific link until we come up with a solution for that.

robdivincenzo (approved):
Awesome, thanks @ramram-mf! R+

Issue #10332, opened 2023-03-15
Add "wide" option to article TableBlock


danielfmiranda:
### Describe the feature request
Currently article page TableBlocks break out the bounds of the content and it would be great to have a checkbox like the one in the image block to display the table at wide or narrow widths. By default, it should align with the content block.

**AC**

* [ ] As an editor, I'm able to check a box to have my table block wider than the content.
* [ ] As a reader, I'm able to see tables the same width as the content block if "wide" box is unchecked.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-144)

simont-cr:
@danielfmiranda Can this ticket be closed per the merged issue #10163?

PR #12843, opened 2024-09-09
Add "wide" option to article TableBlock

robdivincenzo:
# Description

This PR will create a new `WideTableBlock` on article pages with a checkbox so that users can select a table width of either wider than page content (checked by default), or the same width as page content (unchecked by default).

The editor experience remains the same, except the addition of now having a checkbox to change the table width.

Related PRs/issues: TP1-144 / #10332 

# To Test
This is a little tricky because I was only able to find privately published examples of basic tables on articles.

1) Check out `main`
2) Copy staging data via `inv copy-stage-db`
3) Login to the wagtail dashboard via `/cms/`
4) Navigate to the parent public page: http://localhost:8000/cms/pages/16712/edit/
5) Change visibility to `Public` and publish the page
6) Navigate to the view the table on the article page: http://localhost:8000/en/campaigns/personalised-pricing/personalised-pricing/
8) Navigate to the article edit page in order to note the editor screen of the table with no checkbox: http://localhost:8000/cms/pages/16721/edit/
9) Check out this branch `TP1-144-10332-add-wide-table-block` (you may also need a new stage db)
10) Run migrations via `inv migrate`
11) Navigate to http://localhost:8000/en/campaigns/personalised-pricing/personalised-pricing/ in order to confirm the migrated table looks the same
12) Navigate to http://localhost:8000/cms/pages/16721/edit/ in order to view the editor screen of the table and confirm the `Wide` option now exists and is checked by default.
13) Uncheck "Wide" to reduce the width to the content of the page and publish the page.
14) Stop docker and re-run `docker-compose up` in order to get new tailwind classes
15) Navigate to http://localhost:8000/en/campaigns/personalised-pricing/personalised-pricing/ in order to view the now narrow table.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1193)


network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
```diff
@@ -2,41 +2,31 @@
 {% load wagtailcore_tags table_block_tags %}
 
 {% block block_content %}
-    <div class="article-table-block table-responsive">
+    <div class="article-table-block table-responsive{% if not self.wide %} tw-mx-auto tw-max-w-[80%]{% endif %}">
```
danielfmiranda, line 5: Hi @robdivincenzo! 

I noticed that on medium breakpoints and below, the `max-w-80%` class is making the table narrower than the rest of the content. 

Screenshot 2024-09-10 at 12 13 11 PM



Can we double-check with design if that’s what is expected?

If the table should match the width of the surrounding content at those breakpoints, updating the class like this:

```suggestion
    <div class="article-table-block table-responsive{% if not self.wide %} tw-mx-auto medium:tw-max-w-[80%]{% endif %}">
```

should do the trick 👍 thanks!

network-api/networkapi/wagtailpages/pagemodels/customblocks/table_block.py
```diff
@@ -0,0 +1,37 @@
...
+class WideTableBlock(blocks.StructBlock):
+    """
+    A custom block that contains a table and an option to render it wider than the page body content.
+    """
+
+    table = TableBlock()
+
+    wide = blocks.BooleanBlock(
+        required=False,
+        default=True,
```
danielfmiranda, line 14:
```suggestion
        default=False,
```

Hi @robdivincenzo!

After taking a look at the [related issue](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10163) for this PR, it appears that design has requested that the "wide" option be set to `False` as the default. 

Can we please update the default here and also in the custom migration op?  

Thanks!

network-api/networkapi/wagtailpages/migrations/0160_alter_articlepage_body.py
```diff
@@ -0,0 +1,1285 @@
...
+from networkapi.utility.migration.operations import AlterStreamChildBlockDataOperation
+
+
+def migrate_tableblock(source_block):
+    """Copy old table data to new table a TableBlock."""
+    return {
+        **source_block,
+        "value": {
+            "table": source_block["value"],
+            "wide": True,
```
danielfmiranda, line 24:
```suggestion
            "wide": False,
```

Hi @robdivincenzo!

After taking a look at the [related issue](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10163) for this PR, it appears that design has requested that the "wide" option be set to `False` as the default. 

Can we please update the default here and also in the model definition?

Thanks!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
```diff
@@ -64,4 +54,4 @@
             </tbody>
         </table>
     </div>
-{% endblock %}
+{% endblock block_content %}
```
danielfmiranda, line 57: **Kudos** for updating this to comply with Djlint rules!

Since you've renamed and updated it, we can go ahead and remove [this line](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/TP1-144-10332-add-wide-table-block/pyproject.toml#L126) from `pyproject.toml`.

For a bit of background, when we first introduced Djlint to the repo, we encountered a lot of errors from older files that didn’t adhere to the linting rules. Instead of fixing everything in one go,  we created this list of files and errors to ignore temporarily.

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
```diff
@@ -2,41 +2,31 @@
...
 {% block block_content %}
-    <div class="article-table-block table-responsive">
+    <div class="article-table-block table-responsive{% if not self.wide %} tw-mx-auto tw-max-w-[80%]{% endif %}">
         {% comment %}
         Modified table template comes from "table_block/blocks/table.html"
         {% endcomment %}
         <table class="table">
-            {% if table_caption %}
-                <caption>{{ table_caption }}</caption>
+            {% if self.table.table_caption %}
```
danielfmiranda, line 10: **Minor Suggestion**

I’ve noticed that with the new model definition, we’re referencing `self.table` multiple times throughout the template. While this works perfectly fine, I wanted to mention that we could simplify it by wrapping the template in a `with` statement, like so:

```jinja

{% block block_content %}
       {% with self.table as table %}
...
       {% endwith %}
{% endblock block_content %}

```

That said, I’m totally fine with either approach, so feel free to go with whichever option you think works best!

danielfmiranda (changes requested):
Hi @robdivincenzo!!

Thanks for taking this on. The PR is looking great, and I can confirm that everything is working as expected! 👍

I’ve left a question I have regarding the table width on medium breakpoints and below, as well as some other minor suggestions. 

Let me know what you think, thanks again!

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
robdivincenzo: Great catch @danielfmiranda ! I've updated the template with the new targeted viewport. Needed to narrow on`large` to cover it.

network-api/networkapi/wagtailpages/pagemodels/customblocks/table_block.py
robdivincenzo: Thanks @danielfmiranda! So what I noticed was that in [the later comments](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10163#issuecomment-1440244910), the dev/design discussion turned to setting it to `True` by default instead. And I think I agree with that `True` is less disruptive, since the current layout is wide. Do you have any more context on if that changed back after these comments?

network-api/networkapi/wagtailpages/migrations/0160_alter_articlepage_body.py
robdivincenzo: Left a comment on the other thread 🙂

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
robdivincenzo: TYVM, great idea 🔥

network-api/networkapi/wagtailpages/templates/wagtailpages/blocks/wide_table_block.html
robdivincenzo: Definitely did not know about this. Thanks for pointing it out @danielfmiranda!

network-api/networkapi/wagtailpages/pagemodels/customblocks/table_block.py
danielfmiranda: Thanks @robdivincenzo for linking the comments for additional comments. In that case, leaving it set to `True` makes sense to me! 👍

danielfmiranda (approved):
Thanks @robdivincenzo! 
LGTM 🚀

Issue #12598, opened 2024-07-09
[Homepage Refresh] Update hero intro section


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-930)

PR #12858, opened 2024-09-11
Homepage refresh new intro section

ramram-mf:
# Description

This PR updates the homepage model and view, adding a new section below the hero banner, with editable heading, contents, cta and page target. Repositions the `Focus areas` section into it and updates all styling to tailwind utility classes.

image

*Mobile layout ATF*

image

*Mobile layout BTF*

image

*Desktop layout*

image

*Wagtail page form UI*

# To test

1. Login to the [review app CMS](https://foundation-s-tp1-930-ho-hg3if5.herokuapp.com/cms/) with credentials `admin2:admin2`
2. [Update the homepage](https://foundation-s-tp1-930-ho-hg3if5.herokuapp.com/cms/pages/46/edit/) new section `hero bottom` fields, publish.
3. Review the [homepage](https://foundation-s-tp1-930-ho-hg3if5.herokuapp.com/en/) and confirm changes were reflected correctly.

Link to sample test page: https://foundation-s-tp1-930-ho-hg3if5.herokuapp.com/
Related PRs/issues: #12598

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1209)


network-api/networkapi/wagtailpages/migrations/0162_remove_focusarea_page_and_more.py
```diff
@@ -0,0 +1,107 @@
+# Generated by Django 4.2.15 on 2024-09-11 06:29
```
mmmavis, line 1: I’m gonna tag Daniel and Rob to take a look at that file just to double check since they are more experienced in those kind of changes. But from what I understand, I think we can just rely on autogenerated migration files with one dealing with field deletion and one with field additions. We don’t need to set content in migration files either.

@danielfmiranda or @robdivincenzo , could I get your input on this? Thanks!

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/cause_statement.html
```diff
@@ -1,10 +0,0 @@
-<div class="row mb-4">
-    <div class="col">
-        <div class="section-cause-statement tw-dark p-5 text-center">
```
mmmavis, line 3: Let's delete `.section-cause-statement` related SCSS code as well.

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/cause_statement.html
```diff
@@ -1,10 +0,0 @@
...
-            <p class="tw-h4-heading">{{ page.cause_statement }}</p>
-            {% if page.cause_statement_link_text and page.cause_statement_link_page %}
-                <a href="{{ page.cause_statement_link_page.localized.url }}" class="tw-cta-link" id="homepage-cause-statement-cta">{{ page.cause_statement_link_text }}</a>
```
mmmavis, line 6: Let's remove `#homepage-cause-statement-cta` related JS code.

mmmavis (changes requested):
Thanks @ramram-mf . I've left some comments regarding removing old code. I will review the front end code in depth in the my round of review.

network-api/networkapi/wagtailpages/migrations/0162_remove_focusarea_page_and_more.py
robdivincenzo: Hey @mmmavis, I like the intention for it to launch with all the new content, but I agree that you've raised good points. I also like your suggested approach using default values, which would work great for the new hero fields.

Though, I think we'd need a second approach for the `FocusAreas` since they are separate snippets. My idea for it would involve two separate deployments (is that overengineering?). Maybe @danielfmiranda can chime in with a simpler idea or with more context on the tolerance for coordinating content updates like this after a deployment?

network-api/networkapi/wagtailpages/templates/wagtailpages/fragments/cause_statement.html
ramram-mf: Addressing design teams feedback, I did not include the removal of the `cause statement` section in this PR.

nancyt1 (approved):

network-api/networkapi/wagtailpages/migrations/0163_remove_focusarea_page_homepage_hero_intro_body_and_more.py
```diff
@@ -0,0 +1,45 @@
+# Generated by Django 4.2.15 on 2024-09-19 20:09
+
+import django.db.models.deletion
+from django.db import migrations, models
+
+
+class Migration(migrations.Migration):
```
danielfmiranda, line 7: To respond to @mmmavis and @robdivincenzo 's [comments](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12858#discussion_r1757640739):

After looking through the changes made, I do not believe that separate deployments are needed, and we are good to go with the auto-generated migration files in a single PR.


However, I can see what they mean when they mention separating the changes to the focus area and homepage models in separate files.


The easiest way to do so would be:

1. Run `inv new-db`
2. Remove the `page` field from the `FocusArea` model
3. Run `inv makemigrations` and rename the generated migration file to something like "0xxx_remove_page_from_focusarea.py"
5. Add the new fields to the `HomePage` model
6. Run `inv makemigrations` again and rename the file to something like "0xxx_update_homepage_hero_fields.py"

Regarding the need to populate the new fields, I believe the safest approach is to add a post-migration operation. This would ensure that the template remains intact without waiting for someone to manually populate the fields after deploying.


To do so, you can update the migration file generated in step 6 to something like:


```python
from django.db import migrations
...

def populate_hero_intro_fields(apps, schema_editor):
    Homepage = apps.get_model("wagtailpages", "Homepage")

    link_value = {
        "link_to": "external_url",
        "external_url": "<link_button_url>", <-- The URL sent to us from Design
        "label": "<link_button_label>", <-- The Label sent to us from Design
        "new_window": True,
    }

    for homepage in Homepage.objects.all():
        homepage.hero_intro_heading = "<intro_heading_text>" <-- The heading text sent to us from Design
        homepage.hero_intro_body = "<intro_body_text>" <-- The body text sent to us from Design

        # Create the StreamField value for hero_intro_link
        homepage.hero_intro_link = [("link", link_value)]

        # Save changes
        homepage.save()


class Migration(migrations.Migration):

    dependencies = [
...
    ]

    operations = [
... 
        migrations.RunPython(populate_hero_intro_fields),
    ]
```


Once this is deployed, the only remaining task will be to submit the homepage for translation. In the meantime, the translated pages will default to English until the translations are ready. 👍

danielfmiranda (changes requested):
Hi @ramram-mf,

Thank you for taking on this task! I followed the steps in your description, and everything is functioning as expected. Great work! 🙌

I do have one key suggestion regarding the fields `hero_intro_link_text` and `hero_intro_link_page`. Given the recent push to use it wherever possible, we can simplify these fields by combining them into a single field using the [LinkBlock](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/network-api/networkapi/wagtailpages/pagemodels/customblocks/link_block.py#L28), as it handles both link label and link target along with any necessary validation. It also allows users to set the link target to not only internal pages, but also external links, phone numbers, emails, etc: 

Screenshot 2024-09-19 at 11-49-22 Editing Homepage Homepage - Wagtail



I’ve left suggestions in the review to show how you could implement this change in your PR. If you need any help, please feel free to reach out and I would be happy to help via slack or zoom call! 👍

Additionally, you can check out [this PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12629) for reference, where we’ve implemented the LinkBlock in a similar context.

Thanks again, and great job on the progress so far!

network-api/networkapi/wagtailpages/migrations/0162_remove_focusarea_page_and_more.py
danielfmiranda: Hi @mmmavis and @robdivincenzo! My response can be found in the review I left [here](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/12858#discussion_r1767653217) 🙌

network-api/networkapi/wagtailpages/factory/homepage.py
```diff
@@ -102,6 +105,10 @@ def generate(seed):
         print('Generating "who we are" Page (PrimaryPage)')
         wwa_page = PrimaryPageFactory.create(parent=home_page, title="Who we are", show_in_menus=True)
 
+    # Set the hero_intro_link_page
+    home_page.hero_intro_link_page = wwa_page
```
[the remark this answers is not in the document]
ramram-mf: I have addressed those updates and I think third time is the charm. Please let me know if there is anything else I should include in this PR.

PR #12962, opened 2024-10-03
Map ai_is_untrustworthy "yes" and "no" values to "CD"

danielfmiranda:
# Description

This PR updates the `ai_is_untrustworthy` field on the `GeneralProductPage`, mapping any values of "Yes" or "No" to "Can't Determine" (CD) through a custom migration operation. 

Additionally, new options ("Good," "Average," "Needs Improvement," and "Bad") have been added to the `ExtendedYesNoField`, and the "Yes"/"No" options have been removed.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1336)

### DIFF
--- a/network-api/networkapi/wagtailpages/fields.py
+++ b/network-api/networkapi/wagtailpages/fields.py
@@ -38,6 +38,10 @@ class ExtendedYesNoField(models.CharField):
     description = "Yes, No, Not Applicable, or Can’t Determine"
 
     choice_list = [
+        ("Good", "Good"),
+        ("AV", "Average"),
+        ("NI", "Needs Improvement"),
+        ("Bad", "Bad"),
         ("Yes", "Yes"),
         ("No", "No"),
         ("NA", "Not Applicable"),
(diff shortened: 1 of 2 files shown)


network-api/networkapi/wagtailpages/fields.py
```diff
@@ -49,7 +51,7 @@ class ExtendedYesNoField(models.CharField):
     def __init__(self, *args, **kwargs):
         kwargs["choices"] = self.choice_list
         kwargs["default"] = self.default_choice
-        kwargs["max_length"] = 3
+        kwargs["max_length"] = 4
```
danielfmiranda, line 54: Updated so we can use "Good" as an option without abbreviating it. 

I cant quite find why a max length was implemented in the first place 🤔 it does not seem like its being used anywhere, besides making sure that the first item in the `("x", "x")`, is under 3/4 characters.

danielfmiranda:
closing this PR in favor of #12976

Issue #8214, opened 2022-02-09
All external links should open in a new tab


danielfmiranda:
We have added "target='_blank'" to any external links within a rich text block in #8195

However, since external links in other blocks such as link buttons and video/image captions still open themselves in the same tab vs a new one like the rich-text external links, it is providing an inconsistent user experience.

The ask of this ticket is to make sure that **any** external link on the site opens in a new tab.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-260)

simont-cr:
@danielfmiranda Is this still something that needs to be corrected in the platform and if so, would you say we could just keep it with a P3 status?

danielfmiranda:
Hi @simont-cr I would say yes and that this can keep the P3 status. I have updated the labels to reflect this 👍

PR #12980, opened 2024-10-09
Update external links attributes to open in a new tab

ramram-mf:
# Description

This PR updates all external links attributes so that these links open in a new tab.

## To test

1. Open the [review app homepage](https://foundation-s-tp1-260-al-safchz.herokuapp.com/).
2. Click on the hero link cta button, it should open in a new tab.

image

This behavior will be consistent in preview/staging apps, but not in production, since all preview/staging apps hero links point to production urls which are served from a different domain.

Link to sample test page: https://foundation-s-tp1-260-al-safchz.herokuapp.com/
Related PRs/issues: #8214

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1354)


mmmavis (commented):
Hi @ramram-mf , do you think we could address this directly at the template level? While using JS does work, it feels more like a temporary solution, and it might only apply to links available during the initial page load. Let me know your thoughts!

ramram-mf:
> Hi @ramram-mf , do you think we could address this directly at the template level? While using JS does work, it feels more like a temporary solution, and it might only apply to links available during the initial page load. Let me know your thoughts!

Given the decision of renew all content, I feel in this case we can proceed with this approach since the streamblocks wont be inherited to the new project. Is that ok with you @mmmavis?

mmmavis (approved):
Good point and let's make sure we implement this differently for the new site. Thanks @ramram-mf

Issue #12881, opened 2024-09-16
Investigate background/font color customization for EOY banner


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-1234)

PR #12986, opened 2024-10-09
Add `bgcolor` and `textcolor` fields to `DonateBanner` model

ramram-mf:
# Description

This PR updates the `DonateBanner` model with the addition of two new fields `background_color` and `text_color`, allowing the user to customize the donate banner present on all site pages. It also renames the previously incorrectly named `background_image` to `foreground_image` and adds a `background_image` field to store the newly implemented background image feature for the banner.

# To test

1. Open the review app [Wagtail auth page](https://foundation-s-tp1-1234-i-bhpnku.herokuapp.com/cms/), authenticate with `admin2:admin2` credentials.
2. Navigate to the `Settings > Donate banner` [page](https://foundation-s-tp1-1234-i-bhpnku.herokuapp.com/cms/settings/donate_banner/sitedonatebanner/1/), click on the `Choose` in order to pick a banner to show.
image

3. Select a banner from the list by clicking on its name. Save by clicking on the `Save` button on the bottom of the page.
image
image

4. Edit the selected banner by clicking on the three dots next to its name and then selecting the `Edit` menu:
image

5. Note the new model fields `Background image`, `Background color` and `Text color`. Customize the background color of the banner. Save and browse the [preview app homepage](https://foundation-s-tp1-1234-i-hlbzy3.herokuapp.com/en/).
image

6. Confirm that the background color has changed in the [homepage](https://foundation-s-tp1-1234-i-bhpnku.herokuapp.com/en/). 
image

7. Edit again the donate banner and upload a background image. Save and confirm changes reflect on the [homepage](https://foundation-s-tp1-1234-i-bhpnku.herokuapp.com/en/).
image

Link to sample test page: https://foundation-s-tp1-1234-i-bhpnku.herokuapp.com/en/
Related PRs/issues: #12881

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1360)


network-api/networkapi/donate_banner/models.py
```diff
@@ -51,11 +51,42 @@ class DonateBanner(TranslatableMixin, PreviewableMixin, models.Model):
...
+    bg_image = models.ForeignKey(
+        "wagtailimages.Image",
+        models.PROTECT,
+        related_name="+",
+        null=True,
+        blank=True,
+    )
+
+    TAILWIND_COLORS = [
+        ("bg-gray-60", "Gray"),
```
mmmavis, line 68: Would it be possible to include `tw-` in these class names? It would make the classes easier to search for in the future!

network-api/networkapi/donate_banner/models.py
```diff
@@ -51,11 +51,42 @@ class DonateBanner(TranslatableMixin, PreviewableMixin, models.Model):
...
+    )
+
+    TEXT_COLORS = [
+        ("text-white", "White"),
+        ("text-black", "Black"),
+    ]
+
+    text_color = models.CharField(
+        max_length=20, choices=TEXT_COLORS, default="text-white", help_text="Text color for the banner"
+    )
```
mmmavis, line 89: Would it be possible to include `tw-` in these class names? It would make the classes easier to search for in the future!

network-api/networkapi/donate_banner/models.py
```diff
@@ -64,15 +95,19 @@ class DonateBanner(TranslatableMixin, PreviewableMixin, models.Model):
         FieldPanel("subtitle"),
         FieldPanel("cta_button_text"),
         FieldPanel("cta_link"),
-        FieldPanel("background_image"),
+        FieldPanel("foreground_image"),
+        FieldPanel("bg_image"),
+        FieldPanel("bg_color"),
```
mmmavis, line 100:
```suggestion
        FieldPanel("foreground_image"),
        FieldPanel("background_image"),
        FieldPanel("background_color"),
```
to make naming convention consistent

network-api/networkapi/donate_banner/models.py
```diff
@@ -64,15 +95,19 @@ class DonateBanner(TranslatableMixin, PreviewableMixin, models.Model):
...
     ]
 
     translatable_fields = [
         TranslatableField("title"),
         TranslatableField("subtitle"),
         TranslatableField("cta_button_text"),
         SynchronizedField("cta_link"),
-        SynchronizedField("background_image"),
+        SynchronizedField("foreground_image"),
+        SynchronizedField("bg_image"),
```
mmmavis, line 110:
```suggestion
        SynchronizedField("background_image"),
```

mmmavis (changes requested):
Hey @ramram-mf , thanks for adding these customization options for the donate banners. I have left a few comments that I think would help with UX and code maintainability.

network-api/networkapi/donate_banner/models.py
ramram-mf: I opted for using the `bg_`prefix instead of `background_` since when running the migration that renamed and added again the `background_image` field, database issued an error about the relationship already existing, since that field existed previously. I will change that and let you know if it still complains or if it was something related to db state in that moment.

network-api/networkapi/donate_banner/models.py
ramram-mf: Sure. I went without the prefix in this approach contemplating the bootstrap removal would eventually make it go away, and would be portable. But I think you're right, in order to track correctly instances of tailwind classes at the moment we should be consistent in the way we express those in the codebase. Will update.

beccaklam:
Hi @ramram-mf! Thank you for your work on this, it's looking good 👍 Just adding the notes we discussed:

1. I think we should limit the colours to the following for now: black, white, Red 40 #FF4F5E and Blue 40 #595CF3
2. The background colour option should be disabled when a user has chosen to upload a background image so that it's clear the background image will load instead of the background colour. If the user clears the background image choice, the background colour will be enabled again.

beccaklam (approved):
Thank you for making those changes @ramram-mf! This looks good to me.

danielfmiranda (changes requested):
Hi @ramram-mf!

Thanks very much for taking this work on. After testing this out on the review app, I can confirm that everything is working as expected. I appreciate you fixing the indexing issue when it came to repurposing the `background_image`  field name, and transferring over existing data. Great work! 

I liked your approach of disabling the `background_color` field when a background image is selected. 

That said, I have a slight concern that users might find it a bit confusing without immediate context, as there isn’t any help text to clarify why `background_color` is disabled when both fields are set, until they hit “Save.”

To simplify things and add clarity for users, I’d suggest leveraging Wagtail’s `MultiFieldPanel` and `HelpPanel`. This would allow us to provide clear instructions right within the CMS interface. Here’s an example:

```python
MultiFieldPanel(
    [
        HelpPanel(content="Select either an image or a color to serve as the background for the banner."),
        FieldPanel("background_image"),
        FieldPanel("background_color"),
    ],
    heading="Background",
)
```
With this approach, the CMS will display clear instructions before users make their selection: 

Screenshot 2024-10-31 at 00-03-49 Editing donate banner - B2 - Wagtail

We can then handle validation in the models `clean()` method.

Here's what that might look like:

```python
    def clean(self):
        super().clean()

        both_selected_error = "Please select either a background image or a background color for the banner."
        none_selected_error = "Please select a background image or a background color for the banner."

        # Validate that either background_image or background_color is set, not both.
        if self.background_image and self.background_color:
            raise ValidationError(
                {
                    "background_image": ValidationError(both_selected_error),
                    "background_color": ValidationError(both_selected_error),
                }
            )
        if not self.background_image and not self.background_color:
            raise ValidationError(
                {
                    "background_image": ValidationError(none_selected_error),
                    "background_color": ValidationError(none_selected_error),
                }
            )

```

This way, if both or none of the fields are set, they’ll see a styled validation error reminding them to choose just one:

Screenshot 2024-10-31 at 00-08-41 Editing donate banner - B2 - Wagtail

This update could help streamline the user experience and keep the codebase leaner by removing the two new files needed for the custom disabling logic. Let me know what you think!

Thanks again for your hard work!

danielfmiranda (approved):
Thanks @ramram-mf! Great work 🙌 LGTM

mmmavis (approved):
Nice work @ramram-mf ! Also thank you @danielfmiranda for your great suggestion.

PR #12990, opened 2024-10-10
Moved localized logic from take_action.html template to base.py with …

dlopezvsr:
# Description
This PR removes `.localized` from `take_action.html` and moved the logic to `base.py`, implementing `get_localized_take_action_cards` function, that iterates over `take_action_cards` and add every card to a list, as well as the now localized `internal_link`. This PR also updates the `take_action.html` template, adding the new context variable `localized_take_action_cards`.

Link to sample test page: https://foundation-s-tp1-103-re-yv4aeu.herokuapp.com/en/
Related PRs/issues: https://github.com/MozillaFoundation/foundation.mozilla.org/issues/10010

# How to test
- Scroll down Homepage until Take action section.
- Change the code locale in the URL and make sure the external link properly keeps the updated locale.

Example:
image

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1364)


mmmavis (approved):
Woohooo 🎉

danielfmiranda (approved):
Thanks @dlopezvsr for taking on this work 🙌 

The implementation is clear and straightforward, I've also tested it out on my local machine and can confirm that everything is working as expected 👍 

I do have a small suggestion for another approach you might consider. Instead of creating a dictionary with `{"card": card, "internal_link": localized_link}`, you could modify the card object in-memory by updating the `internal_link` field directly with it's localized version, like so:

```python
localized_cards = []
for card in self.take_action_cards.all():
    card.internal_link = card.internal_link.localized
    localized_cards.append(card)
return localized_cards
```
This way, you avoid using a dictionary and can streamline the template by working directly with the card object. 

Here's how the template would be updated:

**Current Template:**
```jinja2
{% for item in localized_take_action_cards %}
     {% with target=item.internal_link %}
           <a href="{{ target.url }}">
               {{ item.card.text }}
           </a>
    {% endwidth %}
{% endfor %}
```
**Updated Template (with in-memory modification):**
```jinja2
{% for card in localized_take_action_cards %}
     {% with target=card.internal_link.url %}
           <a href="{{ target }}">
               {{ card.text }}
           </a>
    {% endwidth %}
{% endfor %}
```

This approach eliminates the need for `item.card`, so you can directly access the card's fields, making the template a bit more readable.

That said, your current approach works perfectly fine, and this is more a matter of preference rather than any performance gain. If you prefer to keep it as-is, feel free to proceed with the merge. Just wanted to share an alternative for your consideration!

Thanks again for the great work!

dlopezvsr:
@danielfmiranda thank you for your suggestion. You are totally right, I also think is best, simpler, and cleaner with the in-memory update. I just commited the changes. Great observation 🙌

Issue #13082, opened 2024-11-07
EOY Banner Button and Background Colour Update


data-sync-user:
Now that the designs have been finalized for the EOY banner for [foundation.mozilla.org](http://foundation.mozilla.org) we need to make some final updates in code. Specifically:

* Update the button colour, size and font-size according to this design: [https://www.figma.com/design/2TK8tXeDIiLC67TMdbMIcO/Fundraising?node-id=2917-6474&m=dev](https://www.figma.com/design/2TK8tXeDIiLC67TMdbMIcO/Fundraising?node-id=2917-6474&m=dev|smart-link) _(Note: can we make it so that the colour of the button is selected through a colour picker like the background colour of the banner? The two colours would be the current styling and this new green colour that we are adding – only if this is not a lot of work to do. If it is we can just change the button colour to green)_
* Add the background colour: #acffe0 to the banner

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1481)

PR #13084, opened 2024-11-08
Update `donate_banner` model adding `button_color` field and updated button styles

ramram-mf:
# Description

This PR implements the current design spec updating the `donate_banner` model to include a new field named `button_color` which will store the main button class name, allowing to select an alternative green color. Also updates styles on the CTA button on mobile and desktop viewports, and fixes the close button color mismatch with the text color.

It includes a new tailwind palette color under the green category `tw-green-110` to allocate the specific shade of green requested, also includes both the CTA button hover class names in the `safelist` section of the tailwind config, since the dynamic nature of the model field value doesn't allow tailwind to detect its use and include them on the final css build as actual utility classes.

# To test

1. Open the preview app CMS auth page and login with `admin2:admin2` credentials.
2. Open the `Settings > Donate banner` page and edit the currently active banner by clicking on the three dots next to its name and selecting `Edit` from the context menu.

image

3. Update the banner new button color field and note the new `Green` background color option. Save.

image

4. Browse the preview app homepage and review the style updates.

image

image

image

image

image

Link to sample test page: https://foundation-s-tp1-1481-e-vll68r.herokuapp.com/en/
Related PRs/issues: #13082


beccaklam (changes requested):
Hi @ramram-mf ! This is looking great, thank you for making those quick changes. The changes you implemented look good, I just noticed some other small details and I was wondering if we could update:

- [x] Can we change the spacing below the body text to be 1.25rem instead of 0.75rem? It feels a little unbalanced right now even though in the design I know it was supposed to be 12px
- [x] The font-weight of the body text is currently **300** and it's actually supposed to be SemiBold. Playing around with it live though I think if we change the font-weight to **400** that will look more balanced
- [x] Can we also update the font-size of the body text at the ~>1440px breakpoint to be 20px instead of 18px?
- [ ] Lastly, since the pop-button has a shadow it looks a little unbalanced right now because the margins on the left and right of the button look uneven. Is there anything we can do to balance it out? In the figma file I played around with reducing the width of the button slightly but keeping the objects left aligned so optically it looks more even

Thank you!

mmmavis (approved):
Looks good to me. Thank you! @ramram-mf

PR #13122, opened 2024-11-15
updated query to grab featured blog pages in the correct order

danielfmiranda:
# Description

This PR refines the method for fetching the Blog Index Page’s "Featured blog posts" to ensure they display in the correct order within the site's navigation.

Previously, the method queried for featured_pages_relationships that were not null, which I believe would work initially, given that only five such relationships would exist in the database. However, this approach introduces the risk of ambiguity in the navigation bar if the featured pages are rearranged or if additional relationships are accidentally left in the database.

With this update, only relationships specifically linked to the Blog Index Page are retrieved, ensuring the correct set of pages is always displayed in the intended order.

### DIFF
--- a/network-api/networkapi/nav/models.py
+++ b/network-api/networkapi/nav/models.py
@@ -190,7 +190,7 @@ class NavMenu(
     def localized_featured_blog_posts(self):
         default_locale = settings.LANGUAGE_CODE
         posts = BlogPage.objects.filter(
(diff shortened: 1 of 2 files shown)


robdivincenzo:
Thanks @danielfmiranda! Confirmed I'm seeing the proper order on a fresh DB. LGTM 👍

robdivincenzo (approved):

PR #13133, opened 2024-11-18
Returning None if no donate_banner is set

danielfmiranda:
# Description

This PR updates the `get_donate_banner` method to return `None` when no donate banner is set on the DonateBannerPage. This resolves an issue where if a DonateBannerPage exists in the DB but no donate_banner is set, an error would be returned when the method attempted to run `.localize` on `None`

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1533)


robdivincenzo (approved):
LGTM, thanks @danielfmiranda

PR #13157, opened 2024-11-19
Add middleware to redirect pt-br to correct locale path

robdivincenzo:
# Description

This PR will add a middleware function to essentially make the pt-BR locale case-insensitive. If the request's locale path is `pt-br` it will redirect users to the correct path `pt-BR` instead of throwing error. It's the most frequent error on our site, and is worthy of accommodation.

To test:
1) Check out branch
2) Navigate to any path with all lower case path '/pt-br/'
3) Confirm you are redirected to `/pt-BR/` locale path instead of getting an error

Related PRs/issues: TP1-1479

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1558)


network-api/networkapi/utility/middleware.py
```diff
@@ -52,3 +54,19 @@ def __call__(self, request):
...
+    def process_request(self, request):
+        # Split the path into segments
+        path_segments = request.path_info.strip("/").split("/")
+
+        # Check if the first segment is 'pt-br'
+        if len(path_segments) > 0 and path_segments[0] == "pt-br":
+            # Change only the country code to uppercase and reconstruct the path
+            path_segments[0] = "pt-BR"
+            new_path = "/" + "/".join(path_segments)
+            return redirect(new_path)
```
danielfmiranda, line 70:
```suggestion
            # Preserve query parameters
            if request.META.get("QUERY_STRING"):
                new_path += f"?{request.META['QUERY_STRING']}"

            return redirect(new_path)
```


As mentioned in my review, I noticed that query parameters (e.g., `?form=donate`) are being omitted during redirects. 

This change should do the trick. 👍

danielfmiranda (changes requested):
Hi @robdivincenzo,

Thank you for taking on this work! It's great to see that this addressing one of the most common errors we're seeing in Sentry. After testing locally, I can confirm that the "update `pt-br` to `pt-BR`" functionality is working well.

I did notice a small caveat: middleware redirects with query parameters (e.g., `?form=donate`) seem to omit the parameters after redirection. I've left a suggestion in the review that should help resolve this issue.


Additionally, it might be helpful to include a test for this middleware. I’ve drafted an example that could work:

```python
class NormalizeLocaleMiddlewareTest(TestCase):
    def test_redirect_with_pt_br_language(self):
        # Simulate a request to a URL with 'pt-br' in the path
        response = self.client.get("/pt-br/destination/subpath?param=value")
    
        # Assert that the middleware issues a redirect to the normalized URL
        self.assertEqual(response.status_code, 302)
        self.assertEqual(response.url, "/pt-BR/destination/subpath?param=value")

```

Appending this to `utility/tests/test_middleware.py` should cover the edge case nicely. Let me know what you think!


**Note**: I also noticed that another locale, `fy-NL`, might face a similar issue. Since `pt-BR` is the primary locale where we’ve observed this error, I’ve created [TP1-1559](https://mozilla-hub.atlassian.net/browse/TP1-1559) to address `fy-NL` separately, so we can get this fix in ASAP.

network-api/networkapi/utility/middleware.py
robdivincenzo: Thanks for the great feedback @danielfmiranda! It seemed quick enough to generalize for both languages so I included that in the PR. Ready again for review!

dlopezvsr (approved):
Thanks for this PR @robdivincenzo! I ran tests and all passed. I also can confirmmed the url path  is properly redirected to the correct locale, also keeping parameters in case it has them. All the code also looks good. Approved 🚀

PR #13180, opened 2024-11-26
Add fallback checks if link_to is not set in certain locales

robdivincenzo:
# Description

This will address Link Block errors happening on some translated pages. The `link_to` field does not seem to be set on certain locales, so this adds a fallback option to try to find the value. While this should be efficient, it'd be best if the link_to values were set already. That warrants further investigation.

Related PRs/issues: TP1-1478

# To Test
1) Checkout `main` branch with a copy of production db
2) Navigate to: http://localhost:8000/fr/campaigns/codemoji/
3) Notice the linkblock error on the `fr` locale page that isn't present on the `en` locale
4) Checkout this branch and notice linkblock value is found via fallback (and the button links properly)

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-1583)


dlopezvsr (approved):
LGTM 🚀 thanks @robdivincenzo!

Issue #13299, opened 2024-12-11
Dropdown Nav Bug


data-sync-user:
Per Genevieve:
The dropdown isn't sticky (or maybe it's too sticky?) when the donate banner is open
(I navigated to the blog nav item with the banner open, scrolled down and the dropdown went with the screen).

┆Issue is synchronized with this [Jira Bug](https://mozilla-hub.atlassian.net/browse/TP1-1703)
┆Attachments: <a href="https://mozilla-hub.atlassian.net/rest/api/2/attachment/content/74406">Screen Recording 2024-12-09 at 1.10.27 PM.mov</a>

PR #13310, opened 2024-12-12
Close menu dropdown on scroll offset update

ramram-mf:
# Description

This PR updates the behavior of the desktop main navigation dropdown so that it closes when scrolled and the top offset changes. This new behavior wont apply when its in the top fixed position, including cases when the donate banner is closed or when it has scrolled past it.

Link to sample test page:
Related PRs/issues: #13299

### DIFF
--- a/source/js/components/nav/desktop-dropdown.js
+++ b/source/js/components/nav/desktop-dropdown.js
@@ -79,6 +79,8 @@ class NavDesktopDropdown extends Accordion {
         sibling.classList.remove("tw-highlighted");
       }
     });
+
+    document.addEventListener("scroll", this.handleScroll);
   }
 
   closeSiblings() {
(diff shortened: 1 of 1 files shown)


source/js/components/nav/desktop-dropdown.js
```diff
@@ -115,7 +117,23 @@ class NavDesktopDropdown extends Accordion {
 
     this.closeAccordion(this.accordion, this.titleButton, this.content);
     this.closeSiblings();
+
+    document.removeEventListener("scroll", this.handleScroll);
+  }
+
+  scrollListener() {
+    const newTopOffset = document
```
mmmavis, line 125: let's be extra careful to prevent runtime error

```js
const newFocusInNav =
  event.relatedTarget &&
  document.querySelector(".wide-screen-menu")?.contains(event.relatedTarget);
```

mmmavis (approved):
Hey @ramram-mf , Thanks for the fix! It works great! I left one suggestion for improvement. If it looks good to you, please apply the code change and merge the PR. Thank you!

source/js/components/nav/desktop-dropdown.js
ramram-mf: Since we're listening to the document's scroll event, I only needed to confirm the element is in the DOM.

PR #13484, opened 2025-02-07
Refactor footer language switcher from `<select>` to seo-friendly `<a>`

ramram-mf:
# Description

This PR replaces the footer language switcher from an auto-submitting form with `<select>` and `<option>` tags, in to a more seo-friendly list of links `<a>` tags.

## Current dropdown:
image

## Updated dropdown:
image

Link to sample test page:
Related PRs/issues: #

### DIFF
--- a/network-api/networkapi/templates/fragments/language_switcher.html
+++ b/network-api/networkapi/templates/fragments/language_switcher.html
@@ -3,27 +3,59 @@
 {% get_current_language as current_language %}
 {% get_local_language_names as languages %}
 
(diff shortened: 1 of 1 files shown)


mmmavis (changes requested):
Thanks for updating the language switcher to enhance our SEO, @ramram-mf ! I noticed that the switcher is not fully keyboard accessible. Could you take a look at the following issues?

1. The switcher cannot be focused (tapped into) using the keyboard.
2. When the list is open, users are unable to navigate through the options using the keyboard.
3. The list cannot be closed using the keyboard.

PR #13520, opened 2025-02-20
Update FRU widget selector to fix a Playwright timeout issue

mmmavis:
Since the FRU widget's title attribute was recently modified, it triggered a timeout issue in CI integration tests. To resolve this, I updated the selector used in our Playwright tests.

image


ramram-mf (approved):
LGTM. Seems that FRU tests will have to be constantly catching up.

PR #13785, opened 2025-04-02
Change Login with Google to Mozilla SSO

robdivincenzo:
# Description

This PR will swap the "Login with Google" via `google-auth2` to Mozilla SSO via `auth0`. Since the Google authentication is tied to `foundation.mozilla.org` and we have no access to the Google project, we won't be able to login with Google in the new `mozillafoundation.org` domain.

A domain swap unblocker, but also a general improvement and step towards best practice, consolidating into Mozilla's SSO w/ 2FA.

Link to sample test page: https://foundation-s-sso-sandbo-98uft4.herokuapp.com/cms/login/

# To Test

Two basic cases covered in these steps

- As an existing user, you should be able to log in with your Mozilla SSO and connect to your Wagtail user / profile as expected
- As a new user, you should be able to get through the Mozilla SSO portion, then get an error saying you don't have permission / access to wagtail. An admin can then grant you an editor or administrator role, and you're in!

### Steps
1) Go to the RA login page via new window (to make sure you're not authenticated already)
2) Click to Login via Mozilla SSO
3) Note you now need to login via Mozilla SSO w/ 2fa and not Google
4) Once logged in you should have every existing access, and profile information you already have (I've cloned staging)
5) Log in as admin/admin
6) Delete your personal profile :(
7) Stay logged in as admin. But now try logging back in via New Window using Mozilla SSO
8) You should receive an error that says you don't have permission to access wagtail
9) In your window where you're logged in as admin, confirm you see your new user / username.
10) Add your role (administrator for most of us)
11) Confirm you've got access
12) One note: new usernames change to the new default which is FirstnameLastname. This is more unique than our pseudo mixed nickname approach, and we could consider standardizing this at a later point.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2376)


danielfmiranda (approved):
Thanks @robdivincenzo! I just followed the testing steps and can confirm that this is working great 👍 LGTM!

PR #13787, opened 2025-04-03
Scaffold Base App

robdivincenzo:
# Description

This PR scaffolds the `base` app with abstract models and moves the former `base` app to `core`. It will also comment out features from unused apps for now (i.e. `blog` which will become articles and `profiles` which is out of MVP scope)

# To Test

1. You should be able to rebuild the site via `inv setup` successfully
2. You should note the homepage is now a subclass from an abstract models which chains up and includes some new things, specifically if you edit it you should now see `tags` and `author` snippets, showing this is working as intended for now.
3. Also note, we do lose some prototype functionality since we've moved things around. But this is just a base scaffold.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2380)


dlopezvsr (approved):
Just finished reviewing changes @robdivincenzo, everything looks good to me. Thanks for having all this ready so quickly to unblock the coming models. Approved. 🚀

PR #13828, opened 2025-04-15
Update domain check for WWW

robdivincenzo:
# Description

This PR updates the cookie banner script to check for www.mozillafoundation.org instead of mozillafoundation.org

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2425)


robdivincenzo:
> LGTM @robdivincenzo! Just one question - shouldn't we check for both? Or the redirect will be from mozillafoundation.org to [www.mozillafoundation.org](http://www.mozillafoundation.org)? Otherwise approved 👍

Thanks @dlopezvsr ! You're right, we'll be redirecting :)

PR #13829, opened 2025-04-15
Domain swap hardcoded urls

robdivincenzo:
# Description

This PR will swap over hardcoded foundation.mozilla.org url to new www.mozillafoundation.org domain url.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2426)


tailwind-plugins/button.js
```diff
@@ -104,7 +104,7 @@ module.exports = [
           "&:focus, &.focus": {
             boxShadow: "0 0 0 0.2rem rgb(130 138 145 / 50%)",
           },
-          '&[href*="//"]:not([href*="foundation.mozilla.org"]):not([href*="donate.mozilla.org"])':
+          '&[href*="//"]:not([href*="www.mozillafoundation.org"]):not([href*="donate.mozilla.org"])':
```
danielfmiranda, line 107:
```suggestion
          '&[href*="//"]:not([href*="mozillafoundation.org"]):not([href*="donate.mozilla.org"])':
```

Hi @robdivincenzo! I noticed that this js file is applying CSS styles to buttons that have links which include `foundation.mozilla.org` (going to be `mozillafoundation.org` soon) somewhere in the HREF. 

Since this is a sort of "wildcard" type of rule that is meant to capture any button linking to `mozillafoundation.org`, would it be best to remove the `www` from here? 

My thinking is that CMS users might not always add the `www` to the HREF, which would cause those links to be ignored.

tailwind-plugins/button.js
```diff
@@ -141,7 +141,7 @@ module.exports = [
                 backgroundColor: "#595cf3",
                 borderColor: "transparent",
               },
-            '&[href*="//"]:not([href*="foundation.mozilla.org"]):not([href*="donate.mozilla.org"])':
+            '&[href*="//"]:not([href*="www.mozillafoundation.org"]):not([href*="donate.mozilla.org"])':
```
danielfmiranda, line 144:
```suggestion
            '&[href*="//"]:not([href*="mozillafoundation.org"]):not([href*="donate.mozilla.org"])':
```

See comment above ^

danielfmiranda (changes requested):
Hi @robdivincenzo thanks for taking this work on! 

Everything looks good to me, however I just wanted an idea by you before approving. I have left it in this reviews comments 👍

robdivincenzo:
> Hi @robdivincenzo thanks for taking this work on!
> 
> Everything looks good to me, however I just wanted an idea by you before approving. I have left it in this reviews comments 👍

Good catch @danielfmiranda, while we hope they would put the www there's no guarentee. Committed your suggestions!

Issue #13216, opened 2024-12-06
Blog app Iteration 3 (related snippets + fields revision based on wireframe)


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-1618)

PR #13948, opened 2025-05-07
Animated Gif Support & Custom Image Model

robdivincenzo:
# Description

This PR creates a new `images` app, introduces a [custom image model](https://docs.wagtail.org/en/stable/advanced_topics/images/custom_image_model.html) for our wagtail instance, migrate all images and their tags to the new custom image model, and introduce functionality behind the scenes to enable animated .gif support via a .webp file conversion.

Using the new `FoundationCustomImage` model, users will be able to upload `.gifs`, which we will then automagickally detect if it's an animated .gif and convert it to `.webp` format (reducing file size by ~50-90%). This file will be stored alongside the image in a new `animated_webp` file field. A custom `get_rendition` method in our `FoundationCustomImage` model will then detect if the image should render normally, or render the animated `.webp` file.

`ffmpeg` is required to convert the `.gif` to `.webp` and while installing `ffmpeg` via our local Dockerfile is quite straightforward, it is not straightforward on heroku since we use buildpacks. We must install a pre-compiled `ffmpeg` package via a custom heroku buildpacks. As a result, I've created two new repos to support this. We're creating our own because there aren't any supported, trusted buildpacks that will work for our purposes.

1. Repo for the static builds themselves - https://github.com/MozillaFoundation/heroku-ffmpeg-static-builds
2. Repo for the buildpack itself -https://github.com/MozillaFoundation/heroku-ffmpeg-buildpack

Both repos have their own Readme with instructions in case we need to install a new version of `ffmpeg` to heroku. We also have a Github action that will check `ffmpeg` for new releases every week (~2x releases / year). In the long term, perhaps we can create a seperate heroku app with Dockerfile deployment to handle just this and convert via API, but this custom buildpack setup will do for our purposes.

We can merge this into `main`, and pull down into `redesign`. There's some coupling consideration that the `legacy_apps` images will use an image model from outside the `legacy_apps`, but seems like there's more risk involved with instead needing to update every template since [wagtail only allows for one image model](https://docs.wagtail.org/en/6.4/advanced_topics/images/custom_image_model.html#migrating-from-the-builtin-image-model), i.e. when you use {% image %} tags. Using the custom image model in `legacy_apps` seems preferable to me because we're not re-writing all the template code in `legacy_apps` and `legacy_apps` should get sunset & drop off over time (so the coupling tech debt will naturally resolve itself).

Review app: https://foundation-s-tp1-1618-a-idd5p3.herokuapp.com/en/

### Simulate a deployment to production:
Tip: keep this list of 20 steps open in a second window!

1. Manually deploy `main` to a review app via the Deploy tab in heroku
2. Copy staging db to review app via `inv staging-db-to-review-app`(make sure heroku CLI is up to date, i.e. `heroku update`)
4. The review app should now mirror staging.
5. Manually deploy `TP1-1618-animated-gif-support` to review app.

### Run a couple of bash commands to help review
6. After release steps, open a bash console via heroku and create an admin via `python manage.py createsuperuser`
7. Also run an update index via `python manage.py update_index` (wait for the images.FoundationCustomImage model to index, then continue to step 8)
9. Log into review app and navigate to Images in the wagtail dashboard.
10.  In a second tab, log into foundation.mofostaging.net and navigate to Images as well

### Compare review app with staging 
11. Compare both Image libraries to review all images are present, in correct order, and searchable.
12. Click into a few images to confirm tags and Collections are exactly the same.
13. Confirm the presence of a new "animated_webp" field in the review app.

### Test out new animated .gif feature
14. Download your favorite animated .gif
15. Go to a blog post in the review app
16. Add the animated .gif into the body via the `/image` block
17. Review the body and notice animated gif
18. Inspect element and confirm this is loading the webp file.
19. Review the Images library and note the animated gif has been uploaded
20. Click into the animated gif and notice the "animated_webp" file has been saved alongside the new animated gif 🎉

## To Deploy
Add the heroku buildpack containing ffmpeg to the staging and production apps (before python):

Example: `heroku buildpacks:add -i 2 https://github.com/heroku/heroku-buildpack-activestorage-preview --app foundation-s-tp1-1618-a-3ynwhv`

Migration is batched and quite quick (~ 1 min), but should schedule and enable maintenance mode / scale dynos before deploying to avoid issues while dynos restart. This also requires a couple custom steps to release, so ideally let's isolate this deployment, and then we can revert out the custom steps.

Might help to run a `python network-api/manage.py update_index` also to update the image index after this is deployed.

## Post-Deploy
Create a PR to remove temporary release steps in `release-steps.sh` and the `migrate_legacy_images.py` command.

Related PRs/issues: TP1-1618 / #13216

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2573)

### DIFF
--- a/app.json
+++ b/app.json
@@ -45,7 +45,7 @@
       "url": "heroku/python"
     },
     {
-      "url": "https://github.com/MozillaFoundation/heroku-ffmpeg-buildpack.git#ffmpeg-v6.1-latest"
+      "url": "https://github.com/MozillaFoundation/heroku-ffmpeg-buildpack.git#ffmpeg-v6.1.2-31"
     }
   ],
   "scripts": {


danielfmiranda (approved):
Amazing!! 

Thank you @robdivincenzo 

https://github.com/user-attachments/assets/bac9aaac-a814-47f7-a875-136ba179026c

PR #13970, opened 2025-05-13
Subpages from new site added to New Home Page  

dlopezvsr:
# Description

This PR introduces the `subpage_types = []` attribute inside the `New Home Page` to avoid legacy pages and block show up inside New Site pages. Additionally as part of the scope of this PR, we were considering an approach for separating New Site Pages and New Site Pages, but after testing multiple options mentioned on this [ticket](https://mozilla-hub.atlassian.net/browse/TP1-2552), and for simplicity, I think the best approach will be to handle pages permissions. So in this way we avoid editors that don't need legacy pages to have see them on the CMS. 

We currently have on the redesign branch the following pages that where added to `subpage_types`:
- GeneralPage
- ArticleAdviceColumn
- ArticleCollectionPage
- ArticleEditorialFeature
- ArticleProductReview

Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2552

**Tests**
- Pull this branch
- Log into the CMS > Go to New Home Page type
- Click + (plus sign) to create a new page > you should only see displayed the pages from new site.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2665)


robdivincenzo:
@dlopezvsr thanks for exploring the `subpage_types[]` for this work! I was just thinking about this some more and I don't know if we can apply this to the home page because we're going to need to move all the old content under this tree. So maybe this approach won't be the catch-all we were hoping for.

I still think this work is valuable though. Do you think we can add this work to the AbstractBaseModel so that all redesign page types have this filtered list by default? And then on the homepage we can override it to allow all via `subpage_types = None`? At least it will get us closer to everything but the homepage for now.

dlopezvsr:
> @dlopezvsr thanks for exploring the `subpage_types[]` for this work! I was just thinking about this some more and I don't know if we can apply this to the home page because we're going to need to move all the old content under this tree. So maybe this approach won't be the catch-all we were hoping for.
> 
> I still think this work is valuable though. Do you think we can add this work to the AbstractBaseModel so that all redesign page types have this filtered list by default? And then on the homepage we can override it to allow all via `subpage_types = None`? At least it will get us closer to everything but the homepage for now.
> 
> I wonder if we absolutely have to, we _might_ be able to move the legacy pages into the homepage tree first, and then put up the subpage_types restriction on the homepage to preventing new ones being created/moved into the tree. But I haven't tried that, so I'm not sure about that possibility or the drawbacks

@robdivincenzo thanks so much for these observations. You are totally right, I did not consider content migration, and this can definitely cause conflicts. I can try moving `ubpage_types[]` to `AbstractBaseModel` and override homepage as you mention. I will think this deeper and try to come up with other potential solutions or scalable workarounds? 

In the meantime I will be prioritizing P1 components, and then get back to this.

danielfmiranda:
Hey @dlopezvsr I am going to remove myself as a reviewer for now but feel free to re-tag me when this is ready! 🙌

PR #13978, opened 2025-05-17
Listing Header Component

dlopezvsr:
# Description

This PR introduces the  listing header component under `foundation_cms/templates/patterns/components/`. Even this ticket was not particularly focused on the listing template, I went ahead and created the file where I initially included the header component. This PR also  includes a simple view  function  under `foundation_cms/core/views.py` that renders the template via `listing_page/` path.

Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2528

# Tests
- Pull this branch and run the application.
- You can create tags on the CMS and they will render on the HomePage
- If you click a tag it will open the listing page with the listing header and the name of it.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2679)


foundation_cms/core/views.py
```diff
@@ -0,0 +1,23 @@
...
+        articles = pages.filter(base_tagged_items__tag__slug=tag)
+    else:
+        articles = pages.all()
+
+    context = {
+        'articles': articles,
+        'active_tag': tag if tag else 'Latest Posts'
+    }
+
+    return render(request, "patterns/pages/core/listing_page.html", context)
```
robdivincenzo, line 23: Updates Abstract model to Page, adds locale check, and updates variable to `pages` for template.

```suggestion
from wagtail.pages import Locale, Page

def listing_page(request):
    # Get the tag from the query parameters
    tag = request.GET.get('tag')
    
    # Get the current locale
    locale = Locale.get_active()

    # Get all pages for the current locale
    pages = Page.objects.live().public().filter(locale=locale)

    # Filter by tag if provided
    if tag:
        pages = pages.filter(base_tagged_items__tag__slug=tag)
    else:
        pages = pages.all()

    context = {
        'pages': pages,
        'active_tag': tag if tag else 'Latest Posts'
    }

    return render(request, "patterns/pages/core/listing_page.html", context)
```

foundation_cms/templates/patterns/pages/core/listing_page.html
```diff
@@ -0,0 +1,4 @@
+{% extends "base.html" %}
+{% block content %}
+    {% include "patterns/components/listing_header.html" with items=pages active_tag=active_tag %}
+{% endblock %}
```
robdivincenzo, line 4: Adds a little snippet to print the pages results.

```suggestion
{% block content %}
    {% include "patterns/components/listing_header.html" with items=pages active_tag=active_tag %}
    <div class="listing-links">
        {% for page in pages %}
            <p>
                <a href="{{ page.url }}">{{ page.title }}</a>
            </p>
        {% endfor %}
    </div>
{% endblock %}
```

robdivincenzo (approved):
Hey @dlopezvsr thanks for your work on this, it's looking great! Have a couple code suggestions in my review for your consideration, but overall approved.

I got a 404 on the listing page because it was querying the Abstract model. Wonder if we could just do the page model here instead for now? We might also need to keep in mind something for later so that we ignore old page types in these results 🤔. I think that's TBD though.

Updated variable names to match what you've got in the template, did a locale filter, and added a small snippet so that we can see an accurate result. I think that snippet might eventually be a pattern template as well, but not required for this PR!

Let me know what you think!

PR #13999, opened 2025-05-21
TP1-2499 Image Card with Title

dlopezvsr:
# Description

This PR introduces the image Card with Title component, considering the variants of this component two blocks were created: `SingleImageBlock` and `DoubleImageBlock`. These two blocks were included under `sized_images` field as  a `StreamField`, added inside of `abstract_general_page.py` and `abstract_article_page.py`. A generic template for general page was also created for testing purposes. Also, simple `scss` style was added which needs to be updated later with the exact sizes and design.

### Variations considered:

- Double: Portrait - Landscape
- Double: Same type
- Single: Portrait
- Single: Landscape

Link to sample test page: https://mozilla-hub.atlassian.net/browse/TP1-2499

**Tests**
- Pull this branch and apply new migrations.
- Access to the CMS and create a General Page.
- You should see the  `sized_images` field with two options, Single and Double.
- Add the images and publish the page.
- Go to the live page and confirm single and double block is rendering correctly.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2700)


robdivincenzo:
Looking good @dlopezvsr 🔥🔥🔥! This is giving me a couple ideas.

1. Instead of creating new fields, what if we added it as a streamfield block option in the body field on the `AbstractBasePage` model so that a user can place it anywhere in the body?
2. I also wonder if we can consolidate this down to just a single ImageBlock and re-use that in a bunch of places.
-- Single Image Portrait Block = your single `ImageBlock`
-- Single Image Landscape Block = your single `ImageBlock`
-- Double Image Block = Update the options in the [TwoColumnContainerBlock](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14001) with your new `ImageBlock`, and the editor can select `ImageBlock` for both columns.
--- Portrait/Landscape variants(s) handled on the ImageBlock level like you have.

If we can do it like that, it would cut down our need to maintain a separate two column layout. And possibly streamline the editor experience. Do you have any thoughts on this?

dlopezvsr:
> Looking good @dlopezvsr 🔥🔥🔥! This is giving me a couple ideas.
> 
>     1. Instead of creating new `sized_images` fields, what if we added it as a [streamfield block option in the body field](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/59a2af93de1a646724b847376a22786d360ccec3/foundation_cms/base/models/abstract_base_page.py#L52-L59) on the `AbstractBasePage` model so that a user can place it anywhere in the body on any page?
> 
>     2. I also wonder if we can consolidate this down to just a single ImageBlock and re-use that in a bunch of places.
>        -- Single Image Portrait Block = your single `ImageBlock`
>        -- Single Image Landscape Block = your single `ImageBlock`
>        -- Double Image Block = Update the options in the [TwoColumnContainerBlock](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14001) (technically `ColumnStreamBlock`) with your new `ImageBlock`, and the editor can select `ImageBlock` for both columns.
>        --- Portrait/Landscape variants(s) handled on the ImageBlock level like you have.
> 
> 
> If we can do it like that, it would cut down our need to maintain a separate two column layout. And possibly streamline the editor experience. Do you have any thoughts on this?

Hey @robdivincenzo thanks so much for your comments. I found them very useful and accurate!

I just finished applying these changes, let me know what you think:
- I removed the previous duplicated blocks I had created, and left only one block: `ImageBlock`
- I included this block inside of the `StreamField` body to make it more flexible and reusable.
- And finally included it inside `ColumnStreamBlock`

Let me know what you think 👍

foundation_cms/core/views.py
```diff
@@ -24,3 +24,7 @@ def listing_page(request):
     }
 
     return render(request, "patterns/pages/core/listing_page.html", context)
+
+def general_page(request):
+    
+    return render(request, "patterns/pages/core/general_page.html")
```
robdivincenzo, line 30:
```suggestion
```

foundation_cms/static/scss/redesign_main.scss
```diff
@@ -3,6 +3,7 @@
 @import "../scss/fonts";
 @import "../scss/components/primary_nav";
 @import "../scss/components/footer";
+@use "../scss/components/images.scss";
```
robdivincenzo, line 6:
```suggestion
@import "../scss/components/images.scss";
```
@mmmavis just had a PR where this is changed to `@import`, so figure we should catch that now!

foundation_cms/blocks/image_block.py
```diff
@@ -0,0 +1,22 @@
+from wagtail.blocks import StructBlock, CharBlock, ChoiceBlock
+from wagtail.images.blocks import ImageBlock
+
+class CustomImageBlock(StructBlock):
```
robdivincenzo, line 4:
```suggestion
from wagtail.blocks import CharBlock, ChoiceBlock
from wagtail.images.blocks import ImageBlock
from foundation_cms.base.models.base_block import BaseBlock

class CustomImageBlock(BaseBlock):
```
New pattern is to inherit from the `BaseBlock`. I have a PR out there do retroactively update mine but let's catch this one now!

foundation_cms/templates/patterns/blocks/image_block.html
robdivincenzo, line 1: Once the `image_block.py` template_name is updated, this will need to live in `/patterns/blocks/themes/default/image_block.html`

robdivincenzo (approved):
Awesome work @dlopezvsr and approved! Smooth editing experience 🚀

Some small tweaks because dev patterns changed a little bit. Once these are in good to merge!!

PR #14003, opened 2025-05-22
Impact Numbers Block Backend

danielfmiranda:
##  Description

This PR introduces a new `ImpactNumberBlock` StreamField block for use on all pages that inherit from `AbstractBasePage`. The block is designed to showcase up to three animated impact stats.

##  New ImpactNumberBlock

- Adds a new reusable block called `ImpactNumberBlock` with a list of `ImpactStatBlock` items.

- Each ImpactStatBlock includes:

  - `stat_number`: required string (supports commas, decimals, prefixes ($,#) and suffixes (K,M))

  - `stat_description`: required short description

## Template Tag

- Introduces a custom template filter: `render_animated_number`, registered in `impact_numbers_tags`.

- This filter takes the raw `stat_number` value and renders each character as its own HTML element. Numeric digits are ready for animation, while characters ($,K,M) remain static.

## Frontend Implementation

- Renders each stat using a responsive grid layout (1–3 items).

- Uses BEM-based class naming.

## Digit Animation

- Animation is done using CSS using [this codepen](https://codepen.io/fungland/pen/GqNWMb) for reference.
- Animations are initialized on scroll, through the use of JS.

## Accessibility

- Uses aria-label to expose the full stat value to screen readers.

- Adds`aria-hidden="true"` to animated digit elements to prevent digit-by-digit reading.

## Screenshots

CMS:
Screenshot 2025-05-30 at 15-56-18 Editing Home Page (new) Age happy area  - Wagtail

FrontEnd:
Screenshot 2025-05-30 at 15-56-53 Age happy area  - Foundation Home Page

Video:

https://github.com/user-attachments/assets/af7ce78a-20fa-4e93-8be9-b58c40cda8b8

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2704)


foundation_cms/base/models/abstract_base_page.py
```diff
@@ -9,10 +9,22 @@
...
+# Shared StreamField block types for use across pages that inherit from AbstractBasePage.
+# Extend this list in specific page models (e.g., HomePage) to add more blocks as needed.
+base_page_block_options = [
+    ("rich_text", RichTextBlock()),
+    ("image", CustomImageBlock()),
+    ("audio", AudioBlock()),
+    ('tabbed_content', TabbedContentContainerBlock()),
+    ("two_column_container_block", TwoColumnContainerBlock()),
+    ("impact_numbers", ImpactNumberBlock()),
+]
```
danielfmiranda, line 25: Not directly related to this PR, but I thought it would be helpful for the future!

foundation_cms/static/js/blocks/impact_numbers_block.js
```diff
@@ -0,0 +1,130 @@
+/**
```
danielfmiranda, line 1: Short explanation of the JS file:

Takes whatever numerical value is set for the Impact Number, and randomizes each digit for 1.5 seconds before landing on the actual value.

foundation_cms/static/js/redesign_main.js
```diff
@@ -1,8 +1,13 @@
...
 import "foundation-sites";
+import { initAllAnimatedImpactNumbers } from "./blocks/impact_numbers_block.js";
 
 $(document).foundation();
 
 console.log(
   "😃 If you see this message, the JS watch and build steps are working properly! 😃",
 );
+
+document.addEventListener("DOMContentLoaded", () => {
```
danielfmiranda, line 11: Hi @mmmavis! I was wondering where the best place to be to init() watchers for blocks that are usable across all pages? For example, I saw that we are doing so in main.js on the old site, however, I wasnt sure if we wanted to take a new approach for this site. 

Thanks!

mmmavis: Hey @danielfmiranda , thanks for checking

This is what we have currently 

```bash
foundation_cms/static/js
├── components
│   └── home_page
│       └── kinetic_type_brand_line.js
├── pages
│   └── home_page.js
└── redesign_main.js
```

I'm thinking we can do

```bash
foundation_cms/static/js
├── components
│   └── home_page
│       └── kinetic_type_brand_line.js
│   └── common (<-- new)
│       └──component_1.js
│       └──component_2.js
├── pages
│   └── home_page.js
└── redesign_main.js (<--- and initialize the common components here)
```

Curious what you think about this.

foundation_cms/static/js/blocks/impact_numbers_block.js
danielfmiranda: Thanks @mmmavis! I was able to implement something close to the codepen provided and its ready for re-review 👍

robdivincenzo (approved):
R+ @danielfmiranda! Very clever use of template tag here to handle digit vs. non-digit numbers. Please double check w/ @mmmavis to resolve the conversations, but this LGTM 🚀

foundation_cms/static/js/redesign_main.js
```diff
@@ -1,8 +1,13 @@
...
+  initImpactNumberStatAnimationsOnScroll();
+});
```
mmmavis:
```suggestion
initImpactNumberStatAnimationsOnScroll();
```
We can drop the DOMContentLoaded listener because our script tag is already placed just before `</body>` and all the relevant HTML is rendered by the server

foundation_cms/static/js/blocks/impact_numbers_block.js
```diff
@@ -0,0 +1,45 @@
...
```
mmmavis, line 1: Minor improvement suggestion -

Let's extract the hard-coded selector strings into a single `SELECTORS` object. This makes it easier to update or rename classes. e.g., 
```js
const SELECTORS = {
  statContainer: ".impact-stat__number",
  digit: ".impact-stat__digit"
};
```

mmmavis (changes requested):
Really clever implementation approach! I've left two minor suggestions below!

mmmavis (approved):
🎉 Awesome work!

PR #14020, opened 2025-05-28
Portrait Cards

robdivincenzo:
# Description

This PR adds Portrait Cards & Portrait Card Set backend components.

Also includes a new OptionalLinkBlock `StreamField` wrapper to make the LinkBlock optional for scenarios like this.

# To Test

1. Check out branch and run migrations
2. Edit homepage
3. Add new `Portrait Card Set`
4. Fill out the fields for Headline(s) and add 3x Portrait Cards, and note the optional link field for the set.

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2723)


foundation_cms/blocks/link_block.py
```diff
@@ -75,4 +75,13 @@ def __init__(self, local_blocks=None, **kwargs):
         self.child_blocks.pop("label")
 
 
+class OptionalLinkBlock(blocks.StreamBlock):
+    link = LinkBlock()
+
+    class Meta:
+        min_num = 0
+        max_num = 1
+        required = False
```
robdivincenzo, line 84: Hey @danielfmiranda, I wanted your opinion on this. When I added just the LinkBlock as the cta_link field, even with required=False, it still threw an error that it was required. This seems to trigger because the `link_to` field via `LinkBlock` and the `label` field via `BaseLinkBlock` are required. But in this usecase, the link should be optional

So I considered two options... one was to edit the `link_to` and `label` fields to be requried=False, and the other was to make an optional `StreamBlock` wrapper where it basically acts as it's own StreamField and only triggers the required fields when an editor adds a link.

My gut tells me we want to leave those base fields required and wrap it in an optional block like this. But I wanted to hear your thoughts on this approach and if there's something else I'm not considering.

danielfmiranda: Hey @robdivincenzo! Thanks for checking.

I agree that leaving those `base_fields` as required would probably be the best approach. 

Also, I appreciate the new `OptionalLinkBlock`, I think its a clever way to make the LinkBlock optional! 

On the old site, we would instead wrap the [LinkBlock in a ListBlock](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/ebc39617c6181f2a727412b0051088b19532c639/foundation_cms/legacy_apps/wagtailpages/pagemodels/customblocks/blog_index_callout_box_block.py#L43-L45):

```python
    link_button = blocks.ListBlock(
        LinkBlock(), min_num=0, max_num=1, help_text="Optional Link Button for the callout box."
    )
```

which would make things a little confusing as we would always have to [refer to the first item in the "list"](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/6d40e1d22df36c1c879d81ffd36af8ea82b7e7c0/foundation_cms/legacy_apps/wagtailpages/templates/wagtailpages/blocks/image_teaser_block.html#L10-L16):

```html
                {% if card.link_button %}
                    {% with link_button=card.link_button.0 %}
                        <a class="tw-{{ link_button.styling }} link-button tw-mt-8" href="{{ link_button.url }}" {% if link_button.new_window %}target="_blank"{% endif %}>
                            {{ link_button.label }}
                        </a>
                    {% endwith %}
                {% endif %}
```

I see that we still need to reference `0` in this new template, however, I think having Optional in the name and referencing the block like that still clears up some confusion, so this looks ready to merge to me. 

However, I have two quick suggestions which I am not sure if they are possible, but they might be worth checking out! (time permitting):

1. Updating LinkBlock to use `required` as a parameter:
  - I see that we are defining `required = False` in the `OptionalLinkBlock`'s meta. I'm curious, could we update the LinkBlock model to have this passed in as a parameter? That way we can just say LinkBlock(required=False). 

If thats not possible:

2. Adding helper method to return link value without using `.0.`:
 - I am curious if this is possible, as we know that there will always be just 1 optional link block.

PR #14029, opened 2025-05-29
Yarn workspaces & inv tasks for new site

robdivincenzo:
# Description

This PR:

- Transitions the dual npm/package-lock.json and yarn/yarn.lock into a single yarn.lock workspaces approach
  - heroku does not accept more than one lock file through buildpacks... resulting in 500 errors on deploy
  - Workspaces should allow for lighter package management with shared node_modules going into the root dir. Workspaces are configured in the root `package.json` file. And allows for multiple `package.json` files elsewhere
  - We can run yarn commands like this now from the root:
    - `yarn install` to run install
    - `yarn build` to build workspaces (this is a custom script in the root package.json)
    - `yarn workspace legacy CMD` will run a yarn specific command for the legacy frontend (ie linting)
    - `yarn workspace redesign CMD` will run a yarn specific command for the redesign frontend (ie linting or adding a new package)
- updates the CI and `inv` commands to use new yarn commands.
- adds global yarn commands to the root `package.json` to run both workspace commands. IE a global `yarn lint` from root. This will re-connect the `inv format` workflow for example.
- moves legacy frontend config code (config.js's, tests, package.json, etc.) to `frontend/legacy` and subdirs the redesign fronend config code into `frontend/redesign`
- updates some documentation
 
# To Test
These are all under the hood, so basically our test scenario is:

1. Note that heroku builds the [review app](https://foundation-s-tp1-2701-i-dewnko.herokuapp.com/en/) successfully
2. Check out this branch
3. Run a fresh build via `docker-compose build` or `inv setup`
4. Build should work as normal
5. Feel free to note and test out yarn commands via `cd` into the new `frontend/legacy` or `frontend/redesign` workspace directories and running `yarn [script_name_defined_in_package_json_file]`. And even try them from the root directory via `yarn lint` or `inv format`. We now use yarn instead of `npm run` to initialize a script run. We _can_ still use npm run within the legacy scripts, and so it did not require a complete refactor of those. I tried to leave it as untouched as possible.
6. Script commands and CI should still work as previous

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2731)


danielfmiranda (approved):
Thanks @robdivincenzo! Excellent work 🙌

dlopezvsr (approved):
@robdivincenzo everything looking great. Amazing work, thank you!

PR #14086, opened 2025-06-13
Video block

danielfmiranda:
# Description

This PR includes the following changes:

 - A new `VideoBlock` model which includes the following fields:
   - `video_url` (required) - The ID of the desired video.
   - `caption` (optional) - the caption to render at the bottom of the video
   - `caption_url` (optional) - the optional URL for the caption to link to
   - HTML and SCSS needed for the block to work as expected.

**Link to review app:** https://foundation-s-redesign-v-cjk59u.herokuapp.com/en/

# Screenshots

CMS: 
Screenshot 2025-06-18 at 18-03-30 Editing Home Page (new) Redesign Homepage - Wagtail

Video Block:
Screenshot 2025-06-18 at 18-29-12 Redesign Homepage - Redesign Homepage

Video Block within Two Column Container Block:

Screenshot 2025-06-18 at 18-31-00 Redesign Homepage - Redesign Homepage


foundation_cms/static/scss/blocks/themes/default/video_block.scss
```diff
@@ -0,0 +1,47 @@
+.video-block {
+  &__container {
+    position: relative;
+    width: 100%;
+    padding-top: 56.25%;
```
danielfmiranda: Found this neat trick to make the video 16:9 while being growing and shrinking with screen size! https://css-tricks.com/aspect-ratio-boxes/

foundation_cms/static/js/blocks/video_block.js
```diff
@@ -0,0 +1,37 @@
...
+
+    video.addEventListener("pause", () => {
+      playButton.classList.remove("is-hidden");
+    });
+
+    video.addEventListener("play", () => {
+      playButton.classList.add("is-hidden");
+    });
+  });
+}
```
kevdmoz: @danielfmiranda I noticed the video is missing the control. Can we set "controls" to true? https://developer.vimeo.com/api/oembed/videos

danielfmiranda:
Thanks for taking a look @kevdmoz!

As mentioned in our standup today, I have updated the PR to embed a vimeo video rather than using a custom HTML video player.

Double checking if you would still like me to enable controls though? I believe @sabrinang mentioned that was intentionally not in the design.

danielfmiranda:
Thanks for taking a look @sabrinang!

Via the Vimeo UI, users are able to customize the appearance of their desired video. This includes UI colors as well as what UI elements they would like to appear. 

For example, I have updated the review app with a different video hosted on our vimeo with a blue play button.

Vimeo also allows us to save settings as presets.

If you would like to take a look with the alternate video, the link is: https://foundation-s-redesign-v-cjk59u.herokuapp.com/en/

Thanks again!

danielfmiranda:
Question for the devs @robdivincenzo @mmmavis @ramram-mf @dlopezvsr : 

For this PR I used other instances of video blocks in the codebase for reference. 

In previous renditions, we allowed users to use either youtube links or vimeo videos, so it makes sense why we would have to create a custom video player HTML/iframe element to account for the both of those.

However, since the ask of this ticket is to only allow vimeo videos, would it make sense to take advantage of Vimeo's "Copy Embed Code" functionality , which automatically generates an iframe, for the video player portion of the block's template?

Here is an example of the code generated from the button:

`<iframe src="https://player.vimeo.com/video/1073235226?title=0&amp;byline=0&amp;portrait=0&amp;badge=0&amp;autopause=0&amp;player_id=0&amp;app_id=58479" width="3840" height="2160" frameborder="0" allow="autoplay; fullscreen; picture-in-picture; clipboard-write; encrypted-media; web-share" title="Join us at Mozilla Festival 2025 - Badges on sale now"></iframe>`

My thinking is that it would be more intuitive to users to copy and paste that code generated, and also allow us to reduce the number of things that could break from our end

sabrinang:
@danielfmiranda Thanks for the info! If we can customize the UI colour, can we change to our brand colour like blue-500 `#006BD4` as our default then?

I wasn't able to see the alternate video styling
image

danielfmiranda:
Thanks for taking a look @sabrinang! The video in the screenshot is actually not using the video block mentioned in this PR. However, i have updated it in the CMS to show the "reference" video using the blue button.

Also, we can update videos to use our brand color, but this is done on the vimeo side. I will add a preset for use in our vimeo settings, similar to the one that the reference video is using, and update the UI color to be #006BD4

foundation_cms/blocks/video_block.py
```diff
@@ -0,0 +1,25 @@
+from wagtail import blocks
+
+from foundation_cms.base.models.base_block import BaseBlock
+from foundation_cms.blocks.link_block import LinkWithoutLabelBlock
+
+
+class VideoBlock(BaseBlock):
+    video_url = blocks.CharBlock(
```
mmmavis, line 11: I included [a custom Vimeo URL validator ](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/redesign/foundation_cms/blocks/hero_accordion.py#L19-L33)for the the `video_url` field in `hero_accordion.py`. I wonder if we can refactor it somehow so both blocks can use it?

foundation_cms/static/scss/blocks/themes/default/video_block.scss
```diff
@@ -0,0 +1,17 @@
...
+  }
+
+  &__iframe {
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: 100%;
+    height: 100%;
+  }
+}
```
mmmavis: We now have a simpler way to set aspect ratio... just use `aspect-ratio`! Note that you only need to specify either `width` or `height` and the other dimension will be calculated automatically. (Disclaimer: double check if my code suggestion below works as expected 😅 )

```suggestion
@use "sass:math";

.video-block {
  $video-aspect-ratio: math.div(16, 9);

  &__wrapper {
    position: relative;
    width: 100%;
    aspect-ratio: $video-aspect-ratio;
    overflow: hidden;
  }

  &__iframe {
    position: absolute;
    inset: 0;
  }
}
```

foundation_cms/templates/patterns/blocks/themes/default/video_block.html
```diff
@@ -0,0 +1,35 @@
+{% load wagtailcore_tags wagtailimages_tags static i18n %}
```
mmmavis, line 1: Nit: Ignore if this is too much work. We can totally get back to this after we have proper linting/formatting rules in place.

It'd be great if we can stick to 2-space indentation instead of 4 :)

foundation_cms/static/scss/blocks/themes/default/video_block.scss
danielfmiranda: Super cool! Thanks for the heads up @mmmavis I have updated the CSS to use the 16,9 setting

foundation_cms/blocks/video_block.py
danielfmiranda: Thanks Mavis! I've extracted this logic into [validators.py](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/4a1359bb2f753e57e08e041cba8560824257e718/foundation_cms/validators.py) and have updated both blocks to use it. Vimeo URL errors now appear like this: 
image

foundation_cms/templates/patterns/blocks/themes/default/video_block.html
danielfmiranda: Done!

mmmavis: Thank you!

foundation_cms/blocks/video_block.py
mmmavis: That's perfect. Thank you @danielfmiranda !

mmmavis (approved):
🥂 🎉

PR #14112, opened 2025-06-17
"Move legacy page" script

danielfmiranda:
# Description
This PR includes the script that we can run via `python manage.py` to move over pages from one location to another. 

This PR also includes an update to the primary page template so it does not render the old navigation once we move it to live under the new homepage. Since this page will not be visible until we move it after we update the site to point to the new homepage, this change is OK to merge in in anticipation of the move.

# Screenshot

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2826)


danielfmiranda:
Thanks for the heads up @robdivincenzo! In that case I have removed the update from this PR, and re-requested review 👍

robdivincenzo (approved):
R+!

PR #14153, opened 2025-06-26
Button Block Styling

danielfmiranda:
# Description


foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
```diff
@@ -0,0 +1,68 @@
...
+    background-color: color(neutral, "600");
+    text-decoration: none;
+    cursor: pointer;
+    width: max-content;
+    overflow: hidden;
+
+    &::after {
+      background-image: url("../_images/icons/button-arrow-right-white.svg");
+      content: "";
+      width: 16px;
```
dlopezvsr: Quick observation, @mmmavis mentioned that we wanted to use rem-calc() custom function to convert pixels, but not sure if it applies also here. Just leaving the comment as an FYI.

mmmavis: Thanks for catching this @dlopezvsr ! For this one since it's pretty obvious that it's 1rem, we can either just put `1rem` for convenience or `rem-calc(16)`

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
```diff
@@ -0,0 +1,68 @@
...
+      height: 16px;
```
mmmavis: let's use rem instead of px

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
mmmavis: Thanks for catching this @dlopezvsr ! For this one since it's pretty obvious that it's 1rem, we can either just put `1rem` for convenience or `rem-calc(16)`

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
```diff
@@ -0,0 +1,68 @@
...
+    }
+
+    &__text {
+      display: block;
+      white-space: nowrap;
+      opacity: 1;
+      transform: translateY(0);
+      transition:
+        transform 0.5s ease,
+        opacity 0.6s ease;
```
mmmavis: Should the two duration match? Also it'd be good if we can extract that as a variable.

foundation_cms/templates/patterns/blocks/themes/default/link_button_block.html
```diff
@@ -0,0 +1,15 @@
...
+    <a href="{{ value.url }}"
+       {% if value.new_window %}target="_blank"{% endif %}
+       class="link-button">
+      <span class="link-button__roller">
+        <span class="link-button__rolltext">
+          <span class="link-button__text">{{ value.label }}</span>
+          <span class="link-button__text link-button__text--roll-in">{{ value.label }}</span>
+        </span>
+      </span>
+    </a>
```
mmmavis: To make this button template code reusable by all other templates (directly in page template or block), let's extract this into a partial. Then we can update this file to use that partial!
```
<div class="link-button-block">
  {% include path_to_partial ... %}
</div>

File: foundation_cms/static/scss/blocks/themes/default/link_button_block.scss [hunk 5]
```diff
@@ -0,0 +1,68 @@
+.link-button-block {
+  margin: rem-calc(16) 0;
+
+  .link-button {
```
mmmavis: Since buttons are shared across the site, it would be better if we move the button styling to its own file (similar to `foundation_cms/static/scss/type.scss`). This file can be as short as 

```
.link-button-block {
  margin: rem-calc(16) 0;
}
```

😁

foundation_cms/templates/patterns/blocks/themes/default/link_button_block.html
danielfmiranda: This sounds like a great idea! I have implemented it 👍 Thanks Mavis

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
danielfmiranda: Thanks @mmmavis! This makes sense to me. I've created a new file, `buttons.scss` which includes the sizing, spacing, and rolling text functionality. That way we are able to add buttons anywhere on the site, and in more specific files, define things such as background color, icon color, etc, as I am doing in `link_button_block.scss`

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
danielfmiranda: Thanks for the reminder @dlopezvsr! I have updated it

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
danielfmiranda: Done!

danielfmiranda:
Thanks for the feedback @mmmavis and @dlopezvsr! I have implemented your suggestions and this PR is now ready for re-review whenever you have a chance. Thanks again!

sabrinang (approved):
Looks good (spacing, size and motion follows Nancy's prototype), thanks!

mmmavis (approved):
Thanks @danielfmiranda !

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
mmmavis: Sorry, one last question 😅 I'm wondering if we should move the styling here to `buttons.scss`? Do the link-button used in the LinkButtonBlock behave and are styled differently than the regular link-button? @danielfmiranda

danielfmiranda: No problem! 
That is a good question 🤔 And after thinking about it some more, I think the answer is no. 

Because after looking at the mock ups, it appears that buttons all over the site are styled the same, even in templates where their locations are hard coded.

So since the link_button_block is just a block that allows us to render a link button in the page body, I think it makes more sense to house this styling in `buttons.scss`. 

I have updated the PR to reflect this in the latest commit!

mmmavis: Thanks for the code update. If these link buttons are just primary button, what do you think about merging the rest of the code in `link_button_block.scss` into `buttons.scss`? I would rename the button to `btn-primary` too to align with Figma

image

danielfmiranda: Thanks Mavis! Just updated the PR to include both styling options:

image
Screenshot 2025-06-27 at 17-19-07 Redesign Homepage - Redesign Homepage

mmmavis (approved):
✨ Thanks @danielfmiranda !

PR #14174, opened 2025-07-01
Apply styling cards timely activations

ramram-mf:
# Description

This PR implements the small, medium and large viewport layouts for the Timely Activations cards component of the homepage.

## Screenshots

Large layout screenshot

*Large layout*

Medium layout screenshot

*Medium layout*

Small layout screenshot

*Small layout*

Mouse hover interactive animation

*Hover animation*

Currently the hover animation is triggered by the whole card surface, while the clickable area is limited to the CTA link nested block. Image zoom is 105%, as in other card zoomed components.

Link to sample test page: https://foundation-s-tp1-2589-a-jzrdj1.herokuapp.com/en/
Related PRs/issues: #13911

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2894)


foundation_cms/blocks/timely_activations_cards_block.py
```diff
@@ -31,24 +30,13 @@ class TimelyActivationsCardsBlock(BaseBlock):
 
     title = blocks.CharBlock(required=False, help_text="Title for the activations section")
```
mmmavis, line 31: Let's remove this field. We've decided to remove `title` field from any blocks to make our blocks more versatile.

mmmavis (changes requested):

foundation_cms/blocks/timely_activations_cards_block.py
mmmavis: We've decided to remove title field from blocks to make our blocks more versatile. I think that applies to this block as well. @ramram-mf  Could you confirm with the designers if we can remove this field from this block?

mmmavis (changes requested):
@ramram-mf , This is looking great! Could you please extract the SCSS for `.activation-card__cta` into buttons.scss so we can easily apply the same styling on other components. I'd also extract the template code for this text link into a fragment for the same reason. 

image

sabrinang (changes requested):
+1 to Mavis comment regarding shared cta link style and this goes for the rest of the heading styles to double check we are using style defined/shared with the rest of the site where we can

Zoom animation: I think the zoom can be a bit more subtle (as it zooms in a lot more than the accordion hero), generally zooming into an image should match with the rest of the site hover image zooms

Not explicitly doing tablet but you have great solutions for the in between breakpoints!

Overall, looking really good Ram

sabrinang (approved):
The zoom is looking good thanks for fixing that. Check that the 'category' is using the defined label style in the design system otherwise, I think it is good to go!

PR #14190, opened 2025-07-03
Add new template filter 'should_wrap_block' to allow full bleed blocks

mmmavis:
Add new template filter `'should_wrap_block'` to allow full bleed blocks.

To test locally,

1. Go to `foundation_cms/templatetags/streamfield_tags.py`
1. Add name of the block you wanna test to `NO_WRAP_BLOCKS`
1. Go to CMS and add the block to homepage
1. Inspect source code in dev tool
1. Verify the block itself is not pre-wrapped in `.grid-container`, `.grid-x`, and `.cell`

Related PRs/issues: N/A

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2905)

### DIFF
--- a/foundation_cms/static/scss/redesign_main.scss
+++ b/foundation_cms/static/scss/redesign_main.scss
@@ -5,6 +5,7 @@
 @import "../scss/colors";
 @import "../scss/type";
 @import "../scss/buttons";
+@import "../scss/utilities";
 
 // Site-wide Blocks
 @import "../scss/blocks/themes/default/impact_number_block";
--- a/foundation_cms/static/scss/utilities.scss
+++ b/foundation_cms/static/scss/utilities.scss
@@ -1,0 +1,9 @@
+.grid-container {
+  // this class is useful if you need to have a full-bleed grid-container on mobile
+  &.mobile-padding-0 {
+    @include breakpoint(large down) {
+      padding-left: 0;
+      padding-right: 0;
+    }
+  }
+}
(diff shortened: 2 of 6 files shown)


foundation_cms/static/scss/redesign_main.scss
```diff
@@ -5,6 +5,7 @@
 @import "../scss/colors";
 @import "../scss/type";
 @import "../scss/buttons";
+@import "../scss/utilities";
```
mmmavis, line 8: Can't think of a better name at the moment. Suggestion welcome!

foundation_cms/static/scss/utilities.scss
```diff
@@ -0,0 +1,9 @@
+.grid-container {
+  // this class is useful if you need to have a full-bleed grid-container on mobile
+  &.mobile-padding-0 {
```
mmmavis, line 3: not used in this PR yet but will be used in block template code once this PR is merged

foundation_cms/static/scss/redesign_main.scss
robdivincenzo: This works!

robdivincenzo (approved):
R+ @mmmavis! Great idea to make a utility for blocks that should be full-bleed 🔥

Longer term, once we better know how we're reusing these blocks across different contexts (like in NP), we _might_ benefit doing this check via a block class method rather than a filter. Thinking of a case where a block is a full bleed in the normal context, but needs the wrapper in a NP themed context kind of thing. But I think it's easy enough to iterate to that if it becomes necessary, and this is ready to go 🚀

mmmavis:
Hey @robdivincenzo , totally agree this will most likely need adjustment in the future!

PR #14213, opened 2025-07-08
Fix newsletter signup block homepage

ramram-mf:
# Description

This PR fixes the double `.grid-container` being applied to the `newsletter_signup` block, resulting in an extra 1rem padding on the sides of the component when rendered.

## Current state:
image

## Fixed state:
image

# To test

1. Clone the `fix-newsletter-signup-block-homepage` branch, run `inv setup`.
2. Create a new `newsletter_signup` snippet with the `expand_on_focus` variant selected.
3. Add a `Newsletter signup` block to the new Homepage, publish.
4. Check in the live page that the double `.grid-container` is no longer present.


dlopezvsr (approved):
LGTM @ramram-mf. Thanks!

PR #14214, opened 2025-07-09
Homepage component fixes

ramram-mf:
# Description

This PR fixes various homepage issues:

## 1. Primary navigation wordmark scroll animation

With the current implementation the scroll animation wasn't triggered on sync when the `kinetic_type_wordmark` scrolled past the navigation bar but until it went over the document's top offset instead. Intersection Observer `rootMargin` property is now correctly calculated and set. This results in a really organic transition which looks and feels better and is less distracting in consequence.

### Current state

*Current wrong offset `large` and `small` viewports layout wordmark scroll animation*

### Fixed state

*Fixed offset `large` and `small` viewports layout wordmark scroll animation*

## 2. Mobile primary navigation layout shift

When the mobile nav menu open due to an outdated body padding value. This value was updated to allow blocks to be immediately positioned after the primary navigation bar in other page types. The spacing was restored by adding a custom rule to the `home_page.scss` stylesheet.

## 3. Explicit vertical spacing in the `featured_card_block`

For both `small` and `large` viewport layouts. Currently the visible space between elements is produced by the default paragraph `<p>` browser style, which is inconsistent with our grid (18px vs 16px) and could vary depending on browser implementation altering the component appearance.

### Current state

image
image

*1: No gap due to paragraph element not present. 2: Paragraph element present and rendering imprecise margins.*

### Fixed state

image

*Native `flexbox` gap*

Link to sample test page: https://foundation-s-fix-featur-df14el.herokuapp.com/en/
Related PRs/issues: #

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2935)

### DIFF
--- a/foundation_cms/static/js/components/primary_nav.js
+++ b/foundation_cms/static/js/components/primary_nav.js
@@ -167,13 +167,14 @@ export function initPrimaryNav() {
  * Hides the navigation wordmark when the kinetic type wordmark is visible in the viewport.
  */
 export function initWordmarkVisibilityOnScroll() {
+  const nav = document.querySelector(SELECTORS.primaryNav);
   const grid = document.querySelector(SELECTORS.primaryNavGrid);
   const wordmark = document.querySelector(SELECTORS.wordmark);
   const kineticTypeWordmark = document.querySelector(
     SELECTORS.kineticTypeWordmark,
   );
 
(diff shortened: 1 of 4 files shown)


dlopezvsr (approved):
Thank you @ramram-mf! R+

PR #14219, opened 2025-07-10
Portrait Card Styling

robdivincenzo:
# This PR is archived and rebased/recreated here: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14332

---------

# Description

This PR contains the code for updated portrait card styling and carousel functionality. Noting that on mobile the static 3 card set became a carousel, this has been refactored and consolidated down to a single element and template that adjusts based on context (viewport and number of cards).

Link to sample test page: https://foundation-s-tp1-2486-p-adceci.herokuapp.com/en/

Credentials:
admin
Qp@QVX0L!W3OJ!mz

# To Test

1. Navigate to the test page on a desktop view
2. Confirm the 3 visible cards match the design req's
3. Shrink browser window
4. Note it's a carousel now (2 cards on tablet, 1 on mobile)
5. Log into the app via https://foundation-s-tp1-2486-p-adceci.herokuapp.com/cms and use the credentials provided above
6. Add a slide
7. Note on desktop it's a carousel and retains carousel on resizing.

Related PRs/issues: TP1-2486

┆Issue is synchronized with this [Jira Story](https://mozilla-hub.atlassian.net/browse/TP1-2942)


foundation_cms/static/scss/blocks/themes/default/portrait_card_set.scss
```diff
@@ -0,0 +1,138 @@
...
+    transition: transform 0.4s ease;
+  }
+
+  .portrait-card {
+    flex: 0 0 auto;
+    display: flex;
+    flex-direction: column;
+    padding: 2.5rem 3rem 4.5rem;
+    min-height: 29.25rem;
+    aspect-ratio: 5 / 8;
```
mmmavis:
```suggestion
    aspect-ratio: math.div(5, 8);
```
because SCSS doesn't like the simple `/` expression anymore 😂

foundation_cms/static/scss/blocks/themes/default/portrait_card_set.scss
```diff
@@ -0,0 +1,138 @@
...
+  );
+
+  &__card-container {
+    display: flex;
+    flex-wrap: nowrap;
+    gap: 1.5rem;
+    overflow: hidden;
+    max-width: 100%;
+    position: relative;
+    -webkit-overflow-scrolling: touch;
```
mmmavis: We have `autoprefixer` in place (see `frontend/redesign/postcss.config.js`) so I just wanna double check if this line is needed?

foundation_cms/static/scss/blocks/themes/default/portrait_card_set.scss
```diff
@@ -0,0 +1,138 @@
...
+    scrollbar-width: none;
+
+    &::-webkit-scrollbar {
```
mmmavis: maybe this line is not needed as well?

foundation_cms/static/scss/blocks/themes/default/portrait_card_set.scss
```diff
@@ -0,0 +1,138 @@
...
+    box-sizing: border-box;
+    width: 80%; // Mobile default
+    scroll-snap-align: start;
+
+    @include breakpoint(medium) {
+      width: 45%; // Tablet
+    }
+
+    @include breakpoint(large up) {
+      width: calc(33.3333% - 1.3333rem);
```
mmmavis: Would be great to have some inline comment for this. I'm assuming 33.3333% comes from `100% / numCardsPerViewport` but can't quite tell what `1.3333rem` is. Could we turn these two numbers into variables for better clarity, please?

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,189 @@
...
+    this.root.addEventListener("keydown", (e) => {
+      if (e.key === "ArrowRight") this.slideTo(this.index + 1);
+      if (e.key === "ArrowLeft") this.slideTo(this.index - 1);
+    });
 
-    track.addEventListener("keydown", (e) => {
-      if (e.key === "ArrowRight") scrollNext();
-      else if (e.key === "ArrowLeft") scrollPrev();
+    // Recalculate position on resize
+    window.addEventListener("resize", () => {
```
mmmavis: Let's have debounce in place to help with performance. Something like https://github.com/MozillaFoundation/foundation.mozilla.org/blob/redesign/foundation_cms/static/js/blocks/spotlight_card_carousel.js#L386-L408

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,189 @@
...
+      { passive: true },
+    );
+
+    this.viewport.addEventListener(
+      "touchend",
+      (e) => {
+        if (!isDragging) return;
+        isDragging = false;
+        const delta = e.changedTouches[0].clientX - startX;
+        if (Math.abs(delta) > 50)
```
mmmavis: Could we have `50` stored as a variable, please.

foundation_cms/static/scss/blocks/themes/default/portrait_card_set.scss
```diff
@@ -0,0 +1,138 @@
...
+    &:not(:last-child) {
+      margin-right: 1.5rem;
+    }
+
+    &:last-child {
+      margin-right: 0;
+    }
+
+    &__image {
+      aspect-ratio: 1 / 1;
```
mmmavis:
```suggestion
      aspect-ratio: math.div(1, 1);
```
because SCSS 🤷🏻

mmmavis (changes requested):
Nice work @robdivincenzo ! I've dropped some inline comments. Happy to discuss anything.

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,189 @@
...
+    const card = this.cards[this.index];
+    if (!card) return 0;
+    const style = window.getComputedStyle(card);
+    return card.getBoundingClientRect().width + parseFloat(style.marginRight);
+  }
+
+  // Move the carousel track by transform
+  updateTransform(index, animate = true) {
+    const offset = this.getSlideOffset() * index;
+    this.track.style.transition = animate ? "transform 0.4s ease" : "none";
```
mmmavis: if `transform 0.4s ease` is shared with the CSS, it would be better if we set it as a CSS variable so both JS and CSS can reference to the same value.

beccaklam (changes requested):
## Firefox
**Overall**
- [x] Image zoom animation is missing on hover
- [x] Can we make the whole card clickable (I know this is under discussion for all cards right now, not sure if this is the right moment to implement, just flagging)
- [ ] On hover, can the heading also be underlined? (this is a new feature design is implementing for all text links. Again not sure right now is the right time for implementation, just flagging)
- [ ] There seems to be something wrong with the using the carousel on my phone. It works fine at first, but then the transition between cards gets choppy before it stops working altogether. I have to reload the page to reset it. The pagination arrows also stop working.

**Mobile Specific**
- [ ] Spacing between cards seems big right now, should be 8px
- [ ] Padding seems too big, should be 24px (seems to be using desktop padding right now)
- [ ] Image corner radii also seem too large, should be 28px

## Chrome
- [ ] All the same issues as stated above under Firefox

**Mobile-Specific:**
- [ ] For some reason, on my laptop the carousel is too tight to the edge of the phone screen when I'm using dev tools but it looks fine on my phone? (margins should be 16px) Not sure if this is a bug but flagging

## Safari:
- [ ] Same issues as stated above under Firefox

## Edge:
- [ ] Same issues as stated above under Firefox

Overall it's looking great though @robdivincenzo! I really like the infinite scroll, thank you for suggesting that. Also always happy to chat through any of this feedback.

foundation_cms/static/js/blocks/portrait_card_carousel.js
robdivincenzo: Great idea, thanks!

foundation_cms/static/js/blocks/portrait_card_carousel.js
robdivincenzo: Got it 👍, and thanks for the example!

robdivincenzo:
Thanks for all the feedback @mmmavis and @beccaklam! The noted items should now be addressed in the PR and reflected in the updated review app. I just added another demo of the portrait carousel in the homepage body.

Quick note for @beccaklam - the two unchecked boxes I did not specifically address, but I wasn't able to recreate the issues. It's possible that the other revisions addressed them, or that I'm missing a step to see the bug myself. If you are able to re-create them, then please flag me again with steps or screenshots to help direct me! If not, I think we can mark them as resolved 😁🤞

EDIT: I may have found it. There was some weird interaction with sliding during a hover / active state. I've added a check to ensure it doesn't do that until the state resolves.

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,225 @@
...
+  setInitialPosition() {
+    requestAnimationFrame(() => {
+      this.updateTransform(this.index, false);
+    });
+  }
+
+  // Navigate to a given index
+  slideTo(newIndex) {
+    const shouldDisable =
+      !this.root.classList.contains("is-carousel") && window.innerWidth >= 1024;
```
mmmavis: Let's store `1024` as a var for easy update in the future!

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,225 @@
...
+    if (shouldDisable || this.isTransitioning) return;
```
mmmavis: I notice that the arrows don't respond to multiple clicks until the sliding animation is done. For example, if I rapidly click the next button twice, the carousel only advances by 1 slide instead of 2. After seeing this line, I just wanna make sure this is an intentional UX behaviour!

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,225 @@
...
+        image.addEventListener("transitionend", animateSlide, { once: true });
+        return;
+      }
+    }
 
-    track.addEventListener("keydown", (e) => {
-      if (e.key === "ArrowRight") scrollNext();
-      else if (e.key === "ArrowLeft") scrollPrev();
+    // Fallback for headline::after or if no hover
+    setTimeout(animateSlide, 300);
```
mmmavis: Could we please store `300` as a var?

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,225 @@
...
+
+    this.track.innerHTML = "";
+    this.applyCardColorDataAttrs(tripled);
+
+    tripled.forEach((card) => this.track.appendChild(card));
+    this.cards = Array.from(this.track.querySelectorAll(SELECTORS.card));
+  }
+
+  // Calculate scroll offset for the current index
+  getSlideOffset() {
```
mmmavis: `getSlideOffset()` calls `getBoundingClientRect()` and `getComputedStyle()` on every slide animation. Since all the cards have uniform width, we're recalculating the same value each time a user navigates (they are expensive operations).

What do you think about caching this value instead? These `getBoundingClientRect()` and `getComputedStyle()` force reflows which could impact animation smoothness. By caching, I meant something calculating the width once during init/resize and storing it in something like `this.slideWidth`, then just using that cached value during animations.

foundation_cms/static/js/blocks/portrait_card_carousel.js
```diff
@@ -1,43 +1,225 @@
...
+    const style = window.getComputedStyle(card);
+    return card.getBoundingClientRect().width + parseFloat(style.marginRight);
+  }
 
-    const scrollPrev = () => {
-      track.scrollBy({ left: -cardWidth, behavior: "smooth" });
+  // Move the carousel track by transform
+  updateTransform(index, animate = true) {
+    const offset = this.getSlideOffset() * index;
+    this.track.style.transition = animate
```
mmmavis: Similar to the `getSlideOffset()` issue, this calls `getComputedStyle()` on every animation to read a CSS variable. If the CSS var doesn't change, I'm thinking this value could also be cached.

beccaklam (changes requested):
## Overall
- [ ] The most recent changes enlarge the whole image component on hover. It should just be the image zooming in on itself (see the cards under "Do good. Do your part." for an animation example

## Mobile Specific
- [ ] I can see that the padding has been changed to 24px but the spacing feels really tight against the rotated image. Is there a way to make it 24px between the tightest point between the rotated image and card edge?
- [ ] The carousel seems to be working better on my phone but the swiping feels delayed. Would it be possible to make it so that the cards move instantly when a user swipes? Almost like the cards move instantly with the user's finger
- [ ] Also would it be possible to make the "cut off" card go to the edge of the display screen? Like this:
Screenshot 2025-07-22 at 10 20 01 AM


Thank you for making all those changes @robdivincenzo I think it's close!

foundation_cms/static/js/blocks/portrait_card_carousel.js
robdivincenzo: Thanks @mmmavis! This was intentional because the infinite scroll effect had required a guardrail somewhere for the animation to end to handle edge cases. Since the wait for transition to finish was not preferred, I've refactored the slide. Should be a lot snappier.

robdivincenzo:
All latest revisions are in! But closing this PR and recreating here: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14332

PR #14243, opened 2025-07-15
Apply style to two column container (Don't merge)

adalcima:
# Description

This PR is about applying styles to a two-column container. The template can be found here foundation_cms/`templates/patterns/blocks/themes/default/two_column_container_block.html` and basically adds the container for this component within the defined template, along with the styles for it. The `title` field was also removed from the template, as it is no longer used in the model.

Files updated: 
```
foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
foundation_cms/static/scss/redesign_main.scss
foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
```
Review app: https://foundation-s-tp1-2645-a-p8lkyq.herokuapp.com/en/

### Tests
- Go to the CMS > Redesign Homepage
-  Select body, add > Two Column Container 
- Add new components for the Two column container (left and right)

## Images
Screenshot from 2025-07-14 17-50-47
Screenshot from 2025-07-14 17-54-28
Screenshot from 2025-07-14 17-56-38

Link to sample test page:
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2645


foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
```diff
@@ -0,0 +1,6 @@
+.two-column-container {
+  padding: 2rem 0;
+  background: $white;
```
mmmavis:
```suggestion
  background: map.get($mofo-colors, white);
```
because `$white` is referencing to a variable the legacy site uses and will be removed once we sunset the legacy code.

You will also need to include 

```scss
@use "sass:map";
@import "../../../colors";
``` 
at the top of this file for it to work I think.

mmmavis (changes requested):
@adalcima Nice first PR on this repo! ✨ I've left a tiny code suggestion but everything else looks great! If this PR is requested by designers, don't forget to also tag them for review. Thanks!

foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
adalcima: Thanks for the feedback @mmmavis. The change has been applied :)

nancyt1:
@adalberto-cima Love how well this is working! I had a few styling requests but not sure if it applies to this block or the individual blocks so let me know. 

#### 1. Is it possible to add the Spacer Block as an option so that we can use it if we want to keep one of the columns empty? And also to adjust spacing within a column if needed.
#### 2. Can we add an asterisks to the Alt-text field for images? It seems like alt-text is required on photos but it's not explicitly shown. 
#### 3. When adding multiple blocks back to back, the spacing between them feels too big on mobile. Can we lower to something around 75px of space between blocks?
**Current:**
image

**Proposed:**
image


#### 4. Can we make the newsletter box a bit wider on mobile so that these 2 blocks are the same width?
image


#### 5. Can we increase spacing between the 2 columns on mobile to be 2rem? Currently it's at ~22px. 
SCR-20250716-luii

sabrinang (changes requested):
^ Please see Nancy's comment above as it captures what design reviewed especially around mobile spacing

adalcima:
Hi @nancyt1 @sabrinang thanks for the feedback. Updating this PR with some observations 

1. Added **SpacerBlock** to the container and tested its use locally. I'm not sure if this is the right approach at the moment, so I'd appreciate your input on this point as well. @mmmavis  @ramram-mf 
image

2. Yes, it can be added. It's a property of CustomImageBlock and is actually optional, but the "Image is decorative" box should be checked. I understand this could be a usability issue, but I'm not sure it's within the scope of this PR.
image

3. Blocks back to back using 0.25rem on mobile, after and before the change (see image below)
image


**_`4. Still in progress. Gettin some help from Ram (since I'm not sure if it has to do with the NewsletterSignupBlock component)`_**


5. Increasing spacing between columns on mobile to 2.5rem.
image

adalcima:
Regarding point 4 (missing in previous comment), the selector was updated to use the one created in the `newsletter_signup_form.scss` file cc @ramram-mf  I would appreciate your comments pls
image

sabrinang:
@adalcima the deployment and/or review app isn't working for us to review/preview

foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
```diff
@@ -0,0 +1,31 @@
+$two-column-mobile-margin-bottom: 0.25rem;
+$cell-block-margin-bottom: 2.5rem;
```
mmmavis: If these vars are only used for this block, let's move them into `.two-column-container { ... }`

mmmavis (changes requested):
Thanks @adalberto-cima ! Code wise I have just one tiny comment regarding SCSS vars. I also noticed the commit history contains commits from other work. Could we have this branch rebased to make sure commit history is clean?

adalcima:
Hi @sabrinang I just updated the link to the review app. There were some migration issues to fix. Sorry for the inconvenience!

foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
adalcima: Hey @mmmavis , sure it can be moved. Regarding the commit history, it was caused by switching the base branch between `redesign` and `main` (since it would now be merged directly into main). Perhaps a better option would be to branch with the clean changes and avoid that new commit history. I'll put it on the table to choose the best option. Thanks for the note!

adalcima: Updated adding vars into the block. Thanks @mmmavis

adalcima:
This PR will be closed for the benefit of this new PR https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14331

PR #14330, opened 2025-07-23
Add responsive image to newsletter signup form success (from main)

ramram-mf:
# Description

This PR implements a style update on the newsletter signup form that displays an image next to the success state message on instances other than the footer form.

Link to sample test page:
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-2963)


beccaklam (changes requested):
Hi @ramram-mf!
- [ ] I'm seeing two hand drawn elements appear at some breakpoints. Can we remove the top hand drawn element until the thank you text is above the image?
- [ ] Would we also be able to center align the hand drawn element with the "Thank you" text?image

🤞 Hoping we'll be done after these changes. Thank you again!

beccaklam (changes requested):
Sorry @ramram-mf one more thing!
- [ ] Could we actually bottom align the hand drawn element and "thank you" when they are side by side? Something about having text underneath makes the center alignment look off.

Thank you!

PR #14331, opened 2025-07-23
[Recreated] Apply styling to the Two column container

adalcima:
# Description

> This PR is a continuation of a PR that had been worked on the redesign branch, and we have opted to use a clean branch based on main to avoid code and commit history inconsistencies.
https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14243
Recent changes added in the previous PR are already added here:
-Spacer Block
-Spaces for mobile
-Newsletter box a bit wider
-Vars inside appropriate css block.

This PR is about applying styles to a two-column container. The template can be found here foundation_cms/`templates/patterns/blocks/themes/default/two_column_container_block.html` and basically adds the container for this component within the defined template, along with the styles for it. The `title` field was also removed from the template, as it is no longer used in the model.

Files updated: 
```
foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
foundation_cms/static/scss/redesign_main.scss
foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
foundation_cms/static/scss/redesign_main.scss
```
Review app: https://foundation-s-main-tp1-2-s5wik7.herokuapp.com/en/

### Tests
- Go to the CMS > Redesign Homepage
-  Select body, add > Two Column Container 
- Add new components for the Two column container (left and right)


mmmavis (approved):
Code looks good to me! Thank you @adalberto-cima !

sabrinang (changes requested):
Overall spacing has improved and I reviewed some combination of components, here is some feedback:

- **Audio Block** — I couldn't log into simplecast to grab an embed code to test this however, the field text are showing up in the block and doesn’t load image. Also, when it is on the left side the columns are misaligned on scrrens larger than >1024px 

image

- **Spacers** — good to have to create an empty column (as columns couldn't be empty as they are a required field) but the spacing differences between small and extra large aren't that different.

**Specific component feedback:**
- **Newsletter signup** — button breaks out of column sometimes
image
image

- **Rich text** — spacing and alignment in general could be adjusted (space after headings, text alignment are not adjustable)

adalcima:
Hi @sabrinang thanks for the feedback, recent updates added:
- For the newsletter, we should be fine using the `Expanded newsletter signup` 
- Columns can now be optional without using the spacer. 
- **_Vertical alignment of the columns content_**
- **_I also couldn't test the audio block in staging with the [dummy](https://iframely.com/domains/simplecast) due to our site's security, but the image also doesn't display if that component is placed outside the 2-column container. I'll try to investigate this further, but I think it has more to do with that individual component._**

foundation_cms/static/scss/components/list_block.scss
```diff
@@ -64,3 +64,12 @@ body.redesign {
     }
   }
 }
+
+/* stylelint-disable-next-line selector-class-pattern */
+.block-two_column_container_block {
```
mmmavis: We should avoid using `stylelint-disable-next-line` as much as possible. Would just targeting `.two-column-container` be enough? We should also move this ruleset to `two_column_container_block.scss`

foundation_cms/static/scss/components/list_block.scss
```diff
@@ -64,3 +64,12 @@ body.redesign {
...
+  .list-block {
+    margin-bottom: 0;
+    margin-left: -1rem;
```
mmmavis: Is 1rem referring to the value in `$grid-margin-gutters`? If so, we should use the var instead.

foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
```diff
@@ -1,14 +1,8 @@
 {% load wagtailcore_tags %}
 
-<div class="grid-container">
+<div class="two-column-container block-two_column_container_block grid-container two-column-container--{{ self.vertical_alignment }}">
```
mmmavis: I could be missing something but my first thought is we don't need to introduce a new class (`block-two_column_container_block`) here?

foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
```diff
@@ -339,4 +339,10 @@
     margin: 0;
     padding: 0;
   }
+
+  .newsletter-signup {
```
mmmavis: Let's get rid of the `block-two_column_container_block` class altogether. If we still need this ruleset (we may be able to get rid of some/all rules I think?), we should move this ruleset to `two_column_container_block.scss`.

foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
mmmavis: I also wonder if we need so many nested `grid-container / grid-x / cell` for this block.  `.two-column-container` is wrapped in `.grid-container` already so I'm thinking `.two-column-container` itself can just be the simple flex container that has two equal width child divs (with `gap` referencing `$grid-margin-gutters`). This way we can simplify both the template code and the SCSS code used to offset the padding/margin which come with Foundation Framework's grid system.

mmmavis (changes requested):
Sorry, just added a few more comments regarding the template code and SCSS code.

foundation_cms/static/scss/components/list_block.scss
adalcima: Using now `-$global-padding;` to match values with vars

foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
adalcima: I was actually reusing styles created in the individual components. This extra class has been removed and the styles have been adapted within` .two-column-container`.

foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
adalcima: Moved to `.two-column-container`

foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
```diff
@@ -1,25 +1,17 @@
...
-        {{ block }}
-      {% endfor %}
-    </div>
-
-    <div class="cell small-12 large-6">
-      {% for block in self.right_column %}
-        {{ block }}
-      {% endfor %}
-    </div>
+  <div class="container-column container-column-left">
```
mmmavis: I know these names are long, but it's better to stay consistent with our BEM naming convention. 😄 
```suggestion
  <div class="two-column-container__column two-column-container__column--left">
```

foundation_cms/templates/patterns/blocks/themes/default/two_column_container_block.html
```diff
@@ -1,25 +1,17 @@
...
+    {% for block in self.left_column %}
+      {{ block }}
+    {% endfor %}
+  </div>
 
+  <div class="container-column container-column-right">
```
mmmavis:
```suggestion
  <div class="two-column-container__column two-column-container__column--right">
```

foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
```diff
@@ -0,0 +1,41 @@
...
+  }
+
+  .container-column {
+    flex: 1;
+    padding: 0 $global-padding;
+
+    @include breakpoint(medium up) {
+      padding: 0 $global-padding;
+    }
+  }
```
mmmavis: The `padding` rules here seem to create extra padding on the left and right edges of the container. What do you think about removing these rules and use `gap` on `.two-column-container` instead.

mmmavis (changes requested):
@adalcima Thanks for making the changes. SCSS and template code are much cleaner now. I have left a few additional suggestions to get us to the finish line.

foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
adalcima: You're correct: using `padding` on each column can create unwanted extra space on the left and right, especially if the parent already has its own padding. Switching to a gap :)

sabrinang (changes requested):
- **Alignment:** overall works however, some components for top/bottom align don’t visually align as I assume some components have some built in spacing (particularly the rich text block I think) which makes elements look misaligned (can separate how to address this on what is deliberate and what is not)
image
image
image

- **List block:** is touching the other column, missing gap, too wide on mobile 
image
image

-  **Audio block:** has extra space on mobile, images don’t render, same issues as before
- **Video block:** video alignment and narrow size (see below)
- **Mobile horizontal alignment is impacted by vertical alignment setting:** top, middle, bottom becomes left, middle, right (I don’t think it needs to be impacted and elements can span the width of grid to match other components)

nancyt1 (changes requested):
Can we add a little gap here when the container has rich text and a list block? To achieve the same spacing as the other blocks. 
 image

Thank you for all your work on this, this was a large block to tackle!

sabrinang:
@adalcima I've tested the Audio block with a Simplecast embed and it still doesn't render. I think since this ticket is quite large, we can separate the Audio component to not block this ticket?

adalcima:
Hi @sabrinang thanks for the observations, regarding this:

`List block: is touching the other column, missing gap, too wide on mobile`

I'm probably wrong; I was following [this use case from figma](https://www.figma.com/design/4vdZ9mWyEwOz1D6Z7x9pqr/%C2%A0%E2%93%82%EF%B8%8F-mozillafoundation.org-MVP?node-id=12966-33895&m=dev). In case we do indeed need that space in the List block, could you please confirm? 
image

adalcima:
@sabrinang And regarding the Audio Block, I just added a missing security policy to that Review app (it's not enabled by default), which is why the Simplecast code wasn't rendering. That policy is already added in Prod, so there won't be any issues there.
image

sabrinang:
@adalcima 
**Regarding List block + Newsletter:**
-  I clarified with designers and that component was meant to built together (same heights) so it does not utilize the 2-column block like it does now. Sorry for the confusion, but this will work for MVP being built like this for now so please format it with the gap like all the other components so it can combined in any combo without space issues.

**Regarding Audio block:**
- Great! Glad the Simplecast embed works now. It still renders the field labels which does not need to show up ("audio title" "audio description" "audio simplecast embed" etc.)

adalcima:
Hi @sabrinang, thank you, one question pls, is there any example design for the `Audio Block `that I could use for reference to apply these styles to the rest of that component (_title, description, etc._), I haven't been able to find any, sorry. On the other hand, perhaps these Audio Block styling adjustments could be addressed in a separate ticket. Please let me know your comments :). Thanks

sabrinang:
@adalcima +1 Let's separate the Audio styling from this ticket to not block this as the 2 column work and also isn't dependent on it

foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
```diff
@@ -386,8 +386,7 @@
   }
 }
 
-/* stylelint-disable-next-line selector-class-pattern */
-.block-two_column_container_block {
+.two-column-container {
```
mmmavis: Could we move this entire block to `foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss`, please?

mmmavis (changes requested):
Thanks @adalcima . Just one tiny code adjustment then the code is good to go!

foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
adalcima: It's been moved to `two_column_container_block.scss`. Fyi, @ramram-mf , this was added recently.

mmmavis (approved):
Thank you for the great work, @adalcima ! ✨

adalcima:
Thanks team. In order to perform the merge, I'll need the reviews for @sabrinang and @nancyt1 due to repository policies. Thanks :). cc @mmmavis @ramram-mf

nancyt1 (approved):
+1 to Sabrina's comment to continue refining audio block in another PR but this is really great, thanks!

sabrinang (approved):

PR #14338, opened 2025-07-24
Added onetrust banner to redesign pages

danielfmiranda:
# Description

This PR adds the JS snippet and template tags necessary for the OneTrust cookie banner to render at the bottom of the redesigned pages.

If a user selects one of the options, a browser cookie is stored with their choice so the banner does not reappear.

 # Screenshots

Screenshot 2025-07-23 at 17-01-21 Redesign Homepage


foundation_cms/templatetags/app_environment_tags.py
```diff
@@ -0,0 +1,19 @@
...
+    app_env = get_app_environment()
+    if app_env == "Staging":
+        env_prefix = "[S]"
+    elif app_env == "Review":
+        env_prefix = "[RA]"
+    return env_prefix
+
+
+def get_app_environment():
+    return settings.APP_ENVIRONMENT
```
danielfmiranda, line 19: While the `environment_prefix` tag is not exactly needed for this work, it seemed like something useful to bring over from the old site for later work!

ramram-mf (approved):
LGTM @danielfmiranda thanks!

foundation_cms/templatetags/app_environment_tags.py
ramram-mf: Definitely useful, thanks!

PR #14341, opened 2025-07-24
Replace ImageChooserBlock instances with ImageBlock

robdivincenzo:
# Description

This PR will replace ImageChooserBlocks with ImageBlock to enable the alt text fields in the redesign site. Swapping to ImageBlock should be [backwards compatible and not require any data migrations or template changes](https://docs.wagtail.org/en/6.3/reference/streamfield/blocks.html#wagtail.images.blocks.ImageBlock).

Login:
admin
&87kZaRDj(6_IAhZ

Link to sample test page: https://foundation-s-tp1-2990-i-8hrtbh.herokuapp.com/cms/pages/3/edit/
Related PRs/issues: TP1-2990

# To Test
1) Log into the review app and edit the Redesign Homepage
2) Find the "Featured Card" and note that the alt text is now present with the option to make image decorative
3) Add an audio block and note that the alt text is now present with the option to make image decorative
4) Add a portrait card set and note the alt text is now present with the option to make image decorative

### DIFF
--- a/migration-cleanup.sh
+++ b/migration-cleanup.sh
@@ -2,7 +2,7 @@
 set -e
 
 # This script can help automate migration conflicts and local environment
-# It stashes your changes, pulls / checks out redesign branch makes a clean db, 
+# It stashes your changes, pulls / checks out main branch makes a clean db, 
 # rechecks out your branch / stash, detects any migrations that haven't run
 # (and deletes them because these are probably just conflicting) and then regenerates them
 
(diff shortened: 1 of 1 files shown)


foundation_cms/templates/patterns/blocks/themes/default/audio_block.html
```diff
@@ -3,7 +3,6 @@
     <div class="grid-container">
         <h3>Audio Block</h3>
         <p>Audio Image: {{ self.image }}</p>
-        <p>Audio Image Alt Text: {{ self.image_alt_text }}</p>
```
robdivincenzo, line 6: Seems like this template isn't totally in use yet, but I'm removing this line for now since it's referencing a field no longer in the model. Alt text attaches to the rendered image

danielfmiranda (approved):
LGTM! 🚀 thanks @robdivincenzo

PR #14345, opened 2025-07-24
Rich text block - ensure it's using the right template

mmmavis:
# Description

Link to sample test page:
Related PRs/issues: #

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

**Merge Method**

### DIFF
--- a/foundation_cms/base/models/abstract_base_page.py
+++ b/foundation_cms/base/models/abstract_base_page.py
@@ -34,7 +34,9 @@ from foundation_cms.blocks import (
 # Extend this list in specific page models (e.g., HomePage) to add more blocks as needed.
 base_page_block_options = [
     # [TODO/FIXME] consider ordering or grouping these blocks
-    ("rich_text", RichTextBlock()),
+    ("rich_text", RichTextBlock(
+        template="patterns/blocks/themes/default/rich_text_block.html",
+    ),),
     ("image", CustomImageBlock()),
     ("audio_block", AudioBlock()),
     ("tabbed_content", TabbedContentContainerBlock()),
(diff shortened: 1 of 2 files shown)


robdivincenzo (changes requested):
Thanks Mavis! One quick question on the approach. What are your thoughts on making a `CustomRichTextBlock` for this? I know it's on the base page and should inherit really well across our other current pages, but if we want to use that elsewhere in streamfield blocks (think wanting to use this custom rich text block template within tabbed content), or if there was a future page that didn't inherit from basepage for some reason, it might help to make this a little more DRY

```
from wagtail.blocks import RichTextBlock

class CustomRichTextBlock(RichTextBlock):
    class Meta:
        template = 'patterns/blocks/themes/default/rich_text_block.html'
        label = 'Rich Text'
```

This way we can reference it more easily in other models and easily more easily add additional features or update template, etc. in the future:

`("rich_text", CustomRichTextBlock()),`

But I don't know if this change will only ever be bespoke to our page models that inherit from BasePage. So if you don't think it's relevant, let me know and I'll approve because the approach is totally fine in that case 😆

mmmavis:
Hey @robdivincenzo I love your suggestion! However, I would like to propose that we make this a post-MVP discussion. Since I'm not familiar with how every component is built, I'm hesitant to introduce code that could potentially break the look of the components this close to our launch. If this sounds good to you, I'll create a ticket for us to follow up.

robdivincenzo:
> Hey @robdivincenzo I love your suggestion! However, I would like to propose that we make this a post-MVP discussion. Since I'm not familiar with how every component is built, I'm hesitant to introduce code that could potentially break the look of the components this close to our launch. If this sounds good to you, I'll create a ticket for us to follow up.

Sounds good me! Let's keep the momentum and explore this post-MVP, R+! 🚀

PR #14350, opened 2025-07-25
Tabbed Content Styling (from main)

ramram-mf:
# Description

This PR implements the styling for the `tabbed-content-container` block, and its children blocks. Featuring `small` and `large` viewport layouts.

Link to sample test page:
Related PRs/issues: #


foundation_cms/static/scss/components/tabbed_content.scss
```diff
@@ -1,91 +1,275 @@
...
-.tabs-container {
+
+$tab-colors: (
+  0: color(blue, "300"),
+  1: color(blue, "200"),
+  2: color(blue, "100"),
+  3: color(orange, "300"),
+  4: color(orange, "200"),
+  5: color(orange, "100"),
+);
```
mmmavis: Great to have these values in a map! If these values are not referenced by other files, let's avoid making this a global variable

foundation_cms/templates/patterns/blocks/themes/default/tabbed_content_container_block.html
```diff
@@ -1,26 +1,24 @@
 {% load wagtailcore_tags %}
 
 {% if self.tabs %}
-  <div class="tabs-container">
+  <div class="tabbed-content">
```
mmmavis: Although it may result in a longer class name, I recommend aligning the class name as closely as possible with the block name for improved clarity and maintainability (e.g., `tabbed-content-container).

foundation_cms/templates/patterns/blocks/themes/default/text_image_block.html
```diff
@@ -1,14 +1,14 @@
 {% load wagtailimages_tags wagtailcore_tags %}
 
-<div class="grid-x grid-margin-x text-image-block">
```
mmmavis, line 3: I'm not familiar with the specs. Do changes in this file produce expected results for the `TextImageBlock` (if used as standalone block or in other compound component)? Can you set up the review app in a way designers can take a look and verify? Thanks!

foundation_cms/static/scss/components/tabbed_content.scss
```diff
@@ -1,91 +1,275 @@
...
+  &__image {
+    order: 1;
+
+    @include breakpoint(large up) {
       order: 2;
-      display: flex;
-      align-items: center;
-      justify-content: center;
+      flex-basis: 1;
+      min-width: 54%;
```
mmmavis: The 54% value looks like a magic number. Could we add some context about how this value was determined? If it can be replaced with a formula and store it as a variable, that would be even better.

foundation_cms/static/scss/components/tabbed_content.scss
```diff
@@ -1,91 +1,275 @@
...
+}
+
+.tab-card {
+  display: flex;
+  flex-direction: column;
+  width: calc(50% - 0.5rem);
+
+  @include breakpoint(large up) {
+    width: calc(25% - 0.75rem);
+  }
```
mmmavis: Could we add some explanation for the `50%` and `25%` values? Defining them as variables could also help improve clarity and maintainability.

foundation_cms/static/scss/components/tabbed_content.scss
```diff
@@ -1,91 +1,275 @@
+@use "sass:math";
```
mmmavis: let's rename the file to match the block name

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
+export function initTabbedContent() {
```
mmmavis: let's rename the file to match the block name

sabrinang (changes requested):
Here's some feedback after reviewing:
- Mobile: remove scroll bar white area so it doesn’t look disconnected _(appears occasionally)_
Screenshot 2025-07-28 at 9 44 59 AM

- No topic tags style for MVP, it can be styled with ‘label’ type style instead
Screenshot 2025-07-28 at 10 01 50 AM

- Portrait image height feels too tall (use a fixed height instead? try same length as horizontal width like ~500px)
- Images get a bit pixelated when it switches to mobile
image

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
...
+          behavior: "smooth",
+        });
+        currentPage.textContent = `${currentPageIndex + 1} / ${pages}`;
+      }
+    };
+
+    nextButton.addEventListener("click", nextPage);
+    prevButton.addEventListener("click", prevPage);
+  });
+}
```
mmmavis: I noticed `initTabbedContentCardSets()` only works on elements created by `initTabbedContent()`. What do you think about calling it internally from `initTabbedContent()` instead of exporting both? That way we'd have a simpler API and anyone using this would only need to call one function

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
...
+
+  const cardSetPanels = document.querySelectorAll(
+    SELECTORS.tabbedContentCardSetPanels,
+  );
+
+  cardSetPanels.forEach((panel) => {
+    const cards = panel.querySelectorAll(SELECTORS.tabbedContentCard);
+
+    // Wrap cards in pages of 4 cards each
+    const pages = Math.ceil(cards.length / 4);
```
mmmavis: let's declare 4 as a variable!

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
...
+    if (initialScrollLeft > 0) {
+      currentPageIndex = Math.floor(initialScrollLeft / pageWidth);
+      currentPage.textContent = `${currentPageIndex + 1} / ${pages}`;
+    }
+
+    const nextPage = () => {
+      if (currentPageIndex < pages - 1) {
+        currentPageIndex++;
+        panel.scrollTo({
+          left: panel.clientWidth * currentPageIndex,
```
mmmavis: This assumes `panel.clientWidth` equals page width, which might break with padding/margins. I wonder if we should use `pageWidth` instead for consistency

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
...
+        });
+        currentPage.textContent = `${currentPageIndex + 1} / ${pages}`;
+      }
+    };
+
+    const prevPage = () => {
+      if (currentPageIndex > 0) {
+        currentPageIndex--;
+        panel.scrollTo({
+          left: panel.clientWidth * currentPageIndex,
```
mmmavis: This assumes `panel.clientWidth` equals page width, which might break with padding/margins. I wonder if we should use `pageWidth` instead for consistency

foundation_cms/static/js/blocks/tabbed_content.js
```diff
@@ -0,0 +1,128 @@
...
+    const buttons = container.querySelectorAll(SELECTORS.tabButtons);
+    const panels = container.querySelectorAll(SELECTORS.tabPanels);
+
+    // Refer content panels by index instead of ID to allow for multiple instances in one page.
+    buttons.forEach((button, tabIndex) => {
+      button.addEventListener("click", (event) => {
+        panels.forEach((panel) => panel.classList.remove("is-active"));
+        panels[tabIndex].classList.add("is-active");
+        buttons.forEach((btn) => btn.classList.remove("is-active"));
+        button.classList.add("is-active");
```
mmmavis: What do you think about making `is-active` a constant? It's used in a few places.

mmmavis (changes requested):
@ramram-mf Nice work on the tabbed content functionality and styling! Thank you for putting so much work into it. I've left a few inline comments with some thoughts. Happy to discuss any of these points if you'd like. Thanks for putting this together.

foundation_cms/templates/patterns/blocks/themes/default/text_image_block.html
ramram-mf: Yes, this TextImage block is only meant to be used as content for the tabbed content container block. This file sets its markup. Designers can review it on the [review app link](https://foundation-s-from-main--flsgd4.herokuapp.com/en/tabbed-content-test-page/) in the description.

image

*`text/image` block*

foundation_cms/static/scss/components/tabbed_content.scss
ramram-mf: Yeah this was inferred from the prototype, calculated to match exactly the width of the text component even by word wrapping. Maybe we can calculate it from a fixed image width value.

foundation_cms/static/js/blocks/tabbed_content.js
ramram-mf: When using `pageWidth` (which doesnt account for padding/margins it looks like this.

image

beccaklam (changes requested):
Hi @ramram-mf:

It looks like you've already implemented 4 profiles per row instead of 6 so that's great! 🙌 I think it actually looks fine as is but if possible ...

- [ ] Instead of jumping straight to two profiles per row can we do 3 profiles per row for the range <1240 range to >640 range?
- [ ] Would it make things easier to make the padding around and between profiles the same number (like how it is on mobile?) We could set it at 32px for >1240, 24px for <1240 to >640, and keep it the same at 16px for <640.

You can see the updates in the figma file [here](https://www.figma.com/design/E3xU27s31av479EJOpnVEO/%F0%9F%9A%A7-MVP-Design-System-%F0%9F%9A%A7?node-id=844-9261&t=QQFbulMKVwf5BHX6-1)

Thank you!

foundation_cms/templates/patterns/blocks/themes/default/text_image_block.html
mmmavis: Got it. Thanks for the explanation. I think post-MVP we should organize our template files and possibly rename `TextImageBlock` so that it's obvious this block is exclusively used in the tabbed component.

foundation_cms/static/scss/components/tabbed_content.scss
mmmavis: Got it. Let's revisit this post-MVP because 54% is too magical to make sense for people who are new to this design and code 😅

mmmavis (approved):
Let's tackle the rest of the non-breaking issues as post-MVP to get this PR landed sooner. Thank you for your great work, @ramram-mf 🚀

PR #14366, opened 2025-07-28
Updated legacy pages to render new nav and footer if they live under the redesigned homepage

danielfmiranda:


foundation_cms/legacy_apps/donate_banner/factory.py
```diff
@@ -12,5 +12,6 @@ class Meta:
 
     name = Faker("sentence", nb_words=3)
     title = Faker("sentence", nb_words=6)
+    foreground_image = SubFactory(ImageFactory)
```
danielfmiranda, line 15: Updated because Donate Banner tests were failing due to no foreground image. This is now a required field

foundation_cms/legacy_apps/donate_banner/models.py
```diff
@@ -178,7 +181,7 @@ def clean(self):
             )
 
 
-class DonateBannerPage(Page):
+class DonateBannerPage(FoundationNavigationPageMixin, Page):
```
danielfmiranda, line 184: Updated to utilize `FoundationNavigationPageMixin`'s new "check whether I live under the legacy or redesign" homepage functionality.

foundation_cms/legacy_apps/wagtailpages/templates/wagtailpages/pages/youtube_regrets_base.html
```diff
@@ -53,8 +53,3 @@
         {% endblock %}
     </main>
 {% endblock %}
-
-
-{% block footer_block %}
-    {% include "fragments/footer.html" %}
-{% endblock %}
```
danielfmiranda, line 60: Removed so we can utilize `base.html`'s "check whether I live under the redesign or legacy homepage" if statement

foundation_cms/legacy_apps/wagtailpages/tests/buyersguide/test_homepage.py
```diff
@@ -230,7 +230,7 @@ def test_get_context_no_products(self):
         request = self.request_factory.get(self.bg.url)
         request.user = AnonymousUser()
         request.LANGUAGE_CODE = "en"
-        query_number = 8
```
danielfmiranda, line 233: It appears that the updates to the mixin introduced 2 new queries. Needed to update the count here so the tests pass.

foundation_cms/legacy_apps/wagtailpages/views.py
```diff
@@ -18,12 +20,24 @@ def custom404_view(request, exception):
     belongs to which site, as "a site" is not tied to "a django
     app" in the wagtail way of things.
     """
-    if Site.find_for_request(request).hostname == "www.mozillafestival.org":
+
```
danielfmiranda, line 23: Does pretty much the same thing as "FoundationNavigationPageMixin". The reason we need to add it here is because 404's technically are not pages in the wagtail sense so they are not able to utilize the mixin.

danielfmiranda:
Thanks for the approval and the help @robdivincenzo @ramram-mf!

@ramram-mf, I was wondering if I could please bug you of a re-review of the latest commit that I added? 

To summarize, I found that some legacy files were updating items in the nav and footer to have some weird styles such:
- Blue background and border on button hover
- Bottom text of the footer on legacy pages was larger than redesign pages
- Line height was slightly off

Nav Button on Hover:

Screenshot 2025-07-30 at 6 09 54 PM

Footer with Donate Button on Hover:
Screenshot 2025-07-30 at 6 10 09 PM


My fix was to add some specific SCSS rules to `redesign_migrated_content.scss` to get these items back to their intended state:

Nav Button on hover:

image

Footer with Donate Button on hover: 

Screenshot 2025-07-30 at 18-13-14 Campaigns - Mozilla Foundation


Could you please take a look and give another R+ if everything looks good to you SCSS wise? 

Thanks in advance!

foundation_cms/static/scss/redesign_migrated_content.scss
```diff
@@ -4,16 +4,34 @@
...
+      border-color: black;
+    }
+
+    &.dark-mode {
+      &:hover {
+        background-color: color(orange, "100");
+      }
+    }
+  }
 }
```
danielfmiranda: This is the latest update I made. I am pretty much just defining the hover state for the buttons and updating some font stuff for the footer text (it appears that bootstrap was targeting  `ul`, `ol`, and `p` elements specifically, as well as `<a>` tags with the class `btn-primary`).  

Note: While we technically have the rule `color(neutral, "600")` at our disposal, it currently does not work with Black or White, as they are not defined with lists of shades.

While I could update that in this PR, I have opted not too in the question of time, and to avoid over-complicating this PR.

PR #14372, opened 2025-07-29
Apply Large Text styling in Rich Text Editor

adalcima:
# Description

This PR addresses styling issues related to Large Text in the Rich Text Editor (Not Reflected on Rendered Page).
Changes included:
- Move the hook for extended `Rich Text `features from legacy_app to core.
- A class called '`body-text-large`' is added to the span that adds this hook, so we keep both classes:
    `<span class="tw-body-large body-text-large">Text</span>`
- A TODO task is added to adjust this hook once we move legacy.

This would allow:
- `tw-body-large` to be used by older styles (Tailwind or legacy).
- `body-text-large` to be used by new styles.

Review app: https://foundation-s-tp1-3036-a-klbdun.herokuapp.com/en/
Related PRs/issues: [#3036](https://mozilla-hub.atlassian.net/browse/TP1-3036)

### Tests

- Go to the CMS 
- Add new Rich Text component in your test page. Use the `body-large` style for "L"-styled rich text.
- Confirm that this styling is applied consistently across pages using rich text.

### Images
Screenshot from 2025-07-29 14-59-16


mmmavis (dismissed):
I'm assuming `.tw-body-large` is referencing to Tailwind body large class? For the redesign site we no longer uses Tailwind. For this PR, I think we want to look into ways on the Django/template level to ensure large text is rendered with the classname `.body-text-large`.

sabrinang (approved):
The rich text "L" applies the body large style and works in the CMS for the block and 2-column container block

nancyt1 (approved):

adalcima:
Hey @mmmavis, some research around this:
Actually, this class `.tw-body-large` for custom rich text feature for "`Large body text`" is defined inside a [hook in legacy_apps](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/107e172e558d47e979cf27980c0ab488667cadf2/foundation_cms/legacy_apps/wagtailpages/wagtail_hooks.py#L63C1-L64C1)


```
@hooks.register("register_rich_text_features")
def register_large_feature(features): 
""" 
Registering the 'large' Draftail feature which 
adds a span around the selected text with its class 
set to 'tw-body-large' 
""" 
...
```

So I can think of 2 options:
1. Move this feature registration to a new `wagtail_hooks.py` (something like `/foundation_cms/core/wagtail_hooks.py`), update the class to `.body-text-large`  and remove that _**legacy**_ snippet to avoid duplication. (Edited) I would just move the rich text feature registration to this new file.
2. Modify the class name in the existing legacy `wagtail_hooks.py` file to use` .body-text-large` instead of `tw-body-large` class (i.e. continue using the current file in `/legacy_apps/wagtailpages/wagtail_hooks.py`)

* Option 1 seems good to me for the long term, since `legacy_apps` will be deprecated. 
* Option 2 would be temporary, in case we don't have plans to move those hooks yet.

I would greatly appreciate your feedback. Thanks!

mmmavis:
Hey @adalcima , thanks for doing the investigation. Looking at the two options, option 1 is def preferred. I'm curious if there is a way to have a new hook just for the redesign site? We want to keep the legacy code in its current state as much as possible.

adalcima:
Hey @mmmavis Yes it is possible to have a new hook just for the redesign, the point is that all the **hooks** that are found within any of the **INSTALLED_APPS** (whether `legacy` or not) will be loaded, so that we cannot repeat features with the same name/functionality (including the existing feature for Rich Text `register_rich_text_features`).

In such a way that if we want to keep the legacy code in its current state, **_any of my 2 previous options will definitely impact_** somehow by changing the `.tw-body-large` class to `.body-text-large`.

Maybe we should plan to **_move/edit_** `legacy hooks` gradually outside the **legacy_apps** route towards another **INSTALLED_APPS** route from **redesign**  when we get rid of `legacy_apps`.

> So probably, this change that just adds this new class `.tw-body-large` to the SCSS file (although it adds duplicity for large text) avoids touching the existing hooks in `legacy` and maintains the current state.

I can continue investigating other alternatives, but I would like to know your opinion. Thank you

mmmavis:
@robdivincenzo - would love to know what you think!

adalcima:
Hey @mmmavis @robdivincenzo 
Some updates for this issue include:
- Move the hook for extended `Rich Text `features from legacy_app to core.
- A class called '`body-text-large`' is added to the span that adds this hook, so we keep both classes:
    `<span class="tw-body-large body-text-large">Text</span>`
- A TODO task is added to adjust this hook once we move legacy.

This would allow:
- `tw-body-large` to be used by older styles (Tailwind or legacy).
- `body-text-large` to be used by new styles.

robdivincenzo:
Hey @adalberto-cima and @mmmavis, there's a lot to consider for the long term and short term getting this launched without affecting old pages. Just saw the approach which adds both classes. If this works without conflicting styles / changing styles then I'm ok with that. But we'd need to verify.

If the class approach doesn't work though, what about in the short term (to merge today) we just add it alongside the other one?

This would mean leaving the `legacy_apps/wagtail_hooks.py` untouched. Maybe just change the description to "Legacy large body text" and the label to "LL".

Then creating a new `core/wagtail_hooks.py` with basically the same code but unique names like `redesign-large` so as to not overwrite the legacy version. And description something like "Redesign large body text", label as "RL", and adding the correct class name.

The short term downside is they'll live alongside each other as "LL" and "RL" until we can invest some more time. As a quick follow-up, we can create a new `CustomRichTextBlock` and specify the rich text features we want (thus removing the legacy large body text but keeping the new large body text feature). And then we can re-label it back to "L", etc, and clean up the editor UX.

adalcima:
Hey @robdivincenzo thanks so much. 
At first glance I don't see style conflicts using these 2 classes, and apparently the `tw-body-large` class is not defined within our new styles, so there should be no rewriting on that side. I will verify again in the old pages to make sure and, if all good, I think we could do the merge with this approach, thoughts @mmmavis ?

mmmavis:
Thanks @adalcima @robdivincenzo . Love that suggestion! Let's go with it.

adalcima:
Hi @mmmavis @robdivincenzo I was playing with some options that use the `"Large Text"` feature in our `old pages`, I have not found possible ISSUES. So I think we can continue with this Approach. Thanks for your help.

Screenshot from 2025-07-30 17-00-40

PR #14396, opened 2025-07-31
P2 Component Fixes

mmmavis:
# Changes

## Pillar Cards
- update heading sizes
- styling fixes

---

Link to sample test page:
Related PRs/issues: #

### DIFF
--- a/foundation_cms/static/scss/blocks/themes/default/pillar_card_set_block.scss
+++ b/foundation_cms/static/scss/blocks/themes/default/pillar_card_set_block.scss
@@ -13,13 +13,16 @@
       rotation: -2.5deg,
       background-color: color(blue, "200"),
       background-image: url("../_images/handdrawn/arrow-up.svg"),
+      background-position: right -1rem bottom,
       background-size: (
-        small: auto rem-calc(110),
-        large: auto rem-calc(150),
-      ),
-      background-position: (
-        small: right bottom rem-calc(12),
-        large: right 1rem bottom 1rem,
+        small: (
+          width: rem-calc(110),
+          height: auto,
+        ),
+        large: (
+          width: rem-calc(120),
+          height: auto,
+        ),
       ),
       border-top-left-radius: $corner-radius,
       border-top-right-radius: $corner-radius,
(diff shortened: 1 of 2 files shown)


mmmavis:
@beccaklam 

> The text seems to be breaking near the 1024px width. I'm wondering if instead of resizing the three cards to fit if we should just switch to the carousel when space runs out (is it possible for the code to sense that?). Or we could try swithcing to Mobile typography after 1200px. I'm also wondering if we could extend this layout (using the page container instead of 16px margins) until the 640px break point.

Thanks for the feedback. Let's go with the second approach because the first approach would require JS and would be an overkill for this. 

> we could try swithcing to Mobile typography after 1200px

Did you mean using mobile type on screens under 1200px?

> I'm also wondering if we could extend this layout (using the page container instead of 16px margins) until the 640px break point.

Because I'm a bit new to this component, could you clarify the desired layout for each breakpoint range? If you have any Figma reference that would be awesome!

nancyt1 (changes requested):
Can we decrease the spacing between here about 2 rem on Desktop?

image

mmmavis:
@nancyt1 spacing updated for the Pillar Cards!

mmmavis:
@beccaklam I assumed you were suggesting that the left edge of the Portrait Card component should always align with the left edge of the grid container, so I’ve updated the PR accordingly.

Regarding the type updates, could you clarify which text element you'd like to update, and for which breakpoint ranges?

nancyt1 (approved):
Amazing, thank you!!

ramram-mf (approved):
LGTM @mmmavis , are there any unaddressed design team comments? If not this is good to go.

mmmavis:
Thanks @ramram-mf . Yes, there's still some clarification I'd like to get from Becca.

beccaklam:
@mmmavis Sorry for the massive delay! Yes, I meant that the left edge of the Portrait Card component should always align with the left edge of the grid container. And confirming that @ breakpoints <1200px, mobile headings should be used for the Portrait Card component.

mmmavis:
@beccaklam Thanks for the clarification! I've updated the PR. Let me know what you think: https://foundation-s-tp1-3033-p-ttwpun.herokuapp.com/test/

beccaklam (approved):
@mmmavis This looks good to me! Thank you! ✨

PR #14432, opened 2025-08-06
Fix homepage accordion hero content specific issue

ramram-mf:
# Description

This PR fixes a specific content overflow presented in the homepage MVP accordion hero second panel. It overwrites typography rules to a lower scale only in `large` viewports (>1024px <1200px wide).

## Current state

image

## Fixed state

image

Link to sample test page: https://foundation-s-tp1-3062-f-tf9pyz.herokuapp.com/en/
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3062)

### DIFF
--- a/foundation_cms/templates/patterns/components/home_page/hero_accordion.html
+++ b/foundation_cms/templates/patterns/components/home_page/hero_accordion.html
@@ -34,6 +34,7 @@
                     height="{{ img_large.height }}"
                     alt="{{ img_large.alt }}"
                     loading="lazy"
+                    class="hero-accordion__video-thumbnail"
                   >
                 </picture>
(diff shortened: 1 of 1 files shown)


ramram-mf:
> At that breakpoint use mobile sizes/styles for all three panels
> 
> Overview of mobile type sizes: hero-accordion__label 18px hero-accordion__heading 24px hero-accordion__description 16px hero-accordion__cta 18px

I've updated the PR with the mobile typography rules applied to the `large` breakpoint exclusively. Still a bit tight on the 1024 mark. What do you think @sabrinang ?

image

sabrinang:
@ramram-mf It looked fine for me? 
image
(the collapsed orange video feature has a gap between the image and heading though)

ramram-mf:
It looks like the production site also displays this image in the same way, based on it has the 100% width rule and an auto height value. Do we want to review that element layout within this ticket scope since we're here? @sabrinang 

image

PR #14451, opened 2025-08-11
Add Alt Text to Hero image

adalcima:
# Description

This PR is to add `alt text` description to images in **Hero Section > Hero image.**

Link to sample test page: https://foundation-s-tp1-2984-c-ju6m6u.herokuapp.com/cms/pages/448/edit/
Test Inspire page: https://foundation-s-tp1-2984-c-ju6m6u.herokuapp.com/en/inspire_test/
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2984

**Main Changes:**
- Change the `hero_image` field type (from `ForeignKey`-based to `StreamField` + `ImageBlock`)
- Migration to change the field type using StreamField + ImageBlock. The current `hero_image` field is renamed as `hero_image_legacy` to add the new one.
- Intermediate migration to move/map the current hero image data to the new field.
- Migration to remove the `hero_image_legacy` (ForeignKey-based) field in favor of the new field. The data migration is performed before removing the ForeignKey-based field.
- Template adjustments to access image data.

**Changes in Models:**
- Changing the field type for `hero_image` in `GeneralPage`. Changing the field type for hero_image in GeneralPage (this will allow us to have the same image in different contexts)

## Images 
image

image


foundation_cms/core/models/general_page.py
```diff
@@ -21,13 +22,14 @@ class GeneralPage(AbstractGeneralPage):
         blank=True,
     )
 
-    hero_image = models.ForeignKey(
-        get_image_model_string(),
+    hero_image = StreamField(
```
mmmavis, line 25: @robdivincenzo - would love hear your thoughts. Do you think we could use `ImageBlock` directly instead of going through `StreamField`?

adalcima: Thanks @mmmavis, yeah we would appreciate your  thoughts @robdivincenzo , according to what I've been testing, using `ImageBlock` directly would not work as it is not a valid `models.Field`, while `StreamField` is (_ImageBlock_ can be used within _StreamField_).

robdivincenzo: Hey @adalcima and @mmmavis! Good ideas all around. Wagtail blocks are unfortunately only accessible via streamfield, but there also is a way to do something like this outside of streamfields by just adding another model field like:

``` 
hero_image = models.ForeignKey(
        get_image_model_string(),
        null=True,
        blank=True,
        on_delete=models.SET_NULL,
        verbose_name="Hero Image",
        help_text="Image for page hero section.",
    )
  hero_alt_text = models.CharField(
        max_length=255,
        blank=True,
        help_text="Override the image alt text for this specific usage"
    )
 ```
 
There are fair points to either approach but I think best practices would recommend using model fields whenever it's something fixed (kinda like a hero) and using streamfields when it's supposed to be more flexible content (kinda like the body).

Although streamfields are amazing to add flexibility (and kinda more DRY-ish here TBH), they do add more clicks for editors and add some complexity for data migrations later on down the road. A quick note that if we go the model field route we also have to write in our own conditional logic to make sure hero alt text is required if there's a hero image and whatnot (or with a fallback to image file description/title?). So a little more work to do model field route, but I _think_ the end result is preferable in this case.

PR #14452, opened 2025-08-11
OneTrust consent settings basic styling

ramram-mf:
# Description

This PR adds basic type and color styling to the OneTrust consent settings dialog

Link to sample test page:
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3067)

### DIFF
--- a/foundation_cms/static/scss/onetrust-override.scss
+++ b/foundation_cms/static/scss/onetrust-override.scss
@@ -203,3 +203,23 @@
     }
   }
 }
+
+// Privacy preference dialog
+#onetrust-consent-sdk,
+#onetrust-pc-sdk {
+  background-color: $white !important;
+
+  #ot-pc-title,
+  #ot-category-title {
+    @include mofo-text-style($header-styles, "h6", $header-font-family);
+  }
+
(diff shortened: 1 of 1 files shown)


danielfmiranda (changes requested):
Hi @ramram-mf thanks for taking this work on! Its looking great so far. I have just noticed to small things that I wanted to call out:

1. It appears that the accept button is touching the left side of the modal. I was wondering if we can add some margin so it aligns with the other buttons? I did want to call out that the copy of this panel does not match whats shown in your screenshots so I am actually not sure if this is the right place for me to look. 
Screenshot 2025-08-12 at 11-55-44 Redesign Homepage


2. The close button seems to be overlapping some text. Would it be possible to space it out a little more so it can be read?
Screenshot 2025-08-12 at 11-55-39 Redesign Homepage



Thanks again! And let me know if you have any questions 👍

nancyt1 (approved):
All looks good to me!

danielfmiranda (approved):
Thanks @ramram-mf for fixing! R+ 🚀

PR #14453, opened 2025-08-12
Fix tabbed content card set paging feature in mobile

ramram-mf:
# Description

This PR fixes a bug in which tab panels with card sets would not scroll through the pages in mobile devices, due to the page width being calculated when element are not visible.

## Current state
tabbed-content-card-set-current

## Fixed state
tabbed-content-card-set-fixed

## To test

1. Open the [preview app Meet Mozilla](https://foundation-s-tp1-3039-q-w2xk3m.herokuapp.com/en/meet-mozilla/) page.
2. Scroll down to the tabbed component instance with the Leadership and Board of directors card sets.
3. Switch to the Board of directors pane by clicking on the labeled pane.
4. Navigate through pages in the card set and confirm the card pages scroll correctly.

Link to sample test page: 
Related PRs/issues: #


ramram-mf:
> Hey @ramram-mf, what do you think about extracting the shared logic to keep things DRY?

Hey @mmmavis , I have refactored the block js code to be a little dryier, I explored the existing carousel implementations at the site and this instance differs a bit in the way that initial elements may not be visible at page load so we have to run some calculations at interaction time. This is a different pattern but we can have a look at how can we encapsulate both use cases in a higher scope component.

mmmavis (approved):
Hey @ramram-mf , thanks for checking the existing code for other elements in the codebase. I really like the refactor with the new `getPanelPageWidth` function. It’s perfect!

PR #14468, opened 2025-08-14
Add language selector to footer

ramram-mf:
# Description

This PR implements the language selector by adding it to the footer component.

## Screenshots

image

*`large` viewport*

image

*`small` viewport*

Link to sample test page: https://foundation-s-tp1-3073-a-nk6xq8.herokuapp.com/en/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3073)

### DIFF
--- a/foundation_cms/base/apps.py
+++ b/foundation_cms/base/apps.py
@@ -1,0 +1,9 @@
+from django.apps import AppConfig
+
+
+class FoundationCMSConfig(AppConfig):
+    name = "foundation_cms.base"
+
+    def ready(self):
+        # Import a patch for wagtail-localize image blocks
+        import foundation_cms.base.patches.wagtail_localize_image_block  # noqa: F401
--- a/foundation_cms/base/patches/wagtail_localize_image_block.py
+++ b/foundation_cms/base/patches/wagtail_localize_image_block.py
@@ -1,0 +1,46 @@
+from wagtail import blocks
+from wagtail_localize.segments.extract import StreamFieldSegmentExtractor
+
+
+def safe_handle_image_block(self, block, image_block_value, raw_value=None):
+    """
+    Patched version of wagtail_localize.segments.extract.StreamFieldSegmentExtractor.handle_image_block
+    Adds a guard so raw_value=None won't break with AttributeError.
+    """
+    # PATCH- bail early if raw_value is None or not a dict
+    # @TODO remove all of this once wagtail-localize is patched for it.
+    if raw_value is None or not isinstance(raw_value, dict):
+        return []
+
+    # Remaining code from wagtail-localize's `handle_image_block` method
+    segments = []
+
+    for field_name, block_type in block.child_blocks.items():
+        if raw_value.get("type") and raw_value.get("value"):
+            # for top-level ImageBlock, raw_value has a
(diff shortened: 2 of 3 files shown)


danielfmiranda (approved):
Thanks @ramram-mf! R+ 🚀

sabrinang:
_There isn't a design ticket so I am leaving my feedback/iterations here._

To get clarity and confirmation on this request, is this a true language picker or is this a locale picker?
- _Language:_ enables users to choose the language they want to see text in
- _Localization:_ allows users to select their preferred regional settings, including language, date/time format, currency and other regional conventions/content we may have

So a locale includes a language but a language does not encompass a full locale

**Alignment/Positioning**
- The picker UI is grouped too close with the donation button and the actions are unrelated. This can be aligned towards the bottom of the columns similarly to where the newsletter field sits
- Fix the alignment of the footer elements at the top 
image

-----------

**For language:** 
- use the language icon
- if we are using 'Language' label + selected language name is generally longer, have these on separate lines
- if we don't want to show label _(however generally perceivable labels are better for accessibility)_, then it could be one line with icon 
image

-----------

**For locale:** 
- use the globe icon
- if we are using 'Locale' label + locale abbreviation (up to 4 letters), than it could be on the same line as an option
- if we are using 'Locale' label + full locale name is generally longer, have these on separate lines
- if we don't want to show label _(however generally perceivable labels are better for accessibility)_, then it could also be one line with icon 
image

sabrinang (changes requested):
It was clarified this is a language selector so we will go with the design with the language icon and label above the dropdown:
image

Please reference the updated the footer component here: https://www.figma.com/design/E3xU27s31av479EJOpnVEO/%F0%9F%9A%A7-MVP-Design-System-%F0%9F%9A%A7?node-id=841-8147&t=M8SyZLquB1flxjbk-1

The ideal UX would be to only show languages that we have content available for the language to avoid confusion if selected and the language does not render. If we must show partial translations we should have a way to indicate language is incomplete or not show it at all.

danielfmiranda:
Thanks again for taking this on @ramram-mf! 

Per request of stakeholders/management, we can merge this in as-is to prevent it from blocking the release of Nothing Personal. I've worked with design to record the improvements/requests in this PR's comments in a [new ticket](https://mozilla-hub.atlassian.net/jira/software/c/projects/TP1/boards/1918?selectedIssue=TP1-3256) so we can implement those post MVP. cc: @sabrinang

ramram-mf:
The review app returns a 500 error when accessing the [root url `/`](https://foundation-s-tp1-3073-a-5nsyfc.herokuapp.com/) and does not redirect to [`/en/`](https://foundation-s-tp1-3073-a-5nsyfc.herokuapp.com/en/) as usually. When updating the environment variable `DEBUG` to `True`, the behavior goes back to the usual redirection. This could impact a lot if the homepage does not redirect correctly when implementing this PR.

ramram-mf:
Merging now the bug has been [fixed](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14468/commits/ecdf8f54bf221213fb5bc4b3fa1b26ac4b8c9507).

PR #14469, opened 2025-08-14
Add video as asset in hero section

adalcima:
# Description

This PR is to add `hero video` as new asset in **Hero Section** of a page in the CMS.
With this, we would have the option to show an image or video within the Hero Section.
Using `wagtailmedia` module based on the example [here](https://github.com/torchbox/wagtailmedia).

Link to sample test page: https://foundation-s-tp1-2995-a-k3hwlt.herokuapp.com/cms/pages/448/edit/
Test Mobilize page: https://foundation-s-tp1-2995-a-k3hwlt.herokuapp.com/en/mobilize_test/
```
Login: admin
Password: nz!hKHauo5uj4QKi
```
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2995

**Main Changes:**
- Option to choose the `media type` to use in Hero Section (_image_ or _video_). **_Default_** image to not affect existing pages.
-  New field for hero `video`. 
- Template and style settings for video.

**Changes in Models:**
- Adding a new `hero_media_type` field as another model field and used to _**choose**_ the media type to show.
- Adding a new `hero_video` field as another model field for the video.

## Images 
- Mobilize section with _**current**_ image option
image

- CMS with video as media type option
image

- Side by side video. Aspect-ratio 16:9 like other current video sections.
image

- Top to Bottom video
image


adalcima:
@mmmavis @robdivincenzo  Since there was a style adjustment for video in this PR, the `hero__media-image` class is indeed not being used, so I have removed it from the template. Reference in previous [PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14464).

sabrinang (changes requested):
For the side by side hero option, having the ability to change video aspect ratio (currently only 1:1 default) to have 16:9 option would be great as we don't always want to crop the video or have a video that fits that size. 

The video also sits a bit high (a bit too close to the navigation) so I would add some space in between the elements as well.

adalcima:
Thanks @sabrinang, updated to keep the side-by-side video `aspect ratio`, as in other existing sections like `video_block` and `video_panel`. Regarding the spacing, there was no change. I think my previous screenshot was taken with a bit more scrolling :smiling_face_with_tear: , so I updated it, as well as the review app ^^.

danielfmiranda (commented):
Hi @adalcima this is looking great! Thanks for taking this work on. I've tested out the functionality on the review app and everything seems to be working well.

I do have one question though: 

If I recall correctly, I believe that the team had mentioned that we want our videos to be hosted on Vimeo whenever possible. Though since the ask for this video is to be a custom shape, I am not sure whether that would be possible via iframe. 

Tagging @mmmavis and @robdivincenzo so they could chime in 👍 

If you would like to see an example of how we are currently utilizing vimeo videos, check out [video_block.py](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/blocks/video_block.py#L11-L15). However, as mentioned this is utilizing vimeos iframe embed functionality so this might not work with styling the container to match the designs.

If we do want to use vimeo videos and don't want use an iframe, I would suggest checking out the legacy site's "[Looping Video Block](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/legacy_apps/wagtailpages/pagemodels/customblocks/looping_video_block.py#L5-L12)".

This is something that I implemented. In short, we take a vimeo-hosted video file URL, and render it in a `<video>` element that gives us control over styling, rendering, playback, etc. 

If this is something you would like to explore let me know and I would be happy to work together or answer any questions!

adalcima:
Hi @sabrinang,  Thanks for your feedback. 

Since these PR changes are related to **_adding the video_** option to the `Hero Section` for use on the "**_Mobilize_**" (Community/Activation) page, what do you think about working on these adjustments related to the _current_ behavior of **_images_** in the `Hero Section` in separate tickets. I'm just not sure if we should treat them as bugs or features and the required priorities. I'll touch base with Simon and continue working on the video observations.

Tasks we could address in separate tickets.
```
1. Hero Section - Side by Side doesn't have a background, so this field is confusing when shown.
Task: Find a way to hide/restrict the "Hero background color" field when the variant is Side by Side.
```
```
2. Hero Section - Top to Bottom crops a 1:1 ratio image into 16:9, which is not how we want it. We can put some help text in the description for users. This hero image is 16:9.
Task: Consult with design if we should keep the 1:1 ratio or display help text.
```
```
3. Hero Section - Additionally, can the image not have a corner round mask? Marketing will often export images with their own shape masks.
Task: (Adal) We need to clarify and try to find out if a "corner round mask" is being applied to the images.
```
And I have a question about the "**Spacing**" (between the navigation and the hero area). 
**Do we only need this change just for "**Mobilize**" or is it required for any other page that uses a "`Hero Section`"? For reference, the "**_Make_**" page, which also uses a "Hero Section," has a similar spacing._**

Screenshot from 2025-08-25 13-55-06

adalcima:
> Hi @adalcima this is looking great! Thanks for taking this work on. I've tested out the functionality on the review app and everything seems to be working well.
> 
> I do have one question though:
> 
> If I recall correctly, I believe that the team had mentioned that we want our videos to be hosted on Vimeo whenever possible. Though since the ask for this video is to be a custom shape, I am not sure whether that would be possible via iframe.
> 
> Tagging @mmmavis and @robdivincenzo so they could chime in 👍
> 
> If you would like to see an example of how we are currently utilizing vimeo videos, check out [video_block.py](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/blocks/video_block.py#L11-L15). However, as mentioned this is utilizing vimeos iframe embed functionality so this might not work with styling the container to match the designs.
> 
> If we do want to use vimeo videos and don't want use an iframe, I would suggest checking out the legacy site's "[Looping Video Block](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/legacy_apps/wagtailpages/pagemodels/customblocks/looping_video_block.py#L5-L12)".
> 
> This is something that I implemented. In short, we take a vimeo-hosted video file URL, and render it in a `<video>` element that gives us control over styling, rendering, playback, etc.
> 
> If this is something you would like to explore let me know and I would be happy to work together or answer any questions!

I'll definitely check it out as an alternative.  This approach is similar to what was discussed in [this PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14451#pullrequestreview-3108029287) regarding the use of `model fields`. In this case, I'm using the `wagtailmedia` module, which we're currently using, as a base ([example implementation](https://github.com/torchbox/wagtailmedia)). 


Thank you very much, @danielfmiranda for reviewing this approach, it would be great to have your input @mmmavis and @robdivincenzo.

robdivincenzo:
Thanks for all your work on this enhancement @adalcima! I'm reviewing the approach and the feedback (in particular editor ux from @sabrinang and block examples from @danielfmiranda). I wonder if the `hero` section scope has grown enough to warrant its own streamfield like the `body` section? It seems like it now takes one of two distinctly different sets of fields 🤔.

To refactor would introduce some complexity and I'm sure a data migration, but might be a better ux in the end by allowing the user to select and see fields for the option they want. We might also be able to re-use or extend existing streamfield blocks (i.e. ImageBlock and the video blocks Daniel mentioned). Curious to hear more thoughts on the approach or feasibility! cc @mmmavis

adalcima:
PR closed, as another solution was implemented in this [PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/15773).

PR #14482, opened 2025-08-18
Add Quote component styling option for bolder article use

adalcima:
# Description

This PR is to add a styling option to the Quote component for bolder article use, as it requires a specific styling setting to allow larger and bolder fonts ([figma link as a reference](https://www.figma.com/design/4vdZ9mWyEwOz1D6Z7x9pqr/%C2%A0%E2%93%82%EF%B8%8F-mozillafoundation.org-MVP?node-id=13369-29584&m=dev)). 

Link to sample test page: https://foundation-s-tp1-3064-q-vhtge4.herokuapp.com/en/
```
Login: admin
Password: 6@t!FGL1L3RWxn#i
```
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-3064

**Main Changes:**
- Added option to have a larger and bolder font style using the existing styles: `larger` and `bold`.
The `Large` option allows us to use a larger font. The `Emphasis` option allows us to use a larger and bolder font.

**Changes in Models:**
- New field quote_style as ChoiceBlock to allow us to choose the style type: **_Regular_**, **_Large_**, **_Emphasis_** (`Regular` by default).

## Images 
- Options in CMS
image

image

- FE styling
image

### DIFF
--- a/foundation_cms/static/scss/blocks/hero_accordion/_base.scss
+++ b/foundation_cms/static/scss/blocks/hero_accordion/_base.scss
@@ -48,7 +48,8 @@ $details-section-padding: (
 );
 $details-subsection-gap: (
   small: 0.5rem,
-  large: 4rem,
+  large: 1rem,
+  xlarge: 4rem,
 );
 $corner-radius: (
   small: 1.75rem,
@@ -176,7 +177,7 @@ $image-hover-scale: 105%;
     margin-top: 0;
     margin-bottom: rem-calc(8);
 
-    @include breakpoint(large up) {
+    @include breakpoint(xlarge up) {
       margin-bottom: rem-calc(21);
     }
   }
@@ -192,7 +193,7 @@ $image-hover-scale: 105%;
       $header-font-family
     );
 
-    @include breakpoint(large up) {
+    @include breakpoint(xlarge up) {
       @include apply-text-style(
         $header-styles,
         "h5",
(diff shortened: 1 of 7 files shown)


danielfmiranda (approved):
Looks good to me! Amazing work @adalcima 👍 approved!

mmmavis:
Hi @adalcima , 👋🏻 Just dropping by to check on something - is this PR meant to give the quote blocks on NP Articles their own unique style compared to the rest of the site? If that's the case, we could actually handle this through the theming layer (like what Rob showed us this morning!). I'm not super familiar with all the details of this ticket, so it might be worth double-checking with the designers to see if they are looking to give users different style options to choose from in the Quote block itself, or would they prefer the styling to automatically come from the theme?

adalcima:
Hey @mmmavis thank you, I would have to explore that alternative with `theming layer,` but maybe it is an option too, maybe something more exclusive for **NP**. This current implementation would in effect enable users to have **_different styles_** to choose from in the `Quote` block itself. I thought this approach would give us flexibility and a slight impact, sorry for that, but it would be good to hear your point of view @nancyt1 @sabrinang. Thanks :raised_hands: !

mmmavis:
Thanks so much for walking me through your thinking on this, @adalcima !

I noticed this PR has already been merged. Would you mind reverting it for now? We typically need designer approval for changes like this, and I think their input would be super valuable here.

We're always trying to find that good balance between giving users flexibility and maintaining our brand consistency. It would great to loop in designers and get their thoughts on the best approach!

adalcima:
Thanks @mmmavis, the revert PR here https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14496

nancyt1:
I think it would be great if this bold style can be part of the NP theming layer but if that's a lot, I don't mind it being used more widely. But CMS wise, it looks great!

PR #14518, opened 2025-08-27
Theming Layer Improvements

robdivincenzo:
# Description

This PR connects the theming layer and fixes some initial issues with it for page types and blocks.

To test and use theming layer:

1. Note that existing/default/non-themed pages should not be affected
2. Edit a redesign page you wish to theme.
3. On the edit page screen navigate to the page Settings tab and select Nothing Personal.
4. Save page and view
5. Inspect the page source and note the body tag should now have a "nothing_personal" class name to enable styling
6. You should also note a new `base.html` template loads via `/tempaltes/base/themes/nothing_personal/base.html` with template block inheritance for nothing personal logos in the precontent and postcontent sections.
7. Subpages from this page should also automatically inherit without needing to set the "theme" field each time.
8. To overwrite a page type template, copy the template to `/templates/patterns/pages/themes/nothing_personal`
9. To overwrite a block template, copy the template to `/templates/patterns/blocks/themes/nothing_personal`
10. Note that non-overwritten blocks simply just grab the default template

One more comment, I've since merged in the code and there seems to be a bit of a confusion getting introduced with the templates/patterns/pages/nothing_personal directory containing the article templates, while templates/patterns/pages/themes/nothing_personal contains themed page types for nothing_personal. We should think of an intuitive way to clear this up since having two similarly named directories could cause confusion even though they do serve different purposes.


adalcima:
Hi @robdivincenzo, thanks for working on this implementation.

Just one question though:

I see that we would have something like `blocks/themes/nothing_personal/` to overwrite a block template. It seems like with this approach, we could simply copy blocks (even if they exist in `blocks/themes/default/`) that require a particular theme or style, 'adapt' that style, and automatically see that _theme_ applied to **subpages** based on the new `nothing_personal/base.html` that use that block?

PR #14526, opened 2025-08-29
Add Quote component styling option for bolder article use

adalcima:
# Description

This PR is to add a styling option to the _**Quote**_ component for bolder Article use, as it requires a specific styling setting to allow larger and bolder fonts ([figma link as a reference](https://www.figma.com/design/4vdZ9mWyEwOz1D6Z7x9pqr/%C2%A0%E2%93%82%EF%B8%8F-mozillafoundation.org-MVP?node-id=13369-29584&m=dev)). Our `Quote` block template currently uses a regular `.quote-text-regular` style option.

**Main Changes:**
- A new style option `.quote-text-emphasis` has been added to allow larger and bolder fonts. The size is based on our current `larger` setting for quote component.

**Theming layer:**
- This approach is based on the theming layer implementation from this [PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14518), overriding the Quote component block template with the new style required for NP Articles.

Link to sample test page: https://foundation-s-tp1-3064-a-apvbcp.herokuapp.com/cms/pages/448/edit/

```
Login: admin
Password: kt(b63q+XI@sQ8R_
```
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-3064

### How to test:
- Create or edit a General page.
- Add a new Quote component block.
- Go to settings and select "Nothing Personal" as the Theme to use. Yo should see the new style applied to the Quote component.

## Images 
- Currently with `default` theme
image

- Applying theme for `Nothing Personal`
image


foundation_cms/templates/patterns/blocks/themes/nothing_personal/quote_block.html
adalcima, line 1: I've added this new file following the new structure for theming layer integration - work in PR(#14518 )

kevdmoz (approved):
looks good to me! @adalcima maybe add @nancyt1 as a reviewer too

sabrinang (approved):
I see the quote block set in _quote style_ for general pages and _quote emphasis style_ appearing on themed Nothing Personal pages work and it switches by default.

I added quote emphasis style in the design system to be used. @nancyt1 I applied it to where I saw it being used but you can double check.

This PR looks good.

**A question for general quotes treatment:**
Currently it looks like quote blocks use the quote (40px) style for general pages. Is it true we no longer use the _Quote Large style_ for general pages then? So I understand if we need this style in the design system anymore or not

foundation_cms/templates/patterns/blocks/themes/nothing_personal/quote_block.html
```diff
@@ -0,0 +1,10 @@
+<div class="quote-block">
+  <figure>
+    <blockquote>
+      <p class="quote-text-emphasis">“{{ value.quote }}”</p>
+    </blockquote>
+    {% if value.attribution %}
+      <figcaption>{{ value.attribution }}</figcaption>
+    {% endif %}
+  </figure>
+</div>
```
robdivincenzo: Hey Adal, great use of theming layer to toggle a class difference inside the template and this works perfectly! Since all we need is an optional class difference, one minor suggestion is that we can keep it a little more DRY and extend the template instead. What are your thoughts on doing something like:

```suggestion
{% extends "../default/quote_block.html" %}

{% block quote_classes %}quote-text-emphasis{% endblock quote_classes %}
```


And then in the `default/quote_block.html` we can make the class render it with something like:

```
    <blockquote>
      <p class="{% block quote_classes %}quote-text-regular{% endblock %}">“{{ value.quote }}”</p>
    </blockquote>
```

This way we can support themed version, but if the general html structure of the quote block changes we only have to modify it in one place.

robdivincenzo (changes requested):
Thanks so much @adalcima! One small idea before approval, but let me know what you think 😄

foundation_cms/templates/patterns/blocks/themes/nothing_personal/quote_block.html
adalcima: Hey @robdivincenzo , thanks for your comments :rocket: , I definitely think it's a better approach, it can help us make blocks maintainable and flexible. I've tested these changes locally and in the review app, and the functionality remains the same, so yeah we could implement this method :).

robdivincenzo (approved):
R+ @adalcima !!

PR #14531, opened 2025-09-03
Consolidate NP styling into theme layer

ramram-mf:
# Description

This PR consolidates the styling approach into the theming layer and removes the redundant template files from Nothing Personal pages.

UX-wise, the editor user will have to specifically select the Nothing Personal theme from the Wagtail Page editor form settings tab in a page or parent page to render the style correctly, in contrast with the template doing it automatically without that requirement.

The theme body classname was updated to replace underscore for dashes in addition to the CamelCase to kebab-case operation already performed (`nothing_personal` becomes `nothing-personal` to adhere to bem).

image

Link to sample test page: https://foundation-s-tp1-3132-c-svlwcr.herokuapp.com/en/this-is-an-article-page/
`admin2`/`3G0i7~ryQ-8&`
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3132)


mmmavis (changes requested):
Thanks @ramram-mf for the PR! I've tested it and have the following feedback:

1. Nothing Personal Article Page ✅ I was able to load the page without errors. After setting the theme to `Nothing Personal`, I could see that NP styling successfully applied on the page.
2. Nothing Personal Homepage ❌ I'm getting `TemplateDoesNotExist` error from the first line `{% extends "base.html" %}` in `foundation_cms/templates/patterns/pages/`. Perhaps we wanna update the path to `{% extends "../../../base/themes/nothing_personal/base.html" %}` now?
nothing_personal/home_page.html`
3. Topics Page ❌ I encountered the same error when accessing `/nothing-personal/topics/name-of-topic/`
4. A general question ❓ Is our goal to eventually move everything in `foundation_cms/templates/patterns/pages/nothing_personal` to `foundation_cms/templates/base/themes/nothing_personal/`

ramram-mf:
@mmmavis I fixed the remaining extends with `theme_base` so those templates will also extend the theme as expected. Using `{% extends "../../../base/themes/nothing_personal/base.html" %}` would break the theme feature since it is hard coded.

Regarding moving the template files, I think we should consider it. Having just the theme base in another path is not optimal. I think we can follow your advice on next steps here.

mmmavis (approved):
@ramram-mf Great call using `theme_base` instead of hardcoded path!

PR #14535, opened 2025-09-04
iframe embed streamfield block

robdivincenzo:
# Description

This PR adds a new streamfield block for embedding iframes from supported urls. It will enable editor ability to embed content from simplecast, but is general enough to support other applications. It is a slimmed down version of the legacy `iFrame` block, and refactored for the redesign codebase using Foundation.

Link to sample test page: https://foundation-s-tp1-3080-v-t6r0lc.herokuapp.com/en/
Related PRs/issues: TP1-3080

Login:
admin
+62WIG1fO2oTQY(z

To Test:
1) Review the test page
2) Note simplecast is successfully embedded with the 8 column width by default
3) Note you can also extend this to full width via the editor (for re-use on other pages or in other contexts as well!) 👍


foundation_cms/blocks/iframe_block.py
```diff
@@ -0,0 +1,34 @@
+from django import forms
+from wagtail import blocks
+
+from foundation_cms.base.models.base_block import BaseBlock
+
+
+class RadioSelectBlock(blocks.ChoiceBlock):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.field.widget = forms.RadioSelect(choices=self.field.widget.choices)
```
danielfmiranda: **Suggestion**: Would it make sense to move this into a separate file, maybe something like `blocks/constants.py`? As it seems like there is nothing too specific to the Iframe block (we pass our options via `choices`/`default` params) 

That way, if another block would like to use a radio select field, we can import it from `constants.py` instead of having to redefine it again.

robdivincenzo: Good call @danielfmiranda, I spun it off into the `blocks/common` directory so it's more visible for reuse later!

PR #14579, opened 2025-09-11
NothingPersonalProductPage "Products Mentioned" section

danielfmiranda:
# Description

This PR introduces the following changes/updates:

- A new `ProductMentioned` Orderable relation
  - This allows users to select up to 3 related NP Product Review pages, in order to populate the page's "Products mentioned" section.
- A new, basic component template for the "Products Mentioned" section. At the moment, this template just loops through the products mentioned found in the page context and renders their title.
- Updates the template `product_review_page.html` to render this new component if mentioned products exist in the page context.
- Copies over the `get_related_items` and `localize_queryset` methods from the legacy site. These are used to return the list of mentioned products.

# Screenshots

Products mentioned section in the CMS:
Screenshot 2025-09-10 at 20-49-29 Editing Nothing Personal Product Review Test1 - Wagtail

Products Mentioned Section (if mentioned products are set): 

Screenshot 2025-09-10 at 20-50-26 Test1 - Mozilla Foundation

Products Mentioned Section (if no mentioned products are set)

### DIFF
--- a/foundation_cms/nothing_personal/migrations/0008_productmentioned.py
+++ b/foundation_cms/nothing_personal/migrations/0008_productmentioned.py
@@ -8,7 +8,7 @@ from django.db import migrations, models
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("nothing_personal", "0006_nothingpersonalarticlepage_hero_image_and_more"),
+        ("nothing_personal", "0007_alter_nothingpersonalarticlecollectionpage_body_and_more"),
     ]
 
     operations = [


foundation_cms/nothing_personal/models/product_review_page.py
```diff
@@ -1,20 +1,88 @@
...
         context = super().get_context(request)
+        context["products_mentioned"] = self.localized_products_mentioned
+        return context
+
+    def get_preview_context(self, request, mode_name):
+        context = super().get_preview_context(request, mode_name)
+        context["products_mentioned"] = self.preview_products_mentioned
         return context
+
+    @cached_property
```
danielfmiranda, line 70: Thought a cached property decorator would be useful for these two, that way we only run these queries once per page visit! [Link to django docs](https://docs.djangoproject.com/en/5.2/ref/utils/#django.utils.functional.cached_property)

foundation_cms/nothing_personal/models/product_review_page.py
```diff
@@ -1,20 +1,88 @@
...
+    def localized_products_mentioned(self):
+        products_mentioned = NothingPersonalProductReviewPage.objects.filter(mentioned_in_pages__page=self).order_by(
+            "mentioned_in_pages__sort_order"
+        )
+        localized_products_mentioned = localize_queryset(products_mentioned, preserve_order=True)
+
+        return localized_products_mentioned.specific()
+
+    @cached_property
```
danielfmiranda, line 79: See comment above ^^

foundation_cms/utils.py
```diff
@@ -13,3 +14,123 @@ def get_default_locale():
         DEFAULT_LOCALE,
         DEFAULT_LOCALE_ID,
     )
+
+
+def get_related_items(
```
danielfmiranda, line 19: This function is copied over directly from the `/legacy_apps/` directory, that way we can delete that directory and continue using this util when the time comes.

foundation_cms/utils.py
```diff
@@ -13,3 +14,123 @@ def get_default_locale():
...
+
+    """
+    # Apply ordering if needed
+    if order_by:
+        queryset = queryset.order_by(order_by)
+
+    return [getattr(relation, related_item_field) for relation in queryset.select_related(related_item_field)]
+
+
+def localize_queryset(
```
danielfmiranda, line 36: See comment above ^^

foundation_cms/nothing_personal/models/product_review_page.py
```diff
@@ -1,20 +1,88 @@
...
 
     def get_context(self, request):
         context = super().get_context(request)
+        context["products_mentioned"] = self.localized_products_mentioned
+        return context
+
+    def get_preview_context(self, request, mode_name):
+        context = super().get_preview_context(request, mode_name)
+        context["products_mentioned"] = self.preview_products_mentioned
         return context
```
danielfmiranda, line 68: We need this `get_preview_context` method as well as the `preview_products_mentioned` property because `Orderable` relationships do not get updated until the page is saved. 

That means if a user is making changes to the order or selection of pages in the "Products Mentioned" section, those changes would not be updated in the preview panel!

PR #14608, opened 2025-09-15
Nothing Personal Homepage and Blocks

robdivincenzo:
# Description

This PR enhances the NP homepage to add various fields and streamfields needed, including a hero image section and body section with a new Text & Image Block (narrow). It adds the two column container to the homepage which will be used for the 50-50 section (text & social block left, newsletter block right). Still to do is the product carousel component. But could unblock some styling in the meantime.

This also moves `hero_title` to the `hero_image` mixin as it seems more appropriate with multiple uses. The text and social block is refactored from the previously merged who-we-are-section page field.

Link to sample test page: 
Related PRs/issues: TP1-3109


ramram-mf (approved):
Thanks Rob, great work as usual!

PR #14632, opened 2025-09-19
Add product review sections

adalcima:
# Description

Link to sample test page:
Related PRs/issues: #

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [ ] Is my code documented?

**Merge Method**


foundation_cms/nothing_personal/models/product_review_page.py
```diff
@@ -42,16 +50,27 @@ class NothingPersonalProductReviewPage(AbstractArticlePage, HeroImageMixin):
...
+        ("good_and_bad", GoodAndBadBlock()),
+        ("bottom_line", BottomLineBlock()),
+    ]
+
+    review_sections = StreamField(
+        review_sections_block_options,
+        use_json_field=True,
+        blank=True,
+        block_counts={
+            "what_you_should_know": {"max_num": 1},
```
adalcima: This is added to ensure that no more than one block is created per section in the CMS.

foundation_cms/nothing_personal/models/product_review_page.py
```diff
@@ -120,24 +119,65 @@ def get_context(self, request):
...
         return context
 
-    def get_anchor_sections(self):
-        """Generate anchor sections based on enabled sections"""
-        section_mapping = [
-            (self.show_what_you_should_know_section, "What You Should Know", "what-you-should-know"),
-            (self.show_reduce_your_risks_section, "Reduce Your Risks", "reduce-your-risks"),
-            (self.show_the_good_and_the_bad_section, "The Good and The Bad", "the-good-and-the-bad"),
-            (self.show_the_bottom_line_section, "The Bottom Line", "the-bottom-line"),
+    def get_ordered_review_sections(self):
```
adalcima: This function is used for some features such as the order of the Newsletter signup and the presence of products mentioned, as these impact how sections are displayed.

foundation_cms/templates/patterns/blocks/themes/default/product_review_bottom_line_block.html
```diff
@@ -0,0 +1,17 @@
+{% if page.products_mentioned %}
```
adalcima, line 1: ⁨⁨Products are selected manually. [If not selected, hide the section](https://www.figma.com/design/4vdZ9mWyEwOz1D6Z7x9pqr/%C2%A0%E2%93%82%EF%B8%8F-mozillafoundation.org-MVP?node-id=14517-4075&m=dev).

robdivincenzo:
Hey @adalcima, thanks for your work on this. It does work nicely, and I particularly like how the anchor nav is more dynamic now rather than a checklist! My only question is if there was a reason to pivot to streamfield blocks here vs. using page fields? My concern here would be that the sections should be in an ordered/fixed position, but with streamfield blocks that isn't always the case as the editor is free to move them around. I see that you've created a helper method that will pass an ordered version of those streamfield blocks to the frontend for rendering (which does work nicely!), but I think it might be confusing for editors if this streamfield functions differently than other streamfields on the site.

adalcima:
Excellent observation, it was really to allow the editor to organize those sections in the future. But since it's not a current requirement (since sections are considered `fixed`), I agree with you that it can be visually confusing. I will update this approach with this recommendation. Thanks @robdivincenzo

foundation_cms/nothing_personal/models/product_review_page.py
adalcima: This feature is no longer needed and has been removed.

adalcima:
@robdivincenzo approach and review app updated, let me know your thoughts. Thanks so much!

PR #14658, opened 2025-09-24
Product Review Scoring Field

robdivincenzo:
# Description

This PR adds a plaintext field for product scoring.

Link to sample test page: https://foundation-s-tp1-3189-a-05adhe.herokuapp.com/en/np/np-product-review/
Related PRs/issues: TP1-3189

Login:
admin
@&pfQPWxi@27k7th

# To Test
1) View the sample test page
2) Confirm new plain text field for privacy scoring


adalcima:
Hey @robdivincenzo thanks for adding this new field. 
It looks good, I've tested the review app and can see the new field. I was just wondering if, post-MVP, we should consider an alternative to having the score be auto-generated or based on some parameter.

robdivincenzo:
Thanks @adalcima, yes we definitely will! Stakeholder suggested they need time to figure out what they want to do with this field, but that a plain text for MVP would be most flexible in the short term.

PR #14691, opened 2025-09-29
new script to reassign site hostnames after a prod DB copy on the staging site

danielfmiranda:
This PR adds a new script which is to be run by `mofo-cron` after the "copy prod DB to the staging site" script has finished running. 

Heres a summary of what the script does:

- Takes two environment variables:
  - `PROD_HOSTNAMES` (EX: "`mofo.net`, `mozfest.mofo.net"`)
  - `STAGING_HOSTNAMES` (EX: "`staging.mofo.net`, `mozfest.staging.mofo.net`")
- The script then loops through the `PROD_HOSTNAMES` variable, and for every hostname in the list, will find the corresponding site on staging, and replace it with the hostname that shares the same index in `STAGING_HOSTNAMES`.
- If the script cant find one of the hostnames (which is unlikely as these pretty much never change), then an error is printed and it moves to the next hostname.

Please note that this is part 1 of 2, and we will need to update `mofo-cron` to run this script once this PR has been merged. (PR for that coming soon!)


robdivincenzo (approved):
R+ @danielfmiranda! Thanks for taking this on, this will make life so much better lol

PR #14710, opened 2025-10-02
Add `CustomImageBlock`, update `TextMediaBlock` in order to support mp4 url video sources in NP Homepage

ramram-mf:
# Description

This PR adds a `CustomMediaBlock` with a `content` choice field  dropdown with `Image` and `Video` options in order to enable video content as source. It also refactors `TextImageBlockNarrow` into `TextMediaBlock`.

custom-media-block

*Updated `TextMedia` block with video*

text-media-block-cms

*Wagtail CMS UI for the `TextMediaBlock`* 

Credentials: `admin`/`+p13KCvtfn36h6jz`
Link to sample test page: https://foundation-s-tp1-3204-e-dcmnu7.herokuapp.com/en/nothing-personal/
CMS edit test page: https://foundation-s-tp1-3204-e-dcmnu7.herokuapp.com/cms/pages/448/edit/
Regressions test for affected ui components: https://foundation-s-tp1-3204-e-dcmnu7.herokuapp.com/en/ui-sandbox/
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3204)

### DIFF
--- a/foundation_cms/templates/patterns/admin/media_block_form.html
+++ b/foundation_cms/templates/patterns/admin/media_block_form.html
@@ -14,11 +14,7 @@
         {% if child.block.name == "image" %}
             <div class="w-field" data-field data-contentpath="image">
                 <div data-media-target="field" data-condition="image">
-                    {% if child.block.label %}
-                        <label class="w-field__label" {% if child.id_for_label %}for="{{ child.id_for_label }}"{% endif %}>
-                            {{ child.block.label }}{% if child.block.required %}<span class="w-required-mark">*</span>{% endif %}
-                        </label>
-                    {% endif %}
+                    {# The ImageChooser renders its own label; avoid duplicating it here. #}
                     {{ child.render_form }}
                 </div>
             </div>


ramram-mf:
> Hi @ramram-mf!
> 
> From a visual perspective everything looks good to go. I did notice though that in the CMS an editor can select "image" but not upload an image (same with video) and not receive an error. I'm wondering if we can make it so that if a user selects "image" or "video" but doesn't upload the correct asset type they would get an error when they try to save or publish. Thoughts?

Yes I will implement validation for those fields, and I'm also looking into using the Stimulus library to show/hide relevant fields but that is getting a bit more complex. Will update with validation rules. Thanks @beccaklam .

ramram-mf:
> Hi @ramram-mf!
> 
> From a visual perspective everything looks good to go. I did notice though that in the CMS an editor can select "image" but not upload an image (same with video) and not receive an error. I'm wondering if we can make it so that if a user selects "image" or "video" but doesn't upload the correct asset type they would get an error when they try to save or publish. Thoughts?

Validation messages added for both content types:

image

image

foundation_cms/templates/patterns/admin/media_block_form.html
ramram-mf, line 1: This one I don't know where to place. As an admin form I think it doesn't belong here but couldn't figure out a better location yet. Any thoughts @mmmavis ?

foundation_cms/static/js/controllers/media_controller.js
```diff
@@ -0,0 +1,64 @@
+export default class extends window.StimulusModule.Controller {
+  static targets = ["field"];
+  static values = { triggerField: String };
+
+  connect() {
+    this.setupTriggerField();
+    this.updateFieldVisibility();
+  }
+
```
mmmavis, line 65: I'm new to Stimulus, but according to [the docs](https://stimulus.hotwired.dev/reference/lifecycle-callbacks#disconnection), I think we should add a `disconnect()` method to this controller to properly clean up event listeners when the controller is removed from the DOM. This would prevent potential memory leaks, especially in dynamic forms.

foundation_cms/static/js/controllers/media_controller.js
```diff
@@ -0,0 +1,64 @@
...
+          t.name === this.triggerFieldValue ||
+          t.name.endsWith(`-${this.triggerFieldValue}`)
+        ) {
+          this.triggerElement = t;
+          this.updateFieldVisibility();
+          this.triggerElement.addEventListener("change", () =>
+            this.updateFieldVisibility(),
+          );
+        }
+      });
```
mmmavis: I noticed the controller could end up attaching duplicate event listeners since it doesn't check if one's already attached. Would it make sense to add a check before attaching listeners, or maybe track which elements already have listeners attached?

foundation_cms/templates/patterns/admin/media_block_form.html
mmmavis: Thanks for flagging this. I wonder something like `foundation_cms/templates/patterns/cms/media_block_form.html` makes sense? Or use `admin` instead of `cms`? What do you think?

mmmavis (changes requested):
Hi Ram, thanks for exploring Stimulus and working to improve the editor experience. As mentioned in Slack, I'll defer to full stack devs for the backend review. For the frontend, I've left some comments on the controller JS.

I have one broader concern - the current implementation relies heavily on specific DOM structures and naming conventions that Wagtail generates. If Wagtail updates their HTML output or field naming patterns, this could break. Maybe we could add guardrails or documentation to the module and its methods?

foundation_cms/static/js/controllers/media_controller.js
ramram-mf: Good catch. I refactored a bit the controller to improve event handling by delegating just the one listener.

foundation_cms/static/js/controllers/media_controller.js
ramram-mf: I've now implemented the `disconnect()` method to adopt the event delegation pattern foreseeing multiple streamblocks being added/removed from the page body in the editor experience. Something like this:

```js
disconnect() {
    // Remove the single delegated listener. This is the only listener we add
    // on connect, so removing it here prevents event-handler leaks when the
    // controller is disconnected (for example, when the block is removed).
    if (this._onDelegatedChange) {
      this.element.removeEventListener("change", this._onDelegatedChange);
    }

    // Clear any stored references so the controller is fully detached.
    this.triggerElement = null;
  }
```

robdivincenzo (changes requested):
Really great work on this @ramram-mf, and an awesome proof of concept for the optional fields! But I do think there will be an issue in the current state without a data migration. Since image is being renamed to media, those blocks are getting wiped out.

For example, if you were to import a staging db locally and view the meet mozilla page you'll see image blocks used in a bunch of places. Then checking out this branch and running migrations, you'll see the blocks wiped out since image blocks don't exist.

Staging:
staging

After running migrations:
after running migrations

So we either need a data migration to migrate all image block / text and image block / etc. instances to the new media version, or make it truly additive in that we leave image blocks as they are and just add a new separate option for media block. Wagtail will sometimes keep the data all fine if the field names and data structure are the same, so there might be another option there.

foundation_cms/templates/patterns/blocks/themes/default/text_image_block.html
```diff
@@ -26,7 +26,6 @@
   {# Right: Image (2/3) #}
   <div class="text-image-block__image">
     {% if value.image and value.image.image %}
-      {% image value.image.image original as img %}
```
ramram-mf, line 29: This was loading the image resource redundantly.

ramram-mf:
> Hi @ramram-mf thank you for implementing those changes! The validation looks good, I just noticed that for some reason when you select "Image" the word image shows up twice. Is there a way to make it so that it only shows up once? 
I also noticed that an editor can upload any url for the video url and save it with no validation error. It won't work but I wonder if there's quick fix where this would trigger a validation error and let the editor know they have to upload a vimeo url. If there's no quick fix though I think we can just go ahead and ship this since the editor will still know something is not right since no video will be showing. We also have the help text that mentions needing to upload a vimeo url.

Hey @beccaklam , good catch. I added a simple validator to enable only vimeo mp4 links.

image

And updated the field labels to only display one image label.

image

PR #14736, opened 2025-10-07
Updated abstract pages `translatablefields`

danielfmiranda:
# Description

The field `body` used to be set as a translatable field on the abstract page models, however, this is not best practice. The reason being is that sometimes pages can overwrite the body field to be `None`, a good example would be the `NothingPersonalProductReview` page. When that is done, an attempt to translate a product review page will be met with a 500 error saying that the page model has no field `body`, yet, since this has been assigned as a translatable field on the models that it inherits from, it is still looking for it! 

Ive removed the field from the abstract page models, as more concrete models such as `ArticlePage`, etc, should be defining it in their translatable fields anyways.


adalcima (approved):
good catch @danielfmiranda :rocket:

PR #14744, opened 2025-10-09
Update NP Homepage translatable fields

robdivincenzo:
# Description

This PR removes a second instance of the `body` field from the NP homepage translatable fields. This resulted in a duplicate error since this inherits from the `AbstractHomePage.translatable_fields` which already contains a `body` field in the translatable fields

Link to sample test page:
Related PRs/issues: TP1-3234


robdivincenzo:
Resolved via #14736

PR #14791, opened 2025-10-15
Add Visual Regression Tests workflow back

mmmavis:
We temporarily removed this file yesterday so Percy wouldn't be a bottleneck for our deploy pipeline. Adding this file back now.


ramram-mf (approved):
Classic Percy.

PR #14799, opened 2025-10-15
Make `NewsletterSignUp` translatable

danielfmiranda:
# Description

This PR updates the `NewsletterSignUp` snippet to make it translatable. It also updates the templates that use the snippet with `.localized` so it will return the localized version when applicable.

# Links to review app
 English: https://foundation-s-make-newsl-x9llz6.herokuapp.com/en/
 Spanish: https://foundation-s-make-newsl-x9llz6.herokuapp.com/es/

# Screenshots

Newsletter Signup Block (In English):
Screenshot 2025-10-15 at 15-54-59 Welcome to Mozilla Foundation - Mozilla Foundation

Footer Newsletter Sign Up Form (In English):
Screenshot 2025-10-15 at 15-54-13 Welcome to Mozilla Foundation - Mozilla Foundation

Newsletter Signup Block (In Spanish):
Note: I did not translate the sign up button

Screenshot 2025-10-15 at 3 52 07 PM

Footer Newsletter Sign Up Form (In Spanish):
Note: I did not translate the sign up button
Screenshot 2025-10-15 at 15-53-21 Te damos la bienvenida a la Fundación Mozilla - Mozilla Foundation

### DIFF
--- a/foundation_cms/snippets/migrations/0005_make_newslettersignup_translatable.py
+++ b/foundation_cms/snippets/migrations/0005_make_newslettersignup_translatable.py
@@ -29,7 +29,7 @@ def backfill_newslettersignup_locale_and_key(apps, schema_editor):
 class Migration(migrations.Migration):
 
     dependencies = [
-        ("snippets", "0003_newsletter_add_cta_description_and_rename_cta_text"),
+        ("snippets", "0004_newsletterunsubscribe"),
     ]
 
     operations = [


foundation_cms/snippets/migrations/0005_make_newslettersignup_translatable.py
danielfmiranda, line 1: Since we already had some instances of this model in the DB (about 4) before making it translatable, just running`inv makemigrations` would return this message: 
```
It is impossible to add a non-nullable field 'locale' to newslettersignup without specifying a default. This is because the database needs something to populate existing rows.
Please select a fix:
 1) Provide a one-off default now (will be set on all existing rows with a null value for this column)
 2) Quit and manually define a default value in models.py
Select an option:
```

However, selecting option 1 and adding the English locale would return this error: 

```
DETAIL:  Key (translation_key, locale_id)=(<UUID>,  1) is duplicated.
```


This migration avoids this error but achieves what we need by first adding the new fields as nullable, then looping through the existing English entries to assign each a unique `translation_key` and set their `locale` to English before making the fields required and adding the uniqueness constraint.

The review app is using a copy of the staging DB data so I can confirm it works!

robdivincenzo: Thanks for confirming w/ staging data!

PR #14832, opened 2025-10-22
Replace "should_wrap_block" filter with a new block decorator "skip_default_wrapper_on"

mmmavis:
# Description

In this PR, I have replaced `should_wrap_block` template filter with a new decorator `@skip_default_wrapper_on()`

The previous approach with `should_wrap_block` template filter had a hardcoded list of block types in the template layer. This PR moves that logic to the block definitions themselves using a decorator. This makes it clearer which blocks manage their own wrapper structure. This flexibility will also enable us to apply page-specific layout/style when we need to.

## What changed
- Created `@skip_default_wrapper_on()` decorator
- Applied it to all blocks previously in `NO_WRAP_BLOCKS`
```python
       "spotlight_card_set_block",
        "portrait_card_set_block",
        "featured_card_block",
        "callout",
        "product_review_carousel_block",
 ```
- Updated `streamfield.html` template to use block context
- Added `streamfield-block-wrapper` div with `data-block-type` attribute for better CSS/JS targeting
- Removed the old filter

## To test

- checkout this branch and run it locally as usual
- go to the homepage (there should not be any visual changes)
- inspect the source code of `<main>` and verify each Streamfield block now has a wrapper div like below
image

---

Related PRs/issues: [Jira TP1-3295](https://mozilla-hub.atlassian.net/browse/TP1-3295) / GitHub #14831

### DIFF
--- a/foundation_cms/templates/patterns/components/streamfield.html
+++ b/foundation_cms/templates/patterns/components/streamfield.html
@@ -7,13 +7,13 @@
     <div class="grid-container">
       <div class="grid-x">
         <div class="cell{% if classnames %} {{ classnames }}{% endif %}">
-          {% include_block block with page_type_bem=page_type_bem %}
+          {% include_block block %}
         </div>
       </div>
     </div>
   {% else %}
     <div class="full-bleed-block-wrapper">
-      {% include_block block with page_type_bem=page_type_bem %}
+      {% include_block block %}
     </div>
   {% endif %}
 
--- a/foundation_cms/templates/patterns/components/streamfield.html
+++ b/foundation_cms/templates/patterns/components/streamfield.html
@@ -1,15 +1,20 @@
 {% load wagtailcore_tags streamfield_tags %}
 
 {% for block in streamfield %}
-  {% if block.block_type|should_wrap_block %}
+  {% get_block_context_value block page 'full_bleed_wrapper' as is_full_bleed %}
+
+  {% if not is_full_bleed %}
     <div class="grid-container">
       <div class="grid-x">
         <div class="cell{% if classnames %} {{ classnames }}{% endif %}">
           {% include_block block with page_type_bem=page_type_bem %}
-        </div>
+        </div>a
       </div>
     </div>
   {% else %}
-    {% include_block block with page_type_bem=page_type_bem %}
+    <div class="full-bleed-block-wrapper">
+      {% include_block block with page_type_bem=page_type_bem %}
+    </div>
   {% endif %}
+
 {% endfor %}
--- a/foundation_cms/templates/patterns/components/streamfield.html
+++ b/foundation_cms/templates/patterns/components/streamfield.html
@@ -8,7 +8,7 @@
       <div class="grid-x">
         <div class="cell{% if classnames %} {{ classnames }}{% endif %}">
           {% include_block block with page_type_bem=page_type_bem %}
-        </div>a
+        </div>
       </div>
     </div>
   {% else %}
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: foundation_cms/templates/patterns/components/streamfield.html [hunk 1]
```diff
@@ -1,15 +1,20 @@
...
 
 {% for block in streamfield %}
-  {% if block.block_type|should_wrap_block %}
+  {% get_block_context_value block page 'full_bleed_wrapper' as is_full_bleed %}
+
+  {% if not is_full_bleed %}
     <div class="grid-container">
       <div class="grid-x">
         <div class="cell{% if classnames %} {{ classnames }}{% endif %}">
-          {% include_block block with page_type_bem=page_type_bem %}
```
mmmavis, line 8: removed `page_type_bem` -- this is old dead code. I don't see `page_type_bem` used in any block template files

foundation_cms/templates/patterns/components/streamfield.html
```diff
@@ -1,15 +1,20 @@
...
+          {% include_block block %}
         </div>
       </div>
     </div>
   {% else %}
-    {% include_block block with page_type_bem=page_type_bem %}
```
mmmavis, line 13: removed `page_type_bem` -- this is old dead code. I don't see `page_type_bem` used in any block template files

foundation_cms/templates/patterns/components/streamfield.html
```diff
@@ -1,15 +1,20 @@
...
 
 {% for block in streamfield %}
-  {% if block.block_type|should_wrap_block %}
-    <div class="grid-container">
-      <div class="grid-x">
-        <div class="cell{% if classnames %} {{ classnames }}{% endif %}">
-          {% include_block block with page_type_bem=page_type_bem %}
+  {% get_block_context_value block page 'skip_default_wrapper' as skip_default_wrapper %}
+
+  <div class="streamfield-block-wrapper" data-block-type="{{ block.block_type }}">
```
mmmavis, line 6: I'm adding this wrapper div. This will make CSS/JS targeting easier

robdivincenzo:
Thanks @mmmavis ! This is a great step to clean things up. It's much clearer how/when wrappers are skipped. I also like the developer UX of a decorator where the allowed page types sit right above the class.

My only hesitation is that the decorator returns a new class and alters the method resolution order. That’s a little unorthodox for Django/Wagtail and could lead to unexpected surprises when other decorators/mixins get involved. Totally works now, but it isn't as predictable of a pattern.

A more conventional alternative is to push this into `BaseBlock` as a keyword argument. Since blocks inherit from `BaseBlock`, they'll all get this ability by default, and we just need to assign at block usage like this:

`("hero", HeroBlock(skip_default_wrapper=True))`

The main trade-off is losing the nice "page types listed above the class" dev UX. But it's more predictable and avoids the potential pitfalls. We also gain the flexibility to skip the wrapper on a per-StreamField basis rather than just a per-PageType basis. Happy to go whichever way the team prefers, just would lean toward a kwarg pattern personally. I have a [PR you can review](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/14837) as a quick example!

mmmavis:
Thanks @robdivincenzo !

Agreed that it’s better to go with the straightforward and conventional approach for this, and avoid introducing a decorator to prevent unnecessary complexity or potential risks given the unknowns in future specs. I’ll close this PR and open a new one to update `BaseBlock`!

PR #14836, opened 2025-10-24
Add Nothing Personal Product Collection Page

ramram-mf:
# Description

This PR implements the Nothing Personal Product Collection Page with a Product Card grid.

image

*Product review collection page grid in `large` viewport*

CMS credentials: `admin` / `k5rOJ+)VYk*3(HxU`
Link to sample test page: https://foundation-s-tp1-3160-p-mebrza.herokuapp.com/en/nothing-personal/product-reviews/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3160)


foundation_cms/blocks/product_review_carousel_block.py
```diff
@@ -28,7 +28,6 @@ class ProductReviewCardBlock(BaseBlock):
 
     class Meta:
         icon = "image"
-        template_name = "product_review_card_block.html"
         label = "Product Review Card"
```
ramram-mf, line 32: By removing the explicit template definition in the model, we can reuse the template file for the collection grid.

nancyt1:
@ramram-mf My cursor doesn't seem to change when I hover over it image or text even tho they are both links, is that just me? (maybe just me, but a clearer hover state would help anyway)

And could we add a hover state to the text as well? Like what Mavis already had on the "More reviews" section or the one you made for the homepage? 

Can you update spacing to be 1rem between title and text on mobile/tablet?
image

foundation_cms/static/scss/pages/nothing_personal/product_collection_page.scss
```diff
@@ -0,0 +1,119 @@
...
+    margin-bottom: 3.5rem;
+
+    @include breakpoint(large up) {
+      margin-bottom: 4.5rem;
+    }
+  }
+
+  .collection-products {
+    // Configuration variables
+    $image-hover-scale: 105%;
```
mmmavis: We now have a `image-hover-zoom` mixin for this! Let's use that to keep things DRY.

foundation_cms/blocks/product_review_carousel_block.py
mmmavis: I believe `template_name` is required or at least it's a convention we have in our codebase. Could you share more about why you think this line can be removed? cc @adalcima

foundation_cms/nothing_personal/models/product_collection_page.py
```diff
@@ -0,0 +1,50 @@
...
+    NothingPersonalProductReviewPage,
+)
+from foundation_cms.utils import get_default_locale, localize_queryset
+
+
+class ProductReviewCard:
+    """Simple object to mimic a block with product_review property"""
+
+    def __init__(self, product_review):
+        self.product_review = product_review
```
mmmavis: @ramram-mf Could you clarify what this new class is for?

cc @adalcima

foundation_cms/static/scss/pages/nothing_personal/product_collection_page.scss
```diff
@@ -0,0 +1,119 @@
...
+    @include breakpoint(large up) {
+      gap: map.get($row-gaps, large)
+        map.get(map.get($grid-breakpoints, large), gap);
+      margin-bottom: 4rem;
+    }
+
+    .product-review-card {
+      $small-config: map.get($grid-breakpoints, small);
+
+      flex: 0 0
```
mmmavis: @ramram-mf What do you think about using `grid` with `gap` instead, so we can avoid complex `flex-basis` calculations?

mmmavis (changes requested):

foundation_cms/blocks/product_review_carousel_block.py
ramram-mf: Yeah grid was in my mind too, got overly complex real fast with flex, refactored.

foundation_cms/nothing_personal/models/product_collection_page.py
ramram-mf: This class was a residual definition not needed anymore.

foundation_cms/static/scss/pages/nothing_personal/product_collection_page.scss
ramram-mf: Good catch, implemented the mixin and removed redundant declarations.

nancyt1:
@ramram-mf Can the product titles is not underlined by default but only on hover?

foundation_cms/nothing_personal/models/product_collection_page.py
ramram-mf: Restored the original template_name property since the error triggered by template parsing was now not showing.

adalcima (approved):
lgmt and test page has worked for me too, thanks @ramram-mf

PR #14837, opened 2025-10-24
Example to swap decorator to argument via BaseBlock

robdivincenzo:
# Description

This PR is a quick example of using kwargs on the `BaseBlock` instead of decorator. For alternative consideration.

Keeps feature within the model, all blocks inheriting from our BaseBlock inherit this ability automatically. Adds flexibility to not only control this on page types, but on different StreamFields within a single page type (i.e. controlling this on the `body` or the different `sections` on the [NothingPersonalProductReviewPage](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/nothing_personal/models/product_review_page.py))

# To Test
1) Check out branch / review code.


foundation_cms/base/models/base_block.py
```diff
@@ -24,6 +32,22 @@ def get_template(self, context=None):
...
+    def get_context(self, value, parent_context=None):
+        context = super().get_context(value, parent_context)
+        if self._skip_default_wrapper:
+            context["skip_default_wrapper"] = True
+        return context
+
+    def deconstruct(self):
+        """
+        Recommended when introducing custom arguments into the constructor. See:
+        https://docs.wagtail.org/en/stable/advanced_topics/customization/streamfield_blocks.html#handling-block-definitions-within-migrations
```
mmmavis, line 44: thanks for including link to Wagtail doc!

PR #14862, opened 2025-10-29
Enable SSO on Review Apps

robdivincenzo:
# Description

This PR will update the default site to a mofostaging.net domain and allow for new users to sign in via SSO and be automatically added as admins only on review app. If a staging DB is cloned down, they can still SSO, but won't be assigned a new role because they aren't new users.

This will also clean up a few things when people review a page from the dashboard it should stay on the review app vs. go to staging or production. It will also ensure and unify only the following URL pattern `REVIEW_APP_NAME`.`mofostaging.net` rather than the `herouapp.com` variant.

# To Test

1. Go to https://foundation-s-tp1-3139-r-e9f16i.mofostaging.net/cms/
2. Login via SSO
3. You should log in successfully with a new user account as an administrator
4. Try to go to the heroku app version of the site: https://foundation-s-tp1-3139-r-e9f16i.herokuapp.com
5. Notice you'll be forwarded to the mofostaging.net domain version
6. Confirm the mozfest and legacy site prefixes work as well
7. https://mozfest-foundation-s-tp1-3139-r-e9f16i.mofostaging.net
8. https://legacy-foundation-s-tp1-3139-r-e9f16i.mofostaging.net

### DIFF
--- a/foundation_cms/legacy_apps/static/js/common/template-js-handles/sticky-cta.js
+++ b/foundation_cms/legacy_apps/static/js/common/template-js-handles/sticky-cta.js
@@ -6,6 +6,8 @@ export default () => {
   let ticking = false;
   let elBurgerWrapper = document.querySelector(`.wrapper-burger`);
 
+  if (!elBurgerWrapper) return;
+
   let adjustNavbar = (scrollPosition) => {
     if (scrollPosition > 0) {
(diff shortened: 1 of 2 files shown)


adalcima (approved):
lgtm, thanks @robdivincenzo , I was able to log in using LDAP :rocket:

PR #14868, opened 2025-10-30
Site-wide Donate Banner Front-end

mmmavis:
**Review app link**: None. I had no luck with review apps after spending quite a bit of time on it 😭 🤷🏻. Please follow the instructions below to test this PR. Commit history should be helpful as well.
**Related PRs/issues**: [Jira TP1-3231](https://mozilla-hub.atlassian.net/browse/TP1-3231) / #14488 

(This PR has passed design review.)

Screen Shot 2025-11-03 at 22 24 51

Screen Shot 2025-11-03 at 22 27 25

### Changes included in this PR:
- implement donate banner frontend. There are two versions, which are essentially the same except for the type of buttons they include:
  - **LEGACY** version: with **close button** (see attached image no.1)
  - **PUSHDOWN** version: with **skip down button**. Scrolling down the page will also gradually hide the banner.  (see attached image no.2)
- introduce a new wrapper `<div class="main-content-wrapper">` in `base.html`. This is essential for the pushdown banner to work

### Changes filled as follow-up PR
  - remove extra styling fields from model (#14877)

# To Test

- check out this branch and run it locally

### Set up Donate Banner Snippet
- Create two versions of [Donate Banner snippets](http://test.example.com:8000/cms/snippets/snippets/donatebanner/?locale=en) - `legacy` style and `pushdown` style
  - You can just fill out `Name`, `Banner style`, `Foreground image`, and hit save.

### Set up Donate Banner Page
- Create a `Donate Banner Page` under `Redesign Homepage`
- Select the banner you want to show
- Publish the page
- Go to any page on the site, you should now see the banner shows up on the page. If you wanna verify this banner on legacy pages, you will need to copy staging or prod db locally first.

## To start an AB test:
 (Instructions copied from Daniel's other PR. Thanks, Daniel!)
  - Publish the page with what you would like to be the "Control" Donate Banner
  - Visit the edit page again, and change the donate banner to what you would like to be set as the "Variant" donate banner
  - At the bottom of the page, click "Publish and start AB test"
  - Add a name for the AB test
  - For the donate Banner Goal, select "Donate Banner Link Button click" from the dropdown menu
  - Set the sample size to what you would like, I usually use 10.
  - Click start AB test

PR #14877, opened 2025-11-04
Remove styling fields from DonateBanner

mmmavis:
This is a follow-up PR of PR #14868 to remove styling related fields we don't need. (I've confirmed with Becca about this decision). This is not a blocker to the Donate Banner work, but it would be good to land soon to avoid showing dummy/inactive fields to CMS users.

@ramram-mf @AdalbertoMoz I'm adding you both as assignees in case I don't have time to merge this PR before Wednesday. Thanks in advance! Note that the base branch might need adjustment (to `main`) depending on the state of its current parent branch/PR #14868

Related PRs/issues: [Jira TP1-3231](https://mozilla-hub.atlassian.net/browse/TP1-3231) / https://github.com/MozillaFoundation/foundation.mozilla.org/issues/14488


mmmavis:
Closing. I opened a new PR for this work #14917

PR #14891, opened 2025-11-10
Add Petition Flow backend functionality

adalcima:
# Description

Implements a petition flow for campaign pages with conditional form fields and social sharing functionality. Users can sign petitions, optionally donate, share on social media.

Link to sample test page: https://foundation-s-tp1-2340-p-xbvdoo.mofostaging.net/cms/pages/448/edit/
Related PRs/issues: https://mozilla-hub.atlassian.net/browse/TP1-2340

**Main Changes:**
- An app has been created in `/campaigns`, with models and templates necessary for the petition flow.
- A petition form based on legacy forms assembly has been added.
- Multi-state flow based on flow.

**How to test:**
1. Create a new Petition: Go to Snippets and search for Petitions (New)
2. Create and publish a campaign page in Redesign Homepage:  search for Campaign Page (New) 
3. Visit the campaign page and fill the Petition form

## Images
image

image

image

### DIFF
--- a/foundation_cms/campaigns/templates/campaigns/campaign_page.html
+++ b/foundation_cms/campaigns/templates/campaigns/campaign_page.html
@@ -2,11 +2,9 @@
 {% load wagtailcore_tags wagtailimages_tags static i18n %}
 {% get_current_language as LANGUAGE_CODE %}
 
-
 {% block extended_head %}
     {% include "campaigns/components/formassembly_head.html" %}
-{% endblock %}
-
+{% endblock extended_head %}
 
 {% block content %}
     <div class="campaign-page" data-page-id="{{ page.id }}" data-state="{{ state|default:'start' }}">
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -235,6 +235,7 @@ ignore="H017"
 "foundation_cms/templates/base/base.html" = "H031"
 "foundation_cms/templates/patterns/components/primary_nav.html" = "D018"
 "foundation_cms/templates/patterns/components/footer.html" = "D018"
+"foundation_cms/campaigns/templates/campaigns/components/formassembly_body.html" = "H021, H026"
 
 [tool.isort]
 profile = "black"


danielfmiranda (commented):
Hi @adalcima!

First thank you so much for taking this on! This PR looks great and is a good foundation for implementing the Petition flow on the redesigned site.

After reviewing the changes, it seems like the `CampaignPage` model and its functionality already cover nearly all the requirements of the new workflow. It’s great to see that, if needed, we could have this up and running quickly with just a few of the follow-up updates you mentioned in the code comments.

Here are a few broader notes and questions:

- **Campaigns app structure:** I think separating the `Campaigns` app into its own directory is an excellent idea. 
- **Inheritance and body blocks:** I noticed that the `CampaignPage` inherits from `AbstractBasePage`. I think that’s a great call — it avoids pulling in unnecessary fields from `GeneralPage`, like the hero section. That said, I believe stakeholders will want all the same body blocks that were available on `GeneralPage` to be available here as well. We should probably add a `body` field to `CampaignPage` and share the same set of available blocks as `GeneralPage`.
- **Block-based content sections:** Using StreamField blocks for `signed_content`, `share_content`, and `thank_you_content` is a really clever approach. I’m curious what you think about instead defining these as page fields — that way, editors can see all the default values right when creating a new page in the CMS. We could then organize these with headings and collapsible sections using Wagtail’s `MultiFieldPanel`.

I'll schedule a meeting with some of the other devs on Monday so we can talk over these questions/ any other work that still needs to be done, but this PR is looking very close to an R+.

Thanks and great job 👍

foundation_cms/campaigns/models/campaign_page.py
```diff
@@ -0,0 +1,198 @@
...
+        default="Would you like to support our work with a donation?",
+        help_text="Donation question shown after petition is signed",
+    )
+
+    signed_body = RichTextField(
+        default="Thanks so much for signing this petition asking "
+        "for transparency on data used to train AI tool! Can you "
+        "add a donation to support Mozilla's  non-profit work for trustworthy AI?",
+        help_text="Content shown after signing",
+    )
```
danielfmiranda, line 53: Hi @adalcima I think that this thank you message was specific to a legacy campaign that was related to AI. Can we check with stakeholders what the default thank you message should be?

foundation_cms/campaigns/models/campaign_page.py
```diff
@@ -0,0 +1,198 @@
...
+                petition_cta = self.cta
+
+        (default_locale, _) = get_default_locale()
+
+        default_articles = (
+            NothingPersonalArticlePage.objects.live()
+            .public()
+            .filter(locale=default_locale)
+            .order_by("-first_published_at")
+        )
```
danielfmiranda, line 185: According to the figma, I think we only need the 2 latest articles. Could we update this with a [:2] please?

foundation_cms/campaigns/templates/campaigns/campaign_page.html
```diff
@@ -0,0 +1,215 @@
+{% extends "base/base.html" %}
+{% load wagtailcore_tags wagtailimages_tags static i18n %}
+
+{% block content %}
+    <div class="campaign-page" data-page-id="{{ page.id }}" data-state="{{ state|default:'start' }}">
+
+        {% if medium == "email" %}
```
danielfmiranda: Just double checking, should this message show when users are sent to the petition via email?

foundation_cms/campaigns/templates/campaigns/campaign_page.html
```diff
@@ -0,0 +1,215 @@
...
+            }).catch(function() {
+                const textArea = document.createElement("textarea");
+                textArea.value = text;
+                document.body.appendChild(textArea);
+                textArea.select();
+                document.execCommand('copy');
+                document.body.removeChild(textArea);
+                alert('Link copied!');
+            });
+        }
```
danielfmiranda, line 200: **Kudos** A great starting point for the front end devs to add the "update html with a copied message" logic. Thanks @adalcima!

foundation_cms/campaigns/templates/campaigns/campaign_page.html
```diff
@@ -0,0 +1,215 @@
...
+                            'page_id': '{{ page.id }}',
+                            'state': '{{ state }}'
+                        });
+                    }
+
+                    console.log(`Button clicked: ${label} (${action})`);
+                });
+            });
+        });
+    </script>
```
danielfmiranda, line 221: **Kudos** thanks for adding the google tag manager event! I am thinking that we should probably remove the `console.log` statement though. Thanks!

danielfmiranda (approved):
Thanks for your patience and making all the updates @adalcima, approved! 🚀

PR #14927, opened 2025-11-17
Revert FA newsletter value

robdivincenzo:
# Description

This PR reverts the form assembly newsletter name change from #14895. Per Kris, the FA api actually does still use `mozilla-foundation`


danielfmiranda (approved):
LGTM!

PR #14941, opened 2025-11-19
Added donate banner dismiss date functionality

danielfmiranda:
# Description

This PR copies over the "Donate Banner Dismiss Day" functionality from the legacy site's donate banner to the redesign sites. 

This functionality only occurs on donate banners using the "legacy" format as that one gives users the option to click "close" at the top right of the banner.

The updated JS then adds a cookie to the users local storage which stores today's date. If this cooker is present and is the current day's date, then the banner will no longer render. However if the cookie does not match today's date, the banner will render again until re-closed. 

# Steps to test

1. Visit https://foundation-s-tp1-3352-d-demhxv.mofostaging.net/ (credentials: admin2/admin2)
2. Note that there is a donate banner
3. Click "Close" 
4. Refresh the page. Note that there is no donate banner anymore
5. Open dev tools, and go into "local storage"
6. Delete "donate banner dismiss day"
7. Refresh the page, the banner should reappear

### DIFF
--- a/foundation_cms/static/scss/blocks/themes/default/fru_element_block.scss
+++ b/foundation_cms/static/scss/blocks/themes/default/fru_element_block.scss
@@ -1,6 +1,5 @@
 .fru-element-block {
   display: flex;
-
   justify-content: center;
 
   &.fru-donate-button {


foundation_cms/static/scss/components/donate_banner/donate_banner.scss
```diff
@@ -17,6 +17,11 @@
   position: relative;
   padding-top: calc(map.get($grid-container-padding, small) / 2);
   padding-bottom: calc(map.get($grid-container-padding, small) / 2);
+  display: none;
```
mmmavis: This `display: none;` hides all banner versions by default. Since the `legacy` version needs special treatment, we should target it specifically with this rule. Alternatively, we could keep all banners visible by default (no SCSS changes) and use JS to call `banner.remove()` when the legacy banner shouldn't be shown. This approach avoids additional operations like:
- Hiding then showing the pushdown version
- Removing an already-hidden legacy banner

foundation_cms/static/scss/blocks/themes/default/fru_element_block.scss
```diff
@@ -1,6 +1,5 @@
 .fru-element-block {
   display: flex;
-
```
danielfmiranda, line 3: Just fixing this so the linter does not return an error!

PR #14952, opened 2025-11-19
Petition flow update tweaks

danielfmiranda:
# Description

This PR makes the following updates to the petition backend branch:

- Updates the `CTA` panel to be a snippet chooser panel. 
- Updates the CTA field to point to a Petition model, that way we can utlize the petition specific fields.
- Makes petitions translatable and returns the localized version on the page template whenever applicable.
- Updates The `thank_you_image` field to be an image chooser panel 
- Updates the form to be the FormAssembly form that we are currently using on the legacy pages, just unstyled so the front end devs can style them however requested.

### DIFF
--- a/foundation_cms/campaigns/migrations/0001_initial.py
+++ b/foundation_cms/campaigns/migrations/0001_initial.py
@@ -861,7 +861,7 @@ class Migration(migrations.Migration):
                 (
                     "signed_body",
                     wagtail.fields.RichTextField(
-                        default="Thanks so much for signing this petition asking for transparency on data used to train AI tool! Can you add a donation to support Mozilla's  non-profit work for trustworthy AI?",
+                        default="Thank you for signing this petition! Your voice matters in the fight for a healthier internet. Would you like to support Mozilla's nonprofit work with a donation?",
                         help_text="Content shown after signing",
                     ),
                 ),
--- a/foundation_cms/campaigns/models/campaign_page.py
+++ b/foundation_cms/campaigns/models/campaign_page.py
@@ -47,9 +47,9 @@ class CampaignPage(AbstractBasePage):
     )
 
     signed_body = RichTextField(
-        default="Thanks so much for signing this petition asking "
-        "for transparency on data used to train AI tool! Can you "
-        "add a donation to support Mozilla's  non-profit work for trustworthy AI?",
+        default="Thank you for signing this petition! "
+        "Your voice matters in the fight for a healthier internet. "
+        "Would you like to support Mozilla's nonprofit work with a donation?",
         help_text="Content shown after signing",
     )
 
@@ -182,7 +182,7 @@ class CampaignPage(AbstractBasePage):
             NothingPersonalArticlePage.objects.live()
             .public()
             .filter(locale=default_locale)
-            .order_by("-first_published_at")
+            .order_by("-first_published_at")[:2]
         )
         latest_articles = localize_queryset(default_articles, preserve_order=True)
 
--- a/foundation_cms/campaigns/templates/campaigns/campaign_page.html
+++ b/foundation_cms/campaigns/templates/campaigns/campaign_page.html
@@ -11,7 +11,7 @@
 {% block content %}
     <div class="campaign-page" data-page-id="{{ page.id }}" data-state="{{ state|default:'start' }}">
 
-        {% if medium == "email" %}
+        {% if medium != "email" %}
             <div class="email-confirmation-notice alert alert-warning">
                 <strong>{% trans "IMPORTANT" %}:</strong>
                 {% if user_email %}
--- a/foundation_cms/static/scss/_layout.scss
+++ b/foundation_cms/static/scss/_layout.scss
@@ -1,5 +1,13 @@
 @use "sass:map";
 
 .main-content-wrapper {
+  padding-top: 1px; // prevents child margin from collapsing
   background-color: map.get($mofo-colors, white);
+
+  // Safari-only fix for sticky element z-index bug
+  // This forces main-content-wrapper to be rendered on its own layer,
+  // which makes Safari respect stacking order with sticky elements
+  @supports (-webkit-appearance: none) and (stroke-color: transparent) {
+    transform: translate3d(0, 0, 0);
+  }
 }
--- a/foundation_cms/static/scss/components/donate_banner/donate_banner.scss
+++ b/foundation_cms/static/scss/components/donate_banner/donate_banner.scss
@@ -64,6 +64,7 @@
     align-content: center;
     background: $white;
     padding: 0 1rem 1rem;
+    align-self: center;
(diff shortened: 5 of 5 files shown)


foundation_cms/templates/base/base.html
```diff
@@ -77,6 +77,9 @@
             };</script>
         {% endblock onetrust_script %}
 
+        {% block extended_head %}
+        {% endblock extended_head %}
```
danielfmiranda, line 81: This portion here allows us to add the Form Assembly head code to the redesign site whenever is needed.

foundation_cms/campaigns/templates/campaigns/components/formassembly_head.html
danielfmiranda, line 1: Form assembly head code, copied over from `Formassembly_head_camo.html`. However, its safe to assume that we are using CaMo now, so we can leave that part out of the filename. Please note that this is pretty much a direct copy **except** for the external css files that were used by FormAssembly themes. That way our front end devs can style these fields however they want!

foundation_cms/campaigns/templates/campaigns/components/formassembly_body.html
```diff
@@ -0,0 +1,300 @@
...
+<!-- FORM: BODY SECTION -->
+<div class="wFormContainer" style="max-width: 100%; width:auto;">
+    <div class="wFormHeader"></div>
+    <style type="text/css"></style><div class=""><div class="wForm" id="187-WRPR" dir="ltr">
+        <div class="codesection" id="code-9"><script nonce="{{ csp_nonce }}">
+            wFORMS.behaviors.paging.isConditionalSubmitAllowed = true;
+            wFORMS.behaviors.condition.isConditionalSubmitAllowed = true;
+        </script></div>
+        <form method="post" action="https://mozillafoundation.tfaforms.net/api_v2/workflow/processor" class=" labelsAbove" id="187" role="form">
+            <input type="hidden" id="tfa_500" name="tfa_500" value="{{ petition_signed_url_tfa_500 }}" class=""><input type="hidden" id="tfa_1" name="tfa_1" value="{{ campaign_id_tfa_1 }}" class=""><input type="hidden" id="tfa_498" name="tfa_498" value="{{ source_url_tfa_498 }}" class=""><input type="hidden" id="tfa_499" name="tfa_499" value="{{ lang_tfa_499 }}" class=""><input type="hidden" id="tfa_501" name="tfa_501" value="mozilla-foundation" class=""><div class="oneField field-container-D    " id="tfa_28-D">
```
danielfmiranda, line 26: This file is pretty much a direct copy over of formassembly_body_camo.html, except since we are safe to assume that we are using camo we removed it from the file name. Also, this file just has one change, which is updating TFA_500 from thank_you_url, to petition_signed_url, just to better reflect the flow of the new petition page. However, everything still works as it should. 
```

File: foundation_cms/campaigns/templates/campaigns/campaign_page.html [hunk 3]
```diff
@@ -28,15 +35,15 @@
         </div>
 
         <!-- PETITION FORM STATE -->
-        {% if page.cta and state == "start" %}
+        {% if petition_cta and state == "start" %}
```
danielfmiranda, line 38: Updated to use the localized petition CTA that is being returned in context.

foundation_cms/campaigns/models/petition.py
```diff
@@ -9,7 +9,7 @@
 
 
 @register_snippet
-class Petition(TranslatableMixin, CTA):
+class Petition(CTA):
```
danielfmiranda, line 12: Since we are calling the translatablemixin on the base classes, we dont need to do so here as that would introduce an error when trying to run migrations and stuff.

foundation_cms/campaigns/models/cta_base.py
```diff
@@ -1,13 +1,15 @@
...
+from wagtail.models import TranslatableMixin
 from wagtail.search import index
+from wagtail.snippets.models import register_snippet
 from wagtail_localize.fields import SynchronizedField, TranslatableField
 
 from foundation_cms.constants import RICH_TEXT_BASE_OPTIONS
 
 
-class CTABase(models.Model):
+class CTABase(TranslatableMixin, models.Model):
```
danielfmiranda, line 12: Updating this class with Translatable Mixin allows classes that inherit from this to be translatable as long as we define their translatable fields 👍

foundation_cms/campaigns/models/campaign_page.py
```diff
@@ -120,7 +123,7 @@ class CampaignPage(AbstractBasePage):
     ]
 
     translatable_fields = AbstractBasePage.translatable_fields + [
-        TranslatableField("cta"),
+        SynchronizedField("cta"),
```
danielfmiranda, line 126: Since CTA's themselves are translatable, and we are returning their localized versions in page context on line 193, this should be a synchronized field 👍

foundation_cms/campaigns/models/campaign_page.py
```diff
@@ -23,7 +26,7 @@ class CampaignPage(AbstractBasePage):
     header = models.CharField(max_length=250, blank=True, help_text="Header for the campaign page")
 
     cta = models.ForeignKey(
-        CTA,
+        Petition,
```
danielfmiranda, line 29: Updating this to Petition for now so we can take advantage of that models additional fields such as "Show country field". If we need to update this at a later date we can do so.

foundation_cms/campaigns/models/campaign_page.py
```diff
@@ -79,7 +82,7 @@ class CampaignPage(AbstractBasePage):
     )
 
     thank_you_image = models.ForeignKey(
-        "wagtailimages.Image",
+        get_image_model_string(),
```
danielfmiranda, line 85: Updated with this so we can use the image chooser field on the CMS! 
image

foundation_cms/campaigns/migrations/0001_initial.py
danielfmiranda, line 1: Ive deleted and re-ran initial migrations to account for my changes. If there are any merge conflicts we are OK to just accept the one from this PR. Its pretty much the same just including my updates.

foundation_cms/campaigns/models/petition.py
```diff
@@ -9,7 +9,7 @@
 
 
 @register_snippet
```
danielfmiranda, line 11: @adalcima **Kudos** for registering these as snippets. This allows us to use the snippet chooser field on the cms rather than a drop down list. I am a fan of the snippet chooser panel because if we dont have a petition created yet it allows the users to create one, its also a little easier to use than a drop down list: 
image

adalcima (approved):
lgtm @danielfmiranda , I like the update of this approach and using the form as the legacy. Thanks :raised_hands:

PR #15027, opened 2025-11-28
[Do not merge] Fix legacy homepage error during template rendering

adalcima:
# Description
PR to fix a `'NoneType' object error: 'object has no attribute 'replace'` during legacy homepage template rendering.
Since the page tree in Polish is mirroring the English legacy homepage, this error is preventing the URL from being retrieved with the removed locale code.

Additionally, the current page tree in Polish(`ID 7955`) has inconsistencies (https://www.mozillafoundation.org/cms/pages/7955/), so we can use fixtree command to perform a DB wide-operation:
`inv manage fixtree`

Related issue: https://mozilla-hub.atlassian.net/browse/TP1-3412
Review app: https://foundation-s-tp1-3412-p-izjxei.mofostaging.net/cms/pages/
Login: admin    |     Password: 8qGR4mRA&0TbckyW

### Images
- Error during template rendering - Local
image

- Error during template rendering - Prod
image

- Polish page tree disappeared from root - Prod currently
image

- Polish page tree root - After fixtree command - Locally with a copy of the Prod DB
image


adalcima:
Hey @robdivincenzo  and @danielfmiranda for now, it's only for review when you have some time; there will be no merge until further notice. Thanks!

adalcima:
> Hi @adalcima! Thanks for taking this work on and for providing a easy to follow PR. I can confirm that your updates do fix the issue and everythinng is looking good to me, I just have one note:
> 
> After investigating the fix_tree command, it appears that even calling it via a page (ex: `page.fix_tree)` preforms a DB wide operation. Since thats the case, I think we are OK to remove the `fix_tree_by_id` file, merge the update to `localization.py` and then we can run `inv manage fixtree` via the heroku bash and all should be well 👍
> 
> Approved!

Hi @danielfmiranda, that's correct, we can use `fixtree` directly to have database-wide operation, and not just a specific ID. So we can use the command through bash as you mentioned. Thanks :rocket:

PR #15038, opened 2025-12-01
Fix Nothing personal featured articles linking to English pages by de…

adalcima:
# Description

PR to update Nothing Personal featured articles since the default links lead to the article pages in English.

Related issue: https://mozilla-hub.atlassian.net/browse/TP1-3415
Test page: https://foundation-s-tp1-3415-n-2h9cr8.mofostaging.net/fr/np-homepage-test/
Review app: https://foundation-s-tp1-3415-n-2h9cr8.mofostaging.net/cms/pages/448/edit/
`Login: admin     |     Password: ))raP4_Qgp2IQhl#`

**Main changes:**
1. Added template tag for localized featured items
2. Updated homepage template

**How it works:**
- CMS editors continue working in English (no workflow changes)
- Template tag detects user's current language
- Automatically finds translated versions of featured pages
- Fallback: Shows original English page if translation doesn't exist

### Images
image

image

### DIFF
--- a/foundation_cms/legacy_apps/wagtailpages/templatetags/localization.py
+++ b/foundation_cms/legacy_apps/wagtailpages/templatetags/localization.py
@@ -56,7 +56,11 @@ def get_local_language_names():
 # Get the url for a page, but with the locale code removed.
 @register.simple_tag()
 def get_unlocalized_url(page, locale):
-    return page.get_url().replace(f"/{locale}/", "/", 1)
+    url = page.get_url()
+    if url:
+        return url.replace(f"/{locale}/", "/", 1)
+
+    return "/"
 
 
 def relocalize_url(url, locale_code):


adalcima:
hey @danielfmiranda  for now, it's only for review when you have some time; there will be no merge until further notice. Thanks

foundation_cms/templatetags/nothing_personal_tags.py
```diff
@@ -32,3 +32,42 @@ def nothing_personal_homepage_url(context):
...
+                )()
+
+                localized_items.append(localized_item)
+
+            except Exception:
+                localized_items.append(item)
+        else:
+            localized_items.append(item)
+
+    return localized_items
```
danielfmiranda: @adalcima This template tag is working great, however, since the `NothingPersonalHomepage` is currently the only page with the featured items functionality, I am not sure if a template tag is needed (yet!).

Instead, what I would propose is to utilize our existing [localize_queryset](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/TP1-3415-Nothin-personal-featured-articles-are-linking-to-english-pages/foundation_cms/utils.py#L36) utility in a helper method on the Home Page class directly, and then passing the localized featured pages to context, similar to the [Legacy Blog Index Page](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/TP1-3415-Nothin-personal-featured-articles-are-linking-to-english-pages/foundation_cms/legacy_apps/wagtailpages/pagemodels/blog/blog_index.py#L508), [PNI's ProductPage](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/TP1-3415-Nothin-personal-featured-articles-are-linking-to-english-pages/foundation_cms/legacy_apps/wagtailpages/pagemodels/buyersguide/products.py#L837-L843) and the [NP Product Page ](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/TP1-3415-Nothin-personal-featured-articles-are-linking-to-english-pages/foundation_cms/nothing_personal/models/product_review_page.py#L224).


Here is an example of what this could look like for the HomePage:
1. First we would need to import the `localize_queryset` util method at the top of the file:
```python
from foundation_cms.utils import localize_queryset
```
2. Update the `page` FK related name:
```python
    page = models.ForeignKey(
        Page,
        related_name="nothing_personal_featured_items",
...
```
**Note:** We would need to run migrations after this. However, since there was no related name before this update, we don't need to worry about overwriting existing data, it should be OK.
 
3. Using this new related name, we can then add a new helper function, right above `get_context`, it could look something like this: 
```python
    def get_localized_featured_pages(self):
        """Return the localized versions of the featured pages"""
        featured_pages = Page.objects.filter(
            nothing_personal_featured_items__home_page=self
        ).order_by("nothing_personal_featured_items__sort_order")

        localized_featured_pages = localize_queryset(featured_pages, preserve_order=True)
        return localized_featured_pages    
```

4. Finally, we can just update the get_context method to include this new QS: 
```python
    def get_context(self, request, virtual_page_name=None):
        context["localized_featured_pages"] = self.get_localized_featured_pages()
...
```

Once this is done, we would just need to make some minor changes to the HTML template to successfully render the localized featured pages. I will leave my comments in that file!

foundation_cms/templates/patterns/pages/nothing_personal/home_page.html
```diff
@@ -1,6 +1,6 @@
 {% extends theme_base %}
 
-{% load static wagtailcore_tags %}
+{% load static wagtailcore_tags nothing_personal_tags %}
```
danielfmiranda:
```suggestion
{% load static wagtailcore_tags %}
```

If my suggestion sounds good to you, we would no longer need to import a new template tag.

foundation_cms/templates/patterns/pages/nothing_personal/home_page.html
```diff
@@ -19,9 +19,12 @@
...
       {% include "patterns/components/nothing_personal/home_page/hero_item.html" with hero_item=page.hero_item orientation=page.hero_item_orientation %}
     {% endif %}
-    {% if page.featured_items.all %}
-      {% include "patterns/components/nothing_personal/home_page/featured_items.html" with featured_items=page.featured_items.all %}
+
+    {% get_localized_featured_items page as localized_items %}
+    {% if localized_items %}
+      {% include "patterns/components/nothing_personal/home_page/featured_items.html" with featured_items=localized_items %}
     {% endif %}
+
```
danielfmiranda, line 26:
```suggestion

    {% if localized_featured_pages %}
      {% include "patterns/components/nothing_personal/home_page/featured_items.html" with featured_items=localized_featured_pages %}
    {% endif %}

```

If my suggestion sounds good to you, we can remove the use of of the template tag here and just update the `if` and `include` statement like so ^

Lastly, we would need to make a minor change to `featured_items.html`:

Instead of:
```jinja2
{% for featured_rel in featured_items %}
  {% if featured_rel and featured_rel.page %}
    {% with featured_item=featured_rel.page.specific %}
...
    {% endwith %}
  {% endif %}
{% endfor %}
```

we can wrap the template contents like so: 
```jinja2
{% for featured_item in featured_items %}
...
{% endfor %}
```

danielfmiranda (changes requested):
Hi @adalcima thanks very much for taking this work on and for the well formatted and easy to follow PR. 

After testing the review app out, this is working great! 

However, I do have some suggestions on how we may be able to simplify this a bit more. Please check out my comments and let me know what you think!

adalcima:
Hi @danielfmiranda I like your approach using this existing utility; I think it will help maintain consistency with the other models. Updated with these changes! Thanks :rocket:

danielfmiranda (approved):
Thanks @adalcima! R+ 🚀

PR #15218, opened 2026-01-14
Update legacy model verbose names

danielfmiranda:
# Description

This PR updates the `verbose_name` and `verbose_name_plural` of all models from the legacy site by appending "Legacy" to the name, so anywhere we see the page model being referenced in the CMS, it will be in the form of `<model_name> (Legacy)`.

This allows us to create our new page models without the need to append "New" to the model verbose name. 

That way, once our "Remove legacy code" work is done, the CMS will be populated with all of our new page models instead of "<model_name> (New)" when there are no old models in the DB. 

# Screenshots 

CMS Page Explorer, note how our new page types no longer have (New) and all legacy pages have (legacy):
Screenshot 2026-01-14 at 13-34-39 Exploring Redesign Homepage - Wagtail

CMS Side Bar, note how all legacy site settings now have (legacy):
Screenshot 2026-01-14 at 1 39 47 PM

# Link to review app

https://foundation-s-update-leg-meivth.mofostaging.net/cms/pages/31532/

Credentials: admin2/admin2

Steps to test:
- The review app is a copy of the staging DB. Please feel free to log in and explore to see what the updates would look like!

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -39,7 +39,7 @@ cffi==1.16.0
     #   pygit2
 charset-normalizer==2.0.12
     # via requests
-cryptography==43.0.1
+cryptography==44.0.1
     # via social-auth-core
 defusedxml==0.7.1
     # via
@@ -200,7 +200,7 @@ pyyaml==6.0.1
     # via django-pattern-library
 redis==4.5.4
     # via django-redis
-requests==2.32.3
+requests==2.32.4
     # via
     #   -r requirements.in
     #   basket-client
--- a/yarn.lock
+++ b/yarn.lock
@@ -232,6 +232,13 @@
   dependencies:
     "@types/json-schema" "^7.0.15"
 
+"@eslint/core@^0.15.2":
+  version "0.15.2"
+  resolved "https://registry.yarnpkg.com/@eslint/core/-/core-0.15.2.tgz#59386327d7862cc3603ebc7c78159d2dcc4a868f"
+  integrity sha512-78Md3/Rrxh83gCxoUc0EiciuOHsIITzLy53m3d9UyiW8y9Dj2D29FeETqyKA+BRK76tnTp6RXWb3pCay8Oyomg==
+  dependencies:
+    "@types/json-schema" "^7.0.15"
+
 "@eslint/eslintrc@^3.3.1":
   version "3.3.1"
   resolved "https://registry.yarnpkg.com/@eslint/eslintrc/-/eslintrc-3.3.1.tgz#e55f7f1dd400600dd066dbba349c4c0bac916964"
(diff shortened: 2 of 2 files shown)


kevdmoz:
@danielfmiranda is this a concern? 
image
 https://github.com/MozillaFoundation/foundation.mozilla.org/actions/runs/21012290282

foundation_cms/legacy_apps/wagtailpages/tests/blog/test_blog_index.py
```diff
@@ -83,7 +83,6 @@ def test_page_templates(self):
         self.assertEqual(response.status_code, HTTPStatus.OK)
         self.assertTemplateUsed(response, template_name="wagtailpages/blog_index_page.html")
         self.assertTemplateUsed(response, template_name="wagtailpages/fragments/entry_cards_item_loop.html")
-        self.assertTemplateUsed(response, template_name="wagtailpages/fragments/blog_card.html")
```
danielfmiranda, line 86: @robdivincenzo As mentioned in my slack DM, this is the assert statement that fails after adding a `Meta` class to the BlogPage model. 

I've investigated for a while, and couldnt find the root cause, the closest thing I could find is that this could be caused because the [page pointer ](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/legacy_apps/wagtailpages/pagemodels/blog/blog.py#L223)we implemented is causing confusion. 

However, I can confirm that all blog pages are still rendering on localhost after adding the meta, and, that the other tests checking the blog index page's get_entries method are still working as well.

Since we are going to deprecate these models anyways, should we just remove this line to keep this ticket from blocking our other work?

Thanks!

robdivincenzo: Thanks for all the work renaming these and the investigation on this one @danielfmiranda! 🚀 Definitely agree that at first glance, it isn't the most important thing to keep included since we're deprecating it. But I think we can solve this one upstream instead of omitting the test 👍.

If you look at the error logs, it looks like it loads a `generic_card.html` template in the test. Looking through the codebase I found the template where we determine where to load `generic_card.html` and `blog_card.html`. There's a conditional in there that looks like it needs update to match the new type / name.

Check out `foundation_cms/legacy_apps/wagtailpages/templates/wagtailpages/fragments/entry_cards_item_loop.html`. I updated the two conditionals for type == `blog page (legacy)`, which then me run the test successfully ✅. Give it a shot adding the test back in and let me know! If not we can revisit your alternative to unblock

```
{% for entry in entries %}
    {% with type=entry.specific_class.get_verbose_name|lower %}
        <li class="tw-m-0 tw-p-0 {% if type == "blog page (legacy)" %} tw-pt-12 {% endif %}">
            {% if type == "blog page (legacy)" %}
                {% include "./blog_card.html" with page=entry %}
            {% else %}
                {% include "./generic_card.html" with page=entry %}
            {% endif %}
        </li>
    {% endwith %}
{% endfor %}
```

danielfmiranda: Thanks @robdivincenzo! I can confirm that this did the trick 👍 its now ready for review whenever you have time. Thanks again

robdivincenzo (approved):
@danielfmiranda- I think this looks great! Thanks for the lift on this 🚀

PR #15220, opened 2026-01-14
[Do not merge] Setup Zine PDF download signup

adalcima:
# Description

PR with suggestions for implementing the Zine PDF Download feature.
`On hold` because we're likely switching from CaMo to another alternative.

Related issue: https://mozilla-hub.atlassian.net/browse/TP1-3528

Main changes considered so far:
- Double API call as suggested by the email team: first to the newsletter and then to the transactional email
- Create a new template, based on the newsletter template (styles can be changed later)
- Reuse the newsletter JS file
- Create a base for signup forms
- Add a new endpoint for CaMo (we might consider a generic link later)


danielfmiranda (commented):
Hi @adalcima thanks for the tag and for taking this work on! 

Also great job on this PR, I think this is a great starting point for the PDF download work and I like your approach of having a base signup form for both the newsletter sign up block and the pdf download block if they are going to share the same fields, and just have the additional PDF functionality. 

That way, whenever the new endpoint is made, all we have to do is update the logic for the PDF download block to hit that endpoint instead of the newsletter one, which I see you have already made a placeholder env variable 😄

PR #15234, opened 2026-01-16
Accordion Block

robdivincenzo:
# Description

This PR contains the new Accordion Block. This will be available as a base item across all page types and is placed in the new "Text" Streamfield Block grouping. Each `AccordionBlock` has an optional title for the whole block and the editor can add many `AccordionItems` within the block. Each `AccordionItem` takes a title and a rich text field which can be expanded for other content later if desired. Dividers will appear between items, matching the figma.

Link to sample test page: https://foundation-s-tp1-3445-a-u2zrhf.mofostaging.net/en/
Related PRs/issues: TP1-3445

### DIFF
--- a/foundation_cms/static/scss/blocks/themes/default/accordion_block.scss
+++ b/foundation_cms/static/scss/blocks/themes/default/accordion_block.scss
@@ -1,9 +1,5 @@
 $accordion-border: color(neutral, "300");
 
-.accordion-block {
-  margin: 2rem 0;
-}
-
 .accordion-item {
   border-top: 1px solid $accordion-border;
 


foundation_cms/templates/patterns/blocks/themes/default/accordion_block.html
```diff
@@ -0,0 +1,13 @@
+{% load wagtailcore_tags %}
+
+<section class="accordion-block" >
+  {% if self.accordion_block_title %}
+    <div class="accordion__title h5-style">{{ self.accordion_block_title }}</div>
```
ramram-mf: I think we could look into BEM class naming with something like:
```html
<section class="accordion-block">
    <div class="accordion-block__title h5-style">...</div>
    <div class="accordion-block__items">
    ...
    </div>
</section>
```
So it gets easier on the scss side to nest those rules like:
```scss
.accordion-block {
    &__title {
        ...
    }
    &__items {
        ...
    }
}
```

foundation_cms/templates/patterns/blocks/themes/default/accordion_block_item.html
```diff
@@ -0,0 +1,28 @@
+{% load wagtailcore_tags %}
+
+<div class="accordion__item">
+  <div class="accordion__item-heading h6-style">
```
ramram-mf: In the same manner, here we could use
```html
<div class="accordion-item">
    <div class="accordion-item__heading h6-style">...</div>
```

robdivincenzo:
Thank you @ramram-mf and @beccaklam! Revisions are ready!

ramram-mf:
Great job @robdivincenzo , I just noticed the caret icon when an accordion item is open does not flip. Is this on the spec?

image

robdivincenzo:
@ramram-mf great catch, had one more class to update. Ready again!

beccaklam (changes requested):
Hi @robdivincenzo a few more change requests:
- [x] Is it just me or is the top of the first line of the body paragraph getting cut off ever so slightly? 
> Screenshot 2026-01-21 at 11 32 25 AM

- [x] The padding top and bottom appears to be 28.8px now but it should be 32
> Screenshot 2026-01-21 at 11 33 14 AM

- [x] The spacing below (see yellow) the accordion body copy should also be 32px but is 25.6 right now
> Screenshot 2026-01-21 at 11 36 32 AM

- [x] Lastly the accordion heading should be bold and I do see font-weight: bold in the CSS but it doesn't look like it's affecting the element
> Screenshot 2026-01-21 at 11 42 17 AM

Sorry one more thing! Can we make the gap between the accordion header and the chevron 32px?
Screenshot 2026-01-21 at 2 30 31 PM


Thank you for removing the extra headings in the rich text block! That looks good.

beccaklam (approved):
Thank you for making all those changes @robdivincenzo and thank you for your patience 😂 Approved!

PR #15318, opened 2026-01-28
Refactor Language Selector

robdivincenzo:
# Description

Previously, the language selector included an extra redirect from the non-localized page to the user selected localized page. This creates a conflict with caching and introduces a redirect chain. By redirecting the user directly to the intended localized page, we can eliminate cache conflict risk, improve page load time, and improve SEO best practices (by eliminating redirect chaining).

Related PRs/issues: TP1-3559

# To Test
1) Go to the production homepage and inspect element, toggle to the "Network," filter by "HTML," and then select a new language
2) Note there are two 302 redirects with a final 200 page found.
3) Check out `TP1-3599-language-picker-bugfix` branch locally
4) Navigate to your local homepage and repeat step 1.
5) Note there is only one 302 redirect with a final 200 page found. There should be no unlocalized redirect path in between.

Before:
Screenshot 2026-01-29 at 10 35 44 AM

After:
Screenshot 2026-01-29 at 10 35 49 AM


adalcima (approved):
LGTM, tested it locally and indeed, now there's only one 302 redirect. Thanks @robdivincenzo 🚀

PR #15324, opened 2026-01-29
Refactor Language Picker Pt. 2

robdivincenzo:
# Description

This PR will restore the `csrf_exempt` on the language picker. The csrf_token route is not getting set properly, and will need additional investigation. But this should get things online for now.

Related PRs/issues: TP1-3599


adalcima (approved):
Approved @robdivincenzo , thanks

PR #15361, opened 2026-02-04
Fix language switcher in non-Page views

adalcima:
# Description

This PR is to add validation in `language_switcher.html` to handle missing `page` context.
Currently, the language switcher component fails in views that don't provide a `page` object to the context (e.g., search view) with:

```
ValueError: pageurl tag expected a Page object, got ''
```

### Main changes:
- With page context: Use specific localized page URL (existing behavior)  
- Without page context: Fallback to home page URL


adalcima:
> Hey @adalcima, thank you, definitely see how this is a problem! Just wondering if always using home is the right solution here. Wouldn't we want the localized URL of the current non-page view?

Hey @robdivincenzo , thanks for the feedback.
I updated the switcher so that non-page views are also localized, not just redirected to home.
For Wagtail pages we still use page.localized, but for non-page views we relocalize the current URL by replacing the locale prefix using a new tag, based on how we did it in legacy [localization tag](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/legacy_apps/wagtailpages/templatetags/localization.py#L66).

`The review app is on its way; I've tested this locally using the feature/search-app branch (which is the branch this fix is ​​now targeting).`

PR #15365, opened 2026-02-05
Fix language switcher in non page views

adalcima:
# Description

This PR is to add validation in `language_switcher.html` to handle missing `page` context.
Currently, the language switcher component fails in views that don't provide a `page` object to the context (e.g., search view) with:

```
ValueError: pageurl tag expected a Page object, got ''
```

### Main changes:
- With page context: Use specific localized page URL (existing behavior)  
- Without page context: Fallback to home page URL


danielfmiranda (approved):
LGTM 👍

PR #15394, opened 2026-02-12
fix: enforce button margin override in search form styles

ramram-mf:
# Description

This PR fixes a button margin error by overriding the default button margin on the search form submit button in the primary nav instance in pages other than the search results page, in the screenshot we can see the extra margin affecting the input field in the homepage.

image

*Current state*

image

*Fixed state*

Link to sample test page:
Related PRs/issues: #


foundation_cms/static/scss/components/search_form.scss
```diff
@@ -31,7 +31,7 @@
 .search-form__submit-button,
 .search-input-container__submit-button {
   line-height: 1.5;
-  margin: 0;
+  margin: 0 !important; // Override default button margin
```
mmmavis: We want to avoid using `!important` unless we need to override styling from 3rd party stylesheets we don't control.

mmmavis (changes requested):
Hi @ramram-mf , could you also specify what styling issue you are trying to resolve in this PR, provide screenshots, and tag @sabrinang for review as well? Thanks!

mmmavis:
image

I can't seem to reproduce the issue 🤔

ramram-mf:
> 🤔

@mmmavis Use the search form in the nav bar in any page but the results page. Try the homepage.

mmmavis:
I think there was some cache issue. I can reproduce the issue now.

foundation_cms/static/scss/components/search_form.scss
ramram-mf: Overriding now with an additional classname rule for the button.

PR #15472, opened 2026-02-25
Enable CMS editable Footer

adalcima:
# Description

This PR implements a CMS-editable footer system using Wagtail snippets, replacing the hardcoded static footer. Users would create a new footer in the admin sidebar, and then apply it to a specific site via the Settings menu.

The system will still show the **static footer** until the `EDITABLE_FOOTER` environment variable is set to `True`. Only then will it display the CMS-managed footer.
`Follow up ticket to remove this check and current static template file once the editable version is stable` 

Link to sample test page: https://foundation-s-tp1-3674-e-dvdzqd.mofostaging.net/en/
Related issue: https://mozilla-hub.atlassian.net/browse/TP1-3674

#### Customizable Elements
-  Internal Links: Drag-and-drop ordering, up to 10 links
-  External Links: Drag-and-drop ordering, up to 10 links
-  Social Media: Platform selection, up to 10 links
-  Legal Text: Rich text with link support

#### How to test
1. Login to the admin panel and verify the "Site Footers" sidebar exists.
2. Click into Site Footers to see the list of footer configurations.
3. Add / Edit a Site Footer.
4. Navigate to Settings → Site Footer in the sidebar and verify a footer is assigned as the active footer for the current site.
5. Enable Editable Footer Feature Flag in Heroku (`EDITABLE_FOOTER=True`)

`Note: Adding a "USE_X_FORWARDED_HOST: True" config variable, removing the "ALLOWED_HOSTS : *" allowed the preview to work in the review app .`

## Images

image
image
image
image
image

### DIFF
--- a/foundation_cms/footer/factories.py
+++ b/foundation_cms/footer/factories.py
@@ -12,7 +12,7 @@ class FooterInternalLinkFactory(DjangoModelFactory):
 
     footer = None
     label = factory.Faker("sentence", nb_words=2)
-    url = factory.LazyAttribute(lambda _: f"/{factory.Faker('slug').generate()}/")
+    url = factory.LazyAttribute(lambda _: f"/{factory.Faker('uri_path')}/")
     sort_order = factory.Sequence(lambda n: n)
     locale = factory.LazyFunction(lambda: wagtail_models.Locale.get_default())
 


kevdmoz:
@adalcima looking great so far. please see my minor notes below

- If I add a new external link field and _don't_ fill out the information (label and url), I can successfully save the draft. It should leave an "Error" to fill in required fields.
- I made updates to the external link section but I do not see the preview or the saved version on the review app. Can you show me what the new update looks like? I want to see if adding too many links will look weird or if we need to cap the number of links that fit into the footer. Thank you!

adalcima:
hi @sabrinang @kevdmoz thanks for the comments:
- The preview for the snippet in the review app isn't working yet. It's rendering a template that shouldn't be used for the preview in the review app. I'm still investigating.

- Added validations for required fields when saving as draft.
image

- Regarding whether we should keep the options to edit the logo and show/hide components like the newsletter signup, language switcher, and donate buttons, it would be good to hear your thoughts, @kevdmoz . The default is to show them and have flexibility over that content, but not sure if it's required or if it would be useful.

foundation_cms/footer/tests/test_models.py
robdivincenzo, line 1: Just one quick comment that I think we should prefix this file with `test_` so that we can avoid confusion with the actual `models.py` file. Other than that, LGTM @adalcima !

adalcima: Thanks @robdivincenzo , filename updated!

PR #15478, opened 2026-02-26
fix: two-column-container alignment

mmmavis:
# Description

- fix: remove top margin from rich-text-block first child in two-column-container ([Jira TP1-3572](https://mozilla-hub.atlassian.net/browse/TP1-3572))
- fix: remove margin-bottom from last block in two-column-container columns ([Jira TP1-3653](https://mozilla-hub.atlassian.net/browse/TP1-3653))

---

Link to review app: https://foundation-s-tp1-3572-m-po2ux1.herokuapp.com/
Link to sample test page: https://foundation-s-tp1-3572-m-po2ux1.mofostaging.net/en/test-page/

--- 

# Additional Design Question

image

`.two-column-container` always has a default margin-bottom applied, which adds extra spacing below the block. Is this intentional, or should we revisit this behaviour?

### DIFF
--- a/foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
+++ b/foundation_cms/static/scss/blocks/themes/default/two_column_container_block.scss
@@ -37,6 +37,9 @@
   }
 
   &__column {
+    flex: 1;
+    min-width: 0;
+
     .grid-container {
       padding-left: 0;
       padding-right: 0;


sabrinang (changes requested):
Seeing that default margin-bottom applied, we can apply it to the top of the section too as long as the content is still aligned within those margins. This can help balance the section I don't think the image/content of the containers need to be at the edges of the section. 

Image

sabrinang:
Let me know if that makes sense as I think the previous issue was there was space added to the rich text block where it made it misaligned with any images or other blocks that was adjacent to it in a 2 column container arrangement.

mmmavis:
PR updated!

PR #15488, opened 2026-03-02
Add id to Title Block for Anchor Nav

robdivincenzo:
# Description

This PR adds a "slugified" id to the Title Block h2 to enable direct linking via anchor nav.

Related PRs/issues: TP1-3090

# To Test
1) Check out this branch locally
2) Navigate to something like http://localhost:8000/#community-spotlight to verify scroll margin is accurate and working
3) Navigate to something like http://localhost:8000/#do-good-do-your-part to verify scroll margin is accurate and working
4) Inspect element on the h2's to verify the title block titles are "slugified" into the id attribute


mmmavis:
Also wondering if we should apply the same treatment to `h2` created by other blocks like rich text block?

robdivincenzo:
> Also wondering if we should apply the same treatment to `h2` created by other blocks like rich text block?

@mmmavis great idea for smooth scrolling, just committed that change and removed the other mentions across our redesign scss!

I also like the idea for enhancing to other blocks. A bit more complicated to do in rich text since it'd be required to parse through it. If you're ok with it, I'd like to split that idea off for further discussion / iteration? I believe the original hope was title block would be sufficient since that represents the major sections of our pages, but maybe we do need to do more. Let me know what you think!

mmmavis (approved):
@robdivincenzo Yup sounds good to me! Maybe we can update the Jira ticket to keep the scope limited to updating title block only.

robdivincenzo:
Sounds good! I made a [new ticket in Jira](https://mozilla-hub.atlassian.net/browse/TP1-3725) to track and updated the original ticket with more context on this being limited in scope to Title Block.

PR #15501, opened 2026-03-03
Enhance CTA Link Block

robdivincenzo:
# Description

This PR adds an alignment dropdown to the Link Button Block to enable center alignment. When center aligned, the button will be full width on mobile.

Link to sample test page: https://foundation-s-tp1-3593-e-kwdqxf.mofostaging.net/en/
Related PRs/issues: TP1-3593

# To Test
1) Review the test page
2) Note the center aligned Link Button underneath the impact numbers
3) Collapse to mobile and notice it goes full width
4) Edit the page and toggle alignment to left
5) Note that it left aligns as before (and not full width on mobile)

### DIFF
--- a/foundation_cms/campaigns/models/campaign_page.py
+++ b/foundation_cms/campaigns/models/campaign_page.py
@@ -15,7 +15,7 @@ from wagtail.fields import RichTextField
 from wagtail.images import get_image_model_string
 from wagtail.models import Orderable, Page, TranslatableMixin
 from wagtail.search import index
-from wagtail_localize.fields import SynchronizedField
+from wagtail_localize.fields import SynchronizedField, TranslatableField
 
 from foundation_cms.base.models import AbstractBasePage
 from foundation_cms.constants import DEFAULT_RICH_TEXT_FEATURES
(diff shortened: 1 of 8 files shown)


foundation_cms/blocks/link_button_block.py
```diff
@@ -13,6 +13,13 @@ class LinkButtonBlock(BaseBlock, LinkBlock):
         ],
         default="btn-primary",
     )
+    alignment = blocks.ChoiceBlock(
+        choices=[
+            ("link-button-block__left", "Left"),
```
ramram-mf: This classname could be more consistent with BEM spec if we replaced the `__left` and `__center` with `--left` and `--center`, since it is a type modifier more than an element rule.

tasks.py
```diff
@@ -429,7 +429,7 @@ def lint_python(ctx):
 def format(ctx):
     """Run formatters."""
     format_html(ctx)
-    format_css(ctx)
+    # format_css(ctx)
```
ramram-mf: I'm curious as to why we need to skip this, If its necessary, can we add an inline comment referencing why?

beccaklam:
@robdivincenzo Is there a way to make it so that this button on mobile always defaults to full width?

robdivincenzo:
Hi @beccaklam - we definitely can! We use the LinkBlock a lot, so I'm just double-checking if that's something that's ok across the board within the design system?

tasks.py
mmmavis: typo here? @robdivincenzo ?

robdivincenzo: Definitely a typo! I ran into an issue and was testing where it came from, I will bump this out

mmmavis: You might want to upgrade your npm packages locally. I bumped a few formatting/linting packages with updated rules, so you'll likely run into errors if you don't.

beccaklam:
Hi @robdivincenzo yeah, I understand the concern and upon reflection, let's leave the other buttons unchanged for now. Thank you for checking in!

foundation_cms/static/scss/blocks/themes/default/link_button_block.scss
```diff
@@ -0,0 +1,15 @@
+.link-button-block {
+  &__center {
```
ramram-mf: We should update this one to `--center` too, being a type modifier aligning with BEM.

robdivincenzo:
Great point @ramram-mf, just updated the class naming!

PR #15512, opened 2026-03-03
fix: prevent flex overflow in two-column container columns

mmmavis:
# Description

Add `min-width: 0` and `flex:1` to two-column container column to prevent flex overflow. This ensure the two columns in two-column-container block is always in equal width.

The "expanded by default" version of newsletter sign up form feels quite squishy in this 2-col layout. If there's any adjustment we want make, let's file a new ticket to track it. Thanks!

---

Link to sample test page: https://foundation-s-tp1-3642-t-faj6oi.mofostaging.net/en/test-page/
Related PRs/issues: [Jira #TP1-3642](https://mozilla-hub.atlassian.net/browse/TP1-3642)


mmmavis:
follow up ticket created: https://mozilla-hub.atlassian.net/browse/TP1-3721

PR #15563, opened 2026-03-12
Add custom TopicSelectWidget for improved page topic selection

danielfmiranda:
# Description

#### New `TopicSelectWidget`
- Created a custom Django form widget in `foundation_cms/base/widgets.py` that replaces the default Wagtail tag picker for the page topics field. The built-in tag picker gives no visibility into what topics are available, making it easy to overlook or skip. The new widget shows all available topics in a searchable dropdown so editors can see their options at a glance.

#### Chip-based UI with searchable dropdown
- The widget renders a chip bar showing currently selected topics and a text input that filters the available topic list as you type. Topics are added as chips and can be removed individually. Keyboard support includes Backspace to remove the last chip, and clicking outside closes the dropdown.
- Styles live in `topic-select-widget.css`; interactivity in `topic-select-widget.js`; markup in `topic_select.html`.

#### Moved topics field below the title
- The `topics` field panel is now positioned directly below the title field, making it more prominent and encouraging editors to fill it in rather than leaving it buried further down the form.

# Screenshots

Screenshot 2026-04-01 at 7 20 45 PM

# Steps to test

1. Visit https://foundation-s-improve-to-uzlmb3.mofostaging.net/cms/pages/31532/edit/ and scroll down
2. Test out the new topics input

### DIFF
--- a/foundation_cms/templates/patterns/admin/topic_select.html
+++ b/foundation_cms/templates/patterns/admin/topic_select.html
@@ -8,7 +8,7 @@
     {% for topic in widget.selected %}
       <span class="ts-chip" data-value="{{ topic }}">
         {{ topic }}
-        <button type="button" aria-label="Remove {{ topic }}">&times;</button>
+        <button type="button" aria-label="Remove {{ topic }}">×</button>
       </span>
     {% endfor %}
     <input type="text"


danielfmiranda:
Hi @kevdmoz! An update to the topic selection experience is ready in the review app. I was wondering if you could take a look and let me know what you think please?

kevdmoz:
Hi @danielfmiranda I think this is looking great so far. One minor change in the help text here, "If a topic is unavailable, you can create it under Snippets > Page Topics." I would change it to "If a topic is unavailable, please reach out to StratComms for help." Since they are the decider on the tags that can be added to the site and we have restricted the access gorup to "Snippet Editor."

danielfmiranda:
Thanks @kevdmoz! I just updated to help text to point users to StratComms 👍

danielfmiranda:
Thanks @adalcima! Yeah confirming that this wont change any of the existing data. On the backend the relationship is the same. The migrations are only needed because we are updating the field's help text. THanks for taking a look!

PR #15572, opened 2026-03-13
Make SEO meta fields required on all pages

danielfmiranda:
# Description

#### Validation for required SEO meta fields

- Added a `clean()` method to `AbstractBasePage` that enforces `seo_title` and `search_description` as required fields on all pages inheriting from this base class.
- Errors surface inline on the correct fields in Wagtail's **Promote** tab when an editor attempts to publish without filling them in.
- Uses Django's `ValidationError` with field-level error mapping so each missing field highlights individually rather than showing a generic page-level error.
- Updated the homepage factory manifest to include `seo_title` and `search_description` so programmatic page creation (e.g. `load_redesign_data`) passes validation.

This was implemented via `clean()` rather than `blank=False` on the field definitions because `seo_title` and `search_description` are defined on Wagtail's core `Page` model (the `wagtailcore_page` table) and cannot be overridden in a subclass.

# Steps to test

1. Visit: https://foundation-s-make-meta--cmc3mx.mofostaging.net/cms/pages/3/edit/#tab-promote
2. Leave **Title tag** or **Meta description** blank in the Promote tab.
3. Attempt to publish or save — you should see inline validation errors on the empty fields.
4. Fill in both fields and confirm the page saves successfully.

### DIFF
--- a/foundation_cms/campaigns/templates/campaigns/components/campaign_page/form_state/02_signed.html
+++ b/foundation_cms/campaigns/templates/campaigns/components/campaign_page/form_state/02_signed.html
@@ -9,8 +9,8 @@
     {{ page.signed_body|richtext }}
 
     <div class="petition__action-wrapper">
-        {% if page.cta.donate_url %}
-            {% include "patterns/components/button.html" with classnames="btn-primary--no-arrow petition__donate-btn" url=page.cta.donate_url label=page.cta.donate_text button_style="btn-primary" %}
+        {% if petition_cta.donate_url %}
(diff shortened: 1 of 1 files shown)


kevdmoz:
@danielfmiranda it's looking great. Can we add the red asterisk to Title Tag and Meta Description, please?

image

danielfmiranda:
No problem @kevdmoz ! I just updated them to show the red asterisk 👍 
Screenshot 2026-03-16 at 18-23-23 Editing General Page Newsletter - Wagtail

PR #15706, opened 2026-04-01
feat: restrict legacy page and snippet access to authorized group

danielfmiranda:
# Description

#### New \`access/legacy content editor\` Wagtail group
- A data migration creates this group, assigns all legacy model permissions to it, and removes those same permissions from the default \`access/content editor: no publishing access\` group. This means running \`migrate\` is all that's needed, no manual permission configuration required in any environment.
- The migration reverse is fully non-destructive: it deletes the legacy group and restores all legacy permissions back to the default content editor group, leaving the DB in the same state as before the migration ran.

#### Snippet access
- The migration removes legacy snippet permissions from the default content editor group, so unless you are a superuser or a member of the \`access/legacy content editor\` group, you will no longer have access to legacy snippets. Wagtail automatically hides the sidebar menu items for any snippets you don't have permission to access.

#### Page explorer and edit view restrictions
- \`construct_explorer_page_queryset\` hook: filters legacy pages out of the page explorer entirely for unauthorized users. Legacy page types are identified by their ContentType ID, using Django's built-in ContentType manager cache.
- \`before_edit_page\` hook: catches anyone who navigates directly to a legacy page edit URL and redirects them to the admin home with an error message.

# Screenshots
User with Legacy Access Role:
Screenshot 2026-04-03 at 9 58 08 AM

User without Legacy Access Role:
Screenshot 2026-04-03 at 9 53 19 AM

Admin:
image

# Steps to test

1. Visit https://foundation-s-legacy-pag-viktmy.mofostaging.net/cms/ without legacy role (UserName: \`testnonlegacyuser\`/ PW: \`WrongPassword123!\`)
2. Observe that no legacy stuff is not rendering in the CMS anymore
3. Visit https://foundation-s-legacy-pag-viktmy.mofostaging.net/cms/ **with** legacy role (UserName: \`testlegacyuser\`/ PW: \`WrongPassword123!\`)
4. Observe that no legacy stuff **is** rendering in the CMS. This user has both "accesscontent / no publishing access" and "access/ legacy content" roles.
5. Visit using SSO, (we all have admin roles)
6. Observe that nothing has changed.

### DIFF
--- a/requirements.in
+++ b/requirements.in
@@ -26,7 +26,7 @@ pygit2==1.14.0
 python-slugify
 requests
 social-auth-app-django
-wagtail==7.1.3
+wagtail==7.2.3
 wagtail-ab-testing
 wagtail-color-panel
 wagtail-factories
--- a/requirements.txt
+++ b/requirements.txt
@@ -70,6 +70,7 @@ django==4.2.29
     #   django-treebeard
     #   djangorestframework
     #   laces
+    #   modelsearch
     #   social-auth-app-django
     #   wagtail
     #   wagtail-footnotes
@@ -109,7 +110,9 @@ django-stubs-ext==5.2.8
 django-taggit==6.1.0
     # via wagtail
 django-tasks==0.8.1
-    # via wagtail
+    # via
+    #   modelsearch
+    #   wagtail
 django-treebeard==4.5.1
     # via wagtail
 djangorestframework==3.15.2
@@ -152,6 +155,8 @@ laces==0.1.2
     # via wagtail
 markdown==3.8.1
     # via django-pattern-library
+modelsearch==1.1.1
+    # via wagtail
 numpy==1.24.4
     # via
     #   scipy
@@ -265,7 +270,7 @@ urllib3==2.6.3
     #   sentry-sdk
 user-agents==2.2.0
     # via wagtail-ab-testing
-wagtail==7.1.3
+wagtail==7.2.3
     # via
     #   -r requirements.in
     #   wagtail-ab-testing


foundation_cms/legacy_apps/wagtailpages/migrations/0174_create_legacy_content_editor_group.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def create_legacy_group(apps, schema_editor):
+    Group = apps.get_model("auth", "Group")
+    Permission = apps.get_model("auth", "Permission")
+
+    legacy_group, _ = Group.objects.get_or_create(name=LEGACY_GROUP_NAME)
+
+    legacy_permissions = Permission.objects.filter(content_type__app_label__in=LEGACY_APP_LABELS)
+    legacy_group.permissions.set(legacy_permissions)
```
danielfmiranda: Assigns all model permissions for the legacy apps to the new group in one query. This is what allows Wagtail to automatically hide legacy snippet menu items for users not in this group.

foundation_cms/legacy_apps/wagtailpages/migrations/0174_create_legacy_content_editor_group.py
```diff
@@ -0,0 +1,54 @@
...
+
+    # Remove legacy permissions from the default content editor group so that only
+    # members of the legacy group (and superusers) can access legacy content.
+    try:
+        default_group = Group.objects.get(name=DEFAULT_GROUP_NAME)
```
danielfmiranda: Strips the legacy permissions from the default content editor group so regular editors lose access. Wrapped in a try/except because the default group may not exist in all environments (e.g. fresh local setups), and we should not fail the migration if it is missing.

kevdmoz:
@danielfmiranda it's looking good. I noticed "Snippets" are missing for both. Is that intentional?

danielfmiranda:
Hi @kevdmoz thanks for taking a look and for the catch. It turns out that was because neither of them were assigned the snippet editor group, which allows them access to the snippets tab which includes our new snippets.

I've updated their permissions. In the screenshots you will see that the **legacy** user now has access to the snippets tab (which will include new and legacy snippets) as well as the other (legacy) snippet groups in the sidebar.

The **non legacy** user now has access to the snippets tab but only the new snippets will show.

If there are any more questions or feedback, let me know! And thanks again

PR #15760, opened 2026-04-14
feat: add management command to remove network API blocks

danielfmiranda:
# Description

#### `remove_network_api_blocks` management command

- Adds a new Django management command that scans all Wagtail pages across all locales and removes blocks that depend on the network/Pulse API (`latest_profile_list`, `profile_by_id`, `profile_directory`).
- Supports a `--commit` flag to write changes to the database and republish affected pages. Runs in dry-run mode by default, printing a summary of what would be removed without touching the DB.
- Recursively handles blocks nested inside `StreamBlock` (value is a list) and `StructBlock` (value is a dict) — not just top-level stream items.
- Fixes a crash when extracting raw stream data from `StreamValue` objects: previously the fallback called `str()` on the value, which triggers Wagtail's HTML rendering path and fails for blocks like `ProfileById` that require a request context. The fix checks `_raw_data` directly and guards against calling `str()` on any `StreamValue` instance.

# Steps to test

1. Run in dry-run mode to verify output:
   ```
   docker compose run --rm backend python manage.py remove_network_api_blocks
   ```
2. Confirm the summary lists pages containing the target block types with no DB changes.
3. Run with `--commit` to apply:
   ```
   docker compose run --rm backend python manage.py remove_network_api_blocks --commit
   ```
4. Verify affected pages no longer contain `latest_profile_list`, `profile_by_id`, or `profile_directory` blocks and have been republished.


danielfmiranda:
Superseded by #15761

PR #15769, opened 2026-04-16
Petition Page Sosha social sharing toolkit integration

danielfmiranda:
# Description

Link to review app: https://foundation-s-sosha-test-u2qjdq.mofostaging.net/en/test-campaign-page/?state=sharing

This PR implements the following changes:
- Replaces share progress related fields in the CMS for a Sosha related field instead, where users can add the embed code generated by a sosha tool kit.
- Updates the templates accordingly to accomodate this change. The legacy share buttons no longer appear, but rather now just a sosha toolkit. 

# Screenshots

 CMS Before:
Screenshot 2026-05-07 at 16-00-52 Editing 2026 Instagram Encryption - Wagtail

 CMS After:
Screenshot 2026-05-07 at 15-57-22 Editing Sign-up form - Wagtail

Share State Before:
Screenshot 2026-05-07 at 16-00-03 Tell Meta Keep Encryption for Instagram Messages - Mozilla Foundation

Share state after:
Screenshot 2026-05-07 at 3 58 35 PM


danielfmiranda:
Hi @kevdmoz @nancyt1! QQ for the both of you:

Previously, the share progress buttons were not required, but, the "Copy to clipboard" button would still render even if both the Facebook and Email buttons were not rendering (since the fields were not filled in).

The Sosha Field is currently not required, but I was curious if I should make it so. Asking because when the field is not filled in, it just renders the "Not Right Now" button but with no options, so users may find it a little confusing: 
image


Thanks!

kevdmoz (approved):
lgtm!

nancyt1:
Can we put a default one right now and make it required? (So that it doesn't break other petitions pages) It doesn't make sense to leave it blank and I would assume the engagement team would always want a share option. cc @kevdmoz

> Hi @kevdmoz @nancyt1! QQ for the both of you:
> 
> Previously, the share progress buttons were not required, but, the "Copy to clipboard" button would still render even if both the Facebook and Email buttons were not rendering (since the fields were not filled in).
> 
> The Sosha Field is currently not required, but I was curious if I should make it so. Asking because when the field is not filled in, it just renders the "Not Right Now" button but with no options, so users may find it a little confusing: image
> 
> Thanks!

kevdmoz:
@danielfmiranda @nancyt1 Do not make it required right now since it'll take time to setup the Sosha toolkit/posts. I worry that live petitions will break when it is released. I think we can decide to make it a required field in the future when Sosha is more mature

adalcima (approved):
Code changes look good. Thanks @danielfmiranda

nancyt1:
@danielfmiranda @kevdmoz If we don't make it required then are we able to not show that share slide? If we can't, then I guess just remove the "maybe later" link.

danielfmiranda:
Hi @nancyt1! Per request of @kevdmoz, I have updated it to be conditional. So for example, if **no** sosha share code is set, the page will render the "Copy to clipboard" button still:
Screenshot 2026-05-11 at 16-48-22 Tell Meta Keep Encryption for Instagram Messages - Mozilla Foundation



Does that work with you? 

Thanks!

nancyt1:
@danielfmiranda yeah that works!

PR #15771, opened 2026-04-16
Django 5.2 Upgrade

robdivincenzo:
# Notice
Still WIP, and ready for next dev to pick up (see final testing note for hopefully last hurdle)

# Description

This PR contains the Django 5.2 upgrade alongside dependent changes for compatibility. Commits are generally grouped in category of changes and the summary is as follows:

- Bump requirements.in, requirements.txt, and dev-requirements.txt
  - `wagtail-localize-git` is now pinned specific commit SHA of a forked version of this that Django 5.2 / Wagtail 7+ compatible.
  - requirements have been compiled and synced, including dev requirements which appears to have not been sync'ed in some time
- Removing `adminsortable` mixin & new Cast() function
  - adminsortable is not maintained and compatible with Django 5.2. There is an adminsortable-2 which is, but this is only used in legacy_apps on the highlights model. It introduces an order field that editors can use. I've removed the library which will eliminate the editor UI to re-order these (which I think is fine for legacy content). I've left the order field since it already exists and will leave the existing order untouched. It will all be sunset later.
  - the `id` field in the legacy_apps profiles models.py needs to be explicitly cast to characters in order to concat
- Remove `wagtail-localize-git patch`
  - This was a necessary change to patch some code in `wagtail-localize-git` for wagtail 7 compatibility. I've patched it in the fork so this patch on the foundation site is no longer needed.
- New formatting via black-26.x removes parenthesis from tuple assignment, so there's a dozen files reformatted this way.
- Fix test compatibility
  - `test_year_choices` returning a list instead of tuple
  - Changing a `SimpleTestCase` to a `TestCase` because Django now requiring DB hit 

PS for devs: probably will need to rebuild your local after this is merged in.

Link to sample test page: https://foundation-s-tp1-3846-d-knhuou.mofostaging.net/en/
Related PRs/issues: TP1-3846

# To Test
1) Review the review app for functionality and ensure nothing is broken.
2) Although it works locally, I am now seeing that our factory images aren't loading on review apps... Something unresolved? Is this a Django 5.2 thing or something related to the new requirements / dev-requirements?


robdivincenzo:
Merged via #15885 , closing.

PR #15803, opened 2026-04-21
Fixed misaligned signup button on two column signup block

Mauricio-RC:
# Description

This PR fixes the newsletter signup block layout when it is rendered inside constrained page contexts, especially inside two-column and three-column containers.

The original issue was visible as a misaligned signup button when the newsletter signup block was nested in a two-column layout. While reviewing the fix, we expanded the coverage to the other places where this StreamField block can be used so the layout behaves consistently across body content, container blocks, Campaign pages, Nothing Personal pages, Product Review pages, Project pages, and the footer newsletter form.

## What changed?

- Added reusable newsletter signup SCSS mixins for constrained/narrow block layouts.
- Moved repeated decorative ray styling into a shared `newsletter-signup-rays-icon` mixin.
- Added a shared narrow block layout treatment for newsletter signup blocks nested inside two-column containers.
- Updated three-column container newsletter signup styling so form controls stack correctly in narrow columns.
- Applied the reusable narrow layout to Campaign page newsletter signup blocks.
- Scoped Nothing Personal article/homepage newsletter signup overrides to `main` so they do not leak into the footer newsletter signup form.
- Preserved the standalone newsletter signup block behavior outside constrained containers.

## Review app pages

Use these pages to review the newsletter signup block in the relevant page/container contexts:

- Homepage: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/
- General page: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/newsletter-signup-matrix-general/
- Campaign page: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/newsletter-signup-matrix-campaign/
- Nothing Personal homepage: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/nothing-personal/
- Nothing Personal article: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/nothing-personal/newsletter-signup-matrix-np-article/
- Product Review page, expand-on-focus layout: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/nothing-personal/newsletter-signup-matrix-product-review-focus/
- Product Review page, expanded layout: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/nothing-personal/newsletter-signup-matrix-product-review-expanded/
- Project page: https://foundation-s-tp1-3729-s-choo0n.mofostaging.net/en/gallery/newsletter-signup-matrix-project/

## How to test

- Review the pages above at desktop, tablet, and mobile widths.
- Confirm both newsletter layouts work:
  - `expand_on_focus`
  - `expanded`
- Confirm newsletter signup blocks inside two-column and three-column containers stack their form controls correctly.
- Confirm standalone/body newsletter signup blocks still use their expected layout.
- Confirm the footer newsletter signup form is not affected by page-specific newsletter signup overrides.

Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3729)


foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
```diff
@@ -291,6 +291,22 @@
   }
 }
 
+.newsletter-signup__container[data-signup-id="3"] {
```
danielfmiranda, line 294: Hi @Mauricio-RC!

The CSS rules are working great. However, I think the `data-signup-id="3"` selector might be a bit too narrow in scope. Hardcoding the ID means this `flex-direction: column` rule will always apply specifically to the [newsletter signup snippet with the ID of 3,](https://foundation-s-tp1-3729-s-gcqzmn.mofostaging.net/cms/snippets/snippets/newslettersignup/edit/3/) which fits the "Join Us" page case perfectly, but could cause issues if we reuse that newsletter snippet elsewhere on the site.

It might be worth making the selector more flexible, something like targeting any newsletter signup container that appears inside a newsletter signup block, within a two-column container block. That way the rule applies based on context/layout rather than a specific ID.

danielfmiranda (changes requested):
Hi @Mauricio-RC! 

Thanks for taking this work on. After taking a look at the review app I can see that the newsletter sign up form on the Join us page is now looking exactly as we would like! 

However, I did have just one comment on how we might be able to update the CSS.

Let me know if it makes sense or if you have any questions, I am happy to help! 😄

foundation_cms/static/scss/components/newsletter_signup/newsletter_signup_form.scss
```diff
@@ -291,6 +291,22 @@
   }
 }
 
+.two-column-container__column .newsletter-signup__container {
```
ramram-mf: Hey @Mauricio-RC its looking good on code so far. I think we could move this style declaration to  the `two_column_container_block.scss` file, since it already has a `.newsletter-signup` nested block around L65, so that way all styling regarding blocks inside the 2-col would be in a single file.

ramram-mf (changes requested):
I left an inline comment regarding the location of the style declarations for this fix. Thanks @Mauricio-RC ! Let me know if you need any further information about this.

danielfmiranda (changes requested):
Hey @Mauricio-RC! Thanks for taking a look at my comment and implementing the fix. I really appreciate it! 

I have taken a look at the review app, and the "Expanded" newsletter variant looks great and exactly as expected. 

I did notice however, that the "Expand on focus" variant goes to the "side by side" layout still on large breakpoints and up:

Image

However on medium and below, it looks to render as expected: 

Image



Is it possible to update that breakpoint? 

As always, feel free to reach out if you need any help 😄 

Thanks again!

Mauricio-RC:
> Hey @Mauricio-RC! Thanks for taking a look at my comment and implementing the fix. I really appreciate it!
> 
> I have taken a look at the review app, and the "Expanded" newsletter variant looks great and exactly as expected.
> 
> I did notice however, that the "Expand on focus" variant goes to the "side by side" layout still on large breakpoints and up:
> 
> Image
> However on medium and below, it looks to render as expected:
> 
> Image
> Is it possible to update that breakpoint?
> 
> As always, feel free to reach out if you need any help 😄
> 
> Thanks again!

Hi @danielfmiranda , 

> Hey @Mauricio-RC! Thanks for taking a look at my comment and implementing the fix. I really appreciate it!
> 
> I have taken a look at the review app, and the "Expanded" newsletter variant looks great and exactly as expected.
> 
> I did notice however, that the "Expand on focus" variant goes to the "side by side" layout still on large breakpoints and up:
> 
> Image
> However on medium and below, it looks to render as expected:
> 
> Image
> Is it possible to update that breakpoint?
> 
> As always, feel free to reach out if you need any help 😄
> 
> Thanks again!

Hey @danielfmiranda , this issue was fixed in my latest commit, here is a preview of the current status for the horizontal signup block, please feel free to let me know if something else needs to be addressed  
image

danielfmiranda (approved):
Hi @Mauricio-RC thank you for taking this work on! The code changes themselves are looking good to me, however there is one thing that caught my eye and would like to confirm with design. 

If I am not mistaking, the form and content inside the two column container block should always be rendering vertically stacked rather than horizontal:

Image

On medium and below breakpoints though they stack as expected: 

Image


tagging @beccaklam from the design team to confirm. Also, @beccaklam  I was wondering what the desired behavior for the two bottom forms should be? The top one is "Expanded" and the bottom one is "Expanded on focus"

beccaklam (changes requested):
Hi @ramram-mf thank you for your thorough work here!!! I'm actually hoping to overhaul this block so having this ticket was helpful to see all the places it's being used. Some change requests/questions:

**Homepage:** ✅ looks good

**General page:** Re: body block expanded - can we fit 'email', 'country' and 'language' into one row with 8px spacing (see image below for an example). Everything else lgtm

Image

**Campaign page:** ✅ looks good

**Nothing Personal homepage:** This is an odd use case. Ideally the grey and orange boxes are always the same height. Is there a way to do this? If this isn't an easy fix we can leave this for now. Like I said, I think we need to overhaul the email block in the future anyway. I think we should always be using the expanded version when using this design next to the grey box.

**Nothing Personal article:** 
- We should only ever need expand on focus right on this page right now (again this may change in the future, so if it's a hard fix just leave)
- Two column container should not be an option on NP article
- Can we switch to the mobile layout starting at the tablet breakpoint? The desktop layout is looking tight for tablet: 
image


**Product Review page, expand-on-focus layout:** ✅ looks good

**Product Review page, expanded layout:** same as request on General Page (can we fit 'email', 'country' and 'language' into one row with 8px spacing? Again, like the example below:

Image

**Project page:** 
- Can we reuse the NP version of the email expand on focus version here?The version on this page in the PR looks too squished.

Not working:
Image

Use this instead?
Image

- We shouldn't need the expanded version on this page (but ditto my comments on the NP Article page, if this is a hard fix, leave for now and wait for overhaul)
- Two column container should not be an option on the Project page type

beccaklam (changes requested):
Hi @ramram-mf, as discussed here are the change requests. Thank you for cleaning this up!!!

- [x] Expanded 2 column is broken (form fields should be full width like the button below): 
Image

- [x] Campaign page email sign up disappears on tablet and mobile breakpoints

- [x] NP article body expanded is broken, should be stacked instead:

Image

beccaklam:
@ramram-mf I'm not sure this is worth fixing since it's not used this way but on the NP homepage in the mobile view: if the email sign up is used on the homepage and it's not the last item, the "sliding" effect hides the email sign up so that a user never sees it.

https://github.com/user-attachments/assets/42615e60-3b61-4c25-9f1f-05e092ebed08

ramram-mf:
> @ramram-mf I'm not sure this is worth fixing since it's not used this way but on the NP homepage in the mobile view: if the email sign up is used on the homepage and it's not the last item, the "sliding" effect hides the email sign up so that a user never sees it.

@beccaklam I think we can create a follow up ticket on this one, since I think that block requires a bit of work too, regarding the equal heights and now the scroll visibility.

beccaklam (approved):
@ramram-mf Sorry for the delay!! I'll create the follow up ticket soon. Approved!

ramram-mf (approved):
R+

PR #15804, opened 2026-04-21
Add character limits for timely activations and portrait card blocks

adalcima:
# Description

This PR adds a `max length` to the text fields of _Timely Activations Cards_ and _Portrait Card Set_ as a guideline for editing limits, since currently some text examples are cut off due to card and text size. A reusable block for links with variable labels has also been added.

Related issue: [TP1-3418](https://mozilla-hub.atlassian.net/browse/TP1-3418)
Review app: https://foundation-s-tp1-3418-a-fdsmhj.mofostaging.net/cms/pages/564/edit/
Test page: [mobilize-test-with-limits](https://foundation-s-tp1-3418-a-fdsmhj.mofostaging.net/en/mobilize-test-with-limits/)
Figma [file](https://www.figma.com/design/E3xU27s31av479EJOpnVEO/%F0%9F%9A%A7-MVP-Design-System-%F0%9F%9A%A7?node-id=2243-5291&m=dev)

**NOTE**: These changes don't affect published pages, so they should continue working as they were published. The new character count limits will display the corresponding alerts in the CMS, so new versions of existing pages cannot be published until the fields are corrected and adjusted to these new character counts.
___
#### How to test
1. Edit/Create a General Page
2. Add a `Timely Activations Cards` and/or `Portrait Card` block
3. Note that there are now limits on the text in Timely Activations and the heading in Portrait Card.

### Images
- Live pages still work and new alerts on existing pages in the CMS need to be fixed for new versions
image

- Timely activation limits
image

- Portrait card limits
image

- Alerts when exceeding limits
image


beccaklam (changes requested):
Hi @adalcima as I mentioned in my other comment I think we need to keep the suggested character counts and not adjust them for existing content. The current content is too long and break the design. Even if the text does not exceed the bounding box, the amount of text is too tight against the edges.

Character limits should be as follows:

**"Timely Activation Cards"**
- [x] Category should be max 20 char
- [x] Title should be max 60 char
- [x] Rich Text (optional description) should be max 180 char
- [x] Link Label should be max 14 char

**Portrait Cards:**
- [x] Headline is max 36 char
- [x] Call to Action Link Label is max 36 char

If we are concerned about breaking any currently published designs can we truncate what currently exists but then add the error validation for new blocks?

adalcima:
> Hi @adalcima as I mentioned in my other comment I think we need to keep the suggested character counts and not adjust them for existing content. The current content is too long and break the design. Even if the text does not exceed the bounding box, the amount of text is too tight against the edges.
> 
> Character limits should be as follows:
> 
> **"Timely Activation Cards"**
> 
>     * [ ]  Category should be max 20 char
> 
>     * [ ]  Title should be max 60 char
> 
>     * [ ]  Rich Text (optional description) should be max 180 char
> 
>     * [ ]  Link Label should be max 14 char
> 
> 
> **Portrait Cards:**
> 
>     * [ ]  Headline is max 36 char
> 
>     * [ ]  Call to Action Link Label is max 36 char
> 
> 
> If we are concerned about breaking any currently published designs can we truncate what currently exists but then add the error validation for new blocks?

Hi @beccaklam thanks for the feedback and the required limits. I have updated the PR with these values, as well as the description and images used in this PR. 
- Live pages continue to function like this [page](https://foundation-s-tp1-3418-a-fdsmhj.mofostaging.net/en/mobilize-test/). In the CMS, we have alerts for new versions of existing pages like [this example](https://foundation-s-tp1-3418-a-fdsmhj.mofostaging.net/cms/pages/534/edit/) (these would need to be corrected before publishing new versions).

beccaklam (commented):
Hi @adalcima thank you for implementing the character count maximums. One question, I see for the rich text field there is a character max countdown indicator. Can we apply something similar to the other fields that have a character count max that are not rich text fields?

adalcima:
> Hi @adalcima thank you for implementing the character count maximums. One question, I see for the rich text field there is a character max countdown indicator. Can we apply something similar to the other fields that have a character count max that are not rich text fields?

Hi @beccaklam this `countdown` is a behavior that Wagtail adds to `Rich Text` and doesn't provide by default for `non-rich text`. The short answer is yes, but I think it would require custom development for other non-rich text fields (probably through a widget or hook). I was wondering if this is something we need for this implementation or if it can be considered an additional task?

mmmavis (approved):
R+ Solid work @adalcima ! The `LinkWithDynamicLabelBlock` wrapper is clean and well-thought-out. 🎉

beccaklam:
> > Hi @adalcima thank you for implementing the character count maximums. One question, I see for the rich text field there is a character max countdown indicator. Can we apply something similar to the other fields that have a character count max that are not rich text fields?
> 
> Hi @beccaklam this `countdown` is a behavior that Wagtail adds to `Rich Text` and doesn't provide by default for `non-rich text`. The short answer is yes, but I think it would require custom development for other non-rich text fields (probably through a widget or hook). I was wondering if this is something we need for this implementation or if it can be considered an additional task?

Ah I see @adalcima thank you for answering my question! We can consider that an progressive enhancement then! I'll create a ticket for it :)

beccaklam (approved):
Approved! ✨ Thank you @adalcima! Really looking forward to having those character count maximums!

PR #15813, opened 2026-04-23
Expert Hub landing page with interactive bubble visualization

mmmavis:
## What changed visually

- Each featured expert is a circular bubble sized by their tier, arranged across the page
- **Desktop** — hover shows a tooltip with name and blurb; bubbles sharing a topic are connected by orange lines, others dim; click navigates to their profile
- **Touch/mobile** — tap slides up a card with photo, blurb, name, and a link to their profile
- Decorative parallax arrows animate on either side of the hero as the user scrolls
- A "Want to collab?" CTA sits below the viz

## Summary

Replaces the static expert grid with a D3-force bubble visualisation and rewrites the page template, models, and assets.

- **`ExpertProfilePage`** — adds a `blurb` field (max 115 chars) surfaced in the tooltip and profile card, registered as translatable
- **`ExpertHubPage`** — caps `description` at 120 chars, enforces a 25-char title limit via a custom admin form and `clean()` backstop, removes topic-sorting from `get_context()` since ordering is now handled by the viz
- **Template** — old placeholder card grid replaced with a hero split into copy and viz container, bubble list, lightbox dialog, and CTA section with page-scoped CSS and JS loaded via dedicated blocks
- **`viz.js`** — D3-force simulation with four breakpoint configs covering desktop hover and click and touch tap interactions
- **`lightbox.js`** — profile card overlay with cross-browser scroll lock (`overflow:hidden` on `<html>` + non-passive `touchmove` preventDefault for iOS) and dialog hoisted to `<body>` so `position:fixed` always resolves to the viewport
- **`viz-configs.js`** — per-breakpoint layout tables covering tier radii, position percentages, and aspect ratios
- **Build** — `d3-force`, `d3-selection`, `d3-timer` added with `expert_hub_page` registered as a bundled esbuild and CSS entry

## Test plan

- [ ] Desktop: hover a bubble → tooltip, topic lines, dimmed peers appear; move away → all clear
- [ ] Desktop: click a bubble → navigates to profile page
- [ ] Mobile/touch: tap a bubble → profile card opens at current scroll position without flicker
- [ ] Mobile/touch: close card → page returns to the same scroll position
- [ ] Mobile/touch: scroll past first viewport, tap → card appears at current viewport (not top of page)
- [ ] Keyboard: Tab trapped inside card while open, Escape closes, focus returns to bubble
- [ ] Resize → viz re-initializes cleanly across all breakpoints

---

Link to sample test page: [landing page](https://foundation-s-tp1-3742-e-1ifvez.mofostaging.net/en/mozilla-expert-hub/) / [CMS edit](https://foundation-s-tp1-3742-e-1ifvez.mofostaging.netcms/pages/4/edit/)
Related PRs/issues: [Jira TP1-3742](https://mozilla-hub.atlassian.net/browse/TP1-3742)


foundation_cms/static/js/components/expert_hub_page/lightbox.js
```diff
@@ -0,0 +1,193 @@
...
+  const linkEl = cardEl.querySelector(SELECTORS.link);
+
+  if (!inner || !imageEl || !blurbEl || !nameEl || !linkEl) {
+    return null;
+  }
+
+  // Move the dialog to be a direct child of <body> so that position:fixed
+  // always resolves to the viewport, regardless of any transforms or
+  // will-change on ancestor elements inside the page content.
+  if (cardEl.parentElement !== document.body) {
```
mmmavis, line 49: Hoisted to `<body>` so `position:fixed; inset:0` always resolves to the viewport. Any ancestor with `transform` or `will-change` would otherwise hijack the containing block and pin the overlay to the top of the page.

foundation_cms/static/js/components/expert_hub_page/lightbox.js
```diff
@@ -0,0 +1,193 @@
...
+
+    cardEl.removeAttribute("hidden");
+
+    // Scroll lock: overflow:hidden stops desktop browsers; touchmove
+    // preventDefault is the only mechanism that reliably stops iOS Safari.
+    savedScrollY = window.scrollY;
+    document.documentElement.style.overflow = "hidden";
+    document.addEventListener("touchmove", preventTouchScroll, {
+      passive: false,
+    });
```
mmmavis: `overflow:hidden` alone is silently ignored by iOS Safari. The `touchmove` listener is the only mechanism that reliably prevents background scroll on iOS. Both are needed: `overflow:hidden` for desktop and `touchmove` for iOS.

foundation_cms/static/js/components/expert_hub_page/lightbox.js
```diff
@@ -0,0 +1,193 @@
...
+
+    // Defer focus to the next frame so the overlay is fully painted before
+    // iOS calculates the focused element's position (avoids scroll-to-top).
+    requestAnimationFrame(() => {
+      if (!cardEl.hasAttribute("hidden"))
+        closeBtn?.focus({ preventScroll: true });
+    });
```
mmmavis, line 164: Without the rAF delay, iOS measures the focused element's position before the overlay is painted and scrolls the page to reach it, snapping the background to the top of the document.

foundation_cms/static/js/components/expert_hub_page/viz.js
```diff
@@ -0,0 +1,482 @@
+import { forceSimulation, forceCollide, forceX, forceY } from "d3-force";
```
mmmavis, line 1: Heads up: some of the math in here (phyllotaxis spiral, force sim tuning, tooltip tail clamping) was co-authored by my AI pair programmer. It works, I promise. 😄

foundation_cms/static/js/components/expert_hub_page/viz.js
```diff
@@ -0,0 +1,482 @@
...
+    .force(
+      "collide",
+      forceCollide((d) => d.r + COLLIDE_PADDING)
+        .strength(COLLIDE_STRENGTH)
+        .iterations(COLLIDE_ITERATIONS),
+    )
+    .force("x", forceX((_, i) => nodes[i].cx).strength(ANCHOR_STRENGTH))
+    .force("y", forceY((_, i) => nodes[i].cy).strength(ANCHOR_STRENGTH))
+    .stop()
+    .tick(SIM_TICKS);
```
mmmavis, line 230: The sim runs headless via `.stop().tick()` rather than animating. We only need the final resolved positions. Running it as a live animation would cause bubbles to visibly bounce around on every page load, which was removed as it looked unpolished.

foundation_cms/static/js/components/expert_hub_page/viz.js
```diff
@@ -0,0 +1,482 @@
...
+    nodes.forEach((node) => {
+      node.el.classList.remove(CLASS_NAMES.overlayActive);
+      node.el.style.removeProperty("--overlay-color");
+    });
+  }
+
+  // ─── Events ────────────────────────────────────────────────────────────────
+
+  const ac = new AbortController();
+  const { signal } = ac;
```
mmmavis, line 348: All listeners are registered with `{ signal }` so `ac.abort()` in the teardown function removes them all at once without having to track individual handler references. This keeps the resize re-init cycle clean.

nancyt1:
@mmmavis 
1. Add 3rem to top margin 
image

2. Can we make the first bubble a bit bigger, like 10% bigger? around 245px?
3. The shape of the bubbles is an actual shape from Pentagram, could we swap to that svg as the frame rather than a regular squircle? 
4. Something is happening on mobile where the bubble doesn't anchor to the bottom if you click an expert that is near the bottom. And you can't click out of it when it's like that.
IMG_B256F58EC71C-1
5. Also just noticed on mobile, the topic tags are yellow instead of orange. Are we also able to click on the topic tag to take us to the filtered view of the index page? If that complicates the whole interaction, I might be ok to not do that.

mmmavis:
> The shape of the bubbles is an actual shape from Pentagram, could we swap to that svg as the frame rather than a regular squircle?

@nancyt1 could you link me to the shape file please?

ramram-mf:
Hey @mmmavis this is looking awesome. Here are some comments from my review:

1. I ran into a small ux hiccup while doing some crash testing: It looks like the scaling of the bubble+label could trigger the mouse in & out routines by mere geometry. This is because the margin below the label is not considered a "hover" are for the scale effect, so when this resizes it can go into a in & out loop effectively.

image

expert-hub-buble-loop

I think we can probably include the label margin into the hoverable area

2. On mobile, the scroll is not working if I drag from the viz element since it only scrolls the expert viz instead of the page's body. This only happens the first time I scroll, subsequent in-viz drag to scroll behavior works fine. This applies both to down and up scrolling, so the first time I try to scroll down, fails, and the first time I try to scroll back up, fails. 

https://github.com/user-attachments/assets/81de370f-92ee-4391-a41b-4c011240e4c9

3. When I click the expert bubble the expert profile page is giving me 500s.

I only tested keyboard for tab focusing on desktop, worked alright.

ramram-mf (approved):
This is looking great @mmmavis . UX is on point too.

mmmavis:
New review app is up now! @nancyt1

nancyt1 (approved):
Looks so good, thanks Mavis!!

PR #15860, opened 2026-05-04
Gallery Hub: Project page frontend

ramram-mf:
# Description

Implements the new Gallery Hub Project Page frontend, including the hero gallery, project metadata, project link, media layout refinements, and the colored-background CTA pattern.

image

image

*Project page*

image

image

*Admin UI*

## What changed

1. Added a responsive hero gallery:
   - Desktop thumbnail selection.
   - Mobile swipe/peek behavior.
   - Optional captions for all hero images.

2. Updated Project Page content support:
   - Added project link.
   - Added gallery images with captions and alt text.
   - Removed the old CTA link field.

3. Refined Project Page body layouts:
   - Project Page media blocks now use the intended streamfield width.
   - Widescreen media can span full width.
   - Left-only colored two-column containers now behave like one-column CTA bands.

CMS for test page: https://foundation-s-tp1-3745-p-39w4b8.mofostaging.net/cms/pages/46/edit/
Link to sample test page: https://foundation-s-tp1-3745-p-39w4b8.mofostaging.net/en/gallery/gallery-project-20/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3745)


frontend/redesign/esbuild.config.js
```diff
@@ -62,11 +67,6 @@ const sources = {
     jsx: false,
     bundle: true,
   },
-  expert_directory_page: {
```
ramram-mf, line 65: This was duplicated since last expert PR merging. Cleaning up.

beccaklam (changes requested):
Hi @ramram-mf ! Some change requests, but I also wanted to say it's looking great! Thank you for your work on this!

### Hero Image:
- [x] Videos need the pause button

### Mobile:
- [x] Can we increase the size of the hero frame? From what I could find, can we go from:

`flex: 0 0 calc(100% - 2.5rem)`
to
`flex: 0 0 calc(100% - 1rem)`
instead?

- [x] There is also a lag for the opacity to go from 50% to 100% when sliding on photos, can we decrease that lag significantly?

### CMS:
- [x] Can we move the **'Expert'** info to be directly after the **'Hero Gallery'** inputs (so before **'Program Label'**). I also wasn't able to test this because I don't think we have expert profiles available yet but please correct me if I'm wrong
- [x] I think we need to move the **'Caption' for the Hero Image** above **'Additional Hero Gallery Media'** and can we add a note that the caption is optional like how we do for the additional hero media items?

Image

ramram-mf:
> Hi @ramram-mf ! Some change requests, but I also wanted to say it's looking great! Thank you for your work on this!

Hey @beccaklam I think i got them all now. Please let me know if there is anything else we should look at.

beccaklam (changes requested):
@ramram-mf Some more requests (sorry!!):

- [x] The 50% opacity in our design system (not just here but also in the pagination PR Mavis is working on) is/was causing some weird transition effects. After a lengthy discussion, the designers agreed to streamline the design and remove the opacity (change the peek-a-boo from 0.5 back to 1) because it doesn't add functional value. (So sorry!)

- [x] Caption for hero image still feels separate in the CMS, can we group it so that it's just one unit like the additional media images:

So change from this:
Image
to this:
Image

- [x] I forgot to mention this here but can we also add infinite scroll back to the hero image carousel? I also forgot to add the carousel pagination (an oversight) but it should also be present here. I've added it to the [figma file](https://www.figma.com/design/tS7xFK32VZMiY8McfF5s14/Gallery-Feature?node-id=3000-6888&t=cCqtW2EmTTv4qXTw-4):

Image

beccaklam (changes requested):
Hi @ramram-mf everything looks good except for the mobile carousel pagination. There are four photos but only three dots. Once that's addressed we should be good to go!

beccaklam (approved):
Good to go! Thank you @ramram-mf ! 🚢

adalcima (approved):
thanks @ramram-mf , lgtm

frontend/redesign/esbuild.config.js
mmmavis: Thanks!

foundation_cms/templates/patterns/components/project_page/hero_carousel.html
```diff
@@ -0,0 +1,193 @@
...
+    {% endif %}
+
+    <div class="hero-carousel__header">
+
+        <h1>{{ page.title }}</h1>
+
+        <div class="project-meta">
+
+            {% if page.author %}
+                <div class="project-meta__author">{% trans "By" %} <a class="project-meta__author-link" href="{{ page.author.profile_url }}">{{ page.author.name }}</a></div>
```
mmmavis: `Author` only has `name`, `image`, `bio`, but no `profile_url`. We should remove the link wrapper so it doesn't renders a broken empty href.

foundation_cms/gallery_hub/models/project_page.py
```diff
@@ -56,35 +178,77 @@ class ProjectPage(AbstractArticlePage, HeroImageMixin):
...
+                        "data-condition": HeroMediaMixin.HERO_CONTENT_IMAGE,
+                    },
+                ),
+                FieldPanel(
+                    "hero_video_url",
+                    attrs={
+                        "data-media-target": "field",
+                        "data-condition": HeroMediaMixin.HERO_CONTENT_VIDEO,
+                    },
+                ),
```
mmmavis, line 205: `hero_video_url` has no URL validator unlike `video_url` above. I think we could benefit from adding `validators=[validate_vimeo_mp4_url]` to the field in `HeroMediaMixin`.

(can be a follow-up if this is out of scope for this PR)

foundation_cms/static/js/components/project_page/hero_carousel.js
```diff
@@ -0,0 +1,463 @@
...
+    this.viewport.addEventListener("pointermove", (event) =>
+      this.handlePointerMove(event),
+    );
+    this.viewport.addEventListener("pointerup", (event) =>
+      this.handlePointerUp(event),
+    );
+    this.viewport.addEventListener("pointercancel", (event) =>
+      this.handlePointerUp(event),
+    );
+    window.addEventListener("resize", () => this.handleResize());
```
mmmavis: This resize handler isn't debounced. The shared carousel.js exports `debounce()` and `RESIZE_DEBOUNCE_MS`, could we use those here?

foundation_cms/templates/patterns/components/project_page/hero_carousel.html
```diff
@@ -0,0 +1,193 @@
...
+                                height="{{ hero_img.height }}"
+                            >
+                        {% endif %}
+                    </div>
+                    {% if page.hero_image_caption %}
+                        <p class="image-block__title hero-carousel__caption">{{ page.hero_image_caption }}</p>
+                    {% endif %}
+                </figure>
+
+                {% for item in page.hero_gallery_media.all %}
```
mmmavis: `hero_gallery_media.all` is called 4 times in this template, which means 4 separate DB queries on every page load. Could we wrap it in `{% with media_items=page.hero_gallery_media.all %}` at the top and reuse `media_items` throughout?

foundation_cms/static/scss/pages/project_page.scss
```diff
@@ -0,0 +1,274 @@
...
+    }
+
+    &__link {
+      color: $black;
+
+      &::after {
+        display: inline-block;
+        content: "";
+        width: rem-calc(16px);
+        height: rem-calc(16px);
```
mmmavis:
```suggestion
        width: rem-calc(16);
        height: rem-calc(16);
```

foundation_cms/templates/patterns/components/streamfield.html
```diff
@@ -6,7 +6,7 @@
       <div class="grid-container">
         <div class="grid-x">
           {% if streamchild.block_type == 'custom_media' %}
-            <div class="cell medium-7 body-text-large">
+            <div class="cell {% if classnames %}{{ classnames }}{% else %}medium-7 body-text-large{% endif %}">
```
mmmavis, line 9: This change could break existing pages. Before this PR, `custom_media` blocks always used `medium-7 body-text-large` regardless of classnames. Now that classnames applies to `custom_media` too, pages like `podcast_page.html` (which passes `classnames="small-12 large-8"`) will get a different layout on their `custom_media` blocks unintentionally.

foundation_cms/templates/patterns/components/project_page/hero_carousel.html
```diff
@@ -0,0 +1,193 @@
...
+                    <div class="hero-carousel__frame">
+                        {% if page.displayed_hero_content == "video" and page.hero_video_url %}
+                            <video
+                                class="hero-carousel__image"
+                                data-project-hero-video
+                                playsinline
+                                muted
+                                loop
+                                preload
+                                autoplay
```
mmmavis: Is autoplaying videos intentional here?

mmmavis (commented):
Hi @ramram-mf ! The carousel implementation is thoughtfully put together and the hero gallery model is clean ✨ ! That said, there are a few things worth addressing before merging. See inline comments for details.

mmmavis (changes requested):

foundation_cms/templates/patterns/components/streamfield.html
ramram-mf: Thanks Mavis! I tested this against the pages that pass custom classnames into streamfield.html:

- Project page
- Nothing Personal article page
- Nothing Personal podcast page

The podcast page doesn’t allow custom_media, so there was no regression target there. The NP article page passes medium-7 body-text-large, which matches the previous custom_media wrapper exactly. The project page is the only place where the change intentionally affects custom_media, so it now aligns with the rest of the project page body content.

np-article-desktop

*NP Article layout with custom media block*

np-podcast-desktop

*NP Podcast layout, no custom media block to show in here*

project-page-classnames-desktop

*Project page layout with custom media block*

foundation_cms/templates/patterns/components/project_page/hero_carousel.html
ramram-mf: Yes

foundation_cms/gallery_hub/models/project_page.py
ramram-mf: Added the validation to the model.

foundation_cms/templates/patterns/components/project_page/hero_carousel.html
ramram-mf: I pivoted to only using the expert field since that is what the model uses semantically for this page type.

mmmavis (approved):
Thanks for addressing the issues. PR looks great now. @ramram-mf 🍻

PR #15870, opened 2026-05-04
Removed char limits from PR 15804

danielfmiranda:
# Description

This PR removes the character limits set in #15804


foundation_cms/blocks/portrait_card_block.py
```diff
@@ -6,12 +6,8 @@
...
 class PortraitCardBlock(BaseCardBlock):
 
-    headline = CharBlock(
-        required=False, max_length=36, label="Headline", help_text="Appears as the main heading (max 36 characters)."
-    )
-    cta_link = LinkWithDynamicLabelBlock(
-        label_max_length=36, required=True, label="Call to Action Link", help_text="Link for the card."
-    )
+    headline = CharBlock(required=False, label="Headline", help_text="Appears as the main heading.")
+    cta_link = LinkWithDynamicLabelBlock(required=True, label="Call to Action Link", help_text="Link for the card.")
```
adalcima: Thanks @danielfmiranda ,I think it would be necessary to use `LinkBlock()` again in the links since `LinkWithDynamicLabelBlock()` uses [label_max_length=36](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/blocks/link_block.py#L98) by default. Another option could be to update `LinkWithDynamicLabelBlock` to not use a default value.

foundation_cms/blocks/timely_activations_cards_block.py
```diff
@@ -12,15 +12,11 @@ class ActivationCardBlock(BaseBlock):
...
-    title = blocks.CharBlock(required=True, max_length=60, help_text="Title for the card (max 60 characters).")
-    text = CustomRichTextBlock(
-        required=False, max_length=180, help_text="Optional description text (max 180 characters)."
-    )
+    category = blocks.CharBlock(required=False, help_text="Optional category for the card.")
+    title = blocks.CharBlock(required=True, help_text="Title for the card.")
+    text = CustomRichTextBlock(required=False, help_text="Optional description text.")
     image = ImageBlock(required=True)
-    link = LinkWithDynamicLabelBlock(label_max_length=14)
+    link = LinkWithDynamicLabelBlock()
```
adalcima: Something similar to the other link [here](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/15870/changes#r3190624036)

adalcima (commented):
hi @danielfmiranda I think it would be necessary to use `LinkBlock()` again in the links since `LinkWithDynamicLabelBlock()` uses label_max_length=36 by default. Another option could be to update `LinkWithDynamicLabelBlock` to not use a default value.

danielfmiranda:
Thanks for the great point @adalcima! 

I have updated the fields to use LinkBlock again and updated migrations. The PR is now ready for review whenever you have a free moment :D Thanks again!

adalcima (approved):
lgtm, thanks for working on this @danielfmiranda

PR #15885, opened 2026-05-06
Django 5.2 upgrade

danielfmiranda:
# Description

This PR is a copy of the [Rob's PR](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/15771) with an [additional commit ](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/15771) that updates our settings file to match whats needed Django 5.2 in order for the images to work on review apps which was the issue that he was experiencing. 

# Testing 

A review app can be found here: https://foundation-s-django-5-2-xqybyq.herokuapp.com/

**Steps to test:**
1. Click around the site, and check if anything is looks out of the ordinary.
2. Visit the CMS using SSO. Check if anything looks out of the ordinary
3. If not, we are good to go!


kevdmoz:
Hi @danielfmiranda 

When I click on pages in the navigation, it leads to a 404. Is that expected? It might be since there are only the Gallery and Expert pages in the review app Example: https://foundation-s-django-5-2-xqybyq.mofostaging.net/en/meet-mozilla/

I was able to log into the CMS using SSO.

Was there anything specific you wanted me to test? It's hard to tell what I should be looking for since navigating to pages are broken and email sign up does not work for me.

danielfmiranda:
Thanks @kevdmoz! That is expected since there are no pages that live at those links right now. In regards to what you should test, nothing really major besides just making sure that the page and the cms seem to be working as they should!

kevdmoz (approved):

nancyt1 (approved):
Seems to work ok for me!

adalcima (approved):
It looks good, thanks @danielfmiranda !
I can see the factory images from Gallery and Expert in the review app. Both the site and the CMS have been working for me 🚀

ramram-mf (approved):
Thanks @danielfmiranda ! All images I could think of were displaying correctly. I think this is solid.

PR #15928, opened 2026-05-12
fix(build): use shx cp -r for fonts copy to fix Linux CI failure

mmmavis:
## Problem

`build:legacy:fonts:fonts` uses native `cp -r` to copy fonts into the compiled directory. On Linux (GitHub Actions), `cp` can exit before its writes are fully visible to subsequent processes, causing the parallel font CSS minification tasks to fail:

```
rm: no such file or directory: .../compiled/_fonts/css/changa.css
ERROR: "build:legacy:fonts:css:changa" exited with 1.
```

## Fix

- `cp -r` → `shx cp -r` in `build:legacy:fonts:fonts` (Node.js fs APIs flush synchronously before process exit)
- `shx rm` → `shx rm -f` in the three CSS cleanup steps as a safety net

## Test plan
- [ ] `yarn workspace legacy build` passes locally
- [ ] CI passes


danielfmiranda (approved):
Thanks @mmmavis! Confirming that this worked for me on my local machine 👍 R+!

PR #15954, opened 2026-05-19
Add gallery hub filter panel

ramram-mf:
# Description

This PR implements the filter feature for the Gallery Hub landing page

image

*Gallery page filter feature*

Link to sample test page: https://foundation-s-tp1-3892-g-gqcovp.mofostaging.net/en/gallery/
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3892)


beccaklam (changes requested):
Hi @ramram-mf this is looking great! Some issues though:

- [x] the program, country and year filters should actually be styled with check boxes (since they are not topic tags and that pill style is particular to topic tags):

Image

- [x] I also can't scroll on mobile when I test the page on my phone?

- [x] The 'display results' button should also having 1rem of padding all around it:

Image

foundation_cms/static/js/components/gallery_hub/overlay.js
```diff
@@ -28,8 +27,95 @@ const FOCUSABLE_SELECTOR = [
...
+  modalCloseTimers.set(modal, timer);
+
+  animationTarget.addEventListener(
+    "animationend",
+    (event) => {
+      if (event.target !== animationTarget) return;
+
+      finish();
+    },
+    { once: true },
```
mmmavis: `animationend` bubbles, so if any child of `animationTarget` finishes its own CSS animation first, the `{ once: true }` listener is consumed before the target element's animation ends. The `event.target !== animationTarget` guard returns early but the listener is already gone, so the real animation end goes unheard and the 360ms fallback timer ends up driving the hide instead.

foundation_cms/static/js/components/gallery_hub/filter_panel.js
```diff
@@ -0,0 +1,307 @@
...
+const FILTER_MODAL_ID = "filter";
+const PROJECT_FILTER_DATA_ID = "gallery-hub-project-filter-data";
+
+/**
+ * Clone active filter arrays.
+ *
+ * @param {Object<string, string[]>} filters - Filter state to clone.
+ * @returns {Object<string, string[]>}
+ */
+function cloneFilters(filters) {
```
mmmavis: `function cloneFilters` is defined twice. We should probably export it from `state.js` and import it into `filter_panel.js` instead of redefining it?

foundation_cms/static/js/components/gallery_hub/state.js
```diff
@@ -31,6 +31,21 @@ const DEFAULT_STATE = {
 
 let state = { ...DEFAULT_STATE };
 
+/**
+ * Clone filter state so array values are never shared between modules.
+ *
+ * @param {Object<string, *>} filters - Filter state to clone.
+ * @returns {Object<string, *>}
+ */
+function cloneFilters(filters) {
```
mmmavis: `function cloneFilters` is defined twice. We should probably export it from `state.js` and import it into `filter_panel.js` instead of redefining it?

foundation_cms/static/js/components/gallery_hub/overlay.js
```diff
@@ -28,8 +27,95 @@ const FOCUSABLE_SELECTOR = [
   "[tabindex]:not([tabindex='-1'])",
 ].join(",");
 
+const ANIMATED_MODAL_IDS = new Set(["project-list", "filter"]);
```
mmmavis: A bit worried about these two magic strings being embedded in the code. Could we extract them into `config.js`?

foundation_cms/templates/patterns/pages/gallery_hub/gallery_page.html
```diff
@@ -189,12 +189,87 @@ <h1 id="gallery-hub-title" class="gallery-hub__title">{{ page.title }}</h1>
...
+                                                <span>
+                                                    {% if category.key == "topic" %}
+                                                        {% trans "Topic" %}
+                                                    {% elif category.key == "program" %}
+                                                        {% trans "Program" %}
+                                                    {% elif category.key == "year" %}
+                                                        {% trans "Year" %}
+                                                    {% elif category.key == "country" %}
+                                                        {% trans "Country" %}
+                                                    {% endif %}
```
mmmavis: The view already builds `filter_categories` as a structured dict per category. The label mapping should belong there too. What do you think about adding a `label` field in `gallery_page.py` and use `category.label` in the template? This means adding a new filter cateogry won't require us to update both the view and the template, and the template has no fallback(renders nothing) if a key doesn't match any condition.

foundation_cms/static/js/components/gallery_hub/filter_panel.js
```diff
@@ -0,0 +1,307 @@
...
+    if (isFilterOpen && !wasFilterOpen) {
+      draftFilters = cloneFilters(state.activeFilters);
+      syncPanel();
+    }
+
+    wasFilterOpen = isFilterOpen;
+  });
+
+  if (Object.keys(draftFilters).length) {
+    commitFilters(draftFilters, false);
```
mmmavis:
```suggestion
    commitFilters(draftFilters, false, true);
```
so it respects the query params in the url on page load?

mmmavis (changes requested):
Love it! Functionality and UX/UI are all very sleek I left a few inline comments for small things. I did find one bug - when the filter panel is open, the site nav and search are still expandable but their dropdowns appear behind the filter overlay. @beccaklam , what should the expected behaviour be? My instinct is the filter should close when the user opens the nav or search, but will rely on your expertise to settle on the decision

beccaklam (changes requested):
Hi @ramram-mf , thank you for making those changes! Just a few more requests upon a second inspection:

- [x] The icons in the filter should be 32x32 instead of 24x24
- [x] On desktop can we decrease the spacing between topic tags (gallery-hub-filter__options) to `gap: 1rem .5rem` from `gap: 1.5rem 1rem`
- [x] Can we also decrease the gap between the check boxes on desktop from 1.5rem to 1rem?
- [x] Also, my apologies, I should have specified that moving the 'Project List' and 'Filter' buttons pertained only to Mobile. Mobile looks good now but Desktop should look like this (note the red lines for alignment):

Image

beccaklam (approved):
This looks good to go 🚢 🚀 !!! Thank you @ramram-mf for implementing the changes so quickly. It looks like you already added the changes @mmmavis mentioned too and those look good to me.

foundation_cms/static/js/components/gallery_hub/overlay.js
```diff
@@ -270,30 +378,40 @@ export function initGalleryHubOverlay() {
 
       lastFocusedElement = toggle;
 
-      if (state.modalOpen === modal) {
-        closeOpenModal(modals);
-        return;
+      if (state.modalOpen !== modal) {
+        closePrimaryNavOverlays();
```
mmmavis, line 409: Spotted a subtle issue when switching between open modals. `closePrimaryNavOverlays()` dispatches `primaryNavWillOpen` synchronously, which triggers the gallery's own listener and calls `setGalleryHubState({ modalOpen: null })` right before `setGalleryHubState({ modalOpen: <new modal> })` runs here. So we get two consecutive `syncModal` calls in the same tick: the first starts the close animation and sets a `WeakMap` timer, and the second cancels that timer and attaches a second `animationend` listener on the same target, which leaves the old modal's animation lifecycle in a weird state.

We could fix this by skipping the `primaryNavWillOpen` dispatch when a gallery modal is already open (since in that case the gallery is the one initiating the close, not reacting to the nav):

```js
function closePrimaryNavOverlays() {
  closePrimaryNavDropdowns();
  if (!getGalleryHubState().modalOpen) {
    document.dispatchEvent(new CustomEvent(PRIMARY_NAV_EVENTS.primaryNavWillOpen));
  }
}

mmmavis (changes requested):
Super close to finish line!

mmmavis (approved):
🎉 🎉 🎉

adalcima (approved):
Thanks Ram, lgtm. It's good to see that some data is generated as tuples, I think that will help the performance of the queries a bit.

PR #15969, opened 2026-05-21
Extend search index to improve and include body content

adalcima:
# Description

This PR fixes duplicate body fields, boost standardization and **_multilingual search quality stemming_** conflicts. All search backends (`en`, `de`, `es`, etc.) shared `AUTO_UPDATE: True`, causing each backend to overwrite the others' `tsvectors` on every page publish.

Related issue: [TP1-3897](https://mozilla-hub.atlassian.net/browse/TP1-3897)
Review app: https://foundation-s-tp1-3897-e-cavbag.mofostaging.net/en/

####Main Changes
1. Duplicate body `search_fields` removed and missing `search_fields` added.
2. Boost standardization on `AbstractBasePage`. Titles, SEO fields, and body contain the highest weights (10 and 8)
3. Fix multilingual stemming conflict with `AUTO_UPDATE: False` on search backends.
4. Per-locale indexing via `signals` 
5. New management command `update_multilingual_index` available (to repair or re-index when necessary and based on locale, it would be an alternative to the native `update_index` command)

#### How to test
1. New command 
2. Create a new page, for example, a `General page`. Add a `title`, descriptions, required fields, and `body` content.
3. Publish the created page and verify the searches using terms added to the page.

### Images
- Pages created for testing
image

image
image
image


foundation_cms/settings/base.py
adalcima, line 1: Auto-update was the mechanism causing each backend to overwrite the others' `tsvectors` on every page publish, so `AUTO_UPDATE: False` is set to fix multilingual stemming conflict.

- The last backend to write wins. In practice, the default (simple/no-stemming) backend consistently overwrote the language-specific ones, so an English page ended up stored as 'donate' instead of 'donat' — making English FTS queries with stemming return 0 results.

foundation_cms/search/signals.py
adalcima, line 1: Connects `page_published` / `page_unpublished` to route each page to its correct backend based on `page.locale.language_code`. Uses dispatch_uid to prevent duplicates.

- Signal handler that reads `page locale` and routes the page to one backend. This means each page gets indexed once, with the right stemming configuration.

danielfmiranda (approved):
Excellent work @adalcima! LGTM 👍

Issue #15967, opened 2026-05-21
 A/B Test: Pushdown Banner vs. Lightbox Popup


data-sync-user:
┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-3945)

PR #15986, opened 2026-05-25
feat(donate-banner): add lightbox as new donate banner option

mmmavis:
Link to sample test page: [review app](https://foundation-s-tp1-3945-d-cvsocf.mofostaging.net/en/) / [CMS edit](https://foundation-s-tp1-3945-d-cvsocf.mofostaging.net/cms/pages/534/edit/)
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3945) / #15967 

# Description

Add lightbox as new donate banner option

This PR also fixes a spacing regression on the "legacy" and "pushdown" versions of banners.

# Test Instructions
- [ ] Ensure a new banner style "lightbox" is shown for [Donate Banner Snippet](https://foundation-s-tp1-3945-d-cvsocf.mofostaging.net/cms/snippets/snippets/donatebanner/add/?locale=en)
- [ ] Visit the review app and verify functionality and design for the the donate lightbox 

# Follow-up

- Move donate-related JS, SCSS, and templates into `donate/` subfolders within their respective asset directories (e.g. `components/donate/` instead of `components/donate_banner/`). Will file follow-up ticket.


foundation_cms/templates/patterns/components/donate_banner/donate_banner_base.html
```diff
@@ -9,31 +9,33 @@
      data-banner-style="{{ banner.banner_style }}"
 >
     {% block top_button_wrapper %}{% endblock top_button_wrapper %}
-    <div class="grid-container donate-banner__inner-wrapper">
```
mmmavis, line 12: Changes in this file is mainly

```
-    <div class="grid-container donate-banner__inner-wrapper">
+    <div class="donate-banner__inner-wrapper">
+        <div class="grid-container">
```
and the rest is just indentation difference.

This is to fix a regression issue with the legacy and pushdown version where the "white box" was narrower than it should be. (White box should extend 64px more on both left and right side)

Image

nancyt1:
@mmmavis Looks great to me but where do i go to edit it? the cms link didn't work

(EDIT: JK I found it ahah)

mmmavis:
Hey @nancyt1 I fixed the link in the PR description. I accidentally pasted the link on my local site 😅

nancyt1:
I'm confused about how the images are cropped. The image I uploaded is really similar ratio to the cat one but it zooms in a lot more. Can you tell me what's happening here? @mmmavis

foundation_cms/static/scss/components/donate_banner/donate_banner.scss
```diff
@@ -50,6 +50,7 @@
 
   &__inner-wrapper {
     background: $orange;
+    margin: 0 auto;
```
mmmavis, line 53: fix regression

foundation_cms/static/scss/components/donate_banner/donate_banner.scss
```diff
@@ -59,6 +60,10 @@
       max-width: 89vw;
     }
 
+    @include breakpoint(xlarge up) {
```
mmmavis, line 63: fix regression

mmmavis:
> I'm confused about how the images are cropped. The image I uploaded is really similar ratio to the cat one but it zooms in a lot more. Can you tell me what's happening here? @mmmavis

Hey @nancyt1 ,

Good question. The cat image is a WebP file, which our image component serve as-is without any cropping (not sure about the historical reason, but we can spin up another convo on this). Your image is a jpeg/png which gets automatically cropped by the CMS to fill a 4:3 frame. So depending on the original image's proportions, it can end up zoomed in quite a bit.

There are two ways to fix it:

1. Set a focal point on the image in Wagtail
2. Use a source image close to 4:3 ratio.

Hope this helps!

nancyt1:
@mmmavis could we use the ratio 3:2 instead? to match the rest of the site?

mmmavis:
PR updated! @nancyt1 @ramram-mf

nancyt1 (approved):
Looks great, thank you!!

ramram-mf (changes requested):
I think we also need to initialize the lightbox in `redesign_migrated_content.js`.

The migrated legacy base template can render the redesigned donate banner via `patterns/components/donate_banner/donate_banner.html`, and this PR already adds the lightbox SCSS to `redesign_migrated_content.scss`. But the migrated JS bundle still only calls `initDonateBanner()`, while `initDonateLightbox()` is only wired into `redesign_main.js`.

That means if a migrated page gets a donate banner with `banner_style == "lightbox"`, the `<dialog>` markup will be present but `showModal()` never runs, so the banner won’t appear. I think `redesign_migrated_content.js` should import and call `initDonateLightbox()` too.

Related small thing: the migrated template include does not pass `cta_button_data`, so the lightbox CTA may also miss `data-donate-banner-cta-button` on that path, which would prevent the lightbox CTA click handler/A-B event from attaching.

mmmavis:
@ramram-mf Thanks for catching those two issues! I def overlooked setup for migrated pages. PR updated!

ramram-mf (approved):
Looking great @mmmavis thanks!

PR #15987, opened 2026-05-26
Build expert profile page frontend

ramram-mf:
# Description

This PR implements the Expert Profile page frontend.

It adds the profile hero, structured bio/quote + image layout, optional curated projects section, curated articles list with show-more behavior, external links section, and the shared Expert Hub CTA.

It also updates the Expert Profile model/admin fields and factory data so `expert-1` has representative bio, quote, articles, projects, and external links for review.

## Screenshots

image

*`large` viewport Profile page*

Link to sample test page: https://foundation-s-tp1-3743-e-ebmf91.mofostaging.net/en/mozilla-expert-hub/expert-1/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3743)


ramram-mf:
One modeling question for review:

The profile intro currently renders from `page.body` for the content to the left of the profile image. Layout, markup, and styling are covered for the current design using rich text + quote streamblocks there.

I want to confirm whether that is the intended content model, or whether this area should instead be driven by dedicated page fields, for example the existing `bio` field plus a separate quote/attribution field.

If the body streamfield is the right source, this is ready as-is. If the intent is for editors to manage this content through structured profile fields, I can adjust the implementation so the frontend maps to that model more explicitly.

kevdmoz:
@ramram-mf my two-cents on the model question: for a first iteration with no content yet, structured fields (bio, quote, attribution) are the better choice. They're simpler for editors to fill out consistently across profiles, and they keep the content reusable — bio text can feed meta descriptions, cards, or future components without parsing StreamField JSON. Since the profile layout is consistent by design, there's no flexibility being lost by moving away from StreamField here. Better to make the switch now than after profiles are populated. Does this swap mean a lot of re-work for you?

nancyt1:
@ramram-mf Dude, this looks exactly like the design! Haha amazing work!! But seeing it now, could we make the name be the first thing? And increase the margin-bottom of the name and title to 1rem?

+1 to what Kevin said. If we can't do that, I don't mind it being streamfields but can we remove the ability to add additional blocks to the body?

image

foundation_cms/profiles/models/expert_profile_page.py
```diff
@@ -35,6 +45,44 @@ class Meta(TranslatableMixin.Meta, Orderable.Meta):
...
+    article = models.ForeignKey(
+        Page,
+        related_name="expert_profile_article_selections",
+        null=True,
+        blank=True,
+        on_delete=models.SET_NULL,
+    )
+
+    panels = [
+        PageChooserPanel("article"),
```
mmmavis: do we want to restrict the page chooser to only allow certain types of pages? if so, we should specify the 2nd argument

foundation_cms/static/js/components/expert_profile_page/article_list.js
```diff
@@ -0,0 +1,61 @@
+const SELECTORS = {
+  list: "[data-expert-profile-article-list]",
+  item: ".listing-content__item",
+  showMoreButton: "[data-expert-profile-show-articles]",
+};
+
+const VISIBLE_ARTICLE_COUNT = 3;
```
mmmavis: The 3 is hardcoded in both the JS and the template. Could we pass it as a `data-` attribute (e.g. `data-visible-count`) from the backend so the JS reads from it instead?

foundation_cms/profiles/tests/test_models.py
```diff
@@ -24,6 +30,90 @@ def test_required_fields_populated(self):
...
+        self._create_project(gallery, "First Project", "first-project", expert=self.page)
+        self._create_project(gallery, "Second Project", "second-project", expert=self.page)
+
+        rows = self.page.get_project_block_rows()
+
+        self.assertEqual(rows, [])
+
+    def test_get_selected_articles_returns_pages_in_editor_order(self):
+        first_article = ExpertDirectoryPageFactory(parent=self.hub, title="First Article")
+        second_article = ExpertDirectoryPageFactory(parent=self.hub, title="Second Article")
```
mmmavis: wrong page type? ExpertDirectoryPage is not an article page 😅

foundation_cms/profiles/models/expert_profile_page.py
```diff
@@ -76,15 +148,52 @@ class Meta:
...
+        project_blocks = [
+            {
+                "project": project,
+                "show_description": True,
+            }
+            for project in self.get_profile_projects()
+        ]
+
+        complete_row_count = len(project_blocks) // 3
+        return [project_blocks[index : index + 3] for index in range(0, complete_row_count * 3, 3)]
```
mmmavis: could we extract the 3 into a named constant so it's clear what it represents?

mmmavis: could we add a comment here noting that the 3 is tied to the `three-column-container` CSS class in the template, so changing it requires a CSS update too?

mmmavis (changes requested):
Nice work @ramram-mf on both frontend and backend! Left a few comments, mostly around keeping things in sync and clarifying intent in a few spots.

mmmavis (approved):
Thank you @ramram-mf !

adalcima (approved):
lgtm, thanks @ramram-mf

nancyt1 (approved):
Looks great, thank you!!

PR #15996, opened 2026-05-28
ci: split backend linting into its own job to reduce test_wagtail duration

mmmavis:
Related PRs/issues: [Jira Ticket](https://mozilla-hub.atlassian.net/browse/TP1-3957)

---

Backend linting and mypy no longer run inside the heavyweight `test_wagtail job` (which needs Postgres, Node, and a full build preroll). They now run in a lean parallel job that only requires Python dependencies, giving faster lint feedback and trimming the `test_wagtail` critical path.

A separate check for backend linting should appear:
image


adalcima (approved):
Lgtm, thanks for working in this improvement @mmmavis

PR #16025, opened 2026-06-03
Tp1 3975 fix bug hero media validation error on translated pages when publishing

adalcima:
# Description
Fix `ValidationError` on page publish caused by inconsistent hero content data from `translated` pages(existing rows in the DB retained the old `default=image` without having an actual `hero_image` set). The error in the logs looks like this::

`django.core.exceptions.ValidationError: {
  'hero_image': ['Image was chosen as displayed hero content, but no image is set.']
}`

Link to review app: https://foundation-s-tp1-3975-f-pwkikd.mofostaging.net/cms/pages/31613/edit/
Related PRs/issues: [TP1-3975](https://mozilla-hub.atlassian.net/browse/TP1-3975)

**Changes in Models:**
- Data migration per affected app that resets `displayed_hero_content` to `""` (No media) for all rows where it is "image" but hero_image is NULL.

**Tests**
1. Publish any page that previously failed — translations should republish without ValidationError.
2. Try using media or not. The page should be published normally.

### Images
image
image


kevdmoz (approved):
Approved! I was able to add "No media" to the hero and was able to publish my changes without a hero image

mmmavis (approved):
Thank you @adalcima !

PR #16099, opened 2026-06-15
Upgrade preview apps to Heroku 24

alfredo-medina-stx:
# Description

 Preparatory, no-op-on-heroku-22 changes to unblock the upcoming Heroku-24 stack upgrade, and clear a Heroku deprecation.

  Three things needed fixing before we can move stacks:

  1. **`runtime.txt` is deprecated.** Heroku's Python buildpack has deprecated `runtime.txt` in favor of a `.python-version` file. It was also pinning an exact patch (`python-3.11.7`), which blocks automatic security patches.
  2. **`app.json` pinned `heroku-22`**, so new review apps would keep building on the old stack.
  3. **`git` at runtime.** Heroku-24 makes `git` a **build-time-only** tool. `wagtail-localize-git` imports GitPython (`from git import Repo`) at `django.setup()` time, so without a runtime `git` binary every dyno boot and release command crashes with `ImportError: Bad git executable`. Confirmed on a heroku-24 review app:
 ```
     File ".../wagtail_localize_git/git.py", line 7, in <module>
         from git import Repo
     ImportError: Failed to initialize: Bad git executable.
```

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented?

**Merge Method**

### CODE REVIEW

File: .gitignore [hunk 1]
```diff
@@ -82,6 +82,8 @@ target/
 
 # pyenv
 .python-version
+# ...but track our committed version: Heroku reads .python-version to pick the Python runtime.
```
mmmavis: Instead of adding new lines, I think we can just remove line 83 and line 84? Thoughts?

mmmavis (changes requested):
Hi @alfredo-medina-stx , thanks for the PR. I just have one question regarding .gitignore. Everything else looks good to me!

.gitignore
alfredo-medina-stx: Yeah, just thought it was a good idea to leave that comment to explain *why* we decided to ignore the ignore (so, stop ignoring? hah). But on second thought, you're right, seems a bit unnecessary. Plus, this PR (*and* the ticket on Jira, *and* the message on Slack) can serve as the explanation.

Deleted the whole thing.

mmmavis (approved):
🎉

ramram-mf (approved):
Thanks Alfredo! Lgtm

PR #16103, opened 2026-06-16
Update autocomplete suggestions backend

adalcima:
# Description

This PR updates the backend search `Autocomplete` behavior by enforcing a minimum query length from settings and exposing that value to templates for frontend.
It also expands `autocomplete` `indexing` coverage so suggestions can be powered by additional page metadata, while keeping current API output unchanged.

Link to sample test page: https://foundation-s-tp1-3442-a-xw9syh.mofostaging.net/en/search/autocomplete/?query=mozilla
Related PRs/issues: # [TP1-3442](https://mozilla-hub.atlassian.net/browse/TP1-3442)

#### Main changes
-  New `SEARCH_AUTOCOMPLETE_MIN_CHARS` env var defaults to 5.
- Added backend guard  to only return suggestions when query length is greater than or equal to `SEARCH_AUTOCOMPLETE_MIN_CHARS`.
- Exposed `autocomplete_min_chars` in the search page context so frontend can use the same threshold in a future UI integration.
- Kept autocomplete response shape unchanged  and result cap at 5 items.

#### Current limitations

- Suggestions are currently based on page title autocomplete only.
- No spelling correction, fuzzy matching, or synonyms in this implementation.
- Frontend behavior must explicitly consume `autocomplete_min_chars` to stay fully aligned with backend threshold.
- `Autocomplete` only works with simple text fields in the model — not with StreamFields or RelatedFields

**Tests**
1. Call endpoint with a short query (below threshold), expect empty results:
GET /search/autocomplete/?query=mo
2. Call endpoint with query meeting threshold, expect suggestions:
GET /search/autocomplete/?query=mozil
3. Validate response format:
JSON includes results list with  `title` and `url`.

###  Images
image

- No results
image


env.default
```diff
@@ -107,3 +107,6 @@ MOZFEST_SCHEDULE_URL=https://schedule.mozillafestival.org/plaza
 
 # Temporary: comma-separated page IDs shown as related content on search no-results page. E.g., 123,45678
 TEMP_SEARCH_RELATED_CONTENT_PAGE_IDS=
+
+# Search
+SEARCH_AUTOCOMPLETE_MIN_CHARS=5
```
mmmavis, line 112: Could we add a new section "Search" in the env var section in README to include this new env var, please?

mmmavis (changes requested):
Sorry for the wait, @adalcima . Code looks great. I just have one ask to update README to include the new env var. Everything else looks sharp 🌟

env.default
adalcima: > Could we add a new section "Search" in the env var section in README to include this new env var, please?

Thanks for the review @mmmavis , I 've updated the Readme with the new variable :+1:

mmmavis (approved):
Thank you!

PR #16134, opened 2026-06-23
TP1-4016 Fix nav search with pushdown banner

ramram-mf:
# Description

Fixes the redesigned primary nav search drawer when a Donation Pushdown Banner is active.

image

Previously, opening search while the pushdown banner was visible caused the search drawer/input to overlap the donation banner area. The search UI was positioned from the default nav height instead of the nav’s actual rendered position after the banner pushed it down.

This update measures the current bottom edge of the primary nav when search opens and stores it in `--primary-nav-search-top`. The search drawer and backdrop now use that value so they anchor directly below the nav. The value is also refreshed on scroll/resize while search is open so it stays aligned as the pushdown banner/page position changes.

A real pushdown donate banner has been created on the review app so this can be tested against the actual CMS-driven banner state.

Link to sample test page: https://foundation-s-tp1-4016-s-ji5r5r.mofostaging.net/en/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-4016)

### DIFF
--- a/foundation_cms/static/js/components/primary_nav/search.js
+++ b/foundation_cms/static/js/components/primary_nav/search.js
@@ -1,17 +1,26 @@
 import { CLASSNAMES, EVENTS, SELECTORS } from "./config.js";
 
+const SEARCH_TOP_CUSTOM_PROPERTY = "--primary-nav-search-top";
+
 /**
  * Controls the open/close state for the primary-nav search drawer.
  */
 class PrimaryNavSearchDrawerController {
   /**
-   * @param {{ searchToggleEl: HTMLElement, searchInputContainerEl: HTMLElement, searchInputEl: HTMLInputElement }} options
+   * @param {{ searchToggleEl: HTMLElement, searchInputContainerEl: HTMLElement, searchInputEl: HTMLInputElement, primaryNavEl?: HTMLElement | null }} options
    */
-  constructor({ searchToggleEl, searchInputContainerEl, searchInputEl }) {
+  constructor({
+    searchToggleEl,
+    searchInputContainerEl,
+    searchInputEl,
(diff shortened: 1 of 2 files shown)


foundation_cms/static/js/components/primary_nav/search.js
```diff
@@ -26,16 +35,25 @@ class PrimaryNavSearchDrawerController {
...
+    this.updateSearchTop();
     this.searchToggleEl.classList.add(CLASSNAMES.searchOpen);
     this.searchInputContainerEl.classList.add(CLASSNAMES.searchOpen);
-    this.searchInputContainerEl.style.maxHeight =
-      this.searchInputContainerEl.scrollHeight + "px";
+    this.updateDrawerHeight();
     this.searchInputContainerEl.setAttribute("aria-expanded", "true");
+    this.addViewportListeners();
     this.searchInputEl.focus({ preventScroll: true });
+
```
mmmavis, line 45: Could we add a short comment here explaining why the rAF pass is needed? This will help clarify the intend for future readers.

mmmavis (approved):
Thanks @ramram-mf ! Giving this a R+ but if we could add inline comment for rAF part, it would be extra awesome!

foundation_cms/static/js/components/primary_nav/search.js
ramram-mf: Good call. The first measurement happens before opening so the drawer has a sane top value immediately, and the rAF pass re-measures after the open classes and any `searchWillOpen` listeners have had a frame to affect layout. That catches cases like the pushdown banner/nav position settling after the DOM update. I will add an inline comment to avoid pr spelunking.

PR #16135, opened 2026-06-23
Fix CSRF token caching issue

alfredo-medina-stx:
# Description

The site is fronted by a CDN that full-page-caches anonymous HTML. Any form that renders `{% csrf_token %}` bakes one visitor's token into the shared cached page, so every subsequent visitor submits a token that doesn't match their own csrftoken cookie, thus giving an HTTP 403 error.

This reproduced reliably on the petition step-2 share/skip flow (`?state=signed`), and was the same latent root cause behind several other forms that had been worked around with `@csrf_exempt` or a custom CSRF middleware. The petition flow was being held together by a temporary Cloudflare `state=signed` cache bypass.

So, the fix was to stop rendering `{% csrf_token %}` into cacheable HTML. Instead, at submit time JS reads the per-user `csrftoken` cookie (minting it via the uncached `/api/csrf/` endpoint when absent) and supplies the token as either:
  - the` X-CSRFToken` header (for fetch-based forms), or
  - a JS-populated hidden `csrfmiddlewaretoken` field (for native form POSTs that navigate/redirect).

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented?

**Merge Method**


foundation_cms/legacy_apps/static/js/common/csrf.js
```diff
@@ -0,0 +1,44 @@
...
+ * `/api/csrf/` endpoint for first-time visitors who landed on a cached page.
+ *
+ * @returns {Promise<string>} The `csrftoken` cookie value (may be "" if minting failed).
+ */
+export async function ensureCsrfToken() {
+  let token = getCookie(CSRF_COOKIE_NAME);
+  if (token) return token;
+
+  try {
+    await fetch(CSRF_MINT_URL, { credentials: "same-origin" });
```
mmmavis:
```suggestion
    await fetch(CSRF_MINT_URL, { credentials: "same-origin" });
    if (!res.ok) return "";
```
`fetch` only throws on network errors, so a non-2xx response (500, 404) would bypass the catch and leave the cookie unset. Would it make sense to add if `(!res.ok) return "";` after the await? Same change might be worth applying to legacy_apps/static/js/common/csrf.js as well.

foundation_cms/legacy_apps/templates/fragments/language_switcher.html
```diff
@@ -20,10 +22,31 @@
 
 
 <script nonce="{{ request.csp_nonce }}">
-    const form = document.getElementById('language-switcher-form')
```
mmmavis, line 23: The `getCookie` + `ensureCsrfToken` logic is duplicated across 5 places (2 JS modules + 3 template inline scripts). Since these tempaltes can't import from the bundle, would it be worth extracting this into a small standalone `type="module"` script that the template can `<script src>` instead? That way there's one source of truth and any future fix (cookie name, mint URL, response handling) only needs to land in one place (or possibly two places if legacy and redesign codebase each needs their own copy)

mmmavis (changes requested):
Thank you @alfredo-medina-stx for addressing this issue. Appreciate the inline documentation too! 

I reviewed the JS and template changes and left a few inline comments. I'll rely on @adalcima 's expertise to review the backend code changes. 😁

mmmavis (commented):
Hi @alfredo-medina-stx , looks like `get_descendants` is broken after the merge with `main` 😬 . Could you take a look?

foundation_cms/urls.py
```diff
@@ -65,7 +65,16 @@ def get_robots_content():
...
+@ensure_csrf_cookie
 def csrf_response(request):
+    # Mints the per-user `csrftoken` cookie (via @ensure_csrf_cookie) so JS can read it
+    # and send it as the X-CSRFToken header / csrfmiddlewaretoken field. The HTML body
+    # (a single hidden input rendered by {% csrf_token %}) is preserved because the legacy
+    # buyers-guide vote JS parses the token out of it.
+    #
+    # NOTE: the Cloudflare Worker overrides origin Cache-Control, so this `no-cache` is only
+    # defense-in-depth — `/api/csrf` must also be in the Worker's URI_BYPASS_SUBSTRINGS or the
+    # endpoint gets cached and hands every visitor the same stale token.
```
robdivincenzo, line 78: Seems like a critical step for this to work, so let's either set this up now or at least have a go-live checklist.

robdivincenzo (approved):
I think the approach here is sound @alfredo-medina-stx ! Please see @mmmavis comment about `get_descendents` but once resolved I'd consider approved from my eyes. Also had one comment about needing the worker path to uncache the csrf api path. Eager to see the proof of concept and test it out with a CDN!

foundation_cms/urls.py
alfredo-medina-stx: I have it ready, don't worry ;)

mmmavis (approved):
@alfredo-medina-stx 🎉 Thanks for the thorough work on this!

foundation_cms/urls.py
robdivincenzo: Thanks for being on top of it! :)

PR #16138, opened 2026-06-24
Add backend for filter and sort search results

adalcima:
# Description

This PR updates the `Search app` backend by adding support for filtering and sorting search results without changing existing templates or search result rendering flow.

Link to sample test page: https://foundation-s-tp1-3964-a-acxfzi.mofostaging.net/en/
Related PRs/issues: [TP1-3964](https://mozilla-hub.atlassian.net/browse/TP1-3964) 

**Main changes**
- Added content type filtering via `content_type` query param.
- Added topic filtering via `topic` query param.
- Added sorting support via `sort` query param: `relevance` (default), `newest`, `oldest`
- Updated result count to reflect the final filtered result set.
- Updated payload based on current query result subset

**Tests**
- Visit the [review app](https://foundation-s-tp1-3964-a-acxfzi.mofostaging.net/en/) and submit a typical query. Results will be returned as before.
- Verify sorting adding query params: 
    - `/en/search/?query=<search_term>&sort=newest`
    - `/en/search/?query=<search_term>&sort=oldest`
- Verify content type filter adding query params (`all` content by default): 
    - `/en/search/?query=<search_term>&content_type=research`
    - `/en/search/?query=<search_term>&content_type=event`
    - `/en/search/?query=<search_term>&content_type=press-release`
    - `/en/search/?query=<search_term>&content_type=what-we-do`
- Verify topic filter adding query params: 
    - `/en/search/?query=<search_term>&topic=<topic_slug>`

### DIFF
--- a/foundation_cms/templates/search/search_page.html
+++ b/foundation_cms/templates/search/search_page.html
@@ -24,9 +24,9 @@
                 </div>
 
                 {# TEMP DEBUG PLACEHOLDER: remove when frontend is finalized #}
-                <div class="cell small-12 large-8 large-offset-2" style="margin: 1rem 0; padding: 1rem; border: 1px dashed">
+                <div class="cell small-12 large-8 large-offset-2 margin-vertical-1 padding-1 border-dashed">
                     <strong>Search debug placeholder</strong>
-                    <ul style="margin-top: .5rem;">
+                    <ul>
                         <li><strong>query:</strong> {{ search_query|default:"(empty)" }}</li>
                         <li><strong>sort:</strong> {{ sort|default:"relevance" }}</li>
                         <li><strong>content_type:</strong> {{ content_type|default:"all" }}</li>
--- a/foundation_cms/templates/search/search_page.html
+++ b/foundation_cms/templates/search/search_page.html
@@ -23,6 +23,33 @@
                     {% include "search/_search_form.html" with form_class="search-form" input_class="search-form__input" submit_classnames="search-form__submit-button" %}
                 </div>
 
+                {# TEMP DEBUG PLACEHOLDER: remove when frontend is finalized #}
+                <div class="cell small-12 large-8 large-offset-2" style="margin: 1rem 0; padding: 1rem; border: 1px dashed">
+                    <strong>Search debug placeholder</strong>
+                    <ul style="margin-top: .5rem;">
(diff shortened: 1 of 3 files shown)


foundation_cms/templates/search/search_page.html
```diff
@@ -23,6 +23,33 @@
                     {% include "search/_search_form.html" with form_class="search-form" input_class="search-form__input" submit_classnames="search-form__submit-button" %}
                 </div>
 
+                {# TEMP DEBUG PLACEHOLDER: remove when frontend is finalized #}
```
adalcima, line 26: Hey @ramram-mf this block is an example of how we can use the new context values. This block is just a placeholder and should be removed with the changes to the frontend. Thanks!

ramram-mf (changes requested):
I’m seeing the displayed count diverge from the actual rendered result list. On the review app, `/en/search/?query=grantmaking` shows `1-5 of 6` but only five cards render and there’s no next page.

It looks like `total_search_results` is calculated from `result_ids` before the final page queryset is hydrated via `Page.objects.filter(id__in=result_ids)`, which can de-dupe IDs. Could we de-dupe the backend IDs while preserving relevance order before counting/pagination, so the summary and rendered pages stay aligned?

Image

adalcima:
> I’m seeing the displayed count diverge from the actual rendered result list. On the review app, `/en/search/?query=grantmaking` shows `1-5 of 6` but only five cards render and there’s no next page.
> 
> It looks like `total_search_results` is calculated from `result_ids` before the final page queryset is hydrated via `Page.objects.filter(id__in=result_ids)`, which can de-dupe IDs. Could we de-dupe the backend IDs while preserving relevance order before counting/pagination, so the summary and rendered pages stay aligned?
> Image

Good catch, @ramram-mf , PR updated to remove duplicates while preserving the order. In this case, `set` couldn't be used, as the order would change. Thanks :raised_hands:

ramram-mf (approved):
LGTM! Thanks @adalcima

adalcima:
@mmmavis @ramram-mf @alfredo-medina-stx will merge this PR, since targets a feature branch, and to accelerate testing and integration with the frontend, future adjustments will be easier to identify and fix in the feature branch. If you have any further comments, please let me know. Thanks team!

PR #16189, opened 2026-07-09
Implement search results filters and sort UI

ramram-mf:
# Description

Implements the Search 2.0 results page UI for desktop, tablet, and mobile.

image

### Changes

- Adds content-type filtering for Articles, Research, Press Release, Event, and All.
- Adds query-relevant topic tag filters.
- Adds sorting by Relevance, Newest First, and Oldest First.
- Updates desktop results immediately when filters change.
- Adds the responsive Filter & Sort drawer for tablet and mobile.
- Updates the mobile “Display results (X)” count before applying filters.
- Adds responsive result cards with topic tags, media, title, author, date, and excerpt.
- Limits displayed card topics to three with an overflow count.
- Prefetches topic relations for the displayed page to avoid per-card queries.
- Prevents mobile preview requests from creating search analytics events.
- Adds modal focus management, keyboard trapping, Escape handling, and focus restoration.
- Matches the Figma hover, selected, spacing, dropdown, and close-control states.

Content types and topic tags currently allow one selection at a time, matching the existing backend contract. Per-content-type result counts remain out of scope.

Link to sample test page: https://foundation-s-tp1-3963-s-cpgqoe.mofostaging.net/en/search/?query=search
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3963)


beccaklam (changes requested):
Hi @ramram-mf, this is looking great overall! I just have a bunch of nit-picky details. Happy to jump on a call to chat anything through:

## General Feedback

- [x] Selected topic tag in the filter should be one shade lighter (yellow-40 or #F7EC50)

> Image
 
- [x] Is the orange color here #FFCEAE? It looks darker

> Image

- [x] Can we align the bottom text of these? (see image below):

>Image

- [x] Even when there's only one page can we show 'Page 1 of 1'? Right now the page count disappears if there's only one page.

- [x] Can the image fill the whole height of the search result card (but still retain its proportions)?
>Image

- [x] Can we increase the spacing between search results to 2rem? Seeing it live, the results feel a bit tight

## Mobile Specific:

- [x] Can we decrease the spacing between the 'filter & sort' button and the results to 16px? Current spacing below:

>Image

- [x] Mobile filter is missing the orange outline (see [figma](https://www.figma.com/design/EjPhm03AdjqXWHh9e4dKs7/Search-Feature?node-id=1454-13509&t=Rlwf1s3wiN51cOcs-4)). The close button also looks really small, can we try this export? 
>Image

- [x] The mobile filter also has a lot of unused white space? This is probably related to the tablet layout. Can we trim it so that on mobile the 'display results' button only has 24px of spacing between it and the filter controls?
 >Image

beccaklam (changes requested):
Hi @ramram-mf thank you for making all those changes! Just a few more 🙇🏻‍♀️ :

- [x] I think the image proportions have been stretched here. I meant can the image fill the vertical height while still maintaining its 3:2 ratio proportions? So e.g. if the height is 146.2px the width should be 219.3px
> Image

- [x] Can we actually make the filter on mobile and tablet full page and not an overlay? It's strange that someone could scroll past the filter only to see a greyed out screen. If the content exceeds the full screen, can we treat the content similar to what we did with the gallery project where a scroll bar appears but the button remains sticky at the bottom of the screen?
> Image

- [x] Can we make the 'display results' button on mobile full width and increase the spacing between the filters and the button to 3rem? Can we also apply this to tablet? The spacing there is also too long.
> Image

- [x] The 'x' in the close button looks off center. Can we center it?
> Image

ramram-mf:
> Hi @ramram-mf thank you for making all those changes! Just a few more 🙇🏻‍♀️ :

Hi @beccaklam, thanks for the detailed follow-up! I’ve addressed the remaining items:

- Search-result media now maintains the requested 3:2 ratio at approximately 219.3 × 146.2px. The desktop title is also clamped to one line so card heights remain consistent.

  image

- The mobile and tablet Filter & Sort experience is now full-page rather than an overlay. Overflow scrolls within the drawer while the results CTA remains sticky.

image

image

- “Display results” is full width on mobile and tablet, with 3rem between the filters and CTA.

image

- The close control is centered. The circle is provided by the outer button border, with only the X supplied by the icon asset.

image

I also verified the live result count in WebKit. Content and topic filters combine correctly, sorting preserves the count, and Reset restores the original count.

ramram-mf:
image

*Tablet/mobile filter count implemented.*

beccaklam (changes requested):
Hi @ramram-mf ! Thank you for implementing the filter count. We actually don't have stand today so I'll just list my tweaks here:

- [x] After seeing it live, I think we can remove the orange border to simplify the design. I was trying to keep the drop down menus consistent with the nav but the nav doesn't have an orange border on mobile/tablet when it's full screen so I think we can remove the orange border.
- [x] If we remove the orange border can we align the 'reset' and 'close' button with the page margins on tablet (and anywhere else they may go outside the page container?)
Screenshot 2026-07-21 at 06-36-51 Search - Mozilla Foundation

- [x] For mobile since the white space is sticky, can we make the sticky portion above the button equal to the other spacing around it? So 3rem to 1.5rem. Can we keep it 3rem between the last filter element and the button though?

I think we should be good after that!! Thank you again!

mmmavis (changes requested):
Hey @ramram-mf , this is what I see on a fresh local build, which is different than the Figma mockup. I didn't dive into code yet but I'm guessing the entire "Latest Stories" section should not show?

Image

I then manually created a page on http://localhost:8000/en/what-we-do/. Did the search again but that page didn't show up in the result.

Also noticed page is missing body content if you try to search without any keyword: 
image

ramram-mf:
> Hey @ramram-mf , this is what I see on a fresh local build, which is different than the Figma mockup. I didn't dive into code yet but I'm guessing the entire "Latest Stories" section should not show?

Thanks for checking this, @mmmavis . I went through each item:

- The “Latest Stories” section is inherited from the base branch. It is removed and replaced by the proper zero-results state in #16250, which builds on this search-results work. It won’t appear in the final integrated UI.
- I reproduced creating a page as a draft and then publishing it under What We Do. Once published, it appeared immediately in search without manually running `update_index`. Pages left as drafts are intentionally excluded. I added regression coverage for this behavior.
- I fixed `/search/` with no keyword so it now renders a search-entry state with a heading and search form instead of a blank body.
  image


Thanks for catching that empty-query case!

ramram-mf:
@mmmavis Thanks, Mavis! I also added a reusable management command for generating deterministic search test data locally or in review apps:
`python manage.py load_search_test_data --update-index`

It creates published pages across all supported content types, with varied topics, dates, descriptions, pagination, and a deliberately long title. It is safe to rerun because it replaces previously generated records.
Cleanup is available with:
`python manage.py load_search_test_data --delete`

For local Docker development, there is also an Invoke wrapper:
`inv load-search-test-data`

This was shaped after the NP article generation routine I introduced earlier.

beccaklam (changes requested):
Hi @ramram-mf, it's so close! Seeing it now, I think it would actually make more sense to align with the items in the filter vs the nav:
- [x] Can we align the "reset" and "close" buttons on mobile and tablet with the `search-filter-drawer__body`? 
- [x] Can we also ensure there's 1.5rem of spacing between the bottom of the nav and the top of filter container?

## Mobile:
Image

## Tablet:
Image

beccaklam:
@ramram-mf Sorry I didn't realize there was already padding built in on mobile for the filter container. 

- [x] Can we remove this 24px of padding on mobile? Tablet looks good though!

image

And sorry one more thing:
- [x] Can we change the 4rem of padding-top on `search-summary` to 1rem instead for tablet? 
image

I realize this PR back and forth is getting long. We can also hop on a call if that's easier!

foundation_cms/search/views.py
```diff
@@ -31,11 +32,16 @@ def search(request):
...
     related_topics = []
     page = request.GET.get("page", 1)
     total_search_results = 0
     current_locale = Locale.get_active()
-    is_initial_search_submit = "page" not in request.GET
+    current_site = Site.find_for_request(request)
+    # Drawer count previews are internal requests and should not be logged as submitted searches.
+    is_preview_request = request.headers.get("X-Search-Preview") == "true"
+    result_state_params = {"page", "content_type", "topic", "sort"}
+    is_initial_search_submit = not result_state_params.intersection(request.GET) and not is_preview_request
```
mmmavis: This now excludes logging for any request with `content_type`, `topic`, or `sort` set, not just `page` like before. Since the filter/sort forms always send these (even at defaults), the first filter or sort click no longer logs a `SearchEvent`/`add_hit`, and a shared URL like `?query=x&sort=newest` never logs on first visit either.

Is that intentional? The comment below (`# Log only on initial submission, not on pagination clicks`) still implies only pagination is excluded.

foundation_cms/search/tests/test_views.py
```diff
@@ -1,4 +1,4 @@
-from unittest.mock import patch
+from unittest.mock import Mock, patch
```
mmmavis, line 1: thanks for adding tests!

foundation_cms/templates/search/search_page.html
mmmavis, line 1: This file has three separate `<form>`s (sidebar, sort dropdown, drawer) with a fair amount of duplicated markup between them. The `content-type`, `sort`, and `topic` filters are duplicated across the forms. Might be worth splitting each form into its own partial, and pulling the `content-type`, `sort` and `topic` fieldsets into shared includes?

foundation_cms/templates/search/_search_results.html
```diff
@@ -9,32 +9,45 @@
...
             {% if result.search_image %}
-              {% responsive_image result.search_image ratio="3:2" base_width=600 sizes="(max-width: 639px) 100vw, 33vw" %}
+              {% responsive_image result.search_image ratio="3:2" base_width=600 sizes="(max-width: 639px) 100vw, 220px" %}
             {% else %}
               <img
                 src="{% cycle listing_page_fallback_1 listing_page_fallback_2 listing_page_fallback_3 listing_page_fallback_4 listing_page_fallback_5 listing_page_fallback_6 %}"
                 alt="{{ illustration_placeholder_alt }}"
                 width="600"
                 height="400"
+                loading="lazy"
```
mmmavis: Nice addition! 

Nitpick: First result is likely above the fold, might not want loading="lazy" on it specifically.

```suggestion
              {% if not forloop.first %}loading="lazy"{% endif %}  
```

foundation_cms/static/js/components/search_page/filter_controls.js
```diff
@@ -0,0 +1,331 @@
...
+        const html = await response.text();
+        const doc = new DOMParser().parseFromString(html, "text/html");
+        const previewRoot = doc.querySelector(SELECTORS.root);
+        const count = previewRoot?.dataset.totalResults;
+
+        if (count !== undefined) setApplyButtonLabel(drawerApply, count);
+      } catch (err) {
+        if (err.name === "AbortError") return;
+
+        // Keep the last known count when the preview request is unavailable.
```
mmmavis: A real fetch failure (e.g., user goes offline, network drops) falls through to the same "do nothing" outcome as an abort, leaving a possibly-stale count on the button. Might be worth falling back to a plain "Display results" (no count) on real errors instead.

foundation_cms/static/js/components/search_page/filter_controls.js
```diff
@@ -0,0 +1,331 @@
...
+
+        event.preventDefault();
+        input.checked = false;
+        label.dataset[DATA_KEYS.topicWasChecked] = "false";
+
+        if (mode === "immediate") {
+          submitForm(form);
+        } else {
+          updateDrawerPreview();
+        }
```
mmmavis: This is the same logic as the change handler further down:

```js
// lines 292-297
if (mode === "immediate") {
  submitForm(form);
  return;
}
updateDrawerPreview();
```

Wondering what you think about creating a shared helper function for this?

foundation_cms/static/js/components/search_page/filter_controls.js
```diff
@@ -0,0 +1,331 @@
...
+ * Search results filter and sort controls.
+ *
+ * @module searchPageFilterControls
+ */
+
+const PREVIEW_DEBOUNCE_MS = 200;
+const LARGE_VIEWPORT_QUERY = "(min-width: 64em)";
+
+const CLASSNAMES = {
+  drawerOpen: "search-filter-drawer-open",
```
mmmavis: `CLASSNAMES.drawerOpen` and the `drawerOpen` variable further down share the same name for different things and can be easily misread. Could we rename the class constant to something like drawerScrollLocked to avoid the ambiguity?

foundation_cms/static/scss/pages/search_page.scss
```diff
@@ -4,15 +4,30 @@
...
+}
+
+body.search-filter-drawer-open .primary-nav-ns {
+  position: fixed;
+}
+
 body.template-search-results {
   $highlight-color: color(orange, "300");
+  $selected-topic-color: #f7ec50;
+  $selected-control-color: #ffceae;
```
mmmavis, line 20: curious if these two are new brand colours?

foundation_cms/static/scss/pages/search_page.scss
```diff
@@ -38,35 +74,553 @@ body.template-search-results {
...
+
+  .search-topic-pill__input:focus-visible + .search-topic-pill__label,
+  .search-topic-pill:hover .search-topic-pill__label {
+    border-color: $black;
+    background: $selected-topic-color;
+  }
+
+  .search-filter-drawer {
+    position: fixed;
+    inset: rem-calc(88) 0 auto;
```
mmmavis:
```suggestion
    inset: $primary-nav-height 0 auto;
```

foundation_cms/search/management/commands/load_search_test_data.py
mmmavis, line 1: That's very thoughtful of you to create this command. I'm thinking we should move it to factories instead of creating a separate command. That way it's reusable and testable. For this PR I'd suggest leaving `foundation_cms/search/management/commands/load_search_test_data.py` (and related files) out and create a follow-up PR for the factory work to keep the scope contained.

tasks.py
```diff
@@ -350,6 +350,35 @@ def load_np_fake_articles(
     manage(ctx, " ".join(parts), stop=stop)
 
 
+@task(aliases=["search-test-data"])
```
mmmavis: Same as the note on `load_search_test_data.py`. This task would move/go along with that command if we pull it into a follow-up PR.

mmmavis (changes requested):
Hey @ramram-mf , 

This turned out really nice, the drawer and filter interactions feel solid and the site-scoping bug you caught in `views.py` was a good find. Left some comments, mostly small stuff plus one bigger ask about moving the test-data command to a follow-up PR. Nothing blocking, let me know what you think on the open questions. Thanks!

beccaklam:
Sorry @ramram-mf I found one more thing! 
- [x] Can we change the spacing here (where the upper yellow is) from 2rem to 1rem on mobile?
image

foundation_cms/search/views.py
ramram-mf: Good catch. Suppressing filter and sort submissions was not intentional. This now logs initial searches and applied filter/sort changes, while excluding pagination and internal drawer-preview requests. I added coverage for each case.

foundation_cms/search/tests/test_views.py
ramram-mf: Thanks! I expanded the coverage around logging, previews, empty queries, filter counts, single-page pagination, published pages, and site-scoped content filtering.

foundation_cms/templates/search/search_page.html
ramram-mf: Agreed. I split the sidebar, sort dropdown, and drawer into separate partials, with shared includes for the content-type, topic, and sort fieldsets.

foundation_cms/templates/search/_search_results.html
ramram-mf: Good point. The first result image now omits loading="lazy", while subsequent images remain lazy-loaded. The responsive-image tag now accepts an optional loading value.

foundation_cms/static/js/components/search_page/filter_controls.js
ramram-mf: Agreed. Aborted requests are ignored, while genuine network or HTTP failures now fall back to the plain “Display results” label instead of leaving a potentially stale count.

foundation_cms/static/js/components/search_page/filter_controls.js
ramram-mf: Agreed. I extracted the shared immediate-submit/drawer-preview behavior into handleFilterChange().

foundation_cms/static/js/components/search_page/filter_controls.js
ramram-mf: Agreed. I renamed the class key to drawerScrollLocked to distinguish it from the drawer element/state variable.

foundation_cms/static/scss/pages/search_page.scss
ramram-mf: These values came directly from the current Search Feature Figma file, but they are not part of the repository’s existing brand-color map. I spoke with Becca today, and the palette is expected to change as part of the upcoming major brand evolution. For this PR I kept the values component-local rather than introducing potentially short-lived global tokens. I think updating the shared palette should be coordinated with the design team as follow-up work.

foundation_cms/static/scss/pages/search_page.scss
ramram-mf: Updated. The drawer positioning and height calculations now use $primary-nav-height.

foundation_cms/search/management/commands/load_search_test_data.py
ramram-mf: Agreed. I removed the management command and its related files from this PR. Reusable factories can be introduced in a focused follow-up.

tasks.py
ramram-mf: Removed along with the test-data command so the factory work can remain contained in the follow-up.

mmmavis (approved):
Really appreciate how thoroughly you went through all the feedback @ramram-mf . LGTM, approving!

beccaklam (approved):
Thank you for @ramram-mf !!! Approved ✅ Excited for this to ship! 🚀

mmmavis:
@ramram-mf Could you create a follow-up ticket for the factory work so we can get your super useful code in? Thanks!

PR #16222, opened 2026-07-16
Fix impact stats count-up animation

OswaldoPinedaMozilla:
## Summary
- Updates Impact Numbers stats to count up to the final value instead of using the random digit wheel animation.
- Removes the retired animated-number template tag.

## Context
The old Impact Numbers animation rendered each digit separately with random fake digits and CSS step animations. That made the counter look like it could overshoot, move up and down, or flicker before landing on the final number.

This PR replaces that behavior with a single count-up animation for the full stat value. It keeps the saved CMS values unchanged, preserves prefixes/suffixes like `+`, `$`, `K`, and `M`, and shows the final value immediately for users with reduced motion enabled.

This approach changes the behavior globally for Impact Numbers instead of adding a CMS option because the requested fix is to remove the confusing digit-wheel behavior, and the old template tag only existed to support that animation.

## Jira
https://mozilla-hub.atlassian.net/browse/TP1-3819

## Preview app
https://foundation-s-fix-tp1-38-lxe7ts.mofostaging.net/en/festival/

## Test plan
- Ran redesign JS lint, SCSS lint, and production CSS/JS builds locally.


ramram-mf (approved):
Great work @OswaldoPinedaMozilla ! This now avoids unnecessary backend work for simple animations.

PR #16225, opened 2026-07-16
fix(csp): allow Civic cookie banner CDN in CI environments

mmmavis:
- Adds `https://cc.cdn.civiccomputing.com` to `CSP_SCRIPT_SRC` and `CSP_STYLE_SRC` in the CI environment configs
- Covers the main CI workflow plus both legacy and redesign visual regression workflows


ramram-mf (approved):
Thanks @mmmavis !

PR #16100, opened 2026-06-16
Add navbar search focus state

ramram-mf:
# Description

Implements the Search 2.0 focused navbar state for TP1-3962.

This adds the expanded search drawer triggered from the primary nav search action, with CMS-managed topic pills and quick links rendered beneath the focused search input.

## What Changed

- Adds a shared search drawer partial for both static and editable primary nav templates
- Adds accessible open/close behavior, focus handling, Escape/outside-click dismissal, and ARIA state updates
- Adds CMS fields to `NavigationMenu` for:
  - Search topic pills, max 5
  - Search quick links, max 3
- Seeds default navigation menu data to match the current static nav structure
- Seeds default search drawer suggestions:
  - Topic pills: privacy, personal data, open source, encryption, ai
  - Quick links: Grantmaking, Mozilla Festival, Common Voice
- Styles the focused drawer, topic pills, quick links, and icons to match the Figma direction
- Omits the suggestion UI entirely when no CMS values are configured

## Review Notes

`EDITABLE_NAV=True` is required for the review app to render the CMS-backed navigation path.

## Screenshots

image

*Focused search drawer*

image

*CMS Navigation Menu search drawer fields*

image

*CMS topic pills / quick links configuration*

Link to sample test page: https://foundation-s-tp1-3962-s-cbtvzq.mofostaging.net/en/
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3962)

PR #16250, opened 2026-07-21
Add zero results page UI

ramram-mf:
# Description

Implements the Search 2.0 zero-results page for TP1-3966 across desktop, tablet, and mobile.

When the base search query returns no matches, the normal results and filter UI is replaced with a dedicated recovery state containing an editable prefilled search field, troubleshooting guidance, and CMS-managed suggested links.

## What Changed

- Adds the responsive zero-results layout from the approved Figma design.
- Displays the submitted query in the heading and prefilled search input.
- Allows visitors to immediately edit and resubmit their search.
- Reuses the CMS-managed topic pills and quick links introduced by the navbar search focus state.
- Extracts the suggestions into a shared component used by both the navbar search drawer and zero-results page.
- Resolves the active navigation menu using the current locale when the search template does not have a Wagtail page in context.
- Omits either suggestion section when no corresponding CMS values are configured.
- Removes the previous “Latest Stories” fallback and its related page lookup.
- Preserves the regular results and filter UI when active filters produce an empty result set.
- Keeps an accessible name on the icon-only mobile search button.
- Adds regression coverage for the zero state, filtered empty states, CMS suggestions, empty CMS configuration, and locale-aware navigation.

## Behavior Notes

The full zero-results page is determined from the backend’s base query results before optional section or topic filters are applied. This means a filter-produced empty state retains the active filters so visitors can adjust them.

## Dependencies

This PR targets feature/search-integration-v2 and builds on the navbar search focus state from #16100 and the search results/filter UI from #16189. Both are now included in the target feature branch.

## Screenshots

image

*Desktop zero-results page*

image

*Mobile zero-results page*

## Testing

- Search view regression tests
- Verified that resubmitting the zero-results form navigates back to the regular results page
- Added regression coverage for populated suggestions, empty suggestion lists, no active navigation menu, filtered empty states, and locale-aware navigation.

Link to sample test page: https://foundation-s-tp1-3966-z-yy5zgm.mofostaging.net/en/search/?query=zzzzzzzzzz
Related PRs/issues: [Jira ticket](https://mozilla-hub.atlassian.net/browse/TP1-3966), #16100, #16189


beccaklam (changes requested):
Hi @ramram-mf just one change request, everything else lgtm!

- [x] Can we increase the right and left padding to 1rem on mobile?

Image

beccaklam (approved):
Thank you @ramram-mf! Everything looks good now! ✨ 🚀

foundation_cms/templates/search/_search_suggestions.html
```diff
@@ -0,0 +1,39 @@
+{% load i18n %}
+
+<div class="search-suggestions">
+  {% if search_topic_links %}
+    <div class="search-suggestions__topics">
+      <p class="search-suggestions__heading">{% trans "Explore our ideas" %}</p>
```
mmmavis: let's use heading tag instead of `p`

foundation_cms/templates/search/_search_suggestions.html
```diff
@@ -0,0 +1,39 @@
...
+            </li>
+          {% endwith %}
+        {% endfor %}
+      </ul>
+    </div>
+  {% endif %}
+
+  {% if search_quick_links %}
+    <div class="search-suggestions__quick-links">
+      <p class="search-suggestions__heading">{% trans "Quick Links" %}</p>
```
mmmavis: let's use heading tag instead of `p`

foundation_cms/templates/search/_search_suggestions.html
```diff
@@ -0,0 +1,39 @@
...
+      <ul class="search-suggestions__quick-link-list">
+        {% for link in search_quick_links %}
```
mmmavis: q: do we need to cap `search_quick_links` as well?

foundation_cms/search/tests/test_views.py
```diff
@@ -78,3 +81,128 @@ def test_no_logging_on_pagination_only(self, mock_add_hit):
...
+        response = self.client.get("/en/search/", {"query": "missing term"})
+
+        self.assertEqual(response.status_code, 200)
+        self.assertContains(response, "Explore our ideas")
+        self.assertContains(response, "Quick Links")
+        self.assertContains(response, f'href="{reverse("search")}?query=privacy"')
+        self.assertContains(response, 'href="/what-we-do/awards/"')
+
+    @patch("foundation_cms.search.views.get_search_backend_for_locale")
+    def test_zero_results_omit_empty_navigation_suggestions(self, mock_get_search_backend):
```
mmmavis, line 276: no test for a site with no `active_navigation_menu` configured at all (only the empty-lists case is covered here). Since this page now depends on `SiteNavigationMenu` for the first time, maybe worth a quick test confirming `/search/` doesn't 500 when that setting is unset?

foundation_cms/search/tests/test_views.py
```diff
@@ -1,9 +1,12 @@
-from unittest.mock import patch
```
mmmavis: Thanks for adding tests!

mmmavis (changes requested):
Great work @ramram-mf ! Left some comments for my first round of review.

foundation_cms/search/tests/test_views.py
ramram-mf: Yes its worth it, we now run `test_zero_results_render_without_an_active_navigation_menu` which unsets `active_navigation_menu`, tests for 200 and expects no suggestions. Good idea!

foundation_cms/templates/search/search_page.html
```diff
@@ -12,11 +12,15 @@
 {% block body_class %}template-search-results{% endblock body_class %}
 
 {% block content %}
+    {% get_settings use_default_site=True %}
+    {% translated_menu settings.navigation.SiteNavigationMenu.active_navigation_menu as localized_menu %}
```
mmmavis: I'm thinking we should move these two lines to the `{% if search_query and is_zero_results %}` block to save an unnecessary settings lookup / translation query on every search request since `localized_menu` is only used there in this template

mmmavis (changes requested):
Great work as usual @ramram-mf ! I left a tiny code suggestion to improve `search_page.html`'s performance. Also wanna flag that `TEMP_SEARCH_RELATED_CONTENT_PAGE_IDS` related code can now be removed as well.

mmmavis (approved):
🍻

Issue #16249, opened 2026-07-21
Set up JS unit testing infrastructure (Vitest) for redesign frontend


data-sync-user:
We don't have any JS unit testing in this codebase today, only linting plus Playwright/Percy for integration and visual regression. That means changes to components have no safety net beyond manual QA and end-to-end checks.

This ticket is pure groundwork: get a test runner installed, configured, and wired into CI, with one proof-of-concept test to confirm the whole path (write a test locally, run it, see it pass in CI) actually works.

## **Scope**

`redesign` codebase only. `legacy` codebase is out of scope.

## **Why Vitest Testing Framework?**

- **Jest**: would need a Babel transform on top of our esbuild-based, native-ESM toolchain; Vitest resolves ESM natively via its own esbuild pipeline, no extra config.
- **Mocha (+ Chai + Sinon)**: no built-in assertions or mocking, means assembling and maintaining several packages. Vitest bundles assertions, mocking (`vi.fn()`/`vi.mock()`), and jsdom in one config.
- **Jasmine**: same gap as Mocha, plus no native ESM/esbuild integration.
- **Cypress Component Testing**: built for component frameworks (React/Vue) rendered in a real browser; `redesign` has no framework, just vanilla DOM modules, so that's heavier than needed.

## **Acceptance Criteria**

- Vitest is installed and configured in frontend/redesign; tests can exercise code that reads/writes the DOM (`document`, `querySelector`, event dispatching, etc.) without a real browser
- A script exists to run the test suite locally
- At least one passing test exists, proving the setup works end-to-end (a simple, fixture-free module is a good first pick)
- A CI job runs the test suite on every PR
- CI job is non-blocking for now (visible status, not required for merge)
- Coverage output is visible from the CI run

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4075)

data-sync-user:
> **Mavis Ou** commented from Jira:

PR filed: [https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16281](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16281|smart-link)

PR #16004, opened 2026-05-29
fix(deps): patch picomatch CVE-2026-33671 and CVE-2026-33672 via resolutions

mmmavis:
## Summary

Pins `picomatch` to patched versions via Yarn `resolutions` to address two medium/high severity CVEs affecting all consumers in the dependency tree.

| CVE | Severity | Type | Related Security Issues
|-----|----------|------| ---------------|
| CVE-2026-33671 | High (CVSS 7.5) | ReDoS via extglob quantifiers | [no. 321](https://github.com/MozillaFoundation/foundation.mozilla.org/security/dependabot/321) AND [no. 324](https://github.com/MozillaFoundation/foundation.mozilla.org/security/dependabot/324)
| CVE-2026-33672 | Medium (CVSS 5.3) | Method injection via POSIX character classes | [no. 322](https://github.com/MozillaFoundation/foundation.mozilla.org/security/dependabot/322) AND [no. 323](https://github.com/MozillaFoundation/foundation.mozilla.org/security/dependabot/323)

Both are fixed in `picomatch` 2.3.2 and 4.0.4.

## Changes

- Adds `resolutions` to root `package.json` forcing:
  - `picomatch` → `^2.3.2` (covers micromatch, anymatch, readdirp consumers)
  - `tinyglobby/picomatch` → `^4.0.4` (covers tinyglobby consumer)
- Lockfile updated accordingly

## Notes

`picomatch` is a transitive-only dependency so a direct version bump is not possible. `resolutions` is the correct Yarn 1 mechanism for this.

PR #16281, opened 2026-07-24
chore(redesign): set up Vitest unit testing infrastructure

mmmavis:
Related PRs/issues: [Jira TP1-4075](https://mozilla-hub.atlassian.net/browse/TP1-4075) / #16249 

## Summary

- Installs and configures Vitest with jsdom for `frontend/redesign`, so tests can exercise DOM reads/writes and event dispatching without a real browser
- Adds one proof-of-concept test (`pillar_card_set.test.js`, 100% coverage) proving the whole path works: write a test locally, run it, see it pass in CI
- Wires a new, non-blocking CI job (`unit_test_frontend_redesign`) that runs the suite with coverage on every push and PR
- Fixes an ESLint config bug (uncovered while testing coverage output) where generated `coverage/` files were being linted

Two deliberate choices worth flagging:
- `test:watch` goes beyond the AC (which only needs one script). Added for the local dev loop, matching this workspace's `watch:css`/`watch:js` convention.
- `pillar_card_set.js` was picked for being simple, not high-value. Future coverage should prioritize logic-heavy modules (`util/*.js`, carousels, accordions, tabs) where bugs won't show up in a Percy screenshot.

## Test plan

- [ ] `yarn workspace redesign test` passes locally with coverage output
- [ ] `yarn workspace redesign lint` and `check-format` pass clean
- [ ] CI job `unit_test_frontend_redesign` runs on this PR and shows as non-blocking (visible status, not required)
- [ ] Coverage output appears in the CI job's log


ramram-mf (commented):
I noticed CI reports:

`Lockfile has incorrect entry for "picomatch@^4.0.4". Ignoring it.`

I see the broad `"picomatch": "^2.3.2"` resolution was introduced in #16004 to address the picomatch CVEs. Was it intended to override v4 consumers as well, or should it apply only to dependencies using picomatch v2?

On this branch, Yarn combines the v2 and v4 selectors under picomatch 2.3.2, causing it to ignore the v4 lockfile entry. A regular `yarn install` reproduces the same result.

I tested scoping the v2 resolution to its current consumers (`anymatch`, `micromatch`, and `readdirp`). After regenerating the lockfile, normal and frozen installs were warning-free and the Vitest tests still passed. Would scoping the resolution that way make sense here?

mmmavis:
@ramram-mf Good catch, thanks for digging into this!  Verified locally: the blanket `"picomatch": "^2.3.2"` resolution was matching v4 consumers too (tinyglobby, vite, vitest, @parcel/watcher), so Yarn merged all the ranges into one lockfile entry and force-resolved everything to 2.3.2, silently dropping the v4 override and printing the warning.

I've updated the PR to scope the v2 resolution to its actual consumers (anymatch, micromatch, readdirp) instead of the blanket picomatch entry.

ramram-mf (approved):
Thanks @mmmavis !

Issue #15475, opened 2026-02-25
Conditionally load Sentry based on user privacy preferences


data-sync-user:
**Context:**

Currently, [sentry-config.js](https://github.com/MozillaFoundation/foundation.mozilla.org/blob/main/foundation_cms/legacy_apps/static/js/common/sentry-config.js#L1) initializes Sentry unconditionally. The [bedrock (mozilla.org) codebase](https://github.com/mozilla/bedrock/blob/030b612360d135103f3a7e6874799b82251120e4/media/js/base/sentry-consent.es6.js#L90-L122) has a more privacy respecting pattern where Sentry is only loaded after checking user privacy preferences.

**Goal:**

Update our Sentry initialization to follow the same pattern as bedrock, only loading Sentry when appropriate based on the user’s privacy preferences

**Note:** 

This ticket only covers the frontend Sentry initialization. Backend Sentry setup is out of scope and may need a separate investigation.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-3695)

PR #16384, opened 2026-08-11
feat(sentry): gate legacy Sentry initialization on analytics consent

mmmavis:
# Description

`sentry-config.js` currently initializes Sentry unconditionally on legacy pages. We want to only load Sentry once the user's privacy preferences allow it, adapted to this codebase's actual consent management (Civic Cookie Control).

Scoped to legacy only because redesign doesn't load client Sentry at all. This does touch the shared `civic_cookie_control/init.js` (loaded by both frontends) to dispatch a consent-change event on accept/revoke, so the Sentry client can react live instead of waiting for the next page load.

Related PRs/issues: [Jira TP1-3695](https://mozilla-hub.atlassian.net/browse/TP1-3695) / #15475 

# Testing steps
- Make sure you have `COOKIE_CONTROL_API_KEY` set in your `.env` (Slack me if you need the key)
- Open a browser window (use an Incognito window with extensions disabled (e.g., ad blockers))
- On http://localhost:8000/cms/sites/ , configure legacy site so the hostname is `localhost`
- Go to http://localhost:8000/
- Clear cookies
- Accept analytics consent in the cookie banner
- Reload the page
- In DevTools Network tab (Fetch/XHR, filter "sentry"), confirm a POST request to the Sentry domain fires automatically . e.g., 
image
- Revoke analytics consent and confirm no further Sentry requests fire


ramram-mf (approved):
Thanks @mmmavis !

PR #16389, opened 2026-08-11
Locale Aliases Not Created On Page Copy

robdivincenzo:
# Description

This PR adds a hook to create localized alias pages on copy. While technically the alias pages were getting created, it only happened once the `sync_locale_tree` ran on cron every 60 min. This will reduce the latency between page copy and alias generation, allowing editors to translate copied pages immediately. Included safeguards if resolved in `upstream` and test cases.

This PR also includes an SEO improvement to exclude aliases from sitemap per best practices. Once translated, they are published as pages and will be included in the sitemap.

Link to sample test page:
Related PRs/issues: TP1-3320

# To Test
1) On `main` branch, reproduce issue by copying a page on your local copy
2) Note there's no locale aliases generated
3) Check out this branch `TP1-3320-alias-pages-not-created` and copy a page
4) Note that the copied page immediately has aliases automatically added for the different locales.


adalcima (approved):
lgtm, tested this locally and I can see the alias pages being created, thanks @robdivincenzo

Issue #16354, opened 2026-08-06
ci: share preroll build artifact + cache apt packages (test_wagtail/test_integration)


data-sync-user:
**Codebase:** Overall
**Impact:** Medium-High
**Notes:** test_wagtail and test_integration both run a near-identical ~6-7min Preroll (yarn build + collectstatic) — shareable as one artifact. test_wagtail's apt-get step also varies 27s-322s across runs; caching would flatten that.
**Decision:** ✅ Should do

Source: 2026 H2 Tech Debt doc.

---

**Update (From PR #16428):** Measured the actual before/after numbers and the original ~6-7min estimate was optimistic, because it didn't account for the setup overhead needed to share the build (checkout, installing python/node). Real net savings across the pipeline is closer to 1m10s.

Apt-cache savings not measured yet: the cache only helps once it has something to restore from. The very first CI run after adding it has no existing cache to pull from, so it downloads everything fresh just like before and only saves a cache afterward for next time. It's really the run after that first one where we'd see whether the cache speeds things up.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4129)

data-sync-user:
> **Mavis Ou** commented from Jira:

PR is up: [https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16428](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16428|smart-link) 

Two things included in the PR:

1. Added a shared preroll job that builds the frontend once and uploads it as an artifact, so test_wagtail and test_integration just download it instead of rebuilding from scratch. 
2. Cached apt packages in both jobs, with the cache key tied to run_id + restore-keys so it stays fresh instead of getting stuck on old package versions.

I’m also updating the target in the ticket description. Measured the actual before/after numbers and the original ~6-7min estimate was optimistic, because it didn’t account for the setup overhead needed to share the build. More details in the PR.

PR #16428, opened 2026-08-13
ci: reduce duplicate work in test_wagtail and test_integration

mmmavis:
# Description

**Problem:**
- `test_wagtail` and `test_integration` each independently rebuilt the frontend (`yarn build` + `collectstatic`)
- Both jobs hit variable apt-get install times (27s-322s per the ticket)

**Changes:**
- Added a `preroll` job that builds once and uploads the compiled assets as an artifact
- `test_wagtail` and `test_integration` now download the artifact instead of rebuilding
- `test_wagtail` drops its Node setup entirely since nothing else in that job needs it
- `test_integration` keeps its Node setup for the later Playwright steps
- Renamed the leftover `Preroll` step in both jobs to `Prepare test environment` to avoid colliding with the new job's name
- Cached each job's apt packages via `actions/cache`, keyed on `github.run_id` + `restore-keys` so it refreshes every run instead of freezing on stale package versions

**Impact:**
- Mainly a compute win (less duplicate work), not necessarily a big wall-clock (time until CI goes green) win

---

Related PRs/issues: [Jira TP1-4129](https://mozilla-hub.atlassian.net/browse/TP1-4129) / #16354 

---

# Test Notes

No new tests, this is a CI workflow change. Check the Checks tab: `Preroll`, `Wagtail CI`, and `Integration testing` all need to pass, since a broken shared artifact would surface as template/manifest errors or visible page failures in those jobs rather than as an obvious diff. Apt-cache correctness isn't verifiable from pass/fail alone (a miss just falls back to a normal install), only timing improves, see the ticket comment for that follow-up.


.github/workflows/continous-integration.yml
```diff
@@ -116,21 +169,18 @@ jobs:
...
       - name: Install Python Dependencies
         run: pip install -r requirements.txt -r dev-requirements.txt
-      - name: Install Node Dependencies
-        run: yarn install --frozen-lockfile
       - name: Install additional tooling
         run: |
           sudo apt-get update
           sudo apt-get install -y gettext ffmpeg
-      - name: Preroll
+      - name: Prepare test environment
```
mmmavis, line 191: renamed so it doesn't collide with the new job

.github/workflows/continous-integration.yml
```diff
@@ -202,10 +257,8 @@ jobs:
           sudo apt-get install -y gettext libgconf-2-4
       - name: Install Playwright
         run: yarn workspace legacy playwright:install
-      - name: Preroll
+      - name: Prepare test environment
```
mmmavis, line 278: renamed so it doesn't collide with the new job

.github/workflows/continous-integration.yml
```diff
@@ -75,9 +75,62 @@ jobs:
       - name: Run type checks
         run: mypy foundation_cms
 
+  preroll:
+    name: Preroll (Build Frontend & Collect Static)
+    runs-on: ubuntu-22.04
+    env:
```
mmmavis, line 81: also filed a follow-up ticket to tidy up `env` sections in this file: https://mozilla-hub.atlassian.net/browse/TP1-4176

ramram-mf (approved):
Thanks @mmmavis this was really needed. 🚀

Issue #16452, opened 2026-08-15
ci: decouple test_backend from preroll via non-manifest static storage in test settings


data-sync-user:
**Codebase:** Overall

**Impact:** Medium-High (rough estimate ~3min off total CI wall-clock time, based on a small sample of recent runs, needs more validation)

**Notes:** `test_backend` currently waits on `preroll` because the configured static storage needs a manifest file that only `preroll` produces. Worth investigating whether pytest can use a simpler non-manifest storage config instead, so `test_backend` can start immediately without waiting.

**Decision:** ✅ Should do

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4179)

data-sync-user:
> **Mavis Ou** commented from Jira:

PR is here: [https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16472](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16472)

PR #16472, opened 2026-08-19
ci: decouple test_backend from preroll via non-manifest static storage in tests

mmmavis:
# Description

`test_backend` waited on `preroll` only because its static storage backend required a `collectstatic` manifest that only `preroll` produces. Added `foundation_cms/settings/test.py` overriding just that backend to a non-manifest one, pointed `test_backend` at it, and dropped the `needs: preroll` dependency and its artifact download step. 

`test_integration` is untouched.

Related PRs/issues: [TP1-4179](https://mozilla-hub.atlassian.net/browse/TP1-4179) / #16452 

# How to verify

Open this PR's **Checks** tab and compare job start times for the `Continuous Integration` workflow:
- `Preroll` and `Backend Test Suite` should start at (or within a second of) the same time — previously `Backend Test Suite` waited for `Preroll` to finish first.
image
image

- `Backend Test Suite` should still pass (confirms tests run fine without the collectstatic manifest).
- `Integration Tests` should behave as before — still starts after `Preroll` finishes, since it needs the real build output, not just the settings override.


ramram-mf (approved):
Thanks @mmmavis !

Issue #16394, opened 2026-08-11
test(vitest): add Vitest coverage for shared JS utilities


data-sync-user:
**Files:** `blocks/util/{carousel,sliding_carousel}.js`, `utils/{csrf,csrf-global}.js`
**Why:** Highest-leverage tests in the codebase — carousel utils are shared logic underlying every carousel block; CSRF utils are pure logic with no DOM dependency.
**Testing notes:** Low jsdom friction, good starting point for the epic.
**Acceptance criteria:** Each listed file has a corresponding `*.test.js` covering its exported functions and primary logic branches. `yarn workspace redesign test` passes. Coverage doesn't regress for files outside this ticket's scope.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4150)

PR #16485, opened 2026-08-20
Add Vitest coverage for shared JavaScript utilities

ramram-mf:
## Description

Adds focused Vitest coverage for the shared redesign carousel and CSRF utilities, including the standalone global CSRF entry point.

## What Changed

- Covers carousel indexing, card cloning, indicator state, debouncing, responsive navigation, loop boundaries, and pointer and keyboard interactions.
- Covers CSRF cookie lookup, token minting and failure handling, native form submission, submitter preservation, and global helper exposure.

Related PRs/issues: [TP1-4150](https://mozilla-hub.atlassian.net/browse/TP1-4150), [GitHub issue #16394](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/16394)

### DIFF
--- a/foundation_cms/static/js/utils/csrf.test.js
+++ b/foundation_cms/static/js/utils/csrf.test.js
@@ -1,0 +1,107 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { ensureCsrfToken, getCookie, initCsrfForms } from "./csrf.js";
+
+function clearCsrfCookie() {
+  document.cookie = "csrftoken=; Max-Age=0; path=/";
+}
+
+describe("CSRF utilities", () => {
+  beforeEach(() => {
+    clearCsrfCookie();
+    document.body.innerHTML = "";
+  });
+
+  afterEach(() => {
+    clearCsrfCookie();
+    vi.unstubAllGlobals();
+    vi.restoreAllMocks();
+  });
+
+  it("reads and decodes a named cookie", () => {
+    document.cookie = "other=value; path=/";
+    document.cookie = "csrftoken=token%20value; path=/";
+
+    expect(getCookie("csrftoken")).toBe("token value");
+    expect(getCookie("missing")).toBe("");
+  });
+
+  it("returns an existing token without fetching", async () => {
+    document.cookie = "csrftoken=existing; path=/";
+    const fetchMock = vi.fn();
+    vi.stubGlobal("fetch", fetchMock);
+
+    await expect(ensureCsrfToken()).resolves.toBe("existing");
+    expect(fetchMock).not.toHaveBeenCalled();
+  });
+
+  it("mints and returns a token when the cookie is absent", async () => {
+    const fetchMock = vi.fn(async () => {
+      document.cookie = "csrftoken=minted; path=/";
+      return { ok: true };
+    });
+    vi.stubGlobal("fetch", fetchMock);
+
+    await expect(ensureCsrfToken()).resolves.toBe("minted");
+    expect(fetchMock).toHaveBeenCalledWith("/api/csrf/", {
+      credentials: "same-origin",
+    });
+  });
+
+  it.each([
+    ["an unsuccessful response", vi.fn().mockResolvedValue({ ok: false })],
+    ["a network failure", vi.fn().mockRejectedValue(new Error("offline"))],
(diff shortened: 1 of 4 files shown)


foundation_cms/static/js/utils/csrf.test.js
```diff
@@ -0,0 +1,107 @@
...
+
+  it.each([
+    ["an unsuccessful response", vi.fn().mockResolvedValue({ ok: false })],
+    ["a network failure", vi.fn().mockRejectedValue(new Error("offline"))],
+  ])("returns an empty token after %s", async (_label, fetchMock) => {
+    vi.stubGlobal("fetch", fetchMock);
+
+    await expect(ensureCsrfToken()).resolves.toBe("");
+  });
+
```
mmmavis, line 72: Could we add a test for the case where `/api/csrf/` returns `ok: true` but the `csrftoken` cookie still isn’t set? `ensureCsrfToken()` re-reads the cookie after a successful response, so this is a separate path from the HTTP/network failure cases and should return `""`.

foundation_cms/static/js/utils/csrf.test.js
```diff
@@ -0,0 +1,107 @@
...
+    await vi.waitFor(() => expect(submitSpy).toHaveBeenCalledOnce());
+    expect(event.defaultPrevented).toBe(true);
+    expect(form.querySelector("[data-csrf-field]").value).toBe("form-token");
+    const mirror = form.querySelector("[data-csrf-submitter]");
+    expect(mirror.type).toBe("hidden");
+    expect(mirror.name).toBe("action");
+    expect(mirror.value).toBe("share");
+  });
+
```
mmmavis, line 101: We should also add a test for what happens when `ensureCsrfToken()` fails during form submission? Right now the implementation still calls `form.submit()` even if the token is empty. It would be good to make the intended behavior explicit, since this would likely result in a CSRF failure/403.

foundation_cms/static/js/utils/csrf_global.test.js
```diff
@@ -0,0 +1,9 @@
+import { describe, expect, it } from "vitest";
+import { ensureCsrfToken, getCookie } from "./csrf.js";
+import "./csrf_global.js";
+
+describe("CSRF global entry point", () => {
+  it("exposes the shared helpers on window.FoundationCSRF", () => {
+    expect(window.FoundationCSRF).toEqual({ getCookie, ensureCsrfToken });
```
mmmavis:
```suggestion
    expect(window.FoundationCSRF.getCookie).toBe(getCookie);
    expect(window.FoundationCSRF.ensureCsrfToken).toBe(ensureCsrfToken);
```

That way the test still verifies both helpers are exposed without requiring `FoundationCSRF` to have exactly those two properties.

mmmavis (changes requested):
Looks good overall! I left a few small suggestions around edge cases and the global test, but nothing major from me. Thank @ramram-mf  👍🏻

foundation_cms/static/js/utils/csrf.test.js
```diff
@@ -0,0 +1,136 @@
...
+  it.each([
+    ["an unsuccessful response", vi.fn().mockResolvedValue({ ok: false })],
+    ["a network failure", vi.fn().mockRejectedValue(new Error("offline"))],
+  ])("returns an empty token after %s", async (_label, fetchMock) => {
+    vi.stubGlobal("fetch", fetchMock);
+
+    await expect(ensureCsrfToken()).resolves.toBe("");
+  });
```
mmmavis, line 71: Could we build the mocks inside the test (or pass a factory into `it.each`) so `restoreAllMocks` can’t strip them? Something like:

```js
it.each([
  ["an unsuccessful response", () => vi.fn().mockResolvedValue({ ok: false })],
  ["a network failure", () => vi.fn().mockRejectedValue(new Error("offline"))],
])("returns an empty token after %s", async (_label, createFetchMock) => {
  const fetchMock = createFetchMock();
  vi.stubGlobal("fetch", fetchMock);

  await expect(ensureCsrfToken()).resolves.toBe("");
  expect(fetchMock).toHaveBeenCalledOnce();
});

mmmavis (changes requested):
One last issue. We are so close!

mmmavis (approved):
🎉

Issue #16395, opened 2026-08-11
test(vitest): add Vitest coverage for accordion/tab/counter blocks


data-sync-user:
**Files:** `blocks/{accordion_block,hero_accordion,tabbed_content_container,impact_numbers_block}.js`
**Why:** Toggle/expand-state and counter-animation logic — high value for unit testing per the codebase's testing priorities.
**Testing notes:** `impact_numbers_block.js` uses `IntersectionObserver`/`matchMedia`/`requestAnimationFrame` — mock all three.
**Acceptance criteria:** Each listed file has a corresponding `*.test.js` covering its exported functions and primary logic branches. `yarn workspace redesign test` passes. Coverage doesn't regress for files outside this ticket's scope.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4151)

PR #16493, opened 2026-08-21
Add Vitest coverage for accordion, tab, and counter blocks

ramram-mf:
## Description

Adds unit coverage for four existing JavaScript block modules without changing production code.

## What Changed

- Covers accordion initialization, open/close exclusivity, and height-transition behavior.
- Covers horizontal hero accordion ARIA state, keyboard activation, responsive width calculations, Vimeo iframe creation, and video-panel restoration.
- Covers tab switching, card pagination, indicators, scrolling, and swipe boundaries.
- Covers impact-number observation, parsing and formatting, reduced-motion handling, count-up animation, invalid values, and completed-value/repeat-animation guards with mocked `IntersectionObserver`, `matchMedia`, and `requestAnimationFrame`.

## QA

Run `yarn workspace redesign test`.

Verify that:

- The accordion, hero accordion, tabbed content, and impact-number suites pass.
- The four existing block modules are covered without regressions elsewhere in the redesign Vitest suite.

Related PRs/issues: [TP1-4151](https://mozilla-hub.atlassian.net/browse/TP1-4151), [GitHub issue #16395](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/16395)


foundation_cms/static/js/blocks/accordion_block.test.js
```diff
@@ -0,0 +1,153 @@
...
+
+  it("leaves panels unchanged when they already have the requested state", () => {
+    const { root, triggers, panels } = createAccordion();
+    const accordion = new AccordionBlock(root);
+    accordion.openAccordion(triggers[0], panels[0]);
+    accordion.closeAccordion(triggers[1], panels[1]);
+
+    expect(requestAnimationFrame).not.toHaveBeenCalled();
+    expect(panels[0].style.height).toBe("auto");
+    expect(panels[1].hidden).toBe(true);
```
mmmavis, line 135: This assertion is already `true` before `closeAccordion` runs (the fixture starts with hidden set, line 25), so it would stay green even if the early-return guard were deleted. It doesn't verify the guard is actually there.

foundation_cms/static/js/blocks/tabbed_content_container.test.js
```diff
@@ -0,0 +1,172 @@
...
+
+  it("does not add carousel interactions when all cards fit on one page", () => {
+    const panel = createCardSet({ cards: 2, cardsPerPage: 3 });
+
+    initTabbedContentCardSets();
+    panel.dispatchEvent(createTouchEvent("touchstart", 100));
+    panel.dispatchEvent(createTouchEvent("touchend", 0));
+
+    expect(panel.querySelectorAll(".tab-card__page")).toHaveLength(1);
+    expect(panel.scrollTo).not.toHaveBeenCalled();
```
mmmavis, line 139: with `pages=1`, `newIndex` still ends up equal to `currentIndex` whether or not the pages <=1 guard is there, so this test would pass even if that guard got removed. It does correctly verify the page-wrapping (`toHaveLength(1)`), just doesn't actually prove it `"does not add carousel interactions"` like the title says.

```suggestion
    expect(panel.scrollTo).not.toHaveBeenCalled();
    
    const indicators = panel.querySelectorAll(".carousel-indicators__item");
    expect(indicators[0].hasAttribute("aria-current")).toBe(false);
```

foundation_cms/static/js/blocks/hero_accordion.test.js
```diff
@@ -0,0 +1,183 @@
...
+    const { panels } = createHorizontalAccordion();
+    panels[0].classList.remove("active");
+    const overlay = panels[0].querySelector(".hero-accordion__video-overlay");
+    panels[0].querySelector("iframe").remove();
+
+    initAllHorizontalAccordions();
+    overlay.click();
+
+    expect(panels[0].querySelector("iframe")).toBeNull();
+    expect(overlay.classList).toContain("hidden");
```
mmmavis: The overlay never had its `hidden` class removed before this click, so this stays true no matter what. The iframe check above it is what's actually doing the work here. Might be worth un-hiding the overlay first?

foundation_cms/static/js/blocks/accordion_block.test.js
```diff
@@ -0,0 +1,153 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
```
mmmavis, line 1: Every test here closes a panel only after its height already settled to "auto" (either from the fixture or a prior transitionend), so this only ever hits the `scrollHeight` side of line 78's ternary (from `accordion_block.js`). Can we add a case that closes a panel mid-open, before it reaches "auto", to cover the `offsetHeight path` too?

foundation_cms/static/js/blocks/hero_accordion.test.js
```diff
@@ -0,0 +1,183 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
```
mmmavis, line 1: Every click test here targets a panel that's not already active (panels[1] or panels[2]). Panel 0 starts active in the fixture, but no test clicks it to check that the guard on line 57 (`if (!panel.classList.contains(active))`) actually skips `_deactivateAll`/`_activate` when a panel is clicked while already open. Can we add a case that clicks the already-active panel and confirms nothing changes?

foundation_cms/static/js/blocks/tabbed_content_container.test.js
```diff
@@ -0,0 +1,172 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
```
mmmavis, line 1: All the tests here only ever set up a single `.tabbed-content-container`, but the comment on line 20 says panels are addressed by index specifically so multiple instances can coexist on a page. Might be worth a test with two containers, just to confirm clicking a tab in one doesn't leak into the other.

foundation_cms/static/js/blocks/hero_accordion.test.js
```diff
@@ -0,0 +1,183 @@
...
+    expect(widthSpy).toHaveBeenCalledOnce();
+    panels.forEach((panel) => {
+      expect(panel.getAttribute("role")).toBe("button");
+      expect(panel.getAttribute("tabindex")).toBe("0");
+    });
+    expect(panels[0].getAttribute("aria-expanded")).toBe("true");
+    expect(panels[1].getAttribute("aria-expanded")).toBe("false");
+  });
+
+  it("activates a clicked panel and restores the previous video panel", () => {
```
mmmavis, line 77: This only ever deactivates the video panel (panel 0), so the early return on line 154 (`if (!wrapper) return;`) never really gets exercised. Might be nice to also test clicking between two non-video panels, just to be safe.

mmmavis (changes requested):
Nice addition overall. Thanks @ramram-mf . I left a few inline comments to flag some assertions that would still pass even without the guard they are meant to test, plus a few branches that are not covered yet.

mmmavis (approved):
🎉

Issue #16398, opened 2026-08-11
test(vitest): add Vitest coverage for carousel blocks


data-sync-user:
**Files:** `blocks/{spotlight_card_carousel,portrait_card_carousel,image_carousel_block,project_block,illustrated_newsletter_signup_block}.js`, `blocks/product_review_carousel/*.js` (a11y, carousel, config, handlers, index, tick)
**Why:** Carousel index/wrap-around logic across all remaining carousel-family blocks; `product_review_carousel/*` kept together as a tightly coupled module cluster.
**Testing notes:** Largest ticket in scope (11 files) — split into multiple PRs if needed. Mix of `IntersectionObserver`, `ResizeObserver`, `matchMedia`, `requestAnimationFrame`, `gettext`, `fetch`/`XMLHttpRequest` depending on file.
**Acceptance criteria:** Each listed file has a corresponding `*.test.js` covering its exported functions and primary logic branches. `yarn workspace redesign test` passes. Coverage doesn't regress for files outside this ticket's scope.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4152)

PR #16495, opened 2026-08-21
Add Vitest coverage for the spotlight carousel

ramram-mf:
## Description

Adds unit coverage for the existing spotlight card carousel without changing production code.

## What Changed

- Covers desktop positioning, active-card accessibility, controls, indicators, and measured layout.
- Covers mobile track cloning, touch movement, swipe thresholds, and infinite-loop rebasing.
- Covers keyboard activation, disabled controls, responsive reinitialization, and empty roots.

## QA

- The spotlight card carousel suite passes.


foundation_cms/static/js/blocks/spotlight_card_carousel.test.js
```diff
@@ -0,0 +1,286 @@
...
+      value: 1280,
+    });
+    resizeHandler();
+    vi.advanceTimersByTime(200);
+    const desktopTransform = slides.style.transform;
+    slides.dispatchEvent(touchEvent("touchstart", 100));
+    expect(slides.style.transform).toBe(desktopTransform);
+  });
+
+  it("rebases mobile clone positions after transform transitions", () => {
```
mmmavis, line 255: Nice coverage on the forward wrap. Could we also add a case for swiping/clicking prev past the first card? That direction's infinite-loop reset never runs in the current tests.

mmmavis (changes requested):
Thanks @ramram-mf ! Just one more test I'm hoping we can add (see inline comment)

foundation_cms/static/js/blocks/spotlight_card_carousel.test.js
mmmavis: Similar to the comment in the other PR - since the button's disabled here, this click never actually fires, so it's not really testing our `hasAttribute("disabled")` guard. Could we test it more directly, maybe by removing the disabled attribute right before asserting?

foundation_cms/static/js/blocks/spotlight_card_carousel.test.js
```diff
@@ -0,0 +1,307 @@
...
+    lastCard.dispatchEvent(keyEvent);
+    expect(keyEvent.defaultPrevented).toBe(true);
+    expect(counter.textContent).toBe("1");
+
+    next.setAttribute("disabled", "");
+    next.click();
+    expect(counter.textContent).toBe("1");
+  });
+
+  it("creates a tripled mobile track and handles swipe movement", () => {
```
mmmavis, line 183: Could we add a case mirroring this one but with clientX moving in the opposite direction?

foundation_cms/static/js/blocks/spotlight_card_carousel.test.js
```diff
@@ -0,0 +1,307 @@
...
+  });
+
+  it("reinitializes only when the responsive breakpoint changes", () => {
+    const root = createCarousel();
+    initSpotlightCardCarousels();
+    const slides = root.querySelector(".spotlight-card-carousel__slides");
+
+    resizeHandler();
+    vi.advanceTimersByTime(200);
+    expect(slides.querySelectorAll(".spotlight-card")).toHaveLength(3);
```
mmmavis, line 306: This assertion passes even if the `viewportChanged` check in `handleResize` were removed entirely, since a full desktop reinit also produces 3 cards. Could we assert on something that'd actually distinguish "skipped" from "reinitialized," like spying on `updateContainerHeight`/`initDesktop`, or checking a mobile-only side effect didn't fire?

mmmavis (changes requested):
We are almost there!

mmmavis (approved):
🎉 Thank you Ram!

PR #16496, opened 2026-08-21
Add Vitest coverage for the project block

ramram-mf:
## Description

Adds unit coverage for the existing project block carousel without changing production code.

## What Changed

- Covers slide initialization, clones, ARIA state, controls, indicators, and active-index updates.
- Covers button navigation, wrap-around normalization, touch and mouse gestures, thresholds, and cancellation.
- Covers video autoplay, pause/play controls, playback failure, and single-slide and empty-root paths.

## QA

- The project block suite passes.


foundation_cms/static/js/blocks/project_block.test.js
```diff
@@ -0,0 +1,260 @@
...
+
+    expect(root.querySelectorAll("[data-project-block-slide]")).toHaveLength(1);
+    expect(root.querySelector("[data-direction='prev']").disabled).toBe(true);
+    expect(root.querySelector("[data-direction='next']").disabled).toBe(true);
+    expect(
+      root
+        .querySelector("[data-project-block-pause]")
+        .classList.contains("is-hidden"),
+    ).toBe(true);
+    root.querySelector("[data-direction='next']").click();
```
mmmavis: Since the button's disabled here, this click never actually fires, so it's not really testing `goTo`'s guard. Might be worth dropping the click line, or enabling the button first if we want to check the guard itself.

foundation_cms/static/js/blocks/project_block.test.js
```diff
@@ -0,0 +1,260 @@
...
+    expect(trackSlides[4].getAttribute("aria-hidden")).toBe("true");
+    expect(
+      root.querySelector("[data-project-block-track]").style.transform,
+    ).toBe("translateX(-100%)");
+    expect(root.querySelector("[data-active-index]").textContent).toBe("1");
+    expect(
+      root
+        .querySelector(".carousel-indicators__item")
+        .getAttribute("aria-current"),
+    ).toBe("true");
```
mmmavis: This only checks `aria-current`, but the active-dot styling is actually driven by the 1carousel-indicators__item--active` class. I'm thinking we should assert that class too, so a class-toggle regression wouldn't slip through.

mmmavis (changes requested):
Thanks @ramram-mf . Two tiny improvement needed :)

mmmavis (changes requested):
Thanks @ramram-mf . Two tiny improvements needed :)

mmmavis (approved):
Thank you!

PR #16497, opened 2026-08-21
Add Vitest coverage for illustrated newsletter signup

ramram-mf:
## Description

Adds unit coverage for the existing illustrated newsletter signup block without changing production code.

## What Changed

- Covers input and select state, localized options, unique error IDs, and validation focus and ARIA errors.
- Covers form expansion, success transitions, reduced-motion paths, and animation cleanup.
- Covers loading state, CSRF-backed request payloads, default and selected languages, success, and network failure.

## QA

- The illustrated newsletter signup suite passes.


foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+
+vi.mock("../utils/csrf.js", () => ({
+  ensureCsrfToken: vi.fn(),
+}));
+
+vi.mock("../components/newsletter_signup/data/country_options.js", () => ({
+  COUNTRY_OPTIONS: [
+    { value: "", label: "Your Country" },
+    { value: "CA", label: "Canada" },
+    { value: "MX", label: "Mexico" },
```
mmmavis, line 13: 🤝

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+
+    initIllustratedNewsletterSignups();
+
+    const email = first.querySelector("input[name='email']");
+    const country = first.querySelector("select[name='country']");
+    const language = first.querySelector("select[name='language']");
+    expect(email.dataset.hasValue).toBe("true");
+    expect(country.options[0].textContent).toBe("Choose a country");
+    expect(country.options[0].disabled).toBe(true);
+    expect(country.options.length).toBeGreaterThan(2);
```
mmmavis: `toBeGreaterThan(2)` will still pass if `populateSelectOptions`'s filter for the empty-value entry in `COUNTRY_OPTIONS` broke and produced a duplicate placeholder (4 options instead of 3). Since the mock data has a known, fixed shape here, this can be exact (e.g, own placeholder + Canada + Mexico = 3 options)

```suggestion
    expect(country.options.length).toBe(3);
```

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+    transitionEnd(form, "opacity");
+    expect(form.style.height).toBe("280px");
+    transitionEnd(form, "height");
+    expect(form.style.height).toBe("");
+    expect(form.style.overflow).toBe("");
+
+    container
+      .querySelector("input[name='email']")
+      .dispatchEvent(new Event("focus"));
+    expect(container.dataset.state).toBe("expanded");
```
mmmavis: The email focus handler is `{ once: true }`, so this second dispatch triggers nothing and the assertion can't fail. Let's drop it or assert something the guard could actually break.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+      >
+        <div class="illustrated-newsletter-signup__signup-view">
+          <form class="illustrated-newsletter-signup__form">
+            <input name="email" type="email">
+            <select name="country"></select>
+            <label>Choose a country</label>
+            <select name="language"></select>
+            <label>Choose a language</label>
+            <input name="privacy" type="checkbox">
+            <div class="illustrated-newsletter-signup__expanded" hidden>More fields</div>
```
mmmavis: Production has two `.illustrated-newsletter-signup__expanded` containers (privacy/submit is a separate one from country/language), but this fixture only has one, so a regression that only revealed the first match wouldn't be caught. Worth adding a second `.expanded` wrapper around the privacy checkbox/submit button to match the real markup.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+    ).toBe(false);
+    transitionEnd(container, "height");
+    expect(
+      container.querySelector(".illustrated-newsletter-signup__signup-view")
+        .hidden,
+    ).toBe(true);
+    expect(document.activeElement).toBe(success);
+    expect(container.style.height).toBe("");
+  });
+
```
mmmavis, line 287: The `setTimeout` fallback for when `transitionend` never fires is never triggered, every animation test completes it manually. Worth adding one test that uses `vi.advanceTimersByTime(...)` instead of dispatching `transitionend`.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
```diff
@@ -0,0 +1,336 @@
...
+      container.querySelector(".illustrated-newsletter-signup__server-error")
+        .hidden,
+    ).toBe(false);
+    expect(button.disabled).toBe(false);
+    expect(button.hasAttribute("aria-busy")).toBe(false);
+    expect(
+      Array.from(button.querySelectorAll("span"), (label) => label.textContent),
+    ).toEqual(["Sign up", "Join"]);
+  });
+
```
mmmavis, line 312: Only the fetch-rejects path is tested for errors. It would be good if we can add a case where `fetchMock` resolves with a non-201 status (e.g. 403 from a bad CSRF token) to cover that branch too.

mmmavis (changes requested):
👀

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
ramram-mf: Updated this to assert exactly three options, matching the fixture’s empty, Canada, and Mexico entries.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
ramram-mf: Removed the redundant second focus event. The test now submits an invalid form after expansion and verifies the expansion guard leaves the height and overflow styles clean without making a request.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
ramram-mf: Updated the fixture to match the production markup with both expandable wrappers, and added assertions that both are revealed in the animated and reduced-motion paths.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
ramram-mf: Added a separate fake-timer test for the success-state timeout fallback. It avoids dispatching `transitionend` and verifies cleanup and focus after the 350 ms fallback.

foundation_cms/static/js/blocks/illustrated_newsletter_signup_block.test.js
ramram-mf: Added a resolved 403 response case that verifies the complete retry state: server error displayed, submit button restored, loading attributes removed, form retained, and success state hidden.

mmmavis (approved):
🍻

PR #16498, opened 2026-08-21
Add Vitest coverage for image and portrait carousels

ramram-mf:
## Description

Adds unit coverage for the existing image and portrait card carousel blocks without changing production code.

## What Changed

- Covers image carousel initialization, infinite-track construction, controls, keyboard navigation, swipe thresholds, and empty pages.
- Covers portrait card design cycling, cloned tracks, margin-based spacing, controls, indicators, multiple roots, and empty pages.
- Exercises the shared sliding-carousel behavior through both block entry points with mocked layout and animation APIs.

## QA

- The image and portrait card carousel suites pass.
- The existing modules are covered without regressions elsewhere in the redesign Vitest suite.


foundation_cms/static/js/blocks/image_carousel_block.test.js
```diff
@@ -0,0 +1,109 @@
...
+    .querySelectorAll(".image-carousel")
+    .item(document.querySelectorAll(".image-carousel").length - 1);
+}
+
+function touchEvent(type, clientX) {
+  const event = new Event(type, { bubbles: true });
+  Object.defineProperty(event, "touches", { value: [{ clientX }] });
+  Object.defineProperty(event, "changedTouches", { value: [{ clientX }] });
+  return event;
+}
```
mmmavis, line 44: This sets touches on `touchend` too, but in an actual touch event it's empty by then. If `touchstart`/`touchend` ever read the wrong list, this wouldn't catch it. Might be worth only setting the list that's actually populated for each event type.

foundation_cms/static/js/blocks/portrait_card_carousel.test.js
```diff
@@ -0,0 +1,101 @@
...
+
+    initPortraitCardSetCarousels();
+
+    const cards = root.querySelectorAll(".portrait-card");
+    expect(cards).toHaveLength(15);
+    expect(
+      Array.from(cards)
+        .slice(5, 10)
+        .map((card) => card.dataset.cardDesign),
+    ).toEqual(["0", "1", "2", "3", "0"]);
```
mmmavis: Since `data-card-design` gets set before the cards are tripled, all three thirds naturally end up with the same `["0","1","2","3","0"]` sequence, so this doesn't quite confirm the originals landed in the middle third specifically. Comparing against `originalItems` directly (or checking node identity) might get closer to what this is meant to verify.

mmmavis (changes requested):
This looks great overall! Left a couple small comments, would appreciate those addressed before merging 😃

mmmavis (approved):
🎉

Issue #16405, opened 2026-08-11
test(percy): add Nothing Personal home/article coverage


data-sync-user:
**Files:** `frontend/redesign/tests/redesign-urls.js`, `.github/workflows/visual-regression-testing-redesign.yml` (path filters + script)

**Why:** `generate_profiles` already creates a Nothing Personal home factory page + 5 curated article factory pages as a side effect (`foundation_cms/profiles/factories.py:98-169`) — these exist today with zero Percy coverage or trigger path.

**Implementation notes:** Add entries for Nothing Personal home (`/nothing-personal/`) and one article detail (`/nothing-personal/expert-profile-article-1/`). Add `foundation_cms/nothing_personal/**` to both path-filter blocks and the script. Do **not** attempt product-review or podcast page-type entries here — those page types have zero factory pages until TP1-4138's `nothing_personal` factory tickets land; track that coverage as a follow-up.

**Acceptance criteria:** `redesign-urls.js` has home + article-detail entries; snapshots pass; `foundation_cms/nothing_personal/**` triggers the workflow; product-review/podcast coverage explicitly deferred.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4161)

PR #16504, opened 2026-08-21
Add Percy coverage for Nothing Personal pages

ramram-mf:
## Description

Adds Percy visual-regression coverage for the Nothing Personal home and a seeded article while preserving the existing homepage snapshot. It also gives the generated Nothing Personal home deterministic hero and featured content without overwriting editorial selections.

## What Changed

- Adds named snapshots for `/nothing-personal/` and `/nothing-personal/expert-profile-article-1/`.
- Includes `foundation_cms/nothing_personal/**` in the redesign workflow's labeled-PR, main-push, and approval-time changed-file filters.
- Seeds one hero and two distinct featured articles from the stable curated-article fixtures, and repairs missing relationships without deleting, reordering, or overwriting existing selections.
- Gives newly seeded curated articles a fixed first-publication date so fresh Percy databases produce stable article snapshots.
- Keeps repeated profile generation idempotent, including custom-hero, missing-hero, reusable-slot, and full-capacity cases, with factory test coverage for those behaviors.
- Keeps product-review and podcast routes deferred until their factory pages are available through TP1-4138.

## QA

Verify that:

- The redesign visual spec includes `Homepage`, `Nothing Personal Home`, and `Nothing Personal Article`.
- Generating profiles gives the Nothing Personal home one hero and two distinct featured articles, while a second run preserves article IDs, publication dates, relationship order, and revision counts.
- Existing custom hero and featured selections remain intact while eligible missing relationships are filled without exceeding the featured-row limit.
- Changes under `foundation_cms/nothing_personal/` qualify the redesign visual-regression workflow, while product-review and podcast routes remain excluded.
- A redesign Percy run for the latest PR head renders meaningful hero and featured content on the home snapshot and a coherent populated article snapshot with a stable publication date.

## Review Notes

Scope is intentionally limited to deterministic Nothing Personal fixture data and redesign Percy coverage. Richer homepage fixture content can be added separately without blocking this baseline.

Related PRs/issues: [TP1-4161](https://mozilla-hub.atlassian.net/browse/TP1-4161), [GitHub issue #16405](https://github.com/MozillaFoundation/foundation.mozilla.org/issues/16405)


mmmavis (changes requested):
Thanks @ramram-mf . I can see the two new snapshots show up in Percy result. However, [NP homepage is quite empty](https://percy.io/Mozilla-Foundation/web/mozillafoundation.org-Redesign-95a4cbce/builds/53282134/changed/2884702932?browser=safari&browser_ids=63%2C73%2C74%2C76&currentBuildItemIndex=1&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_private&viewLayout=side-by-side&viewMode=new&width=1280&widths=375%2C1280) (missing main content) and hence not really useful for visual regression tests. Can we get `ensure_expert_curated_articles` (or `ensure_nothing_personal_home`) in `foundation_cms/profiles/factories.py` to link one of the seeded articles as `hero_item` and a couple as featured items, so the home page actually renders something? Otherwise it might be worth holding off on the home page snapshot until that seed data exists.

foundation_cms/profiles/factories.py
```diff
@@ -158,15 +161,79 @@ def ensure_expert_curated_articles(root, default_locale, topics, expert_pages, f
...
+    empty_featured_items = [item for item in featured_items if not item.page_id]
+    next_featured_sort_order = (
+        max(
+            (item.sort_order if item.sort_order is not None else -1 for item in featured_items),
+            default=-1,
+        )
+        + 1
+    )
+
+    for article in articles[1:]:
```
mmmavis: Small thing here: since the loop already checks `article.pk == hero_item_id` to skip the real hero, slicing off `articles[0]` up front feels redundant, and it actually backfires once the hero isn't `articles[0]` anymore. Might be worth just iterating the full articles list and letting the existing check do its job?

foundation_cms/profiles/factories.py
```diff
@@ -158,15 +161,79 @@ def ensure_expert_curated_articles(root, default_locale, topics, expert_pages, f
...
-    expert.save_revision().publish()
-    print(f"  {len(articles)} curated articles linked to {expert.title}.")
+    if expert_changed:
+        expert.save_revision().publish()
+        print(f"  {len(articles)} curated articles linked to {expert.title}.")
+
+    home_changed = False
+    if not np_home.hero_item_id:
+        np_home.hero_item = articles[0]
+        home_changed = True
```
mmmavis: Noticed one bug here: we skip the hero when picking featured items, but not the reverse, if `articles[0]` is already featured when hero is empty, it becomes both hero and featured, and we burn our last slot creating a replacement. It'd be good that we skip it as the hero candidate if it's already featured?

mmmavis (changes requested):
Good to see content on NP homepage populated!

foundation_cms/profiles/factories.py
ramram-mf: Updated featured repair to consider the complete seeded article list while continuing to skip the effective hero and existing featured selections. Added coverage with a non-default hero to prove article 1 remains an eligible deterministic candidate.

foundation_cms/profiles/factories.py
ramram-mf: Updated missing-hero repair to select the first seeded article that is not already featured, leaving the hero empty when no eligible candidate exists. Added coverage proving featured rows and ordering are preserved, hero/featured duplication is avoided, and a second run causes no revision churn.

mmmavis (approved):
We should probably seed a few more of the homepage's components (tagline, body blocks, the article collection link) so the Percy snapshot looks closer to what's actually live, but this is a solid starting point for getting the NP homepage under visual regression coverage. Approving. Thanks @ramram-mf !

Issue #16430, opened 2026-08-13
ci: shard test_wagtail pytest suite across parallel jobs to reduce CI wall-clock time


data-sync-user:
**Codebase:** Overall
**Impact:** Medium-High
**Notes:** Wagtail CI is our real bottleneck at ~18-20min, since test_integration and preroll both run in parallel with it and finish well before. Right now its pytest suite only parallelizes within one runner via `-n auto`, so it's capped by that runner's core count. Splitting it into a matrix of parallel jobs, each running a slice of the suite, would cut wall-clock further than that. There's no dollar cost since this is a public repo and GitHub-hosted runners are free, and I checked recent runs for queueing and found none, so this shouldn't create contention either. The real cost is engineering time: balancing the shards evenly and merging coverage across them.
**Decision:** ✅ Should do

---

**Terminology (so this doesn't get confusing later):** 

- **Wall-clock time**: elapsed time from CI trigger to finishing (going green); what a developer waiting on a PR actually experiences. Bounded by the critical path (the longest chain of dependent jobs), not the sum of all jobs.
- **Compute / billed minutes**: total runner-minutes summed across every job. This is what gets billed on a paid plan; not applicable here since this repo is public and hosted-runner minutes are free regardless.
- **"Run time"**: ambiguous, avoid using alone. Sometimes means a single job's own duration, sometimes used loosely for either of the above. Say "wall-clock" or "compute" instead.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4177)

PR #16512, opened 2026-08-25
ci: shard Backend Test Suite, cut per-worker migration tax

mmmavis:
**Note for Mavis**: Update branch protection right after this PR is merged into `main`.

Related PRs/issues: [Jira TP1-4177](https://mozilla-hub.atlassian.net/browse/TP1-4177) / #16430 

---

## Summary

TP1-4177 asked for one thing: split the `Backend Test Suite` CI job across several parallel jobs, since pytest's own parallelism (`-n auto`) was already maxed out on a single runner.

While sizing that work, I profiled the job. Most of its ~10.5 minutes wasn't tests at all. Only ~4 minutes was actual test execution. The rest was each parallel test worker rebuilding its own database from scratch, replaying every migration the app has ever had.

Splitting the job further without fixing that wouldn't have helped much. It would have just multiplied the rebuild cost across more workers, not shrunk it. So this PR does two things: 
1. build the database once and clone it cheaply per worker
2. then shard the test run across 4 parallel jobs on top of that fix.

## Measured result

| Shard | Duration |
|---|---|
| 1/4 | 10m05s |
| 2/4 | 12m29s |
| 3/4 | 14m28s |
| 4/4 | 8m22s |

**Overall: ~14m28s, down from ~17m11s (~15% faster)**

Shard spread (8m22s–14m28s) is wider than duration-based balancing should produce. I created a follow-up to regenerate `.test_durations` properly and rebalance: [TP1-4211](https://mozilla-hub.atlassian.net/browse/TP1-4211).

## Commits

1. `build(deps):` add pytest-split
3. `ci:` add `.test_durations` for shard balancing (reconstructed from a real run's logs, not a local `--store-durations` run — one-time seed, not authoritative)
4. `perf(ci):` `prepare_sharded_test_databases` — clone a migrated template DB per xdist worker instead of each replaying migrations. Verified locally against real Postgres: clone takes 0.385s vs. several minutes
5. `ci:` shard `test_backend` into 4 jobs via pytest-split, wire in `--reuse-db`
6. `ci:` combine per-shard coverage into one report
7. `ci:` add `backend_test_suite` gate job so branch protection has one stable check name regardless of shard count


.github/workflows/continuous-integration.yml
```diff
@@ -143,8 +148,68 @@ jobs:
...
+        # itself (see prepare_sharded_test_databases for why this matters here).
+        run: python ./manage.py prepare_sharded_test_databases ${{ steps.workers.outputs.count }}
       - name: Run Tests
-        run: cd foundation_cms && pytest -n auto -v --ds=foundation_cms.settings.test --cov=foundation_cms --cov-report=term-missing
+        env:
+          COVERAGE_FILE: .coverage.${{ matrix.shard }}
+        run: >-
+          cd foundation_cms && pytest --reuse-db -n ${{ steps.workers.outputs.count }}
+          --splits ${{ strategy.job-total }} --group ${{ matrix.shard }}
+          -v --ds=foundation_cms.settings.test --cov=foundation_cms --cov-report=
```
mmmavis, line 165: `--cov-report=` is empty on purpose. We want to suppress this shard's own terminal coverage report since a quarter of the suite's coverage % alone is misleading. `--cov` still writes the data file for `combine_coverage` to merge.

.github/workflows/continuous-integration.yml
```diff
@@ -143,8 +148,68 @@ jobs:
...
+      - name: Upload coverage data
+        uses: actions/upload-artifact@v4
+        with:
+          name: coverage-data-${{ matrix.shard }}
+          path: foundation_cms/.coverage.${{ matrix.shard }}
+          retention-days: 1
+          include-hidden-files: true
```
mmmavis, line 172: This line is required. `actions/upload-artifact@v4` excludes dotfiles by default, hence our new `.coverage.N` files. Without this the upload silently succeeds but uploads nothing.

.github/workflows/continuous-integration.yml
```diff
@@ -143,8 +148,68 @@ jobs:
...
+    # check for branch protection -- see PR description for the required repo
+    # settings change needed alongside this.
+    name: Backend Test Suite
+    runs-on: ubuntu-22.04
+    needs: test_backend
+    if: always()
+    steps:
+      - name: Check shard results
+        run: |
+          if [ "${{ needs.test_backend.result }}" != "success" ]; then
```
mmmavis, line 209: Not 100% sure about this. Does `needs.test_backend.result` reliably come back as "failure" if even one shard in the matrix fails (with `fail-fast: false`)? I think that's how GitHub Actions aggregates matrix results but I haven't verified it myself. Would appreciate a second pair of eyes here since this is the whole gate branch protection relies on. If I'm wrong about this, the check could pass even when a shard actually failed. 😓

foundation_cms/legacy_apps/utility/management/commands/prepare_sharded_test_databases.py
```diff
@@ -0,0 +1,65 @@
+from django.core.management import call_command
```
mmmavis, line 1: I worked with Claude on this file. This approach and code makes sense to me and I verified it end to end against a real local Postgres, but I would still appreciate an extra careful pass here 😀

dev-requirements.txt
```diff
@@ -2,7 +2,7 @@
 # This file is autogenerated by pip-compile with Python 3.11
 # by the following command:
 #
-#    pip-compile --output-file=dev-requirements.txt.tmp dev-requirements.in
+#    pip-compile --output-file=dev-requirements.txt dev-requirements.in
```
mmmavis, line 5: fixing a pre-existing typo here

foundation_cms/.test_durations
```diff
@@ -0,0 +1,964 @@
+{
```
mmmavis, line 1: Reconstructed from real CI log timestamps, not `pytest --store-durations` (see [commit message](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16512/changes/a558bb3e4e2d46e7b68728c950da27b712b76c4e) for how/why). Shards came out uneven on the first real run (8m22s–14m28s). This will be fixed in follow-up to regenerate properly: [TP1-4211](https://mozilla-hub.atlassian.net/browse/TP1-4211).

foundation_cms/legacy_apps/utility/management/commands/prepare_sharded_test_databases.py
robdivincenzo: Approach is sound and also verified locally! Things I was considering sound like they'll be handled in the follow-ups you've already document. 👍

foundation_cms/.test_durations
robdivincenzo: Perfect 🎉

.github/workflows/continuous-integration.yml
robdivincenzo: @mmmavis [Verified!](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16516) It will fail if one shard fails 👍

robdivincenzo (approved):
@mmmavis awesome work, and approved! This will lay a nice foundation for us to continue optimizing our CI and test durations. Left you a comment on the gate question with a verification test, it fails as expected.

.github/workflows/continuous-integration.yml
mmmavis: Yay awesome news! Thanks for verifying it.

Issue #16467, opened 2026-08-18
Remove PNI from the codebase


data-sync-user:
Remove PNI / Buyers Guide from the codebase. Code is fully dead (project sunset, URLs already redirected). I’m going to so split the work into several independent PRs merged directly to `main` rather than one big PR or a long-lived feature branch.

- -Pre-flight: confirm Redirect records cover all old `/privacynotincluded/...` paths-
- -[PR 1](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16518): Remove PNI from CI/Playwright test surface (integration specs, foundation-urls.js, base-urls.js, visual regression baselines) ((/) **Status: Merged into** `main` )-
- [PR 2](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16522): Remove the buyersguide app entirely (cross-references, pagemodels, factory, tests, migration)  (🔍 **Status: PR filed and ready for review. PR should not be merged until we have a** [**PNI DB backup**](https://mozilla-hub.atlassian.net/browse/TP1-4202)  )
- [PR 3](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16524): Remove buyersguide templates (pages, fragments, previews, admin icon) + clean up styleguide.html references (:police_car_light: **Status: PR filed and ready for review. But it should not be merged until Part 2 work (**[**PR #16522**](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16522)**) has actually been deployed to production)**
- PR 4: Remove `buyersguide` static assets (JS, SCSS, images, compiled output) + clean up shared components
- PR 5: Remove PNI/buyersguide settings, env vars, docs, and config references (`settings.py,` README, `tasks.py`, `pyproject.toml`)

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4191)

data-sync-user:
> **Mavis Ou** commented from Jira:

fyi `@Adalberto Cima` , this is how I’m going to break up the work ^

PR #16294, opened 2026-07-29
Notification Banner

robdivincenzo:
# Description

This PR contains a new snippet and block rendering for a notice banner. It is attached to the `Promote` tab of both the current and legacy base pages so that it is available across all page types.

Link to sample test page:
Related PRs/issues: TP1-4058

## Heroku preview app
View  on preview app-> https://foundation-s-tp1-4058-s-vtxrbx.mofostaging.net/en/nothing-personal/expert-profile-article-5/
CMS -> https://foundation-s-tp1-4058-s-vtxrbx.mofostaging.net/cms/
credentials `SSO`

## Jira ticket
https://mozilla-hub.atlassian.net/browse/TP1-4058

## Result
Screenshot 2026-08-25 at 1 03 39 PM

PR #16522, opened 2026-08-27
refactor(legacy): remove the buyersguide app (PNI sunset work 2/5)

mmmavis:
Related PRs/issues: [Jira TP1-4191](https://mozilla-hub.atlassian.net/browse/TP1-4191) / #16467

---

# 🚨 Risks before merging/deploying 🚨

1. **Destructive migration.** This PR's migration permanently deletes every BuyersGuidePage/ProductPage/etc. row in the production database, not just test data. Ensure we have a DB backup before deploying, standard practice for any migration that drops tables. ✅ Confirmed.
2. **Depends on Redirects already covering old PNI URLs.** Once the pages are deleted, an old `/privacynotincluded/...` URL only redirects correctly if a matching Redirect record already exists for that exact path in production. If it doesn't, visitors get a 404 instead. Confirm redirects have been configured already. ✅ Confirmed.

# Description

Part of removing all PNI (Privacy Not Included) / Buyers Guide code (TP1-4191). 
Removes the buyersguide app entirely: cross-references into it, the app code and its tests, and the DB migration dropping its tables/fields.

Note: `wagtailpages/fields.py` looks `buyersguide-only` but stays. A historical migration (`0001_initial_2022.py`) imports from it directly and historical migrations can't be edited.

Note: Easier to review commit by commit: https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16522/commits

Note: this branch was merged with main partway through to pick up other changes. The merge commit resolves a conflict with the new Notice Banner feature (PR #16294), which added fields to buyersguide models after this branch had already deleted them. Kept the deletion.

# How to verify

- `inv docker-djcheck`: Django system check, should show only the 3 pre-existing unrelated template-tag warnings, no errors
- `inv docker-makemigrations-dryrun --args="--check"`: should report "No changes detected"
- Full suite against a fresh test DB. `inv docker-test-python` always runs with `--reuse-db` baked in, which would hide the migration-ordering bug this PR fixes, so force one fresh run first:
  `docker compose run --rm backend bash -c "source ./dockerpythonvenv/bin/activate && pytest foundation_cms/legacy_apps/wagtailpages foundation_cms/legacy_apps/reports foundation_cms/legacy_apps/wagtailcustomization -n auto --create-db"`
  (subsequent runs can go back to `inv docker-test-python`)
- `grep -rln "BuyersGuide\|ProductPageFactory" --include="*.py" foundation_cms/ | grep -v /migrations/` should return nothing


foundation_cms/legacy_apps/reports/tests/test_page_type_report.py
mmmavis, line 1: Swapped `ProductPage` for `BlogPage` here instead of just deleting these lines. These tests actually verify the page-types report's cross-type ordering/counting/locale logic, not anything product-specific, so they need some second page type to stay meaningful. `BlogPage` was just the simplest stand-in with no parent-type restrictions.

robdivincenzo (approved):
R+ @mmmavis , think everything I had is answered or will be handled later!

Worth flagging to the team when this is merged in main so that anyone who deploys is aware of the backup + rollback plan. 🚀

PR #16537, opened 2026-08-31
ci: lean mode default for local copy

robdivincenzo:
# Description

This PR updates the ci to be "lean mode" for legacy content factory generation by default on local copies, with an optional flag to run the full factory generation. This should result in time cost savings per DB creation instance.

What it doesn't do (yet)- generate legacy content factory only on ci:workflows. There's Percy tests still here which would cause a lot of failure. We've discussed retiring those checks and so I recommend we only make ci workflows lean once we no longer run the full legacy test suite. It's simply removing the flag within the PR that flips legacy tests off.

Related PRs/issues: TP1-4144

# To Test
1) Check out the branch locally
2) Run the new `inv new-db` command
3) Note only the bare minimum legacy content created (homepage, main nav, etc.). PNI / Mozfest should not exist.
4) Run `inv new-db --full` command
5) Note the full suite runs and all legacy content is generated, PNI and Mozfest included.


tasks.py
mmmavis:
```suggestion
    initialize_database(ctx, slow=False, full_legacy=True)
```

should we pass `full_legacy=True`?

foundation_cms/legacy_apps/wagtailpages/factory/__init__.py
```diff
@@ -54,6 +54,38 @@ def generate(seed):
...
+    the sections behind them are populated. That fixes the floor for "barebones":
+    the locales, the Homepage and its Site record, the blog (highlights indexes
+    four BlogPages and ideas_posts picks from the same set), profiles to author
+    them, and the homepage section orderables.
+
+    Everything reachable only from a deeper listing is still skipped: the
+    buyersguide/PNI, publications, campaigns, MozFest, donate, the RCC and
+    research hub libraries, the styleguide and youtube-regrets pages.
+
+    Ordering follows generate() above, which is deliberate and not alphabetical.
```
mmmavis: I'm worried about the maintenance burden here since we now have a second copy of the steps that has to be kept in sync manually so the ordering matches `generate()`. What doyou think about turning the steps in `generate()` into a shared list of `(step, required_for_barebones)` tuples, then having `generate_barebones()` just filter that list for the ones marked required?

mmmavis (changes requested):
Thanks @robdivincenzo . Can't wait for this to land so we can speed development process up 🚀 

I left two inline comments + also wondering if we should update `migration-cleanup.sh` line 48 to include the `--full-legacy` flag?

foundation_cms/legacy_apps/wagtailpages/factory/__init__.py
robdivincenzo: Great idea!

mmmavis (approved):
Thanks for the fixes! Verified the STEPS refactor, the schema-snapshot full_legacy fix, and migration-cleanup.sh's new flag all work correctly, ran both inv new-db and inv new-db --full-legacy end to end and everything checks out.

mmmavis (approved):
Thanks @robdivincenzo . This works great 🚀 🎉

Issue #16429, opened 2026-08-13
ci: hoist duplicated CI env vars to workflow-level env in continous-integration.yml


data-sync-user:
**Codebase:** Overall
**Impact:** Low-Medium
**Notes:** `preroll`, `test_wagtail`, and `test_integration` in `continous-integration.yml` each repeat an identical 16-var env block (BASKET_URL, CONTENT_TYPE_NO_SNIFF, CORS_ALLOWED_ORIGINS, DATABASE_URL, DEBUG, DJANGO_SECRET_KEY, DOMAIN_REDIRECT_MIDDLEWARE_ENABLED, NEWSLETTER_SIGNUP_METHOD, PIPENV_VERBOSITY, RANDOM_SEED, SET_HSTS, SSL_REDIRECT, TARGET_DOMAINS, USE_S3, X_FRAME_OPTIONS, XSS_PROTECTION, WAGTAIL_NOTIFICATION_EMAIL, WAGTAIL_NOTIFICATION_EMAIL_PASSWORD). GITHUB_TOKEN and ALLOWED_HOSTS are shared across all 5 jobs that set env at all. Moving these to a workflow-level `env:` block (inherited by every job, with job-level overrides still possible where values differ) removes the duplication and the risk of the copies silently drifting apart over time. 
**Decision:** ✅ Should do

Source: follow-up from TP1-4129 (PR #16428).

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4176)

data-sync-user:
> **Mavis Ou** commented from Jira:

PR is here: [https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16559](https://github.com/MozillaFoundation/foundation.mozilla.org/pull/16559)

PR #16559, opened 2026-09-04
ci: hoist duplicated env vars to workflow-level env in continuous-integration.yml

mmmavis:
## Description

`preroll`, `test_backend`, and `test_integration` in `continuous-integration.yml` each repeated an identical 16-var env block (`BASKET_URL`, `DATABASE_URL`, `DJANGO_SECRET_KEY`, etc.), and `GITHUB_TOKEN`/`ALLOWED_HOSTS` were duplicated across all four jobs that set job-level env at all. This risked the copies silently drifting apart over time.

This PR moves all of these to a single workflow-level `env:` block, inherited by every job. Job-level env still overrides where a job needs a different value:
- `test_backend` keeps its own `COVERALLS_REPO_TOKEN`
- `unit_test_frontend_redesign` keeps its own `PULSE_API_DOMAIN`
- `test_integration` keeps its own `ALLOWED_HOSTS` (needs `127.0.0.1` in addition to the other hosts) and its CSP/security vars

No behavior change intended — this is a pure de-duplication.

Related PRs/issues: #16429


.github/workflows/continuous-integration.yml
ramram-mf, line 217: Should we remove these too? `WAGTAIL_NOTIFICATION_EMAIL` and `WAGTAIL_NOTIFICATION_EMAIL_PASSWORD`

ramram-mf (changes requested):
Great cleanup! I just found another instance of some env vars that you have covered.

.github/workflows/continuous-integration.yml
mmmavis: Thanks for catching these!

ramram-mf (approved):
Thanks @mmmavis !

PR #16564, opened 2026-09-08
feat(careers): add Greenhouse job board block

OswaldoPinedaMozilla:
Adds a Greenhouse job board block so Mozilla Foundation job postings show on our own site. Editors add the block to a page and postings sync from Greenhouse on their own, with no CMS work when a role opens or closes.

Jira: [TP1-4208](https://mozilla-hub.atlassian.net/browse/TP1-4208) (spike: [TP1-4076](https://mozilla-hub.atlassian.net/browse/TP1-4076))

## Pages to test

- `/careers` (redirects to `/en/careers/`)

## Preview app

Heroku app: https://foundation-s-tp1-4208-g-ksb3dw.mofostaging.net/

## Scenarios

Greenhouse's embed has no empty or error state of its own, so a board with no roles, or an outage, would leave a blank hole in the page. The block asks Greenhouse's public API what the board holds before deciding what to render.

| State | When | What renders |
|---|---|---|
| `board` | The board has open roles | The Greenhouse embed |
| `empty` | The board has no open roles | Our "no open roles" panel |
| `degraded` | The API cannot be read, but the token is valid | The embed anyway, plus a link to the hosted board |
| `unavailable` | No token, or a token Greenhouse rejects | Our "temporarily unavailable" panel, no link |

`degraded` renders the embed on purpose. The API and the embed are separate systems, so a failure reading one is not evidence the other is broken, and hiding real openings would be worse than an extra link.

Copy for the panels is editable in the CMS. The board token is not: it comes from an env var, because a typo would take the page down and a text field inside the page body would be handed to translators as a string to translate.

## Design

The listing renders inside a Greenhouse iframe on `job-boards.greenhouse.io`. It is a different origin, so none of our CSS or JS reaches inside it. That is a browser boundary, not a Greenhouse setting, and there are no CSS variables exposed that we could override.

**Ours to change:**

- Page title, intro copy, and anything else on the page
- The two fallback panels (markup, styles, and copy)
- The iframe's box: width, height, margins, and the spacing to the footer

**Greenhouse's:**

- The "Current openings" heading, search field, filters, and the job list
- The application form
- Fonts, colours, and spacing inside the listing

The only place to change any of that is Greenhouse Design Studio. Support for custom CSS ended on 13 March 2026, so there is no other way in.

The board is on defaults right now: the body text falls back to Times New Roman and the heading uses a blue that is not in our palette. Someone needs to own that setup before launch.

## How to test locally

Add to your `.env`:

```
GREENHOUSE_BOARD_TOKEN=mozillafoundation
```

and append the Greenhouse hosts to the CSP lines (see `env.default` for the full values):

```
CSP_SCRIPT_SRC=... https://boards.greenhouse.io
CSP_FRAME_SRC=...  https://job-boards.greenhouse.io
CSP_CHILD_SRC=...  https://job-boards.greenhouse.io
```

Then restart the container (the `.env` is read at container start, not on reload) and seed the page:

```
inv manage "load_redesign_data"
```

The board token `mozillafoundation` is not a secret: it ships in the embed snippet in the page source. If you would rather not copy it from here, ask Oswaldo on Slack.

Each state:

- **board**: the setup above. P&C has a sample job published, so you should see the listing.
- **unavailable**: set `GREENHOUSE_BOARD_TOKEN=` (empty), or to a token that does not exist.
- **degraded**: cut the container's network access, or temporarily lower `REQUEST_TIMEOUT` in `foundation_cms/blocks/greenhouse.py`.
- **empty**: no public Greenhouse board is currently empty, so this one cannot be triggered naturally. Temporarily return `[]` from `fetch_greenhouse_jobs`. Covered by unit tests.

## Deploying

`EDITABLE_FOOTER` is `True` on staging and production, so their footers come from the CMS. Neither the template nor the fixture change in this PR touches an existing footer, so the Careers link has to be repointed by hand, once per environment.

Per environment, staging first:

1. Set the config vars: `GREENHOUSE_BOARD_TOKEN`, plus `https://boards.greenhouse.io` in `CSP_SCRIPT_SRC` and `https://job-boards.greenhouse.io` in `CSP_FRAME_SRC` and `CSP_CHILD_SRC`. Without the CSP entries the embed is blocked and the page renders an empty board.
2. Create and publish the Careers page, then verify the board renders.
3. Wagtail, Snippets, Site Footer, the active footer: remove the Careers external link and add an internal link `/careers/`.
4. Confirm the footer link resolves to the new page.

Staging config vars are already set. Production is not.

## Notes

- CSP is added to `env.default` and `app.json`. Staging and production config vars still need to be set, along with `GREENHOUSE_BOARD_TOKEN`, before the page is published.
- The block is registered for General Pages only, so the migration touches `generalpage.body` and nothing else.
- The console shows a CSP warning about an inline event handler. Greenhouse's loader sets `onload="undefined"` on the iframe it injects. The handler does nothing, and it is not fixable from our side without weakening the CSP.


foundation_cms/blocks/greenhouse_board_block.py
```diff
@@ -0,0 +1,80 @@
...
+        max_length=60,
+        default="View them on Greenhouse",
+        help_text="Label for the link to the Greenhouse-hosted board.",
+    )
+
+    class Meta:
+        template_name = "greenhouse_board_block.html"
+        icon = "list-ul"
+        label = "Greenhouse Job Board"
+        description = "Embeds the Greenhouse job board. Only one per page: the embed uses a fixed element id."
```
ramram-mf, line 56: Could we enforce one `greenhouse_board` per page with a parent StreamField count constraint?

I reproduced this in the review-app Careers page by duplicating the block and opening an unsaved preview. Both containers render with `id="grnhse_app"`, but only the first gets an iframe. The second stays empty at 480px high, pushing the footer down 552px including surrounding spacing.

Please also add a validation test rejecting a second board.

Image

*Unsaved review-app preview: empty duplicate board highlighted.*

ramram-mf (changes requested):
Found one validation issue, reproduced on the review app; details inline.

foundation_cms/blocks/greenhouse_board_block.py
OswaldoPinedaMozilla: thanks!

foundation_cms/core/management/commands/load_careers_test_data.py
```diff
@@ -0,0 +1,95 @@
...
+from django.db import transaction
+from wagtail.models import Locale, Site
+
+from foundation_cms.base.utils.helpers import to_streamfield_value
+from foundation_cms.core.models import GeneralPage, HomePage
+
+CAREERS_SLUG = "careers"
+
+
+class Command(BaseCommand):
```
ramram-mf: Could we follow the factory approach discussed in #16189 for this test-data setup? Move Careers generation into a reusable factory/generator registered with `load_redesign_data`, rather than introducing a separate management command and postdeploy step. The factory README documents that extension point, and the homepage/site dependencies are already available there. If this adds too much scope, we could move the fixture work into a focused follow-up.

.github/workflows/continuous-integration.yml
```diff
@@ -118,6 +118,10 @@ jobs:
             apt-archives-test_backend-gettext-ffmpeg-${{ runner.os }}-${{ matrix.shard }}-
       - name: Install additional tooling
         run: |
+          # The runner image ships Google's Chrome apt repo, which periodically
+          # serves an index whose hash does not match its Release file. None of
+          # these packages come from it, but apt-get update still exits 100.
+          sudo rm -f /etc/apt/sources.list.d/google-chrome.list
```
ramram-mf: Could we move the Chrome apt repository workaround into a separate CI PR? It addresses a general runner issue rather than the Greenhouse integration. If it is needed to unblock these checks, we can merge that fix first and then update this branch.

foundation_cms/core/management/commands/load_careers_test_data.py
```diff
@@ -0,0 +1,95 @@
...
+            stream_block=page.body.stream_block,
+        )
+
+    @transaction.atomic
+    def handle(self, *args, **options):
+        default_locale = Locale.get_default()
+        home_page = self._get_home_page()
+
+        if options["reset"]:
+            GeneralPage.objects.filter(slug=CAREERS_SLUG).delete()
```
ramram-mf: For any generation/reset logic retained here or moved into a factory, could we scope the page lookup and reset to the intended homepage/site? The current slug-based selection can pick an unrelated Careers page, and reset can delete matches across parents, sites, and locales. Please cover repeated generation/reset with an unrelated same-slug page so we know it stays untouched.

foundation_cms/footer/factories.py
```diff
@@ -119,6 +119,7 @@ def generate(seed):
             ("Annual Reports & Financials", "/meet-mozilla/annual-reports-and-financials/"),
             ("Press Center", "/meet-mozilla/press-center/"),
             ("Grantmaking", "/what-we-do/grantmaking/"),
+            ("Careers", "/careers/"),
```
ramram-mf, line 122: Staging and production have `EDITABLE_FOOTER` enabled, and both live English footers currently use the Corp Careers URL. The factory change only populates new footers, so could we document the one-time CMS update of the existing active footer to an internal `/careers/` link, coordinated with publishing and verifying the Careers page in each environment? The template and fixture changes alone will not update those existing links.

foundation_cms/legacy_apps/templates/fragments/footer.html
```diff
@@ -49,7 +49,7 @@
                                             {% with footer_link_classes="tw-dark tw-text-white hover:tw-underline" %}
                                                 <li class="tw-mb-4"><a id="donate-footer-btn" href="?form=donate-footer" class="{{ footer_link_classes }}">{% trans "Donate" %}</a></li>
                                                 <!-- the rest of the links should be listed alphabetically -->
-                                                <li class="tw-mb-4"><a href="https://careers.mozilla.org/listings/?team=Mozilla%20Foundation" class="{{ footer_link_classes }}">{% trans "Careers" %}</a></li>
+                                                <li class="tw-mb-4"><a href="/careers/" class="{{ footer_link_classes }}">{% trans "Careers" %}</a></li>
```
ramram-mf, line 52: The ticket calls for updating the Careers footer link, but does not specify legacy surfaces. Is the intended cutover for the redesign footer only, or should it also include the legacy Foundation and MozFest footers changed here?

ramram-mf (commented):
Thanks for fixing the duplicate-board validation. My remaining feedback covers the fixture-generation approach and reset scope, footer scope and cutover, and separating the unrelated CI workaround.

foundation_cms/core/management/commands/load_careers_test_data.py
OswaldoPinedaMozilla: Moved into `core/factories/careers.py` with a generate(parent) function registered in `load_redesign_data`, following the README's extension point.

.github/workflows/continuous-integration.yml
OswaldoPinedaMozilla: Good call to question it. I checked whether it is still needed and it is not: main has never carried the workaround and has run apt-get update plus the gettext/ffmpeg install successfully across all four backend shards and the integration job since this was written. The upstream apt issue resolved on its own.

So rather than splitting it into its own PR, I dropped it entirely. All three workflows are back to main on this branch, and no separate PR is needed.

foundation_cms/core/management/commands/load_careers_test_data.py
OswaldoPinedaMozilla: The lookup is now `GeneralPage.objects.child_of(parent)` so it is scoped to the homepage it is given. I dropped the reset flag entirely rather than scoping a delete, so there is no deletion left in this path.
  
`core/tests/test_careers_factory.py` covers both cases you asked for: an unrelated same-slug page under a different parent stays untouched, and it still does after generating twice.

foundation_cms/footer/factories.py
OswaldoPinedaMozilla: Confirmed, and thanks for catching it: EDITABLE_FOOTER is True in both, so the live footers come from the CMS and neither the template nor the fixture change touches them.

Per environment, staging first:
1. Publish and verify the Careers page.
2. Wagtail → Snippets → Site Footer → active footer: remove the Careers external link, add an internal link /careers/.
3. Confirm the link resolves and the board renders.

Adding this to the PR description as a deployment checklist.

foundation_cms/legacy_apps/templates/fragments/footer.html
OswaldoPinedaMozilla: Intentionally all three. Legacy Foundation and MozFest currently point at careers.mozilla.org, which is Corp's board, and that stops listing Foundation roles once we move to our own. Leaving them out would leave two live surfaces pointing at a list that no longer includes us.

MozFest gets the absolute https://www.mozillafoundation.org/careers/ since it is a separate site, where a relative URL would 404. Happy to pull them out if you would rather keep this PR to the ticket's literal scope.

ramram-mf (approved):
Thanks @OswaldoPinedaMozilla , looks good now

PR #16567, opened 2026-09-08
Add nothin_personal core page factories

adalcima:
# Description

This PR adds seed data factories for the initial Nothing Personal content hierarchy and registers them in the redesign data loader.

Link to sample test page: https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/cms/pages/26/
Related PRs/issues: [TP1-4157](https://mozilla-hub.atlassian.net/browse/TP1-4157)

**Changes in Models:**
- Added factories for:
    NothingPersonalHomePage, NothingPersonalArticleCollectionPage, NothingPersonalArticlePage, NothingPersonalPodcastPage, NothingPersonalFeaturedItem.
- Creates and publishes the Nothing Personal homepage, article collection, articles, and podcast.
- Registered the factory in load_redesign_data.py.

**Tests**
1. Open the CMS admin.
2. Navigate to Pages -> Nothing Personal.
    Verify that:
    - The [homepage](https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/en/nothing-personal/) is published.
    - The [Articles collection](https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/en/nothing-personal/nothing-personal-articles/) exists.
    - Existing [Nothing Personal articles](https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/en/nothing-personal/expert-profile-article-1/) are preserved.
    - The [podcast](https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/en/nothing-personal/np-podcast-1/) exists and is published.
    - A [hero article](https://foundation-s-tp1-4157-a-zxcpxh.mofostaging.net/cms/pages/27/edit/) is selected.

### Images
image


foundation_cms/nothing_personal/factories.py
```diff
@@ -0,0 +1,139 @@
...
+        podcast.save_revision().publish()
+        print("NothingPersonalPodcastPage created successfully.")
+    else:
+        print("Podcast already exists.")
+
+    # Set hero item to the first article if available
+    if article_pages:
+        home.hero_item = article_pages[0]
+        home.save_revision().publish()
+        print(f"hero_item set to: {article_pages[0].title}")
```
OswaldoPinedaMozilla: Reminder in case this isn't intended: this always replaces the hero, it never checks whether one is already set.

When articles already exist, `article_pages` gets filled with those instead of staying empty, so on a DB that already has content there's always something to assign and the current hero ends up replaced and republished. The description mentions existing articles are preserved, so I'm assuming the hero is an oversight rather than deliberate.

I ran it against a DB with real content (inside a transaction I rolled back) and the hero went from "What's in a 'Friend'?" to "Superficial Intelligence". On a fresh review app it doesn't matter since there's nothing to overwrite, but on a staging copy or a local with content it wipes out whatever an editor picked.

Only assign it when it's empty, same as `ensure_expert_curated_articles` in `profiles/factories.py`:

```suggestion
    if article_pages and not home.hero_item_id:
        home.hero_item = article_pages[0]
        home.save_revision().publish()
        print(f"hero_item set to: {article_pages[0].title}")
```

foundation_cms/nothing_personal/factories.py
```diff
@@ -0,0 +1,139 @@
...
+        print("Nothing Personal Home Page already exists.")
+    else:
+        home = np_models.NothingPersonalHomePage(
+            title="Nothing Personal",
+            slug=slug,
+            locale=locale,
+            seo_title="Nothing Personal",
+            search_description=fake.sentence(nb_words=12),
+            tagline=fake.sentence(nb_words=8),
+            body=_simple_rich_body(fake, paragraphs=3),
```
OswaldoPinedaMozilla: `rich_text` isn't one of the block types this field accepts. `NothingPersonalHomePage` narrows the list down to three blocks (`home_page.py:53-65`), and Wagtail drops unknown block types without raising, so this saves an empty body:

```
>>> p = NothingPersonalHomePage()
>>> p.body = [{'type': 'rich_text', 'value': '<p>hello</p>'}]
>>> len(p.body)
0
```

Only the home page is affected. The collection, article and podcast bodies do accept `rich_text`, so `_simple_rich_body` is fine everywhere else.

Since `body` is `blank=True` and none of the three allowed blocks are a plain text block, dropping it here seems simpler than building a `text_media_block` value just for seed data. Also wrapping the tagline in a `<p>`, it's a `RichTextField` so a bare sentence renders without markup.

```suggestion
            tagline=f"<p>{fake.sentence(nb_words=8)}</p>",
```

This doesn't surface today because the branch never runs (`profiles/factories.py` creates the home page first), but it would the moment that changes.

foundation_cms/nothing_personal/factories.py
```diff
@@ -0,0 +1,139 @@
...
+    """
+    reseed(seed)
+    fake = get_faker()
+
+    # Use site's root page if available, else Wagtail root node (same pattern as gallery_hub)
+    site = wagtail_models.Site.objects.filter(is_default_site=True).first()
+    root = site.root_page if site else Page.get_first_root_node()
+
+    if parent is not None:
+        root = parent
```
OswaldoPinedaMozilla: Small ordering thing: the Site lookup runs first and then gets thrown away two lines later whenever `parent` is passed, so the query happens for nothing. Checking `parent` first avoids it.

```suggestion
    if parent is not None:
        root = parent
    else:
        # Use site's root page if available, else Wagtail root node (same pattern as gallery_hub)
        site = wagtail_models.Site.objects.filter(is_default_site=True).first()
        root = site.root_page if site else Page.get_first_root_node()
```

Worth noting `parent` has no caller today (`load_redesign_data` calls `generate_nothing_personal(seed=42)` and that's the only call site), so dropping the parameter altogether would collapse this back to the original two lines. Either way works, just not the current order.

foundation_cms/nothing_personal/factories.py
```diff
@@ -0,0 +1,139 @@
...
+        article_pages = list(existing_articles_qs)
+        print(f"{len(article_pages)} NothingPersonalArticlePage already exist - skipping factory article creation.")
+    else:
+        article_pages = []
+        for i in range(3):
+            slug_i = f"np-article-{i+1}"
+            existing = np_models.NothingPersonalArticlePage.objects.filter(slug=slug_i, locale=locale).first()
+            if existing:
+                article_pages.append(existing)
+                continue
```
OswaldoPinedaMozilla: This lookup can't ever hit. The branch only runs when there is no `NothingPersonalArticlePage` under the home page, `NothingPersonalArticlePage.parent_page_types` is `["nothing_personal.NothingPersonalHomePage"]`, and the home page has `max_count = 1`. So if none hang off the home, none exist anywhere, and `existing` is always `None`.

```suggestion
            slug_i = f"np-article-{i+1}"
```

OswaldoPinedaMozilla (approved):

PR #16597, opened 2026-09-10
Handle gif to webp conversion asynchronously

alfredo-medina-stx:
# Description

This PR intends to fix an issue when uploading large `.gif` (pronounced *gif*, obviously) files through Wagtail: an `ffmpeg` process would synchronously wait to convert the `.gif` to a `.webp` file, failing after running out of memory trying to do so. 

The solution was to handle this asynchronously in an AWS lambda function, listening for objects falling under the `network/original_images/` path in the target S3 bucket. More details about the lambda integration on the `mofo-infra` repo.

Code changes include waiting for the `.webp` conversion to finish leveraging the cache, adding extra config as safe guards for larger `.gif` files, and preserve the synchronous, inline conversion if run locally.

**Tests are 100% AI-created**

# Checklist

**Tests**

**Changes in Models:**

**Documentation:**
- [x] Is my code documented?

**Merge Method**

### DIFF
--- a/foundation_cms/images/forms.py
+++ b/foundation_cms/images/forms.py
@@ -1,0 +1,52 @@
+from django.conf import settings
+from django.core.exceptions import ValidationError
+from django.template.defaultfilters import filesizeformat
+from django.utils.translation import gettext_lazy as _
+from wagtail.images.forms import BaseImageForm
+
+
+def validate_gif_upload_size(file):
+    """
+    Reject GIFs larger than GIF_MAX_UPLOAD_SIZE.
+
+    Wagtail already enforces WAGTAILIMAGES_MAX_UPLOAD_SIZE across every image
+    type. GIFs need a stricter ceiling of their own: they are converted to
+    animated WebP with ffmpeg, and a GIF costs far more memory and CPU to
+    process than a still image of the same byte size.
+
+    A no-op for non-GIFs, for files with no name or size, and when the limit is
+    unset or zero.
+    """
+    max_size = getattr(settings, "GIF_MAX_UPLOAD_SIZE", 0)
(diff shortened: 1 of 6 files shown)


robdivincenzo:
Hey @alfredo-medina-stx , can you spin this branch up on a review app in Heroku and provide a test script I can follow? Would love to see it in action!

robdivincenzo (changes requested):
Hey @alfredo-medina-stx , can you spin this branch up on a review app in Heroku and provide a test script I can follow? Would love to see it in action!

robdivincenzo:
Thanks @alfredo-medina-stx ! Bummer that it doesn't fully resolve the memory issue with the frames, but I still think this is valuable! For several reasons:

- Precalculating this avoids blowing up memory
- Precalculating this provides helpful editor error messaging
- Performance improvements by offloading the conversion (plus offloading ffmpeg package and dependencies off the heroku slug)
- MVP with a path to fully async later if scope increases

That said, I'm running into some edge case scenarios that need addressed first.

1. Uploading a gif, and then editing that image/choosing new image with a new gif will still point to the old gif and webp.
- This should follow wagtail conventions for edit/replace/delete/etc. Old fields should get wiped, old files, deleted and new fields update upon that change.
- It should also handle a situation where a gif is uploaded but then changed to a non-gif like a jpg or a webp

2. Immediately deleted gif's will orphan the converted_webp
- From Images -> Add Image -> Upload from your computer -> Immediately delete (not-update) -> webp orphaned inside the converted_webp directory of the s3 bucket.

Think we can look at these? Wonder if there are any other edge cases like this that a user might run into or trigger.

PR #16602, opened 2026-09-10
Update locale redirect behavior when non-localized URLs

adalcima:
# Description

This PR adds middleware `PreferredLocaleRedirectMiddleware` to redirect non-localized frontend URLs to the user’s preferred Wagtail locale when a valid language cookie exists.
URLs that already contain a locale, technical endpoints, and requests without a valid preference keep their existing behavior.

Link to sample test page: https://foundation-s-tp1-3643-u-wbzhmt.mofostaging.net/en/
Related PRs/issues: [TP1-3643](https://mozilla-hub.atlassian.net/browse/TP1-3643)

**Main Changes**
- Added `PreferredLocaleRedirectMiddleware`.
- Reads the existing `settings.LANGUAGE_COOKIE_NAME` cookie (default [django_language](https://docs.djangoproject.com/en/6.1/ref/settings/#:~:text=LANGUAGE%5FCOOKIE%5FNAME,-%C2%B6)).
- Redirects only `GET` and `HEAD` requests without a locale prefix.
- Preserves query parameters using `request.get_full_path()`.
- Excludes CMS, API, static, media, documents, JavaScript catalog, and other technical URLs.
- Keeps the existing language fallback through `LocaleMiddleware`.

**Tests**
1. Visit the review app and confirm the existing behavior remains unchanged.
2. Change the language using the `switcher` selector. Confirm language is updated.
3. Visit a non-localized URL and confirm it redirects to the selected locale.
4. Switch languages repeatedly from both localized and non-localized pages.
5. Confirm URLs such as /cms/ are not redirected.


robdivincenzo (changes requested):
Nice job @adalcima ! Just a couple flags. 

The `pt` redirect breaks here. If you navigate to https://www.mozillafoundation.org/pt/ you get redirected to the dialect path `pt-BR`. But if you use language selector and set language to French. You'll start navigating to `/fr/pt`.

I also think we need to exclude robots.txt to the excluded prefixes also. Will get bounced to locale path /fr/robots.txt and a 404 if the cookie is set (low chance of that happening to a real bot... but just covering edge case for anyone trying to review).

adalcima:
> also think we need to exclude robots.txt to the excluded prefixes also. Will get bounced to locale path /fr/robots.txt and a 404 if the cookie is set (low chance of that happening to a real bot... but just covering edge case for anyone trying to review).

Thanks for the nice catch @robdivincenzo . I updated the PR for `pt` with a `locale_aliases` dict and added `robots.txt` to the `excluded_prefixes`.

PR #15458, opened 2026-02-23
feat: update rich text block styles and streamfield layout

mmmavis:
# Styling changes

- constrain rich text block (standalone and the one within accordion block) width to 2/3
- update all rich text field text to 18px by default with line-height 1.5. `L` large text is now 20px big and with line height 1.5.

# Code changes

## feat: constrain rich text block width to 2/3 using block-wrapper mixin

- Add a new mixin `block-wrapper` to `_mixins.scss`
- Use this mixin to apply rich text width constraints (2/3) within the **standalone rich text block** and **accordion block** 

## refactor: scope accordion border variable to its selector

Small SCSS improvement

## feat: add rich-text-large class and update Wagtail hook for large text style

- Add `rich-text-large` style with 18px default and 20px large variant
- Update Wagtail rich text hook to use new rich-text-large class instead of `tw-body-large`, and set default rich text block font size to 18px.

## feat: add legacy class name aliases in _type.scss to ensure visual consistency before data migration

- Add `.tw-body-large` and `.body-text-large` as aliases for `.rich-text-large` within `.rich-text-block`
- These legacy class names are still stored in the database for existing content until a follow-up data migration is completed
- Ensures the L style renders consistently regardless of which class name is stored

# Follow-up ticket
Data migration to backfill legacy class names (`tw-body-large`, `body-text-large`) to `rich-text-large` is tracked in [TP1-3685](https://mozilla-hub.atlassian.net/browse/TP1-3685).

---

# To test 👨🏻‍🔬 👩🏻‍🔬 👨🏻‍🔬 👩🏻‍🔬 👨🏻‍🔬 👩🏻‍🔬 👨🏻‍🔬 👩🏻‍🔬 

Related PRs/issues: [Jira TP1-3099](https://mozilla-hub.atlassian.net/browse/TP1-3099) and [Jira TP1-3571](https://mozilla-hub.atlassian.net/browse/TP1-3571)

Link to review app: https://foundation-s-tp1-3099-3-sctpzn.mofostaging.net

Some examplepages that use rich text blocks:
- [Review app "join us" page](https://foundation-s-tp1-3099-3-sctpzn.mofostaging.net/en/join-us/) vs [staging](https://foundation.mofostaging.net/en/join-us/)
- [Review app "imagine" page](https://foundation-s-tp1-3099-3-sctpzn.mofostaging.net/en/what-we-do/imagine/) vs [staging](https://foundation.mofostaging.net/en/what-we-do/imagine/)
[...]

PR #16677, opened 2026-09-14
Update richtext classnames

robdivincenzo:
# Description

Adds a management command that backfills legacy rich text "large" class names (`tw-body-large`, `body-text-large`, and their combined form) to `rich-text-large` across the site, per [TP1-3685](https://mozilla-hub.atlassian.net/browse/TP1-3685). The Wagtail rich text hook was updated in #15458 to write the new class name going forward, but existing content across both `legacy_apps` and redesign models still carries the old classes because the same rich text feature is registered globally, not scoped to one app.

The command walks every model in the project, finds `RichTextField` and `StreamField` fields, and rewrites matching class names in place. It also fixes each object's latest revision so a pending draft publish can't reintroduce the old classes, and processes rows from a finished query rather than an open streaming cursor to avoid interleaving reads and writes on the same table. A `--dry-run` flag reports what would change without saving.

- Added `backfill_rich_text_large_class` management command in `foundation_cms/core/management/commands/`
- Verified field discovery and replacement logic against real data in the local dev database (dry run)

Link to sample test page:
Related PRs/issues: TP1-3685

# Test URL Paths
- /what-we-do/imagine/
- /mozilla-festival/highlights/mozfest-house-amsterdam-2024/
- /common-voice/platform-and-dataset/
- /insights/strengthening-data-ecosystems-for-education-in-india/appendix/
- /nothing-personal/test-article-1013/

# To Test
1) Clone a staging DB copy to local
2) View source of some pages with the old class names (`tw-body-large` and `body-text-large`)
3) Run `python manage.py backfill_rich_text_large_class --dry-run` (this will show you how much will be updated)
4) Run `python manage.py backfill_rich_text_large_class` (this will actually change things)
5) Refresh the pages and note the class name is now `rich-text-large` and no visual regression has occured

# Launch Notes
- Must be run it for real against staging, spot-check a few affected pages render unchanged visually


mmmavis (changes requested):
Hi @robdivincenzo , I ran this against a local staging clone, and the backfill itself worked really well. It covered all live fields, revisions, and translation memory. Awesome work!

One thing to check before this goes out: I found visual regressions on the legacy pages. For example, on `/en/common-voice/platform-and-dataset/` the large text got a bit bolder and taller, and slightly bigger on mobile. It looks fine to me, but we should probably check with designers to see if they are okay with it before we merge this PR. If not, let's add a small override in `redesign_migrated_content.scss` to bring back the old look.


Image

Image

robdivincenzo:
The two URLs for design review on the large rich text "Common Voice is the largest crowdsourced multilingual speech dataset in the world"

Current -> https://www.mozillafoundation.org/en/common-voice/platform-and-dataset/

The regression in review app -> https://foundation-s-tp1-3685-u-uqycjs.mofostaging.net/en/common-voice/platform-and-dataset/

Looks like the new class removes the thin `font-weight: 300` rule and changes line-height from `1.3` to `1.5`

PR #16683, opened 2026-09-14
Improve SEO Canonical and Lang Tags

robdivincenzo:
# Description

This PR adds canonical URL or hreflang tags. It contains new seo template tags and tests.

- Added `content_language` and `seo_links` template tags (`seo_tags.py`) to the redesign base template. `content_language` reports the actual rendered locale, following `page.alias_of` for aliases, instead of the URL prefix. `seo_links` emits a canonical URL for every page and hreflang alternates for pages with live, non-alias translations.
- Alias pages (English content served under another locale's URL as a fallback) get a canonical tag pointing at their source page instead of a self-referencing one, and no hreflang entries.
- Blanked the new SEO block in the footer and navigation snippet previews, since their preview context stubs in the site homepage rather than the content being edited. Extracted a shared `previews/_base.html` so future previews inherit that instead of repeating it.
- Confirmed the sitemap already excludes aliases (`alias_of__isnull=True` in `sitemaps.py`), so this PR doesn't touch sitemap generation.

Related PRs/issues: TP1-4292 TP1-4293

# To Test
## Verifying current state
1) View source on a default /en/ translated page: https://foundation.mofostaging.net/en/tp1-3980-image-grid-caption-qa/
2) Do a find and note the html tag lang attribute is set to en (correct) `<html lang="en"`
3) Do a find and note that the canonical link tag is missing (incorrect, we will fix this) `<link rel="canonical"` 
4) View source on the /de/ page (Wagtail calls this an alias and it is not a real page with translation) - https://foundation.mofostaging.net/de/tp1-3980-image-grid-caption-qa/
5) Do a find and note the html tag lang attribute is set to de (incorrect, the page is not translated and should still be set to en. We will fix this) `<html lang="de"`
6) Do a find and note that the canonical link tag is missing (incorrect, we will fix this) `<link rel="canonical"` 
7) View source on a translated /fr/ page (I translated this, and so Wagtail now considers this a real page with translation and not an alias): https://foundation.mofostaging.net/fr/tp1-3980-image-grid-caption-qa/
8) Do a find and note the html tag lang attribute is set to fr (correct) `<html lang="fr"`
9) Do a find and note that the canonical link tag is missing (incorrect, we will fix this) `<link rel="canonical"` 

## Verifying the fix
1) View source on a default /en/ translated page: https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/en/tp1-3980-image-grid-caption-qa/
2) Do a find and note the html tag lang attribute is set to en (correct) `<html lang="en"`
3) Do a find and note that the canonical link tag is present (corrected, and points to itself) `<link rel="canonical"`
4) Note the source also provides the alternate real translations (not aliases) as `<link rel="alternate"`
5) View source on the /de/ page (Wagtail calls this an alias and it is not a real page with translation) - https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/de/tp1-3980-image-grid-caption-qa/
6) Do a find and note the html tag lang attribute is set to en (corrected, the page content is `en` and now says so) `<html lang="en"`
7) Do a find and note that the canonical link tag is present (correct, and points back to `en` as source of truth) `<link rel="canonical"` 
8) View source on a translated /fr/ page (I translated this, and so Wagtail now considers this a real page with translation and not an alias): https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/fr/tp1-3980-image-grid-caption-qa/
9) Do a find and note the html tag lang attribute is set to fr (correct) `<html lang="fr"`
10) Do a find and note that the canonical link tag is missing (correct, and points to itself) `<link rel="canonical"`
11) Note the source also provides the alternate real translations (not aliases) as `<link rel="alternate"`


foundation_cms/templatetags/seo_tags.py
```diff
@@ -0,0 +1,71 @@
...
+    site_url = context.get("CANONICAL_SITE_URL")
+
+    if page.alias_of_id:
+        canonical_url = _absolute_url(site_url, page.alias_of.get_url(request=request))
+        return {
+            "canonical_url": canonical_url,
+            "hreflang_links": [],
+        }
+
+    canonical_url = _absolute_url(site_url, page.get_url(request=request))
```
ramram-mf: Could we preserve the routable subpage path here? On `/en/topics/privacy/`, the page context is still `HomePage`, so `page.get_url()` produces `/en/`. The rendered canonical points to the homepage, and the hreflang links likewise point to locale homepages.

The canonical should retain `/en/topics/privacy/`, with alternates pointing to equivalent topic routes where available. Could we cover this with a rendered topic-route test? The current ordinary-page tests don’t catch it.

[Reproduce on the Privacy topic page](https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/en/topics/privacy/).

Privacy topic source showing homepage canonical and locale homepage hreflang URLs

*Source of `/en/topics/privacy/`: canonical and hreflang URLs point to locale homepages.*

ramram-mf (changes requested):
The alias and translation examples look correct. Found one issue with canonical and hreflang URLs on topic routes; details inline.

foundation_cms/templatetags/seo_tags.py
robdivincenzo: Great catch @ramram-mf . Updated the PR to handle routablepages and a test for this as well. Redeployed to review app: https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/en/topics/privacy/

foundation_cms/templatetags/seo_tags.py
```diff
@@ -0,0 +1,81 @@
...
+def seo_links(context, page=None):
+    page = page or context.get("page")
+    request = context.get("request")
+    if page is None or request is None:
+        return {"canonical_url": None, "hreflang_links": []}
+
+    site_url = context.get("CANONICAL_SITE_URL")
+
+    if page.alias_of_id:
+        canonical_url = _absolute_url(site_url, page.alias_of.get_url(request=request))
```
ramram-mf: The real-page topic canonical and hreflang URLs now retain the route—thanks! Could we apply the same suffix handling in this alias branch?

On the deployed [alias Privacy topic](https://foundation-s-tp1-4292-4-wgalmo.mofostaging.net/en/tp1-4292-alias-topic-qa/topics/privacy/?utm_source=review), the canonical is `/en/`; it should be `/en/topics/privacy/`. The page renders `lang="en"`, omits the query string, and emits no hreflang links as expected.

This fixture is HomePage alias 37841, sourced from and placed beneath English HomePage 31532, with no copied descendants. Could we preserve the requested alias route suffix when building the source canonical and add a rendered alias + topic-route test?

Alias Privacy topic HTML canonical points to the source homepage

*Deployed alias Privacy topic: the HTML uses English and incorrectly canonicalizes to the source homepage.*

ramram-mf (changes requested):
The real-page topic fix looks correct. One remaining alias + topic-route issue is reproduced on the review app; details and evidence inline.

foundation_cms/templatetags/seo_tags.py
robdivincenzo: Thanks @ramram-mf ! Alias topic page should now capture the canonical as intended.

ramram-mf (approved):
Thanks @robdivincenzo !

PR #16689, opened 2026-09-15
fix(legacy): delete orphaned buyersguide page rows

mmmavis:
## Description

#16522 dropped the buyersguide page models but never deleted the corresponding `wagtailcore_page` rows, so any leftover page still points at a now-deleted content type. That breaks Wagtail admin actions (draft, publish, preview) since resolving `.specific` needs `content_type.model_class()`, which returns `None`. Found this on staging via a 500 when drafting the leftover PNI page.

Adds a follow-up migration that deletes any `Page` rows for the seven buyersguide content types, querying the base `Page` model so it can't hit the same error while running. No-op anywhere already clean.

Related PRs/issues: #16522


robdivincenzo (approved):
R+, thanks for the quick debug!

PR #16421, opened 2026-08-12
Comment out expert hub cta for launch

robdivincenzo:
# Description

This PR comments out the Expert Hub CTA for launch. Will have a follow-up PR via TP1-4172 to make it reusable.

Related PRs/issues: TP1-4171

## To Test
1) Go to your local copy's expert hub page on `main` branch: i.e. http://localhost:8000/en/mozilla-expert-hub/
2) Note the CTA at the bottom
3) Check out this branch `TP1-4171-remove-expert-hub-cta`
4) Refresh page, and note it's now removed

Issue #16420, opened 2026-08-12
Make Expert Hub CTA reusable


data-sync-user:
[From slack convo](https://mozilla.slack.com/archives/C98GW3P7X/p1786529299509119), Lauren would like the Expert Hub CTA removed for now, but editable later. This ticket is to track the work to make the component reusable.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4172)

PR #16692, opened 2026-09-16
Make the Expert Hub CTA a reusable CMS block

OswaldoPinedaMozilla:
## Description

Makes the Expert Hub CTA an editable Wagtail block. It was hardcoded in the template and commented out for launch in #16421, so editors could neither change it nor bring it back once the Fellows team settles on a contact process.

## What Changed

- Adds a Call to Action block with an editable heading, optional body, and button link, listed under Engagement in the editor.
- Gives the Expert Hub its own body block list, the shared base blocks plus the new one, so the block reaches the hub without altering any other page type.
- Moves the CTA styles out of the Expert Hub stylesheet into the block's own, and clips the decorative shapes horizontally so they cannot widen a page that lacks the hub's own clipping.
- Removes the commented-out include and the `_expert_hub_cta.html` partial.
- Adds `profiles.0015` as a schema-only migration. No existing content is transformed.

What it doesn't do: the block is registered but not placed, so the Expert Hub keeps showing no CTA until an editor adds it. Other page types cannot pick it yet. Adding it to `BASE_BLOCK_NAMES` would have rewritten the body schema of ten page models across five apps, so that is left for whoever needs it. Expert Profile pages stay untouched for the same reason; if the CTA belongs there, a snippet is the better shape since editing it profile by profile would be tedious.

## To Test

1) Check out the branch and run `inv migrate`
2) Open the Expert Hub in the CMS, add a Call to Action block to its body, fill in the heading, body, and button link, and publish
3) Visit http://localhost:8000/en/mozilla-expert-hub/ and note the CTA renders as it did before launch, with the yellow square and orange sunburst
4) Narrow the viewport to mobile width and note the decorative shapes do not add horizontal scroll
5) Run `inv test-python --file=foundation_cms/blocks/tests/test_call_to_action_block.py -n 1 --verbose`

Related PRs/issues: [TP1-4172](https://mozilla-hub.atlassian.net/browse/TP1-4172), #16420, [TP1-4171](https://mozilla-hub.atlassian.net/browse/TP1-4171)


ramram-mf (approved):
LGTM, thanks @OswaldoPinedaMozilla!

Issue #16690, opened 2026-09-15
Flaky test: TestApplePayDomainAssociationView leaks Django settings across test runs


data-sync-user:
`legacy_apps/tests/test_views.py` mutates `settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION`/`_MOZFEST` directly instead of using `override_settings`, with no reset. Since settings are process-global and pytest-xdist reuses processes across tests, a value set by one test can leak into another if execution order shifts, causing intermittent failures like:

```

Seen in [CI run 34875023577|https://github.com/MozillaFoundation/foundation.mozilla.org/actions/runs/34875023577/job/104079963028] and other CI runs.

*Fix:* use {{@override_settings(...)}} in that test class instead (auto-reverts per test). Verified locally, all 5 tests pass.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4312)

PR #16702, opened 2026-09-17
fix: use override_settings for Apple Pay domain association tests

mmmavis:
# Description

- `TestApplePayDomainAssociationView` in `foundation_cms/legacy_apps/tests/test_views.py` mutated `settings.APPLE_PAY_DOMAIN_ASSOCIATION_KEY_FOUNDATION`/`_MOZFEST` directly instead of using `override_settings`, with no reset
- Since settings are process-global and pytest-xdist reuses processes across tests, a value set by one test could leak into another if execution order shifted, causing intermittent failures like `test_mozfest_site_request_with_no_key_set - AssertionError: 200 != 501`
- Replaced direct mutation with `@override_settings(...)` decorators on the 4 affected test methods, which auto-revert per test

Link to sample test page:
Related PRs/issues: [Jira TP1-4312](https://mozilla-hub.atlassian.net/browse/TP1-4312) / #16690

## Test plan
- [ ] `inv test-python --file foundation_cms/legacy_apps/tests/test_views.py -n 1 --verbose` — all 5 tests pass


adalcima (approved):
R+, thanks @mmmavis

Issue #16052, opened 2026-06-08
Component Library Integration


data-sync-user:
Implement LP's global component library in the codebase. Apply typography, color, spacing, and layout tokens. Ensure components are structured so new pages can be added without LP involvement.

*Acceptance criteria:* All 15 components implemented, documented, and usable by internal team independently.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-3989)

data-sync-user:
> **Mavis Ou** commented from Jira:

All four PRs are merged into the `site-refresh-2026` feature branch:

- [**#16721**](https://github.com/mozilla/foundation.mozilla.org/issues/16721) Token build setup: LP's JSON becomes CSS vars and a Sass `$tokens` map, and a CI check catches files that are out of date.
- [**#16745**](https://github.com/mozilla/foundation.mozilla.org/issues/16745) Synced the token JSON with LP's repo, which adds `content.max` (1144px).
- [**#16742**](https://github.com/mozilla/foundation.mozilla.org/issues/16742) Swapped existing values to tokens where they already matched exactly (5 colors, 4 font weights, `$global-radius`).
- [**#16749**](https://github.com/mozilla/foundation.mozilla.org/issues/16749) Token usage docs in docs/stack.md, plus a `mobile-only` mixin for building desktop-first.

Checklist: Color, Typography and Spacing tokens are all available, and other components will move to them as they're reskinned. Breakpoint is a single token (`size-breakpoint-mobile`, 768px). Layout: the layout tokens are in the `size-*` set (e.g. `content.max`, 1144px). The 12-column grid isn't part of the refresh, so new components build against LP's layout directly.

PR #16721, opened 2026-09-18
feat(tokens): add token generation infrastructure 

mmmavis:

# Description

Sub-task 1 of TP1-3989 (Token Library Integration). Vendors the LP design system's token JSON, generates two consumable outputs from it, and wires both into the build pipeline.

## Why two generated files

CSS custom properties (`_tokens.scss`) resolve in the browser at paint time, fine for anything a component reads directly. Sass and Foundation-Sites' settings resolve once at compile time, before any CSS exists, so they can't read a custom property, media queries are the clearest case: `@media (min-width: var(--x))` isn't valid CSS anywhere. So breakpoints and Foundation overrides need a real Sass value, which `settings/_tokens.scss` (a `$tokens` map) provides.

## What's in this PR

- `frontend/redesign/tokens/`: the 7 raw JSON files vendored from LP's delivered design system.
- `frontend/redesign/tools/build-tokens.js`: flattens and merges them, resolves `{dot.path}` references, generates both outputs.
- `foundation_cms/static/scss/_tokens.scss`: CSS custom properties for runtime use.
- `foundation_cms/static/scss/settings/_tokens.scss`: a Sass `$tokens` map for compile-time use.
- `build:tokens` wired into `build` only. `watch:css` doesn't touch tokens at all, editing a token JSON file means running `build:tokens` by hand. Tried watching it too at first, but that watcher's own output matched its own glob and could retrigger itself, not worth the complexity for something this rare.
- `_redesign_base.scss` imports the generated custom properties globally.
- `docs/stack.md` documents the token build step, including that it's manual.

## Keeping the generated files from drifting

Both files are added to stylelint's `ignoreFiles` and `.prettierignore`, same pattern as `customized-settings.scss`. They're written verbatim from LP's JSON, no reason to make the generator chase hand-written formatting rules. A `check:tokens` script regenerates and diffs against what's committed, wired into the existing Lint & Format Check job, so a stale or hand-edited generated file fails the build instead of going unnoticed.

## What's deliberately not in this PR

Nothing consumes a token yet. No component, Foundation setting, or color/type/breakpoint value changes anywhere. Purely additive infrastructure.

## Verification

- Ran the full build, confirmed `build:tokens` executes in the right order.
- Diffed compiled CSS before and after: identical except the new `:root` block appearing on every page.
- Confirmed reference resolution against real data: `color-border-default` resolves from `{color.gray.1000}` to `#161616`.

**Next up (PR 2):** wires the Sass `$tokens` map into `customized-settings.scss` and swaps in the values already verified byte-identical to today's (5 colors, `$global-radius`, a font-weight subset).


ramram-mf (approved):
LGTM Thanks @mmmavis !

PR #16724, opened 2026-09-18
Fix Search termns report hits counter when using sort or filter on ex…

adalcima:
# Description

This PR fixes the **Search Terms Report** so the views (`hit`) counter only reflects genuine new searches, not repeated sort/filter interactions on an already-executed search. Adds a `Refinements` metric to retain visibility into that engagement without inflating the main hit count.

Link to sample test page:
Related PRs/issues: [TP1-4229](https://mozilla-hub.atlassian.net/browse/TP1-4229)

**Changes in Models and Templates:**
- **SearchEvent** model: added `is_refinement` boolean field.
- **search/views.py**: now only calls `add_hit()` for genuine initial searches. Requests coming from the sort/content-type/topic forms are tagged `is_refinement=true` and logged to `SearchEvent` without incrementing the native hit counter.
- **search/wagtail_hooks.py**: new _custom_ Search Terms Report that extends Wagtail's built-in report with a `Refinements` column, sourced from SearchEvent. Replaces the native **Search terms** menu item with this extended version (same label/position).
- **Templates**: the `sort`, `content-type`, and `topic filter` forms now submit a `hidden is_refinement=true` field so the backend can distinguish refinements from new searches.

**Tests**
1. Log in to the CMS admin and go to Reports -> Search terms.
2. Confirm the report loads with columns: `Views` and `Sort/Filter Refinements`.
3. Open the site frontend and perform a search for a new/unique term.
4. Return to Reports -> Search terms and confirm that term now shows Views = 1, Refinements = 0.
5. Go back to the search results for that same term and change the `sort order/content-type filter/pagination` a few times.
6. Refresh the report and confirm:
    - Views for that term did not increase.
    - Refinements increased by the number of changes you made.
    
### Images
image

image


foundation_cms/search/wagtail_hooks.py
```diff
@@ -0,0 +1,71 @@
...
+        NumberColumn("_refinements", label=_("Sort/Filter Refinements"), sort_key="_refinements"),
+    ]
+    export_headings = {**SearchTermsReportView.export_headings, "_refinements": _("Sort/Filter Refinements")}
+    list_export = SearchTermsReportView.list_export + ["_refinements"]
+    index_url_name = "foundation_search_terms_report"
+    index_results_url_name = "foundation_search_terms_report_results"
+
+    def get_base_queryset(self):
+        refinement_counts = (
+            SearchEvent.objects.filter(query_string=OuterRef("query_string"), is_refinement=True)
```
ramram-mf: Could we apply the report’s selected date range here too? Filtering to today still counted a refinement from 30 days ago in my test.

ramram-mf: Could we use the same query normalization for both models? With two spaces in `climate  change`, Wagtail stores `climate change`, but SearchEvent keeps both spaces. This comparison then misses the refinement and reports 0 instead of 1.

ramram-mf (changes requested):
Thanks Adal! The initial-search/refinement split looks good. I left two inline comments about the date filter and query normalization in the report.

adalcima:
> Thanks Adal! The initial-search/refinement split looks good. I left two inline comments about the date filter and query normalization in the report.

Thanks @ramram-mf , updated the PR to include the same native `normalise_query_string` function for normalization and a new `_restrict_to_date_range` function to maintain date ranges.

ramram-mf:
Thanks @adalcima , both earlier issues are addressed. One remaining case: if a search happens yesterday and gets refined today, filtering the report to today drops the term entirely because it has no initial hit in that period. Should we include it with 0 searches and 1 refinement, or is the report intentionally limited to terms with initial searches within the selected dates?

ramram-mf (approved):
Thanks @adalcima !

PR #16726, opened 2026-09-21
Fall back to the hero image on article listing cards

OswaldoPinedaMozilla:
## Description

Makes article listing cards fall back to the hero image when no Share Image is set, so an article with a hero image no longer renders an empty card on the Nothing Personal articles listing.

## What Changed

- Adds a `hero_image` fallback to the shared listing card partial. `search_image` stays primary, so cards that already render an image are unchanged.
- Adds template coverage for the three card states: the Share Image wins when both are set, the hero image is used when the Share Image is blank, and no image tag is emitted when both are missing.
- The fallback is deliberately not gated on `displayed_hero_content`. That field controls the hero on the article page itself, and gating on it would reproduce the same bug for articles whose media type is left at the default "No media".
- No model or migration changes. Placeholder illustrations for cards with neither image, and passing `hero_image_alt_text` through to the card, are left out on purpose; both are worth their own tickets.

## QA

Run:

```bash
inv test-python --file=foundation_cms/nothing_personal/tests/test_templates.py
```

Verify that:

- An article with a Hero Image and a blank Share Image shows the hero image on the articles listing card instead of an empty orange box.
- An article that has a Share Image still shows the Share Image, with no visible change.
- Topic listing pages under `/topics/<slug>/` render as before, including pages that have neither image.

Related PRs/issues: [TP1-4323](https://mozilla-hub.atlassian.net/browse/TP1-4323)


ramram-mf (approved):
Thanks @OswaldoPinedaMozilla !

PR #16742, opened 2026-09-22
feat(tokens): swap verified byte-identical values to token references

mmmavis:

# Description

Sub-task 2 of TP1-3989, follows #16721. Wires the generated Sass `$tokens` map into the redesign settings files and swaps in the values already verified byte-identical to LP's tokens.

## What's deliberately not in this PR

Grid or breakpoint values. After discussing with designers, we have decided to drop the 12-column grid from the refresh, so there's no reconciliation left to do there.


ramram-mf (approved):
Thanks @mmmavis !

Issue #16743, opened 2026-09-22
inv lint/inv format don't check redesign's Prettier formatting


data-sync-user:
`inv lint`/`inv format` only run stylelint for the redesign workspace's SCSS, never Prettier's `check-format`/`format`. CI's "Lint & Format Check (Frontend)" job checks both as separate steps, so a dev can run `inv format`, see no changes, push, and still fail CI on something Prettier would have caught. There's no wiring locally for that half of the check. Wire `lint_css`/`format_css` in `tasks.py` to also run redesign's `check-format`/`format` scripts, so local tooling actually matches what CI enforces.

┆Issue is synchronized with this [Jira Task](https://mozilla-hub.atlassian.net/browse/TP1-4339)

PR #16744, opened 2026-09-22
fix(tasks): run redesign's prettier check in inv lint/format

mmmavis:
Related PRs/issues: [Jira TP1-4339](https://mozilla-hub.atlassian.net/browse/TP1-4339) / #16743

# Description

`inv lint`/`inv format` only ran stylelint for CSS, never redesign's separate Prettier `check-format`/`format` scripts. CI's "Lint & Format Check (Frontend)" job checks both as separate steps, so `inv format` could report success while CI still failed on something only Prettier catches (this is what happened on [#16742](https://github.com/MozillaFoundation/foundation.mozilla.org/actions/runs/35670249547/job/106564870318)).

Legacy has no Prettier check in CI, so this only adds the redesign side.


ramram-mf (approved):
Thanks @mmmavis !

PR #16749, opened 2026-09-22
docs(tokens): document reskin token usage and add mobile-only mixin

mmmavis:

# Description

Adds a "Using tokens in a reskin" section to `docs/stack.md`: how to read tokens at runtime (`var()`) vs compile time (`map.get($tokens, ...)`), the single breakpoint token, and the decision to build new/reskinned components against LP's design system directly rather than Foundation's grid.

Also adds a `mobile-only` mixin to `_mixins.scss`. New/reskinned components are authored desktop-first to match LP's own CSS, so base rules are the desktop styling and this mixin wraps the mobile override, the opposite direction from the rest of the codebase's mobile-first Foundation styles.


ramram-mf (approved):
Thanks @mmmavis . Good to have updated docs on this!

